PR #108, opened 2019-06-27
Update first screenshot to use image, and compress files

jancborchardt:
New screenshot1.png with a more interesting picture than a whiteboard, which makes it look nice and obvious there’s image support. :)

Please review @nextcloud/text @georgehrke


georgehrke:
Maybe change Christoph's "asfökasd" in the second screenshot?

jancborchardt:
> Maybe change Christoph's "asfökasd" in the second screenshot?

That’s ok – something for next time @jospoortvliet @ChristophWurst ;)

PR #169, opened 2019-07-16
Adjust chunking

juliushaertl:
- [x] Commit bundles for shipping with release


juliushaertl:
I removed the check for committed bundles for now since I'm currently unable to figure out why the builds from within drone have a different contenthash than the ones from my local setup. Needs some further investigation but nothing blocking the app shipping I'd say.

PR #351, opened 2019-10-09
Rich workspaces

juliushaertl:
## ToDo

Requires https://github.com/nextcloud/server/pull/17482

- [x] Add endpoint for checking availability of files and creating short term sessions

66461155-b2e77880-ea78-11e9-88f6-b7551c614dc4


juliushaertl:
@jancborchardt for some design input (screen recording is in the first post)

jancborchardt:
Cool stuff! Some first design points:
- Very nice on only fading in the controls and date once the textarea is active!
- [ ] The placeholder text is italic, but it’s enough when it’s greyish like already.
- [ ] Instead of "Click to enter notes, lists or links" the placeholder could be "Add notes, lists or links …"
- [ ] The textarea is a bit far to the left. It could be left-aligned with the filetype icons instead?
- [ ] The width of the textarea should have a max-width for readability just like we do in the Text app
- [ ] I would put the Recommendations below the textarea. Currently they seem floating in mid-air a bit, and that way the recommended files are more connected to the file list.
- [ ] When adding text to an empty folder, can we prevent the "README.md" row being highlighted? It takes a lot of attention.

juliushaertl:
@jancborchardt Thanks for the input, all fixed :wink: 

image

image

juliushaertl:
@jancborchardt @nextcloud/designers Some things that we should consider:

- On mobile you end up with just seeing the text, so we probably should collapse it by default then or limit the height, if it is not focused
- I expect people want to disable this, so maybe a setting per instance or even per user makes sense here
- The limited with is currently the same as the text app uses, but when having a text block it looks a bit strange to me, maybe we should increase that?

What do you think?

jancborchardt:
>     * On mobile you end up with just seeing the text, so we probably should collapse it by default then or limit the height, if it is not focused

Yup, like 3.5 rows would be good, with a fade to white at the bottom. When you click it, quick animation to open.

>     * I expect people want to disable this, so maybe a setting per instance or even per user makes sense here

Yep. I see this as very similar to **Recommendations toggle in bottom left Files settings #108**. We should have the setting saying:

> Show rich workspace

Enabled by default, and it hides it everywhere, per user. Because the README.md is still in the files, you can still access it.



>     * The limited width is currently the same as the text app uses, but when having a text block it looks a bit strange to me, maybe we should increase that?

I’d say let’s go for the same width for now, just to keep it consistent as it’s specifically for readability.

One thing: The "a few seconds ago" should move further to the left so it’s kind of flush right with the end of the text instead of floating in space there. (Cutting that specifically to "seconds ago" would also help.)


>     * Should we handle more file names than README.md (like HEADER.md or even FOOTER.md)? I feel it gets a bit to crowded if we also have the option to add a footer.

Let’s keep it simple for now. README.md is the standard that GitHub and GitLab also use now, so this is established and some folders might even have that. We should of course also make sure that files like readme.md or Readme.md are shown – and maybe actually we should have the default casing be "Readme.md" to look less imposing to people who are not in the know about this?

jospoortvliet:
I'd not have a footer with this, just a header. And using just readme.md makes sense, too. The rest gets complicated.

The use case of a fixed footer for example - that can be done with the current README.MD app anyway.

I think it's important to think about rendering for public links, though - if a folder is editable, it should be editable, otherwise it should just show. A great way to add a note to a shared folder! Sadly it can't replace the existing 'add a note' as that is share-specific and of course also works on files.

karlitschek:
great. makes sense!

PR #375, opened 2019-10-24
Reduce 'Removed X inactive sessions' severity to debug

ovk:
Signed-off-by: ovk <koshelnikov@gmail.com>

* Resolves: #261
* Target version: master 

### Summary

This simply reduces 'Removed X inactive sessions' log message severity to debug. Eventually I decided not to wrap it in `if ($removedSessions > 0)` as there are some other messages logged for debug level and doing so would not be consistent.


juliushaertl:
Thank you very much

PR #397, opened 2019-11-04
Distinguish second and third-level list entries

jancborchardt:
Can’t test at the moment because my local setup is broken, but this should distinguish the sublevels in lists like so:

- 1st level is a filled circle
  - 2nd level is an empty circle
    - 3rd level and below is a square
      - 4th level

### DIFF
--- a/css/prosemirror.scss
+++ b/css/prosemirror.scss
@@ -132,6 +132,16 @@ div.ProseMirror {
 		list-style-type: disc;
 	}
 
+	// Second-level list entries
+	ul > li > ul > li {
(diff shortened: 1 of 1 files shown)


juliushaertl:
@jancborchardt Please add this to https://github.com/nextcloud/text/blob/999fabbb497b2ca05f168a51aa7f8a5fe0fa9b0d/css/prosemirror.scss#L133 where the general editor styling is located.

jancborchardt:
@juliushaertl right, thanks! Done :)

PR #415, opened 2019-11-13
Add dav plugin for rich workspaces

juliushaertl:
cc @tobiasKaminsky


juliushaertl:
Aparently the commit was already included in #423

PR #478, opened 2019-12-17
Remove dot from direct editing file extension

juliushaertl:
As discussed with @tobiasKaminsky


juliushaertl:
cc @LinneyS this should be adjusted in the OnlyOffice integration as well

PR #485, opened 2019-12-20
DoumentId should be a string when using as a filename

juliushaertl:
https://github.com/nextcloud/server/issues/18497

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -172,7 +172,7 @@ class DocumentService {
 	 * @throws NotFoundException
 	 */
 	public function getBaseFile($document): ISimpleFile {
-		return $this->appData->getFolder('documents')->getFile($document);
+		return $this->appData->getFolder('documents')->getFile((string) $document);
 	}
 
 	public function get($documentId) {


lib/Service/DocumentService.php
```diff
@@ -171,7 +171,7 @@ public function createDocument(File $file): Document {
 	 * @return ISimpleFile
 	 * @throws NotFoundException
 	 */
-	public function getBaseFile($document): ISimpleFile {
+	public function getBaseFile(string $document): ISimpleFile {
```
kesselb: Please update the phpdoc

juliushaertl:
Let's actually use a regular cast as you suggested. I'll look into adding proper type hints in general for 19.

kesselb:
https://github.com/nextcloud/text/blob/4f23339115bb27300c9058aa018707b4ec87cd8b/lib/Controller/DirectSessionController.php#L105-L107

Type cast above makes sure that documentId is a integer. That's fine as long as you're using it to fetch entities. For getting the file node cast it to a string first. Probably getBaseFile should do this. Otherwise you have to maintain this "documentId to fileNameDocumentId" conversion at any place calling getBaseFile.

Issue #317, opened 2019-10-02
Images in subfolder are not displayed


severin-lemaignan:
Images in my markdown document are placed in a subfolder of the working directory, like:

```
./figs/fig1.png
./document.md
```

When inserting an image in the document, eg:
```
![my image](figs/fig1.png)
```

the image is not displayed.

Text version: 1.0.2

juliushaertl:
Let's track this as an enhancement since relative paths in general are not supported at the moment.

wenzhuoz:
+1

It's such a nice feature to have. Markdown Editor (files_markdown) supports relative paths. I use them a lot in my own markdown documents, e.g.

    ![Certificates on System Keychain](Screenshots/Keychain_Access_System_Certificates.png)

QOwnNotes also inserts images by copying them to the `Notes/media` folder and adds relative paths like:

    ![IMG_20160629_103917](file://media/4726.jpeg)
    ![Screenshot from 2019-04-15 15-48-05](../media/390017047.png)

HappyClint:
looking forward to this feature.

Zaxzeng:
Any progress on this really useful feature?

azul:
#504 addressed this and has been merged.
It should fix this. Please reopen if it does not.

Issue #54, opened 2019-06-13
Visibility of images, placeholder, permissions (images don’t show for others)


jancborchardt:
I inserted a picture from the Nextcloud and it shows fine if I open the share link in a new tab of the same window where I’m authenticated in:
![image shown when opened from same window](https://user-images.githubusercontent.com/925062/59424558-79597600-8dd4-11e9-9905-0c3bd5bb18db.png)

In a private window though it doesn’t show:
![Image not displayed in shared link view](https://user-images.githubusercontent.com/925062/59424559-79f20c80-8dd4-11e9-9676-c94b794f27c6.png)

szaimen:
A possible solution might be putting the pictures and the text-file (with those embedded pictures) inside the same folder. When sharing the folder e.g. over a public link, the pictures are then also accessible and therefore should be automatic visible inside the text-file for everyone who opens the txt-file over this public link. (And not just the internal link to the file, as it is today)

On the other hand: if you share a text-file with embedded pictures, you probably always want the pictures inside the text-file to be visible for everyone you are sharing the file with. (But also just inside the text-file) So this isn't a data leak. (Its just like sharing e.g. a docx-file with embedded pictures.

juliushaertl:
I though a bit more about this. The approach I would take for now is that the text app has a dedicated preview endpoint. When inserting an image we store the file id of the image and the text file for example in a new oc_text_embedded table. If the  preview is requested then by the user though a share/share link we check if the table has a link between those two files and in that case return a preview of the file on the custom endpoint. The endpoint of course also needs to verify that the user actually has access to the text document.

jancborchardt:
Does this seem good @rullzer @MorrisJobke? ^

rullzer:
Yes a dedicated endpoint makes sense.
However it brings the bigger question to document editing. If I add a picture to a document. Do I expect that to be a link or the image at that time. In collabora/onlyoffice you insert the image at that time and not a link to the image.

juliushaertl:
Makes sense for images. But for adding a link to a office document for example it would make sense to still reference the original file.

jancborchardt:
Yeah, and we can do it in steps. For sure what’s expected for images is that the image shows in the document, so inserting it is probably easiest. Any linking to the original is something for a later step.

danimo:
[...]

PR #504, opened 2019-12-30
Allow embedding relative image paths

juliushaertl:
https://github.com/nextcloud/text/issues/317


jancborchardt:
@severin-lemaignan @wenzhuoz @HappyClint could you review this as it fixes the issue you reported at https://github.com/nextcloud/text/issues/317 – thanks a lot! :)

juliushaertl:
Still needs some more work to be done. :wink:

azul:
@juliushaertl I'm thinking about picking this one up.
I'll take a look at rebasing it and try it out.

What else is missing?

juliushaertl:
@azul That is great to hear. 

The most blocking thing here is basically to figure out how we can handle relative images during insert and later on when the text file might be accessed though a share or public link by another user. We could probably always use the relative path, but that would of course have the effect that it will only show an image that is in the same or in a child directory as the text file and it work if only the file is shared.

However that is also somehow a limitation of the current image insertion process with the preview. The current approach is takes the fileId as an argument which means that it will still show the image if it is somewhere available to the current user (also though a different share for example) but it also breaks when sharing though a public link. cc @jancborchardt 

There is also https://github.com/nextcloud/text/issues/54#issuecomment-524225866 with the idea of providing some additional image endpoint that provides images that were embedded in text files but that would of course be rather a next step as I assume that will be a pretty large amount of work.

azul:
@juliushaertl I see :eyeglasses ... lot's of things to consider here. I've been playing with this and thinking about it for a while... :thinking: 

In the long run i think it would be nice to have relative sources plus fileids in nextcloud itself:
```markdown
![Alt-Text](relative-path/image.jpg?fileid=1234)
```
That would provide the better readability in plain markdown of relative links and the flexibility of fileid based links. It would also allow resolving the image sources outside of nextcloud - for example when using the same content for a single page generator.

Do you agree that is a meaningful goal?

azul:
Obviously this does not answer all the other questions.
Images on publicly shared pages do not show up right now if only the page is shared. I'd like to not try and solve this as part of this pull request.

Having both the `filename` and the `fileid` in the `src` introduces redundancy and questions about what to do when the two conflict.

- [ ] From nextclouds point of view treat the filename as decorative as long as `fileid` is present. Do not attempt to sync these in any way as relative paths may differ between different sharing situations.
- [ ] When adding an image set the `filename` to the relative path the user adding the image is seeing.
- [x] If the `fileid` is absent display the image as your pull request already does. That's way better than only displaying a broken image.
- [ ] drop the ability to edit the relative source. If people have an image without a fileid it was added with some other mechanism. They can probably use that mechanism to fix the src if they understand what they are doing. If they don't they can remove the image and add the one they want.

So when only using nextcloud itself nothing changes. Only when looking at the markdown source or trying to reuse markdown files in other contexts the filename part of the source will be more meaningful.

azul:
@juliushaertl I pushed a new commit illustrating what i am aiming for.

Please let me know whether you think this is going in the right direction.

juliushaertl:
> That would provide the better readability in plain markdown of relative links and the flexibility of fileid based links. It would also allow resolving the image sources outside of nextcloud - for example when using the same content for a single page generator.

Yes that sounds like a sane way for that.

> Having both the filename and the fileid in the src introduces redundancy and questions about what to do when the two conflict.

> From nextclouds point of view treat the filename as decorative as long as fileid is present. Do not attempt to sync these in any way as relative paths may differ between different sharing situations.

Yes, the file id is the unique identifier we can use across shares.

> When adding an image set the filename to the relative path the user adding the image is seeing.
> - If the fileid is absent display the image as your pull request already does. That's way better than only displaying a broken image.
> - drop the ability to edit the relative source. If people have an image without a fileid it was added with some other mechanism. They can probably use that mechanism to fix the src if they understand what they are doing. If they don't they can remove the image and add the one they want.

:+1: 


>  Images on publicly shared pages do not show up right now if only the page is shared. I'd like to not try and solve this as part of this pull request.

Totally fine, let's do this step by step. Just for completeness, I thought about two different approaches for that, but I'm not yet sure about what the best one would be.

- Automatically create a shared link for the inserted image that could then be used for inserting (this somehow conflicts with the whole relative image path solution, but if we consider using relative image paths only for existing image that could be a way)
- Try to obtain a preview though the public preview endpoint based on the relative path of the share link. This will then only work with shared directories and not with single file share links of the markdown file

Another idea was to basically provide two modes of image insertion, like a "embed mode" that would create a share link and a "link mode" that would then only work if the user has already access.

As mentioned, a lot more things to consider and think about so all good with just implementing the relative support for now. This is already a huge improvement imo.

azul:
@juliushaertl I added another commit that makes use of the new src format when adding new images.

There's a few things i would still like to tweak:
- [ ] use relative path if shorter than absolute path.
- [ ] maybe get rid of some of the meta info.

For the relative path I will need to get the path of the current file. Not sure how to achieve this yet. I can't use the browser url because that will not work in apps like talk.

For the reduced meta info i would like to use defaults. Could we have `hasPreview` default to `true` and only add it if it's false? 

For the mimetype it seems natural to rely on the file extension for a default. So `.jpeg` files would be assumed to be of mimetype `image/jpeg`. That would probably require a lookup table for default mimetypes based on file endings. Seems somewhat ugly.
Right now the mimetype seems to only be used to determine if it is supported. If none is specified in the url it is assumed to be supported. So one could also only put mimetypes down if they are not supported (yet). However that would imply that support for certain mimetypes will not go away in the future.

What else do you think is missing for this to be merged?

azul:
@juliushaertl I cannot request your review because you initiated the pull request.
Please review it never the less. From my point of view this is a self consistent step ahead.
Please let me know what's missing. If i find some time I'll also look into reducing the metadata in the fragment of the url. But let's not wait for that.

juliushaertl:
Sorry for the delay here, I'll make sure to find some time to read though your messages and review the PR.

juliushaertl:
@azul Awesome work. This is really a huge step forward in embedding images. I'd say we get that in now.

I pushed a small fix to make the insert image dialog also work in rich workspaces, since the path was not passed properly to the editor in that case.

> For the reduced meta info i would like to use defaults. Could we have hasPreview default to true and only add it if it's false?

We actually can just try to load the preview and otherwise show the fallback. Not sure why it was there in the first place.

juliushaertl:
:+1: (I cannot self-approve my pr but let's get that in once CI is happy)

dincsi:
Hi! I'm just trying to migrate my Joplin notes to Nextcloud and I'm wondering if should this '' type of image markdown work would in Text? If I open in web or in ios app the image is not displaying and link is pointing to https://[SERVERNAME]/nextcloud/index.php/_resources/001.jpg

dincsi:
Sorry the markdown was:` `

dincsi:
After a little more investigation I realized, that: in IOS app - no images at all. If you go to nextcloud app you can see some images but not that ones which name consists some acented or special characters like á. `` is shown but `` is not.

Issue #452, opened 2019-12-06
Workspace polishing


juliushaertl:
Some stuff brought up by @jospoortvliet 

- [x] Inserted images and attachments need some size adjustments
- [ ] ~images and attachments should be arranged in a row if possible~
- [x] Checklists work but you first have to make a list, THEN make it a checklist
- [x] code block doesn't work on a single word, only on the entire block (works in Text)

dacoex:
Please add the issues w.r.t. Readme.md:
https://github.com/nextcloud/server/issues/19938

juliushaertl:
Let's close this and track follow up stuff in individual issues.

PR #505, opened 2019-12-30
Rich workspace polishing

juliushaertl:
- Fix menu bar popover positioning (https://github.com/nextcloud/server/issues/18494)
- Fix checkbox toggle behavior (https://github.com/nextcloud/text/issues/452)
  - Checklists work but you first have to make a list, THEN make it a checklist
  - Make sure that only a click on the checkbox toggles the state


jancborchardt:
@kesselb could you review if this fixes https://github.com/nextcloud/server/issues/18494 for you? :)

@jospoortvliet and also can you check if it fixes https://github.com/nextcloud/text/issues/452 for you? ;)

kesselb:
> Fix menu bar popover positioning ([nextcloud/server#18494](https://github.com/nextcloud/server/issues/18494))

:+1: 

Peek 2020-01-01 20-45

1) Type some text
2) Try to select all with CTRL + A
3) See "Add link" element

juliushaertl:
@kesselb There is no quick fix for the link popover positioning when all text is selected as the editor does not only have text selected internally but also the paragraph nodes that are wrapping around it.

kesselb:
> @kesselb There is no quick fix for the link popover positioning when all text is selected as the editor does not only have text selected internally but also the paragraph nodes that are wrapping around it.

OK. I was more confused about the button shown then the wrong position.

PR #510, opened 2020-01-02
Revert "Add more mimetypes for plain text editing"

juliushaertl:
Reverts nextcloud/text#509


juliushaertl:
broke collabora ... reverting for rc1 of 18

PR #512, opened 2020-01-03
Fix design of menububble for adding links

jancborchardt:
Before the alignment was quite broken, margin and text. Now fixed:
menububble before after

Please review @nextcloud/text


gary-kim:
Going to need a rebase :)

PR #516, opened 2020-01-03
Cypress test skeleton

juliushaertl:

* Resolves: # 
* Target version: master


juliushaertl:
Merging as it already has some basic coverage now. Can be extended further during development.

PR #618, opened 2020-01-26
Lint on github actions

rullzer:


juliushaertl:
Cypress build failure is unrelated. That is fine on master now with the newer cypress version.

PR #655, opened 2020-02-04
Check if the accessibility app is enabled

juliushaertl:
Fixes https://github.com/nextcloud/server/issues/19177


src/views/RichWorkspace.vue
```diff
@@ -71,7 +71,7 @@ export default {
 			loaded: false,
 			ready: false,
 			autofocus: false,
-			darkTheme: OCA.Accessibility.theme === 'dark',
+			darkTheme: typeof OCA.Accessibility !== 'undefined' && OCA.Accessibility.theme === 'dark',
```
skjnldsv: Why not just  ?
```suggestion
			darkTheme: OCA && OCA.Accessibility && OCA.Accessibility.theme === 'dark',
```

juliushaertl: Well that still throws an ReferenceError

skjnldsv: OCA is not defined?

juliushaertl: Ah never mind, to early :see_no_evil:

Issue #326, opened 2019-10-04
Code linebreaks & editor width


Gmentsik:
As mentioned in #320  there are line-breaks in the code-section. I'd prefer horizontal scrollbars instead because that's what most IDEs have as default setting and most devs are used to I guess.

I'm sure there are a lot of non-devs using this app, but having an option to switch between floating and fixed width (like eg. gitlab) would make working with code or tables also lot easier.

ahstro:
Created a PR to use horizontal scrolling instead of wrapping. I can expand on it and add the ability to switch, but I'm not sure if that's a desirable feature? 😇

PR #686, opened 2020-02-25
Use horiziontal scrolling instead of wrapping in code block

ahstro:
* Resolves: #326 
* Target version: master 

### Summary
If a line in a code block exceeds the width, hide overflow and add horizontal scrollbars instead of wrapping the line.

Edit: I didn't manage to find another way to create `pre` tags, but CSS is a devious mixtress, so keep that in mind when reviewing 😅


juliushaertl:
Thanks a lot :+1:

juliushaertl:
/backport 6bcca64380d97f614ad9774edfb6b922e1f2b696 to stable18

PR #695, opened 2020-03-05
Bugfix/noid/rich workspace empty height

juliushaertl:
- Remove static height and make the rich workspace adjust to the content with a defined max height
- Maximum height is 30vh
- Once focused it will increase to 50vh
- The height is reset only when changing directories


jancborchardt:
Got Cypress failing – seems cause it makes a request to `http://localhost:8081/index.php/apps/files` but my Nextcloud is at `http://localhost/nextcloud/index.php/apps/files/`.

Guess I can ignore the failures, also because the bot has them passing? Just for info. :)

juliushaertl:
> Got Cypress failing – seems cause it makes a request to http://localhost:8081/index.php/apps/files but my Nextcloud is at http://localhost/nextcloud/index.php/apps/files/.

Ah yes, those need some extra setup when running locally, we should not run them by default with make. Can be ignored for now ;)

jancborchardt:
It’s not related to this, but since you brought up the focusing – can we also only show the scrollbar when the rich workspace is focused? Currently it looks quite off:

scrollbar on text

Although we can’t do `overflow-y: hidden;` cause that would also get rid of the ability to quickly scroll through it while not focused. Is there a way we can have both: Ability to scroll but not scrollbar, only on that specific rich workspaces container, while not focused? Would look much cleaner.

juliushaertl:
> Although we can’t do overflow-y: hidden; cause that would also get rid of the ability to quickly scroll through it while not focused. Is there a way we can have both: Ability to scroll but not scrollbar, only on that specific rich workspaces container, while not focused? Would look much cleaner.

I don't see a way to do this. But scrolling though the collapsed workspace is also only partly working because of the fade out (this is scrolled to the bottom and not all text is currently visible):

image

danxuliu:
Backport to stable18 in #759

PR #746, opened 2020-04-02
Add index on the oc_text_sessions table

rullzer:
Just to make the queries index. Esp on big installations with lots of users editing text files this helps.


lib/Migration/Version030001Date20200402075029.php
```diff
@@ -0,0 +1,35 @@
...
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('oc_text_sessions');
```
juliushaertl, line 27:
```suggestion
		$table = $schema->getTable('text_sessions');
```

Ah, no need for the prefix :wink:

rullzer: right..

PR #748, opened 2020-04-02
Only return 404 for workspace property if it is not avilable

juliushaertl:
This way mobile clients can disable the use of rich workspaces when it is disabled either by the user or though `occ config:app:set text workspace_available --value=0`. When workspaces are enabled the dav property will now always contain an empty string.

I've tested that it still works fine with android. After that is merged, the android client can hide the "Add folder info" in the file creation menu.

cc @tobiasKaminsky @marinofaggiana

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -28,8 +28,10 @@ use OC\Files\Node\File;
 use OC\Files\Node\Folder;
 use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Files\FilesHome;
+use OCA\Text\AppInfo\Application;
 use OCA\Text\Service\WorkspaceService;
 use OCP\Files\IRootFolder;
+use OCP\IConfig;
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Hm, I am unsure about this idea, as this is totally unrelated code-wise.
Can't we have a user-based capability for this?

juliushaertl:
> Can't we have a user-based capability for this?

I also thought about this but with using this approach the usage of the rich workspace is properly blocked at the API level, so I went for this approach. As an App you don't need to care about the general state of the setting. This would also work fine if we at some point have folders where you can create workspaces and others where it is not possible. Any concrete concerns?

tobiasKaminsky:
We can release this independent, right?

juliushaertl:
> We can release this independent, right?

Aw no, text is shipped with the server so bound to the maintenance releases :/

tobiasKaminsky:
Ok, no problem, as client change is backward compatible.

PR #749, opened 2020-04-02
Use full list of supported filenames when obtaining the name to create

juliushaertl:
This makes sure that newly created files use the translation of "Readme.md" and the translated files are properly picked up for opening.

cc @jancborchardt

### DIFF
--- a/lib/Controller/WorkspaceController.php
+++ b/lib/Controller/WorkspaceController.php
@@ -85,11 +85,6 @@ class WorkspaceController extends OCSController {
 	/** @var IURLGenerator */
 	private $urlGenerator;
 
-	private const SUPPORTED_FILENAMES = [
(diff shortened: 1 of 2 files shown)


gary-kim:
I thought the file name for new files was on the client side. This is the file I patch for my instance.
https://github.com/nextcloud/text/blob/732da03d72fb2e20d4c9b666a5d2c0709213d011/src/views/RichWorkspace.vue#L152
Sorry, I might be misunderstanding something.

juliushaertl:
Ah, indeed, why did we have this in multiple places :see_no_evil: 

Let me see if we can reuse the workspace ocs api for that as well.

juliushaertl:
Let's get this in first then and fix the javascript case in a follow up.

PR #776, opened 2020-04-20
Remove unused preview plugin

juliushaertl:
The files app does no longer support javascript based preview plugins since Nextcloud 18


skjnldsv:
Icon changed
CypressError: Timed out retrying: Expected to find element: '#viewer-content .modal-header button.icon-menu-sidebar-white-forced', but never found it.

icon-info-force-white

juliushaertl:
Yep, that is fixed in a separate PR. 

I think we could even drop that test as for text it should be fine to assume that the viewer does its stuff properly, however I added a case for it back then since we were the only app integrating at that point in time.

juliushaertl:
/backport 69e0eca688ebabe5c0c185da92db75989aeaae89 to stable18

PR #798, opened 2020-04-27
Handle 403 errors the same as not found

juliushaertl:
There is no reason to retry since 403 error is most likely not temporarily,
so the editor connection can be closed whenever this happens.


MorrisJobke:
🙄

PR #829, opened 2020-05-16
emit `ready` signal via parent

azul:
Emit the `ready` signal via `$parent` as well.

In a wiki app we render a preview of the markdown content
while loading the text editor.

In order to know when the text editor is ready
we need `ready` to be emitted by the parent.

Alternatives:
* We are currently using the ViewerComponent in our app.
  Just reemitting `ready` from it would work as well.
* Ideally we would be able to build our own component
  based on the EditorWrapper component.
  Making the `EditorWrapper` public would work as well -
  but i guess that is not really an option
  because it may change or disappear in the future.

* Resolves: # 
* Target version: master


juliushaertl:
> Ideally we would be able to build our own component based on the EditorWrapper component. Making the EditorWrapper public would work as well - but i guess that is not really an option because it may change or disappear in the future.

You might be interested in https://github.com/juliushaertl/nextcloud-text which aims to extract the main editor functionality into a vue library. I started this a while ago with the goal to be able to use the editor in other apps as well, unfortunately I didn't have enough time yet to finish. As a first version this would not contain collaborative editing as that requires a backend (provided by the text app for now), but unifying that across apps is probably also something worth to consider.

juliushaertl:
@azul Mind to rebase this again and update the committed bundles? Feel free to push your branches directly to the repo in the nextcloud organization next time, that makes it a bit easier to get this merged then ;)

azul:
@juliushaertl, I rebased and updated. Will use branch on main repo next time :smiley:

PR #835, opened 2020-05-19
Test running cypress against stable19

juliushaertl:

* Resolves: # 
* Target version: master


juliushaertl:
Let's do this for now since otherwise cypress fails due to https://github.com/nextcloud/server/pull/20949#issuecomment-631057018

PR #853, opened 2020-06-05
dont use global moment

icewind1991:
Fixes some log spam


juliushaertl:
Pushed package-lock update and bundles.

PR #878, opened 2020-06-16
Show tooltip on link hover

juliushaertl:
Otherwise the user cannot determine the URL of the link before clicking.

This is a temporary fix until https://github.com/nextcloud/text/pull/687 is resolved, unforunately tiptap does not yet offer a way to show the menububble that we use for link editing if the link just has a cursor in it when clicking. Therefore we cannot use that to show the url in there [as shown here](https://user-images.githubusercontent.com/3404133/75259860-facd2900-57e8-11ea-943b-09e15389bc91.png)


src/marks/index.js
```diff
@@ -74,10 +74,39 @@ class Strike extends TipTapStrike {
...
+				{
+					tag: 'a[href]',
+					getAttrs: dom => ({
+						href: dom.getAttribute('href'),
+					}),
+				},
+			],
+			toDOM: node => ['a', {
+				...node.attrs,
+				title: node.attrs.href,
```
juliushaertl, line 97: This line is the only difference to the tiptap implementation

juliushaertl: https://github.com/scrumpy/tiptap/blob/master/packages/tiptap-extensions/src/marks/Link.js

nickvergessen:
Seems to need a rebase?

Issue #308, opened 2019-09-24
Easily add links to other nextcloud files


Klakla2:
**Is your feature request related to a problem? Please describe.**
Adding links to other files in my Nextcloud instance is tedious.

**Describe the solution you'd like**
Add an 'Add Nextcloud link' icon to the right of the current 'Add picture' icon to easily add a link to a file in your Nextcloud instance just how you do it to add a picture.

**Describe alternatives you've considered**
It is already possible to select a file using the 'Add picture' icon, but that one adds a big picture like rectangular shape and the link is not to the file itself, but to the directory in which it is contained. But the code should be very eaasy to adapt to add in this case a proper link to a file in my text.

**Additional context**
## Current tedious:

In order to add a link to another file in your Nextcloud you need to

 1. Have additional files tab open in your browser showing the files in Nextcloud

 2. Click on the files tab

 3. Go into the directory with the file to be linked

 4. In order to avoid spelling issues right-click the file and attempt to rename it

 5. Copy the name of the file to the clipboard

 6. Click on the text editor tab

 7. Paste the name of the file into your text editor

 8. Highlight the copied name

 9. Click on the files tab

10. Right-click the file to be linked again

11. Select Details

12. Click on the icon to do a direct copy of the Nextcloud Link to the clipboard

13. Click on the text editor tab

14. Paste  the Nextcloud Link as a link in the text editor

15. Done - you have a link like [NextcloudApps.md](https://cloud.adfc-berlin.de/index.php/f/193740)

## Proposed:

1. Click on the 'Nextcloud link' icon right to the current 'Add picture' icon

2. Select the file in that GUI as you would select a picture

3. Done - you have a link like [NextcloudApps.md](https://cloud.adfc-berlin.de/index.php/f/193740)

juliushaertl:
Sounds like a nice idea, we probably could also add a file picker option when a text is selected in addition to the existing url field for adding a link.

twkrol:
Great idea. An more, I would like to make a link to another nextcloud file, and when clicked to let it open the file automatically as I pointed to file, not a directory.

gorn:
It would be great to be able to use relative links - this way the editor can turn your filesystem in a very basic wiki. Please add another option when a word is selected - currently there is "add link" only. Something like "link to nexcloud file". In line with wiki style it would be great if the file is red if it does not exist and by clicking on it you would created nex .md file.

azul:
[...]

PR #879, opened 2020-06-16
Link to local files

azul:
* Resolves: #308 
* Target version: master 

### Summary

Add a button to the menu bubble that links to local files.
See #308 for descriptions of the approach.

- [x] open in new tab on middle click and ctrl-click.
- [x] change dir param according to relative Path.
- [x] Open the file directly in viewer
- [x] Create urls that work everywhere - even from talk.
- [x] Make all strings translateable.
- [x] ~~Fix creating links in workspace (currently uses wrong url for `PROPFIND`)~~ - also seems to be an issue with images


azul:
Here's a little demo of the basic functionality:
link-demo

azul:
I think this is enough for a first PR.

Things that i know are missing:
* changing the browser title when following links
* history / back button to go back after following a link
* linking from within the workspace area.
  (links work but creating them is broken just like inserting images is)

juliushaertl:
Wow, amazing contribution :heart: 

I'll give this a detailed review later, the general approach seems all good to me. 

> changing the browser title when following links
> history / back button to go back after following a link

Might be that those two are things we need to fix in viewer or the files app.

src/marks/index.js
```diff
@@ -74,10 +77,69 @@ class Strike extends TipTapStrike {
...
+						if (attrs.href && event.target instanceof HTMLAnchorElement) {
+							event.stopPropagation()
+							const htmlHref = event.target.href
+							if (event.button === 0 && !event.ctrlKey && htmlHref.startsWith(window.location.origin)) {
+								const query = OC.parseQueryString(htmlHref)
+								const fragment = OC.parseQueryString(htmlHref.split('#').pop())
+								if (query.dir && fragment.relPath) {
+									const filename = fragment.relPath.split('/').pop()
+									const path = `${query.dir}/${filename}`
+									OCA.Viewer.open({ path })
```
azul, line 122: This does not have the `fileId` as a fallback.
Sharing may lead to different directory layouts for different users.
If the calculated path is wrong for the current user... there is no way for the Viewer to recover from that.

We could find out the right path by looking up the file properties by id.
But that adds complexity and another roundtrip.
Since this is an optimization to open links more smoothly then with window.location it would be nice to avoid the roundtrip.

Not sure what the best approach here is.

azul:
@juliushaertl You are right about the history navigation. There's no `onPopState` handler in the files app yet.

nextcloud/viewer#497 is an attempt to move in that direction.

juliushaertl:
@azul Just a sidenode to make the rebasing process a bit easier, feel free to not commit the bundles until it is approved, and then we just add them right before the merge.

src/marks/index.js
juliushaertl: Hm, I'm also not sure what to do in that regards.

src/helpers/links.js
```diff
@@ -0,0 +1,84 @@
...
+
+const domHref = function(node) {
+	const ref = node.attrs.href
+	if (!ref) {
+		return ref
+	}
+	if (ref.match(/^[a-zA-Z]*:/)) {
+		return ref
+	}
+	const match = ref.match(/^([^?]*)\?fileId=(\d*)/)
```
juliushaertl, line 58:
```suggestion
	const match = ref.match(/^([^?]*)\?fileId=(\d+)/)
```

src/marks/index.js
```diff
@@ -74,10 +77,69 @@ class Strike extends TipTapStrike {
...
+			...inherited,
+			parseDOM: [
+				{
+					tag: 'a[href]',
+					getAttrs: dom => ({
+						href: parseHref(dom),
+					}),
+				},
+			],
+			toDOM: node => ['a', {
```
juliushaertl, line 94: When running into a conflict during rebase you mainly need the title to be added here:
https://github.com/nextcloud/text/pull/878/files#r440731486

azul: thanks for the heads up. Did just that. :heavy_check_mark:

azul:
@juliushaertl I rebased everything and added the regex fix you proposed.

jancborchardt:
Wow @azul, that is amazing! Especially the linking to other markdown files – it’s the Hypertext! :)

azul:
@juliushaertl I rebased again.

I think i addressed your requests. Please approve or let me know what else is missing.

Thanks :smiley:

jancborchardt:
By the way @azul, we have a chat channel "Text team" on our Nextcloud Talk instance. I can invite you for a guest account via your email address if you like? :)

azul:
> I can invite you for a guest account via your email address if you like? :)

@jancborchardt, please do :smiley:

juliushaertl:
@azul Mind to do one last rebase dance ? :see_no_evil:

juliushaertl:
Cypress failures are unrelated.

jochenstu:
Linking to a docx file leads to an "There is no plugin available to display this file type" Error with OnlyOffice installed.

azul:
@jochenstu I created a new issue for this so it does not get lost here.

kevinde91:
Great work so far!
As I was looking for relPaths in Nextcloud I came up here.

So the case that a file is moved is allready mentioned, but what if the file is copied?
A new fileid will be generated and the link will redirect to the old file, right?

Is it possible to link a file with unknown id while the relPath (including file name) is known?

Thanks for your help!

azul:
@kevinde91 I am not sure i understand the situation you are describing. Let me try to rephrase what i understood:
* I create a file called `original.md` - it gets some file id say 123
* I link to that file from another file in the text editor - the link will point to `original.md?fileId=123`
* I copy the file to `copy.md` - the new file gets a id say 234.

Now the link will still point to `original.md` with it's file id. This seems intuitive to me. Would you like this behaviour to change?

Another thing we were discussing was allowing the creation of links to yet to be created files:
* I link to `copy.md` even though it does not exist yet.
* When clicking the link it will create the file and allow me to edit it.
* When copying `original.md` to `copy.md` the link will start pointing to the copied file.

Is that what you were talking about?
My understanding is that this currently does not work with the text app. It lacks 
* The ui for generating links to arbitrary (possibly non-existing file names)
  (It should very much be possible by editing the markdown with other means.)
* Handling links to missing docs by creating the doc.

kevinde91:
@azul thanks for the fast reply, and sorry for the confusion.

I tried to recreated my problem and this time it behaved exactly how it should:

1. Create a folder 'Original'
2. Create a folder 'Subfolder' in 'Original'
3. Create file 'linked_file.md' in 'Subfolder' with some text and create file 'test.md' with a link to 'linked_file.md'
4. Create a folder 'Copy' in the root dir
5. Copy 'Subfolder' from 'Original' to 'Copy'
6. Change text in 'Copy/Subfolder/linked_file.md'
7. Open 'Copy/Subfolder/test.md' and click on the link, the new 'linked_file.md' with changed text will be openend.

Sorry again for wasting your time, not sure what I did wrong yesterday.

PR #888, opened 2020-06-20
fix: path for workspaces

azul:
We need the relative path inside a users home directory
to fetch the folder content for the filepicker.

$file->getPath() gets the absolute path
including username and files.

Instead use $path which we already know and append the file name.

Frankly I have no idea if this is the right way to do this in the Controller
as I am not familiar with the `OCP/Files` API. This works for me and i hope it illustrates the problem. Feel free to fix in a different way.

### DIFF
--- a/lib/Controller/WorkspaceController.php
+++ b/lib/Controller/WorkspaceController.php
@@ -107,7 +107,8 @@ class WorkspaceController extends OCSController {
 	 */
 	public function folder(string $path = '/'): DataResponse {
 		try {
-			$folder = $this->rootFolder->getUserFolder($this->userId)->get($path);
+			$userFolder = $this->rootFolder->getUserFolder($this->userId);
+			$folder = $userFolder->get($path);
 			if ($folder instanceof Folder) {
 				$file = $this->workspaceService->getFile($folder);
 				if ($file === null) {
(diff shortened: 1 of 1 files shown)


azul:
Split this one from #879 as it's a fix for an existing problem and should be easier to review and get in.

lib/Controller/WorkspaceController.php
```diff
@@ -124,7 +124,7 @@ public function folder(string $path = '/'): DataResponse {
 						'id' => $file->getId(),
 						'mimetype' => $file->getMimetype(),
 						'name' => $file->getName(),
-						'path' => $file->getPath()
+						'path' => $path . '/' . $file->getName()
```
juliushaertl, line 127: Just to be on the save side and not end up with double slashes:

```suggestion
						'path' => '/' . trim($path, '/') . '/' . $file->getName()
```

juliushaertl: Ah actually let's better use something like the following, since that then takes care about normalization of the path:

`$this->rootFolder->getUserFolder($this->userId)->getRelativePath($ile->getPath())`

azul:
@juliushaertl updated according to your comment.

lib/Controller/WorkspaceController.php
azul: Yes... that is what i was looking for :heart_eyes:

Issue #819, opened 2020-05-11
Enhance checkbox behavior


stefan-niedermann:
# Current behavior

One has to explicitely use the menu button to add checkboxes.

![text](https://user-images.githubusercontent.com/4741199/81563563-b2101f80-9396-11ea-8e3d-5af9b4cc0d9e.gif)

# Wanted behavior

This can be automated, so one can focus on *typing* and does not need to make a break to hit the menu button.

| Entered Key | Display |
| --- | --- |
| <kbd>-</kbd> | `-` |
| <kbd>Space</kbd> | `- ` |
| <kbd>[</kbd> | `- [` |
| <kbd>x</kbd> | `- [x` |
| <kbd>]</kbd> | `- [x]` |
| <kbd>Space</kbd> | :ballot_box_with_check:  |

After pressing backspace, the checkbox should disappear and the markdown displayed again

| Entered Key | Display |
| --- | --- |
|  | :ballot_box_with_check:  |
| <kbd>Backspace</kbd> | `- [x]` |

Valid combinations for checkboxes are of course with leading `-` as well as `*`, the value (within the brackets) can be ` `, `x`, `X`.

This is a video that demonstrates the wanted behavior (Nextcloud Notes app `3.3.0`)

![t7h37-kcjwy](https://user-images.githubusercontent.com/4741199/81562983-f5b65980-9395-11ea-870d-89e0b495d261.gif)

cc @korelstar since i know, that he aims to use Nextcloud Text in the future for the Notes app and cc @nextcloud/designers for a great UX :slightly_smiling_face:  :rocket:

juliushaertl:
Work in progress at https://github.com/nextcloud/text/pull/901

PR #901, opened 2020-06-22
Add input rule for checkboxes

juliushaertl:
- Allows to convert the current node checkboxes by typing "- [ ]"
- Fixes some positioning issues compared to regular lists

* Resolves: #819
* Target version: master 

Peek 2020-12-22 15-44

Before:
image

After:
image


juliushaertl:
This currently has the side effect that regular lists will only be converted after typing the first character after `- `, I'm not sure if there is an easy way around that yet.

jancborchardt (approved):
Great stuff, looks good design-wise! :)

stefan-niedermann (commented):
Nice! (Sorry, didn't check the code, but only the screen record)

- Does it also work if there is a space character between the brackets like `- [ ]`?
- Wouldn't it make more sense to replace the graphical checkbox when pressing <kbd>backspack</kbd> with the text representation `- [ ]` instead of removing it completely? At least for me this is the expexted behavior, but i don't insist to keep it, it's just well-known to me.

juliushaertl:
> * Does it also work if there is a space character between the brackets like `- [ ]`?

Yes, that is the case.

> Wouldn't it make more sense to replace the graphical checkbox when pressing backspack with the text representation - [ ] instead of removing it completely? At least for me this is the expexted behavior, but i don't insist to keep it, it's just well-known to me.

Moved to https://github.com/nextcloud/text/issues/1307

Issue #920, opened 2020-07-08
Inserting images does not work when used in a call


azul:
**Describe the bug**
When i am editing a markdown file during a call i cannot insert an image.
The filepicker opens fine.
I pick the file.
Nothing happens.

**Insights from debugging**
Console says:
```
Uncaught (in promise) TypeError: Cannot read property 'fileList' of undefined
    at files.js:67
    at s (runtime.js:45)
    at Generator._invoke (runtime.js:274)
    at Generator.forEach.A.<computed> [as next] (runtime.js:97)
    at m (editor-rich.js?v=4c36b80e8788de5f1c33:23)
    at a (editor-rich.js?v=4c36b80e8788de5f1c33:45)
```

The [code in question](https://github.com/nextcloud/text/blob/master/src/helpers/files.js#L67) tries to use 
`OCA.Files.App.fileList` : 
```javascript
const files = OCA.Files.App.fileList.filesClient._client.parseMultiStatus(response.data)
```
This is not available when using text in the viewer inside a call.

**To Reproduce**
1. Start a call
2. Share a '.md' file
3. Edit it
4. Try to insert an image

**Expected behavior**
The image shows up in the text just like it does when editing in file browser

azul:
This works for me:
```javascript
const client = OC.Files.getClient();
const files = client._client.parseMultiStatus(response.data)
	return files.map(file => {
		const fileInfo = client._parseFileInfo(file)
		fileInfo.href = file.href
		return fileInfo
	})
```
Not sure if that's the right way to approach it though.

PR #921, opened 2020-07-08
fix: use OC.Files API to get file info

azul:
This api is also available in other apps while
the internal structures used in fetchFileInfo are not.

* Resolves: #920 
* Target version: master 

### Summary

The [old code in question](https://github.com/nextcloud/text/blob/stable19/src/helpers/files.js#L67) tried to use 
`OCA.Files.App.fileList` : 
```javascript
const files = OCA.Files.App.fileList.filesClient._client.parseMultiStatus(response.data)
```
This is not available when using text in the viewer inside another app such as spreed during a call.


azul:
@juliushaertl this should be a quick review. :eyes: 

Could you please take a look so I can then adjust #879 to it?

MorrisJobke:
🙄

PR #924, opened 2020-07-09
Use IBootstrap for the app bootstrap

MorrisJobke:
I tested it and it seemed to work, but please have another look.

Also @ChristophWurst must be proud of this :)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -24,85 +24,33 @@ declare(strict_types=1);
 
 namespace OCA\Text\AppInfo;
 
-use OCA\Text\DirectEditing\TextDirectEditor;
+use OCA\Files\Event\LoadAdditionalScriptsEvent;
+use OCA\Text\Listeners\FilesLoadAdditionalScriptsListener;
+use OCA\Text\Listeners\FilesSharingLoadAdditionalScriptsListener;
+use OCA\Text\Listeners\LoadViewerListener;
+use OCA\Text\Listeners\RegisterDirectEditorEventListener;
 use OCA\Viewer\Event\LoadViewer;
 use OCP\AppFramework\App;
+use OCP\AppFramework\Bootstrap\IBootContext;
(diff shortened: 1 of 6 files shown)


lib/AppInfo/Application.php
```diff
@@ -25,84 +25,56 @@
...
-			\OCP\Util::addStyle('text', 'icons');
-		});
+		$initialStateService = $server->query(IInitialStateService::class);
+		$userSession = $server->getUserSession();
+		$config = $server->getConfig();
 
-		if ($this->userSession->isLoggedIn()) {
-			$eventDispatcher->addListener('OCA\Files::loadAdditionalScripts', function () {
+		if ($userSession->isLoggedIn()) {
+			$eventDispatcher->addListener('OCA\Files::loadAdditionalScripts', function () use ($config, $userSession, $initialStateService) {
```
MorrisJobke, line 62: For this one it would be super awesome to have a dedicated "if user is logged in" event to listen for, then we could move this as well.

MorrisJobke: cc @rullzer to avoid the user session creation.

rullzer: If only somebody had the same idea... https://github.com/nextcloud/server/blob/master/apps/files/lib/Event/LoadAdditionalScriptsEvent.php

rullzer: Just use that
and reverse the logic?

rullzer: Always listen. and when it is emmitted check if the user is logged in

MorrisJobke: I just noticed that this wasn't really used how it should be. Let me refine it a little bit.

MorrisJobke:
Ready for review and it feels so much better than before :)

juliushaertl:
Hm cypress fails seem random, and not really related. Let's get this in :rocket:

PR #929, opened 2020-07-13
Enh/noid/sharing additional scripts event

juliushaertl:
Follow up to #924 

`OCA\Files_Sharing::loadAdditionalScripts` is still an old event, so this one will use the newly introduced event class from https://github.com/nextcloud/server/pull/21815

Other than that the second commit will make sure that text app functionality is also available for logged in users on share links.


lib/Listeners/FilesSharingLoadAdditionalScriptsListener.php
```diff
@@ -47,9 +47,6 @@ public function __construct(IConfig $config, IInitialStateService $initialStateS
 	}
 
 	public function handle(Event $event): void {
-		if (!$this->userSession->isLoggedIn()) {
```
MorrisJobke, line 50: Then drop the user session also from the constructor.

PR #930, opened 2020-07-13
Fix event register method for files_sharing event

MorrisJobke:
Followup to #924 because it caused issues on public share pages.


MorrisJobke:
Let's close this in favor of #929

Issue #353, opened 2019-10-09
consolidate header and paragraph buttons?


brainchild0:
I tried the "paragraph" button many times, attempting to discover its function.  It never seemed to have any effect.  I almost gave up before realizing that paragraph means not header.  

Maybe the paragraph and header buttons should be consolidated.  Regular body text might just be one of the drop down options, in addition to the header levels.  

I never saw this kind of interface before.  I think it is probably confusing since there are so many rich text editors that have a different solution.

jancborchardt:
> Maybe the paragraph and header buttons should be consolidated. Regular body text might just be one of the drop down options, in addition to the header levels.

Yeah, I’d say that’s a good point. But then we have to have an `.icon-caret` below the icon to indicate it’s a menu. @juliushaertl?

brainchild0:
Whatever you have in the header button seems   clear, at least on the question of a drop-down menu.  You can probably just move the paragraph icon into the header menu.   Even the existing icon for the header menu is probably fine. The paragraph icon is not intuitive shown in isolation, but the  juxtaposition  to the header options  probably expresses the concept with adequate clarity.

jancborchardt:
The issue is that the default action icon of the menu also functions as the icon of what is currently active.

And since everything is basically normal paragraph by default, it will show the paragraph icon and you might never click it. This is why it needs a caret.

brainchild0:
I see.  I definitely support an indication of the drop-down menu whenever there is one.

Regarding the issue of icon, perhaps the user would be best guided by continued use of the header icon.  The menu has three functions: to make a regular paragraph into a header, to make a header into a paragraph, and to change a header level. All three cases involve a header, so the header icon carries the needed information.

brainchild0:
Another option is to use buttons (and keyboard shortcuts) that promote or demote header level (with a body paragraph having the position of level-zero header), rather than setting the level to a chosen value.  Most invocations of commands related to header level will be to promote or demote a header level one position, or to set a paragraph to a header immediately after it is typed.

azul:
I landed here because i was searching the repo to learn if the para button had any function. I think the current icon is really confusing. For me it's a symbol of a line break / para break. Not the unformatted text itself.

I'd just drop the Paragraph button completely. You can turn a heading into a para by clicking the name of the heading again. Just like you can turn bold into normal text by clicking bold again.

If you just turned something into a heading accidentally undo will also work.

juliushaertl:
Sounds good to me :+1:

PR #968, opened 2020-08-12
cleanup: remove confusing paragraph button

azul:
The paragraph button was confusing for a number of reasons:
 * it only acts on headings
 * the symbol indicates a line break
   rather than the removal of formatting
 * there are other ways to turn headings back to normal text
 * all other formatting is removed
   by clicking the same button again

* Resolves: #353 
* Target version: master 

### Summary

See #353 for the discussion leading to this pull request.


jancborchardt:
Sounds good! :)

As said in the chat too, this needs the heading action menu (the dropdown trigger) to be highlighted too when you selected a heading so you see what to open to deselect the formatting again.

asaage:
> there are other ways to turn headings back to normal text

@azul how would i do that now?
I've copy&pasted text from an arbitary Website and headings are carried over as such.
I really can't find a way to turn them into normal text.

susnux:
@asaage 
 > I really can't find a way to turn them into normal text.

Just select the heading and click on the heading button again (e.g. `click: heading 1` will make the text a heading of level 1, clicking the button again makes it normal text again):
Create heading and revert to normal text again

jancborchardt:
@asaage @susnux in a design review we also found that it’s confusing how to go back to normal formatting. We will add back a "Body text" entry in that menu, see https://github.com/nextcloud/text/issues/2836

> Add a separate item in the heading menu "Body text" as the first item (but still only show dot indicator when any heading level is selected)

Issue #802, opened 2020-04-30
'Show rich workspaces' breaks CTRL+F 


bertrand-benoit:
**Describe the bug**
CTRL+F does not toggle Nextcloud textarea search anymore, when 'Show rich workspaces' is switched ON (which is the default).

**To Reproduce**
1. Go to Nextcloud Files
2. Click on CTRL+F
4. Browser is handling the keystroke instead of Nextcloud textarea search 

**Expected behavior**
Nextcloud textarea search handling CTRL+F keystroke like before.

**Client details:**
 - OS: Fedora 31
 - Browser: Firefox
 - Version: 75.0 (64-bit)
 - Device: desktop

<details>
<summary>Server details</summary>

**Text app version:** (see Nextcloud apps page)
2.0.0

**Operating system:**
Gandi // Linux myWebSites 4.9.182-paas-110c975

**Web server:**
Gandi WebSites / Percona Server

**Database:**
MySQL / 5.6.41-84.1-log - Percona

**PHP version:**
PHP Version 7.3.13

**Nextcloud version:** (see Nextcloud admin page)
18.0.4
</details>

PR #988, opened 2020-08-25
Only overwrite Ctrl-f when text is focussed

juliushaertl:
* Resolves: #802 
* Target version: master 

### Summary

Otherwise the unified search does not trigger properly if a rich workspace is shown. With this PR the browser default search will be only triggered by text if the keydown event is triggered on the editor input.

Using the regular fallback with hitting Ctrl-F twice would not work here since when editing in the viewer the unified search is hidden in the background and therefore the user doesn't get any feedback when hitting ctrl-f for the first time. Also when having the text focused I think that searching in it is a sane default.


azul:
Code looks good to me.

:monocle_face: Minor nitpick: `Keymap` class sounds very generic and behaves generic in that all of it's behaviour is specified through the options during construction. At the same time it now implements special treatment for `ctrl-f`. Seems counter intuitive that by adding a keymap to the editor that handles `mod-s` you also get special treatment for `ctrl-f`.

Maybe put the plugin into a separate class like `BrowserSearch` or so?

azul (approved):
Confirmed that it fixes the issue in files.
Keeps using browser search when opening a file in the viewer and looking at the sidebar. I think that is the behavior we want.

I have not played with this in talk.

PR #1014, opened 2020-09-05
Harden check when using token from memcache

rullzer:
Explicitly comparing never hurt anybody.

### DIFF
--- a/lib/Service/SessionService.php
+++ b/lib/Service/SessionService.php
@@ -152,7 +152,14 @@ class SessionService {
 
 		$data = $this->cache->get($token);
 		if ($data !== null) {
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

juliushaertl:
Session cache is only available starting with Nextcloud 19 https://github.com/nextcloud/text/pull/761

juliushaertl:
@rullzer It might make sense to also backport that one, any opinions?

MorrisJobke:
> @rullzer It might make sense to also backport that one, any opinions?

🙄

juliushaertl:
Not urgently needed, so I'd say we skip this for now.

PR #1015, opened 2020-09-06
Use to correct cache to obtain version info

rullzer:
* The documentservice writes to the text prefix. Not the textSession
  prefix. Hence we never found any of the cache entries. Thus always
  doing all the work.

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -46,14 +46,13 @@ use OCP\Lock\LockedException;
 
 class ApiService {
 
-	protected $cache;
+	protected $request;
 	protected $sessionService;
 	protected $documentService;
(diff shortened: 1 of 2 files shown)


rullzer:
ok. so this actually does boom...

rullzer:
@juliushaertl I guess we either need to improve the caching logic or rip it out in this case. Because right now it does nothing ;)

juliushaertl:
I moved the cache handling entirely to the DocumentService now as there is no need to have that in two places. It was basically broken because we returned to early and missed the document and session information.

juliushaertl (approved):

juliushaertl:
Doesn't save much indeed, but also doesn't hurt to have a cache for that case.

Issue #1005, opened 2020-09-04
Store last_contact in memcache


rullzer:
I noticed on some systems people keep text files open all the time. Hence there is a lot of traffic to the session table. Because we keep updating the last contact timestamp.

If we have a memcache available we should utilize this to store the last contact time. That way we could reduce the write load of the system quite a bit.

@juliushaertl as discussed.

juliushaertl:
Done with https://github.com/nextcloud/text/pull/1026/files

PR #1026, opened 2020-09-07
Refresh lastContact only once every 30 seconds

rullzer:
Improvements for #1005
Just storing this less often should be a bit easier on the database.


juliushaertl:
Pushed a fix for the idle/disconnect times in the frontend. It will now take longer for the view to update but i think that is still quite acceptable considering the performance impact.

juliushaertl (approved):

PR #1032, opened 2020-09-09
Yet another IE11 fixing round

juliushaertl:
Fixes various issues with IE11:

- Since some time IE11 was broken due to dependencies not being compatible anymore. Therefore I decided to for now just let all dependencies pass babel as I'm way to lazy to investigate which depedencies are culprit for an almost dead browser here and the build time is still ok
- Multiple styling issues because IE doesn't support sticky positioning and does weird things in general


skjnldsv:
Eager to reach nc22? :D

skjnldsv (approved):
No runtime-generator?
Not needed without the babel/polyfill ?

Didn't try but code looks good!

ChristophWurst (approved):

juliushaertl:
> No runtime-generator

I couldn't see any issue with that, so let's get this in for now.

PR #1035, opened 2020-09-11
IE11 madness

juliushaertl:
- Backport of #1032 
- Additionally downgrade core-js since for some reason this version causes issues with String.split with the regex that is used by `prosemirror-model` where `"block+".split(/\s*(?=\b|\W|$)/)` would return `["b", "l", "o", "c", "k", "+"]` instead of `["block+"]` 
 However the exact same version combinations work fine on master, so I assume there are some wild polyfills overriding each other

Related issue for core-js https://github.com/zloirock/core-js/issues/810


skjnldsv (approved):

juliushaertl:
Tests needs adjustments as well:

```
Error: Cannot find module '@babel/polyfill'
Require stack:
- /home/runner/work/text/text/src/tests/setup.js
```

danxuliu:
I can not add anything valuable, so I will just leave this here (and I can not believe that @skjnldsv has not done it yet):

this-is-madness-madness-this-is-ie11

rullzer:
sooo 18.0.10? or...?

juliushaertl:
Yes, please.

Rebased and updated bundles

juliushaertl:
Cypress failure currently expected

rullzer (approved):
:elephant:

Issue #600, opened 2020-01-21
Issue with Link Tooltop in Firefox


nickian:
NextCloud version: 17.0.2
Text version: 1.1.1
Firefox: 72.0.1

I can't add links to text in Firefox. This is what it looks like. When I try to click the tooltip it just goes away.

<img width="631" alt="Screen Shot 2020-01-21 at 1 11 10 PM" src="https://user-images.githubusercontent.com/590678/72839538-b251a700-3c4f-11ea-9718-783f1a862180.png">

As a side note, it would also be nice to have a source editing option in the browser, not just wysiwyg.

ahstro:
Is this still happening and/or can someone else reproduce it? I tried in both Firefox 73.0.1 and Firefox Developer Edition 74.0b4, but it worked fine, and being on Arch, I can't easily install a previous version.

nickian:
Yes, still an issue for me on Windows 10 and MacOS, Firefox 73.0.1. The input comes up, but doesn't look right. It's cut off on the left side and just doesn't behave the same as Chrome.

![34234234](https://user-images.githubusercontent.com/590678/75222001-65865200-5760-11ea-8661-55f103daf2c1.jpg)

ahstro:
Hmm, that looks like a different issue to me, specifically the menu bubble not showing up correctly on too narrow editor windows. I'll look into that right away 😊

But in which way is it behaving differently? I'll try to help with that as much as possible, but unfortunately(?) I'm getting the same behavior from all my browsers 😅

ahstro:
The alignment issue seems to be an upstream one: https://github.com/scrumpy/tiptap/pull/239
It is allegedly fixed, but since that doesn't seem to be the case here, I asked the creator of the original PR.

Still, if something other than the alignment is off, share! 😊

nickian:
On Mac OS, I am not able to get the textbox to appear:

![text](https://user-images.githubusercontent.com/590678/75294206-0a4a7300-57e5-11ea-9a33-fb3d81577b21.gif)

On Windows, it works, but if I'm trying to add a link that happens to be on the left side of the document, the popup ends up going too far left and getting cut off. Maybe some conditional logic that says if the selected text is x pixels away from the left edge, then increase left positioning by however much is needed to not cut off.

nickian:
Actually the left positioning issue seems to be universal. This is in Chrome on Mac OS, as an example:

[...]

PR #1045, opened 2020-09-16
Fix link menu positioning

juliushaertl:
### Summary

Fixes some styling issues with the menu bubble positioning and giving it a fixed height of 44px since it otherwise would change height when switching to the link input. 

I'm not entirely sure about the 44px height since it looks kind of huge compared to the rest, but for now I picked this since those are buttons and we have a minimum click target of 44px

#### Before
image

#### After
image

### Notes for backporting 

:warning: When backporting we need to specify an additional fixed width for this since tiptap doesn't handle resizing the menu bubble properly. With Nextcloud < 20 the single button is smaller than the input which caused that the input could be cut of due do missing repositioning. A width of 180px seems to be fine as a fix for older versions for this.


azul (approved):
Looks good to me. I had a look both at the code and tested it (rebased on current master) locally. The height change makes sense to me.

I was not able to reproduce the alignment issue discussed in #600 - with current master. 
Maybe this has already been fixed. The changes in here do not seem to target it anyway. Just the branch name made me read the bug description and see if it had any effect.

juliushaertl:
> I was not able to reproduce the alignment issue discussed in #600 - with current master.

Yeah #600 was somehow mitigated on master/stable20 since with the two buttons the initial width of the button container was higher, therefore the initial position calculations work better when switching to the input for links.

jancborchardt (approved):
Looks good! Would be even better with white background for the menu and the divider in color-border, but we can also do that separately. :)

jancborchardt:
@juliushaertl needs rebase, but then let’s get it in? :)

juliushaertl:
Rebased and also added a design fix as discussed in our call: 
image

MorrisJobke:
🙄

Issue #820, opened 2020-04-27
HomePage Loading Circle


nunofmds:
### Steps to reproduce
1. Login and open homepage

### Expected behaviour
Circle progress should disappear after work done

### Actual behaviour
Circle progress always spinning 

### Server configuration

**Operating system:** Docker Image

**Web server:** Docker Image

**Database:** Latest

**PHP version:** Latest

**Nextcloud version:** (see Nextcloud admin page)
Nextcloud 18.0.4

**Updated from an older Nextcloud/ownCloud or fresh install:**
Updated from Nextcloud 18.0.3

**Where did you install Nextcloud from:**
Docker Image

![2020-04-27 15_56_24-Files - Nextcloud](https://user-images.githubusercontent.com/8095169/80386750-b8cf6a80-889f-11ea-8bff-3153008d020b.png)

kesselb:
404 is the correct response if there is no README.md found. Yeah it's a bad choice (because the webserver will log it). 

It works fine over here. 404 response, spinner is gone.

nunofmds:
The spinner is still there...

nunofmds:
![2020-05-12 10_38_24-Window](https://user-images.githubusercontent.com/8095169/81668581-ca3a7a00-943c-11ea-866c-82457e9e9c4a.png)

kesselb:
The response for the failing xhr request is?

nunofmds:
404 !? Where is the property "data"?

kesselb:
![image](https://user-images.githubusercontent.com/3902676/81669837-1094d680-9447-11ea-9e7f-c0a46850127b.png)

Click on Response ...

nunofmds:
![image](https://user-images.githubusercontent.com/8095169/81671421-e3452a00-9440-11ea-8aa7-41de52a4f766.png)

kesselb:
You disabled the text app?

nunofmds:
![image](https://user-images.githubusercontent.com/8095169/81674538-4769ed00-9445-11ea-8e59-b51a32863b6c.png)

It is enabled and I can use it to edit md txt files...

kesselb:
Anything in `nextcloud.log`?

Actually logs, configuration and apps are already part of the issue template and I already added the "needs info" label earlier because those information are missing. You see there is a reason we ask that information.

nunofmds:
Just did a tail -f /data/nextcloud.log and nothing new appears even after a force CTRL+F5 on the page

[...]

PR #1053, opened 2020-09-21
Check if error is ocs response on workspace request

kesselb:
* Resolves:
* Target version: master 

### Summary

On setups like #820 (with error_page 404 /some_error.html) now the spinner is gone. Creating a new file is not possible because the folder data is missing but that's fine (at least for me).


juliushaertl (approved):
Thanks :+1:

PR #1057, opened 2020-09-25
Fix images in public links

juliushaertl:
Otherwise manually embedded images for public share previews will never be shown

As reported by @jospoortvliet


juliushaertl:
Cypress failure is currently expected.

azul:
Code looks good to me. I have not been able to test this properly though. Here's what i tried:

I created a folder `public` and moved one of the images from `Media` into it.
Then i inserted the image into `Readme.md` via the workspace menu bar.

The image shows up fine for me but not if i share the folder publicly and open that link in a private window.

I suspect that's another issue that needs fixing after this.

@juliushaertl How do i test the issue this is solving?

azul (approved):
Code looks good.
`hasPreviewUrl` now has one very long line that i would break in two - but that's a matter of taste i guess.

I tested this with #1058 since i did not know hwo to create an image with a public src otherwise.
The public link works will in #1058 for me. So i think this is good to go.

PR #1058, opened 2020-09-25
Use public link previews for relative paths

juliushaertl:
Follow up to #1057 to make sure inserted images are accessible if they are in the same or sub directory of a share link

- [ ] Wait for #1057 to be merged
- [ ] Add tests

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -106,12 +106,24 @@ export default {
 			if (this.hasPreviewUrl) {
 				return this.src
 			}
-			if (this.fileId) {
-				return generateUrl('/core/preview') + `?fileId=${this.fileId}&x=1024&y=1024&a=true`
+
+			const isPublic = document.getElementById('isPublic')?.value === '1'
+			const sharingToken = document.getElementById('sharingToken')?.value
+			const previewPath = (!isPublic ? generateUrl('/core/preview.png') : generateUrl('/apps/files_sharing/publicpreview/' + sharingToken))
+
+			if (this.fileId && !isPublic) {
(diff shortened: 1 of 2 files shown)


azul:
I tested this as follows:
* I created a folder public and moved one of the images from Media into it.
* I inserted the image into Readme.md via the workspace menu bar.
* The image shows up fine on teh public share
* However the image fails to show for the logged in user.

The relevant part of the html source for the logged in user looks like this
```html
<div data-v-336b1e4e="" data-src="photo-1495962637988-4be9db2af01f.jpeg?fileId=10#mimetype=image%2Fjpeg&amp;hasPreview=true" class="image" contenteditable="false" draggable="true">
  <div data-v-336b1e4e="" class="image__placeholder">
    <div data-v-336b1e4e="" class="image__main" style="">
      <a data-v-336b1e4e="" href="/index.php/f/10" target="_blank">
        <div data-v-336b1e4e="" class="icon-image" style="background-image: url(&quot;/index.php/apps/theming/img/core/filetypes/image.svg?v=0&quot;);"></div> 
        <!---->
      </a>
    </div>
    <div data-v-336b1e4e="" class="image__caption" style="">
      <input data-v-336b1e4e="" type="text">
    </div>
  </div>
</div>
```

azul (changes requested):
With the one adjustment mentioned above this works for me.

the `imageUrl` function is getting a bit complex and repetetive towards the end.
I suspect replacing the temp `previewPath` and `f` [with a query](https://refactoring.guru/replace-temp-with-query) / computed property would help here.

juliushaertl:
Obsolete through #1900

PR #1060, opened 2020-09-25
Fix alignment and design of checkboxes

jancborchardt:
As discussed @juliushaertl – also found some more stuff than just the alignment of the checkbox. :) 

Before | After
-|-
checkboxes before|Checkboxes after


jospoortvliet:
Very nice Jan! I can't test but the css looks sane.

juliushaertl:
Looks much nicer, one small suggestion, how about also adjusting the regular lists, so that the bullet points are centered to the checkboxes?

image

rullzer:
master is 21.

jancborchardt:
> Looks much nicer, one small suggestion, how about also adjusting the regular lists, so that the bullet points are centered to the checkboxes?
> 
> image

I would agree, however this is not possible without hacking one way or the other:
https://stackoverflow.com/questions/4373046/css-control-space-between-bullet-and-li

And this is something which – if at all – we should do separately. :) In the usual cases, bulleted lists and task lists might not immediately follow each other, so it’s not so bad for now.

juliushaertl (approved):

eneiluj (approved):
Super nice

PR #1107, opened 2020-11-06
[stable18] Update axios to 1.3.2

PVince81:
Fixes an issue where file names called "JavaScript.md" would cause errors in the console.

The PR is messy because npm wanted to update a lot of other things, I did not manage to prevent it...
For a stable release I wouldn't want to update that much just for that simple fix...

But if we're ok with all the other dep updates, even @nextcloud/auth and @nextcloud/vue on stable18, then we can move forward and revert the package.json sticky versions I put in.

Also I'm surprised that `make build-js-production` modified so many JS files. Maybe it's related to all the other dep updates...

Thoughts ? @juliushaertl @ChristophWurst


PVince81:
I tried pinning the versions in package.json to whatever they were set to in package-lock.json, but still npm install allows itself to update sub-deps. From what I read here https://medium.com/the-guild/how-should-you-pin-your-npm-dependencies-and-why-2b8d545c7312 it seems it's not possible to pin sub-deps.

Let's try the depandabot route and see how it does it...

PVince81:
dependabot work started here: https://github.com/nextcloud/text/pull/1110#issuecomment-723861115

Issue #1027, opened 2020-09-07
Close idle documents


rullzer:
I noticed a lot of times that documents stay open in peoples tabs. And they just keep pinging the server until that one faithful day where their browser crashes and they need to restart.

I would suggest we close documents after 30 minutes of inactivity. With this I mean. No activity from the user themselves but also no activity from the server. So if you are in a meeting and have the document open you do not have to type something every 10 minutes. Just as longs as somebody is typing.

After these 30 minutes of inactivity we close the document gracefully.
Then we show a modal with that the document was closed due to inactivity. Then the user can chose to close or reopen the document.

This should avoid a lot of useless text polling on the servers.

PR #1162, opened 2020-11-09
Disconnect session after 30 minutes of idle timeout

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/1027

As discussed with @rullzer this is mainly to avoid people leaving browser tabs open causing a high load on the instance.

Can easily be tested with the following patch:

```diff
diff --git a/src/services/PollingBackend.js b/src/services/PollingBackend.js
index 78831d94..aef6f964 100644
--- a/src/services/PollingBackend.js
+++ b/src/services/PollingBackend.js
@@ -40,7 +40,7 @@ const FETCH_INTERVAL_MAX = 5000
  * Interval to check for changes when there is only one user joined
  * @type {number}
  */
-const FETCH_INTERVAL_SINGLE_EDITOR = 5000
+const FETCH_INTERVAL_SINGLE_EDITOR = 2000
 
 const MIN_PUSH_RETRY = 500
 const MAX_PUSH_RETRY = 10000
diff --git a/src/services/SyncService.js b/src/services/SyncService.js
index 8ee1a031..aa13d4b2 100644
--- a/src/services/SyncService.js
+++ b/src/services/SyncService.js
@@ -36,7 +36,7 @@ const defaultOptions = {
  * The session will be terminated and the document will stay open in read-only mode with a button to reconnect if needed
  * @type {number}
  */
-const IDLE_TIMEOUT = 30
+const IDLE_TIMEOUT = 0.25
 
 const ERROR_TYPE = {
        /**
```

Furthermore the second commit makes sure that if the browser is considered hidden (becase another tab is visible for example) through the browser visiblity api we only fetch changes once per minute. We could even further reduce the time of course but we would run into issues with displaying active users in the document then quickly, as there is no ping back to the server otherwise.

@jancborchardt for the wording and UI:

image


skjnldsv (approved):

rullzer (approved):
cool stuff!
seems to do the trick!

PR #1165, opened 2020-11-12
Validate link on click

juliushaertl:
Some hardening to avoid opening links that could occur during editing sessions which are usually stripped out by markdown-it.


nickvergessen (approved):

rullzer (approved):
looks good

Issue #1171, opened 2020-03-13
Find another solution for editing the readme.md


dartcafe:
The current inplementation and display of the readme.md inside a folder has some confusing side effects. 

# Steps to reproduce

-  Don't have a Readme.md
-  click inside the "free space" above the file list
-  click anywhere outside the editor

# What happened
![grafik](https://user-images.githubusercontent.com/26707476/76640141-b6af8780-654f-11ea-8fa4-f7eb6ed97526.png)

# Expected behaviour
![grafik](https://user-images.githubusercontent.com/26707476/76640137-b0b9a680-654f-11ea-8e3f-e85f236f72e2.png)
- Clicking in a "free space" of the file list should not open an editor.
- No grown "white space"
- No magically appeared file in my folder. 

Some users got confused and complained about of the empty space and the fact, that an unknown and unwanted file appears in their folders.

Possible solution could be, to add a button for explicit adding/editing readme.md.

dartcafe:
My thoughts about this: 
* Do not use an editable area, this confuses many users.
* Add a CTA which says "Add folder description" and let the user edit a new readme.md
* If a readme.md exists, 
  * show it's content in the current way (without useless whitespace) 
  * and add a CTA which says "edit folder description"

* Optionally remove empty readme.md, but this could be also surpringly to users.
* Optionally let the user edit the readme in a modal.

The handling of readme.md is not very common for 'normal' users.

lars-sh:
I just created a Pull Request, which deletes the Readme.md file if its content should be set to an empty value and recreates a file if it does not exist, yet.

Warning: This Pull Request is not proven to work. It is a first try to help out on that problem. I knowing anything specific about Nextcloud development nor the Text app.

Does #1175 help to handle this problem?

azul:
I think people are clicking this by accident because:
* the clickable area is too big
* there is no change in the cursor - so it seems like an okay place to click to say dismiss a menu somewhere else.

My understanding is that any click within the blue area will create the readme:
![Bildschirmfoto von 2022-01-05 18-37-20](https://user-images.githubusercontent.com/102902/148263095-0ff29047-366e-4bee-8810-704169116944.png)

Only the area around the text with a height of 44 px should be clickable and it should also change the cursor to a pointer to indicate clicking it will have an effect.

There's a lot more that could be done here. Right now the first click will create the readme... but it takes another click to edit it. The second click has to be more precise than the first. Making this a single click operation would be more smooth - but is probably not trivial.

[...]

PR #1175, opened 2020-11-14
#1171 Delete readme file when saving empty content

lars-sh:
* Resolves: #1171
* Target version: master 

### Summary
Users complain, that empty Readme.md files are created, when accidentally clicking above the files list.

This Pull Request deletes the Readme.md file if its content should be set to an empty value and recreates a file if it does not exist, yet.

Warning: This Pull Request is *not* proven to work. It is a first try to help out on that problem. I knowing anything specific about Nextcloud development nor the Text app.

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -282,8 +282,12 @@ class DocumentService {
 	 */
 	public function autosave($file, $documentId, $version, $autoaveDocument, $force = false, $manualSave = false, $shareToken = null, $filePath = null): Document {
 		/** @var Document $document */
-		$document = $this->documentMapper->find($documentId);
-
+		try {
+			$document = $this->documentMapper->find($documentId);
+		} catch (DoesNotExistException $e) {
+			$document = $this->createDocument($file);
+		}
+		
 		if ($file === null) {
 			throw new NotFoundException();
 		}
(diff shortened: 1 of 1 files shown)


lib/Service/DocumentService.php
```diff
@@ -318,7 +322,11 @@ public function autosave($file, $documentId, $version, $autoaveDocument, $force
 		}
 		$this->cache->set('document-save-lock-' . $documentId, true, 10);
 		try {
-			$file->putContent($autoaveDocument);
+			if (empty($autoaveDocument)) {
+				$file->delete();
```
juliushaertl, line 326: This will not work, if the file gets deleted the file id will be invalid and therefore the editing session that users might be currently connected to will fail.

juliushaertl: I'm not sure if there is an easy way to solve this other than introducing a more decoupled session handling for rich workspaces that will only use the target file for storing the results, however we don't have a unique identifier then for concurrent editing sessions anymore (as this is what the file is used for). If a file gets deleted and recreated it will get a new file id.

lars-sh: Well, in this case (the user did not enter anything/removed content) we want the file to be deleted.
If it were deleted in the meantime however, that should be fine. So do we simply need to handle DoesNotExistExceptions here?

lars-sh:
OK, I just tried to dig a little deeper into this. To me it seems, the client requests creating a file using DAV and then separately requests creating a text file session.
I ran into multiple kind of problems while trying to handle this nicely and would probably need just too much more information about Nextcloud development and therefore simply close this PR.

PR #1176, opened 2020-11-16
Add migration to fix oracle issues with the database schema

juliushaertl:
- Fix columns that had a notnull constraint set but used a default integer value which fails on oracle
- Remove autoincrement from oc_text_documents primary key since it is a file id that is inserted here
  - For oracle we need to drop the old table and recreate it. This is fine since it was never working before so the table doesn't have any useful information.

### DIFF
--- a/.github/workflows/oci.yml
+++ b/.github/workflows/oci.yml
@@ -1,0 +1,74 @@
+name: PHPUnit
+
+on:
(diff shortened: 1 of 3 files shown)


rullzer (approved):
:elephant:

nickvergessen (approved):

PR #1202, opened 2020-11-25
Catch possible database exceptions when fetching document data

juliushaertl:
Fixes an issue where an already cleaned up document in the database caused an uncatched exception (most likely due to the session being still valid in the cache but the document got cleaned up already.

A case where this could happen is the third commit where when a document was reset due to conflict resultion the sessions were not cleaned up immediately.


skjnldsv (approved):

rullzer (approved):
Looks good but didn't test

Issue #91, opened 2019-06-22
Color annotations for author of text


juliushaertl:
It should be possible to see who edited which part of the document, similar to etherpad where the text has different background colors.

Limitation: Since we only store markdown in files, editor information will be lost after save once all users have left the document

ge-fa:
Any chance of getting this implemented? Is there a Bountysource or something similar available for this? This feature is pretty important IMHO, especially compared to Etherpad etc.

Thanks for your work on Nextcloud!

varac:
I second this, imo it's the most missing feature which hold us back migrating from etherpad.

juliushaertl:
While I appreciate the integration work, this is not related to the issue discussion, therefore marking as spam.

hansimausisteffiwurstbrot:
> Any chance of getting this implemented? Is there a Bountysource or something similar available for this? This feature is pretty important IMHO, especially compared to Etherpad etc

I second the importance, we'd like to get rid of etherpad.

as bountysource shouldn't be used any more but we'd like to support this feature: please contact us if financial incentives help :)
it's not much but we could offer 200 € I guess.

juliushaertl:
Implemented with #1296 to be available during collaborative editing sessions.

hansimausisteffiwurstbrot:
Hey,
great to hear there is an implementation!

Just to get it clear: Due to the usage of markdown files it's not possible to save the colors?

I think it would be really usefule to achieve an implementation as in Pads (like Etherpad).

Cheers!

juliushaertl:
> Just to get it clear: Due to the usage of markdown files it's not possible to save the colors?

Yes, we have some ideas how that could still work (#1296) but this is a next step and requires a lot more effort.

PR #1296, opened 2020-12-22
Temporary editor colors during editing sessions

juliushaertl:
First attempt towards implementing https://github.com/nextcloud/text/issues/91 

For now the author annotation is only available during editing sessions

image

- [x] Make sure that color contrast ratio for text works properly
- [x] Check if we can indicate the author colors better in the avatar list
- [x] Since we currently use the session id as an identifier author annotation might also be lost as soon as one leaves the document and the session is cleaned up
	- [x] should at least persist the session metadata as long as the document is open
	- ~otherwise we might be able to persist decorations as marks in the document which would be also useful for persisting them later on~

## Follow up
- [ ] decorations are always rendered into the dom which might have a performance impact on larger files -> check if there might be a way to only render decorations if actually needed

cc @jancborchardt for design feedback :)


jancborchardt:
Aaawesome! :) So if we use the same colors as for the avatars, the contrast should work, as @skjnldsv made sure that contrast is enough. :)

Even if it's only enough for WCAG A or AA level that is fine though, since it can just be turned off and full text contrast is there.

----

About the indication of the color in the avatar list, you are referring to the issue that when people have avatars, the placeholder color is not visible, right?

I'd say a fix here is to give the avatar a 2px border of the author color. That would work both in the list of avatars-only as well as in the menu including the names.

jancborchardt:
Alternative for the highlight color as discussed: Reduce opacity of the author color so we can keep the text black, would make it more readable. And then the highlight color looks more like a text marker pen. :)

juliushaertl:
Ready for review and testing

ChristophWurst (approved):
Code looks good as far as I can judge :+1:

PR #1308, opened 2020-12-29
Add PHP-CS

rullzer:


juliushaertl (approved):

.php_cs.dist
```diff
@@ -0,0 +1,19 @@
...
+
+use Nextcloud\CodingStandard\Config;
+
+$config = new Config();
+$config
+	->getFinder()
+	->ignoreVCSIgnored(true)
+	->notPath('build')
+	->notPath('l10n')
+	->notPath('lib/Vendor')
```
ChristophWurst, line 15: looks like someone stole from Mail

ChristophWurst (approved):

.php_cs.dist
rullzer: borrowed

PR #1352, opened 2021-01-12
Implement support for creating files from templates

juliushaertl:
> - [x] With the corresponding text branch I get 2 entries to create text documents?


lib/AppInfo/Application.php
```diff
@@ -53,5 +56,16 @@ public function register(IRegistrationContext $context): void {
 	}
 
 	public function boot(IBootContext $context): void {
+		$context->injectFn(function (ITemplateManager $templateManager, IL10N $l) {
+			$textFile = new TemplateFileCreator(Application::APP_NAME, $l->t('New plain text file'), '.txt');
```
juliushaertl, line 60: Should be merged to one entry for that.

ChristophWurst (approved):
Tested & works

Issue #1075, opened 2020-10-06
✨ Text app design review


jancborchardt:
Thanks everyone for attending! @nickvergessen @ma12-co @juliushaertl @PVince81 @rullzer @skjnldsv @jancborchardt 

Also cc @azul @mejo-

## Overall

* [x] Saving feedback: Text saying "Saved" or "Saving …"
  * [x] Offline state: "Offline, changes will be saved when online" https://github.com/nextcloud/text/pull/2027
  * [x] We shouldn’t expose the difference between "unsaved" and "unpushed"
* [x] When zooming with sidebar open, the Files app / background shows (on zooming in), or the Text app overlaps the sidebar (on zooming out)
* [x] Header white and title/action icons not fade out → Pull request in progress in the Viewer component

## Errors

Moved to https://github.com/nextcloud/text/issues/2240

## Formatting bar

* [ ] Slight whitespace between "sections", 22px (half clickable space):
  * [ ] Undo/redo
  * [ ] Bold/Italic/Strikethrough/Headings
  * [ ] the 3 list types
  * [ ] quote, code, image
* [x] Increase height of formatting bar to 50px, size of avatars to 44 (might fix the status bubble display issue too) https://github.com/nextcloud/text/pull/1960
* [x] A little transparency and blur for more open feel https://github.com/nextcloud/text/pull/1960
* [x] Undo and redo buttons should be shown at full opacity when there is something to undo/redo respectively
* [x] Better feedback on what is the currently active formatting, except just full opacity cause it’s not so easy to see (possibly little circle below, or border around)

## Floating marking bar

**To be clarified in https://github.com/nextcloud/text/issues/2211#issuecomment-1063094981**

* [ ] Only show on mouseup https://github.com/nextcloud/text/issues/1059
* [ ] When clicking a linked file, going back via browser history (or viewer) does not work
* [x] Icons don’t need tooltip
* [x] action backgrounds white, divider grey
* [x] "Remove link" action in the bar when a link is set https://github.com/nextcloud/text/pull/1962

## Content

* [ ] Think about what to do with "Insert an image" cause it can be used to insert links to files too → possibly just wording
  * [x] If a file is inserted just add a regular link https://github.com/nextcloud/text/pull/1966
* [x] Numbers of numbered lists don’t show
* [x] If you mark a whole list and switch it to a task list, only the first item gets tasklist-ified https://github.com/nextcloud/text/pull/1973
[...]

PR #1353, opened 2021-01-14
Simplify save indicator

juliushaertl:
Some fixes from the design review in https://github.com/nextcloud/text/issues/1075

Based on #1296 

- Simplify save indicator to Saving/Saved
- Add some space to the menu bar so that button circles do no longer look cut off
- ~Add disabled styling for undo/redo buttons~ Needed to revert this since it had a major performance impact and requires more investigation
- Add improved indicator for active styling
- Show full menubar on public share links (where the talk sidebar overlapped before)
- Simplify wording
- Add spacing to color annotations


juliushaertl:
>  cypress: default-group — 1 test failed 

Failure from a previous run, but cypress doesn't update the state if the github action is just retriggered.

rullzer:
For me it shows all the way to the left. Is that intented?

rullzer (approved):
Works as adverstied on regular md files.

juliushaertl:
Fix for that is in https://github.com/nextcloud/text/pull/1409

Issue #787, opened 2020-04-24
Text tries to open EPub files while it cannot display them correctly


e-alfred:
**Describe the bug**
Text tries to open Epub files while it cannot display them correctly. The user only gets garbled text shown.

**To Reproduce**
1. Open an Epub file in Nextcloud

**Expected behavior**
Stop opening Epub files or add support for them. There is a viewer for Epub files out there already: https://github.com/e-alfred/epubreader Onlyoffice supports viewing them too.

**Screenshots**

![image](https://user-images.githubusercontent.com/10073629/80207923-6a1b9980-862f-11ea-946e-a8cbf01e2c6c.png)

PR #1384, opened 2021-01-24
Remove EPUB mimetype

leifurhauks:
* Resolves: #787
* Target version: master 

### Summary
An EPUB is a zip archive, so opening it for editing as plaintext is unexpected behavior and risks corrupting the file.

### DIFF
--- a/src/helpers/mime.js
+++ b/src/helpers/mime.js
@@ -29,7 +29,6 @@ const openMimetypesPlainText = [
 	'application/cmd',
 	'application/x-empty',
 	'application/x-msdos-program',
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

leifurhauks:
Hi @juliushaertl , I think the bundles weren't rebuilt in the most recent 20.x releases, so this bug is still occurring in 20.0.7 and 20.0.8. Another user recently reported the same behavior in nextcloud/server#25921. 

When I build the app myself, clicking an epub file initiates a download as expected.

PR #1385, opened 2021-01-26
Track authors by step rather than transaction

juliushaertl:
Since applying transactions has quite a big performance impact the initial approach of tracking the author in a blame map based on a per transaction level didn't work out well in terms of memory usage and computation time when loading a document with a somehow bigger amount of steps, since before this pr each step would be dispatched as its own transaction. 

Now with this we pass through the author information for all steps in one transaction, apply it in one go and build the blame map of authors based on the individual steps.

Should fix the performance and memory usage regression that was reported by @jancborchardt and @jospoortvliet


juliushaertl:
Cypress test that failed at the first attempt is known to be somehow unstable

ChristophWurst (approved):

PR #1386, opened 2021-01-26
Only increase refetch timer if we have all initial steps loaded

juliushaertl:
This will speed up the initial loading process of a document to no longer wait for one sync interval to show the document but rather fetch all initial steps immediately.

Found when looking into #1385


rullzer (approved):
hehe nice once.

PR #1395, opened 2021-01-26
Bump patch releases of dependencies in one run

juliushaertl:

* Resolves: # 
* Target version: master


juliushaertl:
Seems npm update did weird things, lets do the individual steps.

PR #1449, opened 2021-02-10
Improve design of idle/reconnect message

jancborchardt:
Was a bit too busy before:
- Removed the icon, unless it’s an error message
- Removed the border below
- Set the text to color-text-maxcontrast

Before | After
-|-
Nextcloud Text document idle warning looks bad|Text reconnect message new


jancborchardt:
@juliushaertl also wondering if we need that big present "Reconnect" button, when just clicking in the document would work.

mejo- (commented):
Same here. Overall, I like the changes a lot. I wonder whether the brighter text color will make the notice less visible, but I'd have to see it appearing in live to comment on that.

azul (commented):
Looks good to me. I'd also make the button less important. Removing the `primary` sounds like a good option.

scaredmushroom:
> @juliushaertl also wondering if we need that big present "Reconnect" button, when just clicking in the document would work. At least it could be non-primary, if it’s just that notice instead of an error.

I like the idea to just click into the document a lot.

But I think a less visible idle timeout is a problem. Multiple users reported to me, that they already overlook the idle notification often and get confused why they can't edit the document. (especially because the check boxes can still be marked but will not be synced, but this needs it's own issue)
From my perspective it would be nice to have a greyed out window with a banner notification which I just have to click to get back to editing.

Thank you all so much for developing such a nice add-on. It's an great enrichment for us.

jancborchardt:
> Multiple users reported to me, that they already overlook the idle notification often and get confused why they can't edit the document. (especially because the check boxes can still be marked but will not be synced, but this needs it's own issue)

So then I’d say this visual improvement with the still primary "Reconnect" button is a good intermediate step.

But ideally yes, we should have the simple logic of `just click into the document to continue editing` as you say. And in that case I we also don’t need a big banner overlaying it talking about timeouts, cause if you click and it just works then that’s perfectly good.

But that would be a separate pull request. :) Could I get 2 approvals for this one then to just fix the design? ;D

juliushaertl (approved):

PR #1505, opened 2021-03-08
[stable20] Disable cypress recording for now

juliushaertl:
- Revert action icon selector in cypress tests since it is not used in stable20
- Disable sending records to cypress.io since we cannot have the token in our public org
- Use testing mode of viewer to avoid timeouts


skjnldsv:
Oh come on, cypress too?
What is the risk? Aside from pushing fake runs?

juliushaertl:
I mainly followed the cypress recommendation there as for our org with that much contributors we cannot make sure that the key is not exposed or abused:

https://docs.cypress.io/guides/dashboard/projects.html#Record-key
> Exposing a record key
> 
> Anyone that has access to both the projectId and the record key of a project can record runs to that organization’s project in the Dashboard.
> 
> You would not want people outside of your team running tests because:
> 
>     This could drive up the number of tests you’ve run. Since Cypress bills based on the number of recorded tests - this means they can use up all allotted tests and there would be consequences to that.
>     They can write any data they want to your Cypress Dashboard. They could edit the test suite so that recorded tests record different things than the project’s original intent. This could including visiting different websites and generating videos of visiting those websites, for example.
> 
> If a record key is exposed you should delete it and create a new record key. Deleted keys will be invalid; If a project is run with a deleted key it will not be able to record.
> 
> You can set your record key as an environment variable to help protect it. Learn more here.

Maybe @LukasReschke can comment if that is an acceptable risk for test results.

rullzer (approved):
fine by me

juliushaertl:
> Is there anything sensitive on Cypress? Or has Cypress access to any other sensitive data?
> Sensitive being "anything not public".

No, just the test results which will be uploaded there by running the tests (from the sources in the public repo)

LukasReschke:
I'm okay with keeping it then.

skjnldsv:
I'llallowit

juliushaertl:
OK, thanks, then I'm readding it and will do another PR for the cypress fixes.

PR #1574, opened 2021-04-26
fix: rich workspaces overlap with new file dropdown

gary-kim:
Fix https://github.com/nextcloud/server/issues/27588
* Target version: master 

### Before
Screenshot 2021-04-26 at 14-03-29 Talk

### After

Screenshot 2021-04-26 at 14-00-00 Talk


gary-kim:
Rebased

gary-kim:
@juliushaertl Could you take a look when you have time? Thanks.

src/components/EditorWrapper.vue
```diff
@@ -638,7 +638,7 @@ export default {
 			#editor {
 				overflow: auto;
 				// Fix for IE11 issue where the menubar sometimes was positioned under the text
-				z-index: 1000;
+				z-index: 20;
```
juliushaertl, line 641: Mind to drop the comment above since with the removed IE11 support this is no longer required and caused side effects the bug the pr fixes ;)

juliushaertl:
Sorry for taking so long for the review, I have a small comment, but tested and works fine otherwise :+1:

szaimen:
I wonder what the difference to this PR is: https://github.com/nextcloud/text/pull/1631

juliushaertl:
The `position: unset;` that is currently set on `#editor-wrapper` basically moves the element into the regular stacking context and will be above any element due to its order in the dom. Since the `#editor` has 10000 (which is adjusted in this PR), the 10000 are in the regular stacking context. The other PR creates a new stacking context (which would have an implicit z-index of 0 as non is set on that element) so the 10000 no longer are taken into the global but the local stacking context of the parent only.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index and 
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context might be a good first read on the topics.

The second PR would be a good possible addition to avoid further overlay issues, but I need to do some further testing on that.

src/components/EditorWrapper.vue
gary-kim: Done

juliushaertl (approved):

szaimen:
Created the backport manually in https://github.com/nextcloud/text/pull/1769

szaimen:
Created the backport manually in https://github.com/nextcloud/text/pull/1770

Issue #1104, opened 2020-11-04
richWorkspace in PROPFIND


marinofaggiana:
I have found an issue using the PROPFIND with the Depth 0, this return the richWorkspace to nil (NULL) otherwise the  PROPFIND with Depth 1 for the same directory return correctly the richWorkspace string (when present).

PR #1590, opened 2021-05-14
Only return workspace property for top node in a propfind request

### LINKED ISSUE #1507: PROPFIND with Depth "0"
The PROPFIND with the Depth "0" on directory return richWorkspace always NULL 

Sorry if this is not the right place where write the this issue. (Was better the Server ?)

juliushaertl:
Fixes #1507 and #1104  while preserving the compatiblity with the previous fix for Android https://github.com/nextcloud/text/pull/994

This makes sure that no matter what depth is requested for the propfind we only return the rich workspace property for the current requested base directory.


ChristophWurst (approved):
:+1:

marinofaggiana:
@juliushaertl when will be available ?

PR #1645, opened 2021-06-09
ViewerComponent: pass on autofocus to EditorWrapper

azul:
The `ViewerComponent` is reused in the collectives app.
Allow disabling autofocus when using the viewer component.

* Resolves: # 
* Target version: master, 21, 20


juliushaertl (approved):

juliushaertl:
Hm, where has the node check gone :/

juliushaertl:
Ah it only runs on push which is then probably not happening on a fork.

@azul Could you build and commit the updated bundles as well?

azul:
@juliushaertl like so?

juliushaertl:
Yes, looks good :+1:

PR #1658, opened 2021-06-16
Try enabling apcu on cli for cypress

juliushaertl:
Let cypress pass again


artonge (approved):
Sweet ! I hope this works for the activity app as well !

PR #1659, opened 2021-06-16
Move to shared babel/webpack config

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>* Resolves: # 
* Target version: master


webpack.js
```diff
@@ -0,0 +1,30 @@
...
+const webpackConfig = require('@nextcloud/webpack-vue-config')
+
+webpackConfig.entry = {
+	text: path.join(__dirname, 'src', 'main.js'),
+	files: path.join(__dirname, 'src', 'files.js'),
+	public: path.join(__dirname, 'src', 'public.js'),
+	viewer: path.join(__dirname, 'src', 'viewer.js'),
+}
+
+webpackConfig.output.chunkFilename = '[id].js?v=[contenthash]'
```
skjnldsv, line 12: Any reason?

webpack.js
```diff
@@ -0,0 +1,30 @@
...
+webpackConfig.optimization.chunkIds = 'named'
+webpackConfig.optimization.splitChunks.cacheGroups = {
+	defaultVendors: {
+		test(module) {
+			return module.resource && module.resource.includes(`${path.sep}node_modules${path.sep}`) &&
+				!module.resource.includes(`${path.sep}highlight.js${path.sep}`)
+		},
+		name: 'vendors',
+	}
+}
```
skjnldsv, line 23: What does this do?

juliushaertl: it makes sure that vendor files get into a separate chunk, except the highlight js ones which should stay at one chunk per supported syntax highlighting language.

webpack.js
juliushaertl: Mainly to keep chunk names persistent with the `highlight/{lang}.js` ones, but I'm also fine to remove that.

artonge (approved):

webpack.js
skjnldsv: What benefit do we get from this?

skjnldsv: Vendors are nice to split as long as they don't get all used at the same time, right?
Like dynamic loading?
But if we always end up loading 90% of the chunks, we just create more bandwith and doesn't help cleaning the bundles imho :)

juliushaertl: The main issue i have is that I fail to tell webpack to only split off the highlight.js chunks and the ones that are dynamically imported (e.g. for only loading the files app init code and not the whole viewer component right away). Therefore I went for that hybrid approach that unfortunately leaves one extra chunk.

juliushaertl: I'm happy to look into that into a follow up again :wink:

skjnldsv: You can do that with an ignore plugin and do a dynamic import of the highlight lib :)
https://github.com/nextcloud/contacts/blob/017026a8e42717cc8e1935af5ce61aa46b3e9f79/webpack.js#L8

skjnldsv: Iirc, any direct runtime import will overrule compile time bundling

webpack.js
skjnldsv: The idea is to have all apps preceding their appName on files names, so it's easier to check what is what on network logging and such :wink: 
I'm always happier when we stick to the origal config files for less maintenance, but it's personal opinion, so I'm good with whatever you prefer :wink:

Issue #987, opened 2020-08-24
Make it easy to insert emojis


azul:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to add an emoji to a text. I need to find some emoji website, pick the one i like, copy and paste it.

**Describe the solution you'd like**
I'd like an easy way to pick an emoji to insert it into the text.
Currently I imagine another button in the menu bar that opens the emoji selector.
Picking an emoji from the selector would insert it into the text at the location of the cursor.

**Describe alternatives you've considered**
Opening the emoji selector when starting an emoji shortcut with `:` like github supports it.
Emoji button in the menubar seems more accessible and more consistent with what exist so far.
In contrast to github the editor currently has little support for typing magic like `@` mentions.

**Additional context**
The Nextcloud Vue components have a nice [emoji picker](https://nextcloud-vue-components.netlify.app/#/Components/EmojiPicker) already.

juliushaertl:
A shortcut like `:` would definitely be nice, but for the ease of keeping the implementation simple I'd stick with the proposed addition button in the toolbar for now. What do you think @nextcloud/designers is that to present?

jancborchardt:
Both sounds good – let’s start with the action menu addition of the menu icon, and as a second step we can look into the `:` shortcut. :)

mejo-:
Ok, so I couldn't resist and started playing around with adding shortcuts for inserting emojis via keyboard (using tiptap input rules). That's something I personally miss way more than the emoji picker :wink: 

The implementation is just a rough prototype for now and I'd like to get emoji picker (PR #1707) settled in text first, but still I wanted to share a screencast with you so that you can give early feedback on the design/concept:

![recording2](https://user-images.githubusercontent.com/3582805/124615443-922e9b80-de75-11eb-9f09-2d63a15ea210.gif)

PR #1707, opened 2021-07-02
Add emoji picker to menubar (Fixes: #987)

mejo-:
* Resolves: #987
* Target version: master 

### Summary

This is a first attempt to add the emoji picker to the text editor menubar. It works for me, but still has some rough edges.

Here's a screencast: recording

I'm curious to hear your thoughts @juliushaertl and @azul :blush:

The way menu actions are organized in `src/mixins/menubar.js` at the moment unfortunately is a bit limited. Since the emoji picker isn't a simple `<button>` or `<ActionButton>` element and I prefer to reuse the `<EmojiPicker>` component from nextcloud-vue, I went a different route: I added the emoji action in `src/mixins/menubar.js` but special-treated it in the `MenuBar` component. I'm not particularly happy with this approach though. Maybe you have a better idea for a clean implementation?


src/components/MenuBar.vue
```diff
@@ -127,6 +138,7 @@ export default {
 			forceRecompute: 0,
 			submenuVisibility: {},
 			lastImagePath: null,
+			showEmojiPicker: true,
```
juliushaertl, line 141: Seems unused

src/nodes/Emoji.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { Node } from 'tiptap'
+import { insertText } from 'tiptap-commands'
+
+export default class Emoji extends Node {
```
juliushaertl, line 26: I think this should also work with using an extension which would be a more cleaner way from my perspective to handle only the command, see https://github.com/ueberdosis/tiptap/blob/v1/packages/tiptap-extensions/src/extensions/History.js#L41-L48 for example. Unless there is a reason for keeping a dedicated node type of course that I might not see ;)

juliushaertl:
> I'm not particularly happy with this approach though. Maybe you have a better idea for a clean implementation?

Yes, that one is indeed a bit tricky. The reason I went for the array instead of having the buttons directly in the component is that this allows to do some straight forward subset for rendering the popovermenu if the space exceeds the width. I might need to think about that a bit more.

szaimen:
And I think a rebase is probably needed...

src/components/MenuBar.vue
mejo-: Good catch!

src/nodes/Emoji.js
mejo-: You're right, an extension is enough here - no need to make it a node. Changed now :blush:

mejo-:
> Yes, that one is indeed a bit tricky. The reason I went for the array instead of having the buttons directly in the component is that this allows to do some straight forward subset for rendering the popovermenu if the space exceeds the width. I might need to think about that a bit more.

After thinking about it a second time, I'm more confident with my solution of special-treating the EmojiPicker in the for-loop.

Beware though that the way I implemented it results in the EmojiPicker being always displayed as the last non-hidden tool. That's on purpose, given that adding an emoji would require three clicks otherwise (open popovermenu, open emojipicker, select emoji).

Let me know what you think about it :)

juliushaertl:
> > I'm not particularly happy with this approach though. Maybe you have a better idea for a clean implementation?
> 
> Yes, that one is indeed a bit tricky. The reason I went for the array instead of having the buttons directly in the component is that this allows to do some straight forward subset for rendering the popovermenu if the space exceeds the width. I might need to think about that a bit more.

We could probably have a dedicated component with a slot that takes all the actions and then does conditional rendering based on the slot elements but not fully sure if that would be successful.

Something similar how it is done in the actions component https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue#L649

juliushaertl:
Anyways I'd be fine with the current workaround for now. Doesn't seem too bad ;)

mejo-:
> We could probably have a dedicated component with a slot that takes all the actions and then does conditional rendering based on the slot elements but not fully sure if that would be successful.

To be honest, I'm not sure whether it's worth the effort at the moment :wink:. But you decide. If you'd like it to be that way, I'd give it a try.

> Anyways I'd be fine with the current workaround for now. Doesn't seem too bad ;)

Yay :blush: Feel free to approve/merge. I just changed the PR from draft to ready to merge.

mejo-:
Rebased on latest master.

juliushaertl (approved):

mejo-:
:heart_eyes:

juliushaertl:
Awesome contribution @mejo- :clap:

PR #1771, opened 2021-07-16
allow to insert text after trailing codeblock

### LINKED ISSUE #109: When a code block is the last thing in a document, then you can't insert text after it
**Describe the bug**
When a code block is the last thing in a document, then you can't insert text after it (at least I can't figure out how to :)).

**To Reproduce**
1. Go to the last line of the document
2. Insert a code block and enter some text in it
3. Attempt to insert a new line of text after the code block

**Expected behavior**
I would expect the same behavior as inserting images, where a newline is inserted after the image or the ability to double click after the last line to create a new.

**Client details:**
 - OS: Windows
 - Browser: Firefox
 - Nextcloud version: 16.0.1

szaimen:
Fix #109
Signed-off-by: szaimen <szaimen@e.mail.de>


szaimen:
@juliushaertl I am honest: I don't know if this is the way to solve the mentioned bug or not but I that's what I understood from the mentioned ticket, though I am not able to compile the code because the compiple bot doesn't work...
I need some helping hands here...

juliushaertl (approved):
Tested and works :rocket:

PR #1778, opened 2021-07-19
Extend mimetypes for direct editing

### LINKED ISSUE #1709: Extend mimetypes for direct editing
The web currently has a longer list of mimetypes https://github.com/nextcloud/text/blob/master/src/helpers/mime.js#L27 which would make sense to support for the direct editing endpoint as well in https://github.com/nextcloud/text/blob/master/lib/DirectEditing/TextDirectEditor.php#L83-L97

szaimen:
Fix #1709

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/lib/DirectEditing/TextDirectEditor.php
+++ b/lib/DirectEditing/TextDirectEditor.php
@@ -83,7 +83,28 @@ class TextDirectEditor implements IEditor {
 	public function getMimetypes(): array {
 		return [
 			'text/markdown',
-			'text/plain'
+			'text/plain',
+			'application/cmd',
+			'application/x-empty',
+			'application/x-msdos-program',
+			'application/javascript',
+			'application/json',
+			'application/x-perl',
+			'application/x-php',
+			'application/x-tex',
+			'application/xml',
+			'application/yaml',
+			'text/css',
+			'text/csv',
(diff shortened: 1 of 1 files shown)


lib/DirectEditing/TextDirectEditor.php
```diff
@@ -83,7 +83,28 @@ public function getName(): string {
...
+			'text/csv',
+			'text/html',
+			'text/org',
+			'text/x-c',
+			'text/x-c++src',
+			'text/x-h',
+			'text/x-java-source',
+			'text/x-ldif',
+			'text/x-python',
+			'text/x-shellscript',
```
juliushaertl, line 107: Not sure if we should rather move those to be the optional opener for direct editing. I assume a local code editor would be more suitable for mobile editing if present.

@tobiasKaminsky What is the behavior on mobile if there is no local editor, would it fall back to the optional ones then or are those really only used through the context menu options?

tobiasKaminsky: We do not differentiate between optional and normal ones, if there is no local app.
if there is a local app and file is downloaded, it is used.

juliushaertl (approved):

PR #1795, opened 2021-07-26
Avoid creating invalid URIs from user input

juliushaertl:
Make sure that the link we create is still valid in markdown, even if a user wrongly enters spaces into the input when adding a link through the UI.

To reproduce.
- Insert a link into a text file with "mailto: test@example.com" as the url
- Close the document
- Make sure to reset the text state for the document so that the document is refetched from the markdown source `occ text:reset FILEID -f`
- Reopen


azul (approved):
Looks good to me.

I checked that a mailto link with `%20` still works. Also confirmed that [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A) does not allow for spaces to occure anywhere in urls.

I tried to use other characters that usually do not appear in urls but these days are encoded and still displayed by the browser. In particular i tried: https://はじめよう.みんな . The browser opened the link just fine. So limiting the encoding to space charactes seems like a good approach as not everything needs to be encoded that is not a strictly valid url char.

juliushaertl:
Yep, markdown-it which we use to transform the markdown to the document seems to explicitly stop parsing links with ascii control characters including whitespace while allowing anything else: https://github.com/markdown-it/markdown-it/blob/e5986bb7cca20ac95dc81e4741c08949bf01bb77/lib/helpers/parse_link_destination.js#L50

PR #1798, opened 2021-07-28
Custom input rule to add first character after bullet

### LINKED ISSUE #1448: creating a bullet with * eats the first character that follows the space
On the current dev version (Nc 21 first RC as deployed on tech-preview and cloud.nc.c) creating a bullet seems a bit broken.

type:

```
* Bla bla
```

should result in

* Bla bla

But it results in:

* la bla

mejo-:
The `ListItem` input rule resulted in the first character after the
bullet being eaten. The custom input rule `listInputRule` fixes
this by inserting the character after creating a bullet list item.

Fixes #1448

* Resolves: #1448
* Target version: master


azul (approved):
Looks good to me.
I had a look at using a regexp with lookahead instead - but what we want is to insert the character inside the `li` the wrappingInputeRule creates. Lookahead could be used to only transform the `* ` part - but it would not put the character inside.

I also tried matching the char with the gegexp - but it turns out the match in the regexp does not get inserted in the wrappingInputRule either. So i think this is the best approach without changing tiptaps and therefore prosemirrors `wrappingInputRule`.

juliushaertl (approved):

PR #1800, opened 2021-07-30
fix: resolve relative image path based on currentDirectory option

### LINKED ISSUE #1655: View webp Pictures in Markdown Files
Hi,
it would be nice if webp images could also be displayed in the markdown files 
markdown webp
Best regardes

### LINKED ISSUE #1780: Images that have a relative url without fileId do not render ouside the files app
**Describe the bug**
When opening a document with relative image src without the fileId the images are not displayed because `FileList` is not available outside the files app:

```
TypeError: FileList.getCurrentDirectory is not a function
    imageUrl ImageView.vue:112
```

**To Reproduce**
1. Create a document with an image
2. Edit the document in a different markdown editor and remove the fileId
3. Open the document in talk

**Expected behavior**
The image should still be resolved and displayed.

**Screenshots**

Bildschirmfoto von 2021-07-20 15-30-42

**Client details:**
 - Browser: firefox and chromium
 - Version: stable22 branch, also current 21 release
 - Device: desktop

<

azul:
* Resolves: #1780 and #1655

Summary:

### Display relative Images without fileId in talk and collectives

Hand `currentDirctory` to `Image` plugin.
This approach is more reliable because it does not rely on the files app.
Therefor it also works in `talk` and `collectives`.
Fixes #1780.
We already know the current files path in the Editor.
Just need to pass it down like we do for the `MenuBubble`.

### Display animated gifs

Use the `currentDirectory` to determine dav paths for gif files
and display the animated gif via dav rather than the preview
that lacks animation.

### Support webp
Also display webp files with dav urls. Fixes #1655.


juliushaertl:
@azul Mind to do a rebase :)

juliushaertl (approved):

PR #1821, opened 2021-08-12
Fix header popover in file browser

CarlSchwan:
* Target version: master 

### Summary

Previously the popover for the header in the file browser was cut off. Now instead of making the whole text editor scrollable, only the content is.

image

Missing to make this patch perfect is changing the z-index from 61 to 62 in the server repo, which I will do shortly


src/views/RichWorkspace.vue
```diff
@@ -211,8 +211,12 @@ export default {
 	}
 
 	#rich-workspace::v-deep #editor {
+		overflow: visible !important;
+	}
+
+	#rich-workspace::v-deep .content-wrapper {
 		overflow: scroll !important;
-		max-height: 50vh;
+		max-height: 40vh;
```
jancborchardt, line 219: Reducing the rich workspace height more would box it in a bit much, no? Otherwise the change seems good.

CarlSchwan: I probably should use calc(50vh - 50px) so that it use a similar size has before

juliushaertl: Is that needed for the fix?

CarlSchwan: The new code doesn't apply the max-height on the editor but on the content instead so using calc(50vh - 50px) make it so that the height of the editor is the same as before

juliushaertl (approved):

PR #1835, opened 2021-08-25
[stable22] Dependency updates

juliushaertl:
~:warning: DO NOT MERGE until 22.1.1 is out~

Bumping patch releases to avoid a long rebase dance


azul (approved):
Looks good to me. Only updating patch level makes sense unless there are security issues and then we can do updates manually.

PR #1860, opened 2021-09-14
make collabora on top of text idle message

luka-nextcloud:

* Resolves: # 
* Target version: master


src/views/RichWorkspace.vue
```diff
@@ -175,6 +175,7 @@ export default {
 		text-align: left;
 		max-height: 0;
 		transition: max-height 0.5s cubic-bezier(0, 1, 0, 1);
+		z-index: 110;
```
juliushaertl, line 178: The cypress test failure seems expected due to this z-index being larger than the one of the file creation menu which would cause some overlapping like this:

image

In addition it might be good to add a `position: relative;` here as well to make sure the z-index always applies.

juliushaertl (changes requested):
The cypress failure actually catched a case that where it has some visual regression :wink:

juliushaertl:
@luka-nextcloud Could you check that again for the comment above?

Issue #46, opened 2019-06-12
Options for image like delete and alt text?


jancborchardt:
Currently it seems you can only delete images via Delete on the keyboard, right? And there’s no way to set alt text?

I would expect something like a 3-dot menu in the top-right corner to show up when the image is tapped, with:
- Add alt text
- Delete image

![how to delete an image with mouse, would expect a delete icon in the corner on tap](https://user-images.githubusercontent.com/925062/59391917-1df89b00-8d76-11e9-8431-1fb7f492d6f6.png)

PR #1872, opened 2021-09-30
add image removing on hover

luka-nextcloud:
* Resolves: https://github.com/nextcloud/text/issues/46
* Target version: master 

### Summary

Should be able to remove images inside text when hovering over them.

image


jancborchardt (approved):
Looks good design-wise. As mentioned, as soon as the button component comes, would be best to do it with that (to make sure it works on all sorts of colors of images):
https://github.com/nextcloud/nextcloud-vue/pull/1808 cc @marcoambrosini

szaimen:
How to do that on mobile?

src/nodes/ImageView.vue
```diff
@@ -1,34 +1,62 @@
...
-			<transition name="fade">
-				<img v-show="loaded"
-					:src="imageUrl"
-					class="image__main"
-					@load="onLoaded">
-			</transition>
+		<div v-if="imageLoaded && isSupportedImage"
+			class="image__view"
+			@mouseover="showTrash = true"
+			@mouseleave="showTrash = false">
```
eneiluj, line 28:
```suggestion
		<div v-if="imageLoaded && isSupportedImage"
			v-click-outside="() => showTrash = false"
			class="image__view"
			@click="showTrash = true"
			@mouseover="showTrash = true"
			@mouseleave="showTrash = false">
```
@szaimen This could make it mobile-compliant :grin:.
@luka-nextcloud To use the `v-click-outside` you need to:
``` js
import ClickOutside from 'vue-click-outside'
```
and add:
``` js
directives: {
	ClickOutside,
},
```
just before the `props` component attribute.

src/nodes/ImageView.vue
```diff
@@ -1,34 +1,62 @@
...
+						<path
+							d="M632.32 667.306667a33.979733 33.979733 0 0 1-24.2176-10.0864L512 560.452267l-96.1024 96.768a34.133333 34.133333 0 1 1-48.4352-48.093867l96.3584-97.041067-96.273067-96.273066a34.133333 34.133333 0 1 1 48.264534-48.264534L512 463.7184l96.187733-96.170667a34.133333 34.133333 0 1 1 48.264534 48.264534l-96.273067 96.273066 96.3584 97.041067A34.133333 34.133333 0 0 1 632.32 667.306667z" />
+					</svg>
+				</div>
```
eneiluj, line 52:
```suggestion
				<a
					v-if="showTrash"
					class="trash-icon"
					title="Delete this image"
					@click="deleteImage" />
```
It's better to put images in the `img` directory so they can be optimized and/or opened with a local image viewer. You could just put this `<svg>` in `img/trash-icon.svg`.

src/nodes/ImageView.vue
```diff
@@ -232,6 +261,15 @@ export default {
 		onLoaded() {
 			this.loaded = true
 		},
+		showTrashIcon() {
+			this.showTrash = true
+		},
```
eneiluj, line 266: This is never used and can be removed.
```suggestion
```

src/nodes/ImageView.vue
```diff
@@ -302,4 +341,13 @@ export default {
...
 	}
+
+	.trash-icon {
+		cursor: pointer;
+		position: absolute;
+		display: inherit;
+		width: 50px;
+		top: 12px;
+		right: 8px;
+	}
```
eneiluj, line 352:
```suggestion
	.trash-icon {
		position: absolute;
		display: inherit;
		top: 12px;
		right: 8px;
		width: 44px;
		height: 44px;
		padding: 0 !important;
		background: url('../../img/trash-icon.svg');
	}
```

eneiluj (changes requested):
Very nice!

Built files are included in this repo (in `js/`). Please include the changed compiled files in the PR.

marcoambrosini (approved):

src/nodes/ImageView.vue
```diff
@@ -35,6 +35,28 @@
 						type="text"
 						:value="alt"
 						@keyup.enter="updateAlt()">
+					<div
```
juliushaertl, line 38: Would be great if we could switch to use https://www.npmjs.com/package/vue-material-design-icons directly here, so you could just import the icon from there and then we can easily switch to the button component later.

juliushaertl:
@jancborchardt Should we just always show the icon for mobile and keyboard navigation?

jancborchardt:
@juliushaertl we could, but:
- @luka-nextcloud mentioned it is difficult to do the keyboard focus on that, and also if you use the keyboard, you can just backspace it away
- Style-wise it of course looks much better without the icon there all the time. I'd also say that at some point we should remove the file name below for files with preview.

juliushaertl:
OK, fine with me then.

eneiluj:
@jancborchardt @juliushaertl How about showing/hiding the icon on click/click-outside? As we would keep the hover events, click events would only be effective in a mobile context.

jancborchardt:
> Not sure if you mean the width of the text or the width of the page of the window.

Width of text, like on Medium etc. :)

azul:
I think the image size was fixed with #2002. So my understanding is that this is good to go.

azul:
I rebased everything on current master.
unit tests are failing due to jest complaining about vue syntax in the `vue-material-design-icons`.

I updated `vue-material-design-icons` and tried to install `@vue/vue2-jest` but without any success. I think it might not transform files inside the node modules folder.

azul:
I resolved all conversations that seemed to have been addressed.
There's one open where i am unsure if it was fully addressed.
The PR is still marked as changes requested
@eneiluj could you check if it is okay now and approve?

eneiluj (dismissed):

eneiluj (changes requested):
Can't find getPos in ImageView.

eneiluj:
@luka-nextcloud @azul I found a fix. What's the best way to push it in your opinion?
* interactive rebase to remove changes on compiled javascript, rebase on master, add my changes force push this branch
* or a simple merge with master

juliushaertl:
> interactive rebase to remove changes on compiled javascript, rebase on master, add my changes, force push this branch

:+1: for that ;)

eneiluj (approved):

PR #1883, opened 2021-10-04
Add automatic line break on txt files

luka-nextcloud:
* Resolves: # 
* Target version: master 

### Summary

image
image


juliushaertl (approved):
Looks good, could you add the signed off message to the commit ?

Issue #1170, opened 2020-11-12
Properly handle linked files with direct editing


juliushaertl:
Currently we have no special handling for linking files with direct editing, so this is not working as described in https://github.com/nextcloud/android/issues/7192.

I currently see two ways of implementing support for opening linked files:
- We send a post message to the mobile clients with the path/file id so they can directly navigate to it and open it with direct editing
- We open a browser with the private link of the file. This has less effort on the client side, but will need the user to be logged in in the browser of course.

In addition there is also no support for adding file links through direct editing, since the user is not authenticated via a session. To solve this we'd need to implement something similar to the Collabora insert image functionality where a postMessage is emitted by the text app. Then the mobile clients would show a file picker and pass the selected file to the webview again.

cc @tobiasKaminsky @marinofaggiana What do you think. Does that make sense?

tobiasKaminsky:
On web ui one is then simply redirected, right?
Isn't the same possible? As we are already within a webview?

Regarding adding file link:
Post message sounds good :+1:

juliushaertl:
> On web ui one is then simply redirected, right?
> Isn't the same possible? As we are already within a webview?

Yes, but on the web UI we have an authenticated session, while the direct editing uses a one-time token just for the file. Therefore we can only redirect to the normal files app, which might require the user to log in again.

tobiasKaminsky:
Hm. 
Then your first suggestion is good, I think.

So we have three new interface methods:
- openFile(String remotePath) -> open new file in clients, needed for link handling
- insertImage() -> client opens file picker and sends file to editor via
- postImage(String filename, String linkToUploadedFile)

juliushaertl:
In addition we would also need those two for inserting links to any file type: 
- insertFile() -> client opens file picker and sends file to editor via
- postFile(String filename, String linkToUploadedFile)

machiav3lli:
Any progress on this?
Thanks

Raudius:
Related to: https://github.com/nextcloud/text/issues/1169

In 1169 I suggest we handle the opening in the browser, as this would also work for direct editing from desktop. But we could additionally send a PostMessage to the clients in case they can handle the open request more gracefully.

juliushaertl:
Yep, sounds like something for a `openFile` message that is emitted for that case :)

PR #1900, opened 2021-10-18
Insert images from local device (or link)

### LINKED ISSUE #58: Ability to insert images from local device (or link)
It would be nice if it’s possible to also insert images from the local device into the document. For example:

- The image icon shows a popover having 2 options: "Image from Nextcloud" and "Image from device"
- "Image from Nextcloud" shows the current picker"
- "Image from device" shows a native file picker
- The image is uploaded in the background – we have to probably decide where to. We shouldn’t necessarily make the same mistake as in Talk and dump it into the root folder – maybe it doesn’t even need to be visible at all?

And as @juliushaertl said about remote images:

> We cannot add remote images in general because they are blocked by the CSP as mentioned in the error. So for now only images from the same domain work for that.
>
> @rullzer Does it make sense to to fetch remote images and deliver them from the local domain, similar to the proxy we have for the app store images? Or should we just block remote images for security reasons.

eneiluj:
Fixes #58

Here is a basic implementation of 2 new options to insert images:
* Upload from local device
* Download from http link

In both cases, the image is then stored in the `/Text` directory which is created if it does not exist.
An alternative mentioned in https://github.com/nextcloud/text/issues/58#issuecomment-540081015 would be to store uploaded images in a directory next to the markdown file.

### Todo
- [x] Menu action for local file insertion
- [x] Handling file upload
- [x] Display uploaded image
- [x] Handle and display upload errors
- [x] Menu action for image link insertion
- [x] Handling link download/store
- [x] Display image from link
- [x] Handle and display link insertion errors
- [x] Change menu item naming
- [x] Change attachment folder name
- [x] Use the attachment folder when inserting images from Files
- [x] Handle .md file copy/move/delete with File events
- [x] Cleanup attachment folder when images are removed from markdown content
### Gifcording :grin:
link


src/components/MenuBar.vue
```diff
@@ -23,6 +23,13 @@
 <template>
 	<EditorMenuBar v-slot="{ commands, isActive, focused }" :editor="editor">
 		<div class="menubar" :class="{ 'is-focused': focused, 'autohide': autohide }">
+			<input
+				ref="imageFileInput"
+				type="file"
+				accept="image/*"
+				aria-hidden="true"
+				class="hidden-visually"
+				@change="onImageFilePicked">
```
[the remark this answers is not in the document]
mejo-: Nevermind, you already solved this :)

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,86 @@
...
+	private $imageService;
+
+	public function __construct(string $appName,
+								IRequest $request,
+								ImageService $imageService,
+								?string $userId) {
+		parent::__construct($appName, $request);
+		$this->userId = $userId;
+		$this->imageService = $imageService;
+		$this->request = $request;
```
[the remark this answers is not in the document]
mejo-: Forget about that one, `$request` is incorporated from the parent class `Controller`.

src/components/MenuBar.vue
```diff
@@ -264,33 +314,137 @@ export default {
...
+			}).then((response) => {
+				this.insertImage(targetFilePath, this.imageCommand)
+			})
+			*/
+		},
+		onImageLinkUpdateValue(newImageLink) {
+			// this avoids the input being reset on each file polling
+			this.imageLink = newImageLink
+		},
+		onImageLinksubmit(command) {
```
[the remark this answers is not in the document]
eneiluj: Yep! Thanks.

eneiluj:
This now includes "attachment folder" changes. It's been rebased on master (so it must be used in NC 24).

The TODO list in this PR's description has been updated.

src/components/MenuBar.vue
```diff
@@ -34,6 +41,41 @@
 							:aria-label="t('text', 'Insert emoji')"
 							:aria-haspopup="true" />
 					</EmojiPicker>
+					<Actions v-else-if="icon.class === 'icon-image'"
```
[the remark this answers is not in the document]
eneiluj: I'll have a look.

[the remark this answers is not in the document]
eneiluj: It's fine on my side with NC 24 :thinking: 
col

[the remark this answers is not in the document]
mejo-: Interesting. I'm on latest NC 24 too :thinking: Will have to further look into it then.

lib/Controller/ImageController.php
```diff
@@ -74,6 +74,26 @@ public function __construct(string $appName,
...
+	public function insertImageFile(int $textFileId, string $imagePath): DataResponse {
+		try {
+			$insertResult = $this->imageService->insertImageFile($textFileId, $imagePath, $this->userId);
+			if (isset($insertResult['error'])) {
+				return new DataResponse($insertResult, Http::STATUS_BAD_REQUEST);
+			} else {
+				return new DataResponse($insertResult);
+			}
+		} catch (Exception $e) {
+			return new DataResponse(['error' => 'File insertion error: ' . $e->getMessage()], Http::STATUS_BAD_REQUEST);
```
juliushaertl, line 93: Let's not expose the actual exception message and rather log it or use a [HintException](https://github.com/nextcloud/server/blob/b78f3a57d1d05fb5f81ec8361139b64a3e54d7b0/lib/public/HintException.php#L36) for known cases. Otherwise exception messages might leak internal data.

lib/Service/ImageService.php
```diff
@@ -394,6 +452,26 @@ private function getOrCreateAttachmentDirectoryForFile(File $textFile): ?Folder
...
+	 * @param int $filePath
+	 * @param string $userId
+	 * @return Folder|null
+	 * @throws \OCP\Files\NotPermittedException
+	 * @throws \OC\User\NoUserException
+	 */
+	private function getFileFromPath(string $filePath, string $userId): ?File {
+		$userFolder = $this->rootFolder->getUserFolder($userId);
+		if ($userFolder->nodeExists($filePath)) {
+			$file = $userFolder->get($filePath);
```
juliushaertl, line 467: Will throw a NotFoundException so would be good to also annotate that if not catches

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+		$textFile = $this->getTextFile($textFileId, $userId);
+		$attachmentFolder = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+		if ($attachmentFolder !== null) {
+			try {
+				$imageFile = $attachmentFolder->get($imageFileName);
+			} catch (NotFoundException $e) {
+				return null;
+			}
+			if ($imageFile instanceof File) {
+//				return $imageFile;
```
juliushaertl, line 109: - [ ] To be discussed if we still need the original file for gif support for example

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+		$textFile = $this->getTextFilePublic($textFileId, $shareToken);
+		$attachmentFolder = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+		if ($attachmentFolder !== null) {
+			try {
+				$imageFile = $attachmentFolder->get($imageFileName);
+			} catch (NotFoundException $e) {
+				return null;
+			}
+			if ($imageFile instanceof File) {
+//				return $imageFile;
```
juliushaertl, line 139: - [ ] To be discussed if we still need the original file for gif support for example

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 */
+	private function copyImageFile(File $imageFile, Folder $saveDir, File $textFile): array {
+		$mimeType = $imageFile->getMimeType();
+		if (in_array($mimeType, ImageController::IMAGE_MIME_TYPES)) {
+			$fileName = (string) time() . '-' . $imageFile->getName();
+			$targetPath = $saveDir->getPath() . '/' . $fileName;
+			$targetFile = $imageFile->copy($targetPath);
+			$path = preg_replace('/^files/', '', $targetFile->getInternalPath());
```
juliushaertl, line 260: Might not work well with other storages than the user home (e.g. group folders)

I think something like this should also work to resolve the file within the user folder:

`$userFolder->getRelativePath($file->getPath())`

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+				'name' => $fileName,
+				'path' => $path,
+				'id' => $targetFile->getId(),
+				'textFileId' => $textFile->getId(),
+			];
+		} else {
+			return [
+				'error' => 'Unsupported file type',
+			];
+		}
```
juliushaertl, line 272:
```suggestion
	return [
		'error' => 'Unsupported file type',
	];
```

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+	 * @return array|string[]
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotPermittedException
+	 * @throws \OCP\Lock\LockedException
+	 */
+	private function downloadLink(Folder $saveDir, string $link, File $textFile): array {
+		$fileName = (string) time();
+		$savedFile = $saveDir->newFile($fileName);
+		$resource = $savedFile->fopen('w');
```
juliushaertl, line 356:
```suggestion
		$resource = $savedFile->fopen('wb');
```

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+				$response = $client->get($url, $options);
+			} elseif ($method === 'POST') {
+				$response = $client->post($url, $options);
+			} elseif ($method === 'PUT') {
+				$response = $client->put($url, $options);
+			} elseif ($method === 'DELETE') {
+				$response = $client->delete($url, $options);
+			} else {
+				return ['error' => 'Bad HTTP method'];
+			}
```
juliushaertl, line 572: There doesn't seem to be much use of other methods than GET here, right?

lib/Service/ImageService.php
```diff
@@ -0,0 +1,596 @@
...
+		} catch (ServerException | ClientException $e) {
+			//$response = $e->getResponse();
+			//if ($response->getStatusCode() === 401) {
+			$this->logger->warning('Impossible to download image: '.$e->getMessage(), ['app' => Application::APP_NAME]);
+			return ['error' => 'Impossible to download image'];
+		} catch (ConnectException $e) {
+			$this->logger->error('Connection error: ' . $e->getMessage(), ['app' => Application::APP_NAME]);
+			return ['error' => 'Connection error'];
+		} catch (Throwable | Exception $e) {
+			$this->logger->error('Unknown error: ' . $e->getMessage(), ['app' => Application::APP_NAME]);
```
juliushaertl, line 592:
```suggestion
			$this->logger->error('Unknown error when downloading the file', ['exception' => $e]);
```

No need for specifying the app if the LoggerInterface is injected through DI as it will be resolved to an instance that has the app already set. Also losing the exception directly will preserve the trace which might be helpful.

src/components/MenuBar.vue
```diff
@@ -225,6 +294,10 @@ export default {
 			return this.lastImagePath
 				|| this.filePath.split('/').slice(0, -1).join('/')
 		},
+		token() {
```
juliushaertl, line 297: Might be good to call it sharingToken, as there is a separate token for the session within the text codebase as well, to avoid any confusion

lib/Service/ImageService.php
eneiluj: Oh right. I'll keep that in mind. In this specific case, the `path` attribute is actually not used anymore in the frontend. It has been removed.

lib/Service/ImageService.php
eneiluj: Yes and we most likely won't use other methods later.

lib/Service/ImageService.php
eneiluj: Ok, fixed in many places.

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param int $textFileId
+	 * @param string $imagePath
+	 * @return DataResponse
+	 */
+	public function insertImageFile(int $textFileId, string $imagePath): DataResponse {
```
juliushaertl, line 82: Thinking about that it might be useful to just make this a public endpoint for both logged in users and public share links and use the session token to authenticate. That way the same methods can also be used when opening a file through direct editing (ref https://github.com/nextcloud/server/pull/17625) through the desktop/Android/iOS clients.

FYI @tobiasKaminsky For inserting local files/files from an URL this should then work without any client change. The existing feature to insert an image from Nextcloud would though depend on https://github.com/nextcloud/text/issues/1170

eneiluj: @juliushaertl Yes sure, seems better. Do you have any example where authentication is checked like that?

lib/Service/ApiService.php
```diff
@@ -47,12 +47,18 @@ class ApiService {
 	protected $sessionService;
 	protected $documentService;
 	protected $logger;
+	private ImageService $imageService;
```
mejo-, line 50: That breaks on PHP7.3 :)

```suggestion
	protected $imageService;
```

src/components/MenuBar.vue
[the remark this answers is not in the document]
mejo-: I took another look and it seems like `vertical-align: top` is missing. The other action icons have it either through `menubar.button`, `menubar.submenu` or `menubar.menuitem-emoji` styling.

```suggestion
					<Actions v-else-if="icon.class === 'icon-image'"
						class="submenu"
```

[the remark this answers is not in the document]
eneiluj: Oh I realized there actually is an alignment issue on my side too but it's the 3 dots menu icon rather than the image one. Anyway, using the submenu is fine. Done, pushed. Thanks!

lib/Service/ApiService.php
eneiluj: Right, I though Php 7.3 support had been dropped in NC 24 but it's actually not.

eneiluj:
Hey there, some news:
* Attachment **cleanup** is implemented. It happens when closing Text if there are no more open sessions. It's also done by the Cleanup cron job when cleaning up the inactive sessions.
* The attachment dir is **deleted** when a markdown file is deleted.
* The attachment dir is **moved** to follow the markdown file when it's moved. This is only done if the moved .md file was in the same directory than the attachment dir. We make sure the attachment dir stays next to the .md file in the owner's storage. This covers all the use case I could think of:
    * the .md file's owner moves it: the attachment dir is moved
    * the .md file is shared (single file share) and moved by the share target user: the attachment is **not** moved
    * a parent directory of the .md file is shared, the share target user moves the .md file: the attachment dir is moved.
* File **copy** is implemented but there is a major issue: The target file is locked when catching the event so we can't make changes to its content (which is needed to update the attachment links). @juliushaertl @mejo- Any idea how we could proceed here?

eneiluj:
Some news here:
* Attachment cleanup is done:
    * when closing the document and there is no other active session
    * in the Cleanup cron job if an inactive session is being deleted and there is no other active ones
 * On file copy, the attachment folder is copied too. 
 * On file move, the attachment folder is moved only if the .md file was in its owner's storage.
 * On file deletion, the attachment folder is deleted.

So I think everything we discussed is implemented. There still are some things to check and improve:

- [ ] We need to edit the target .md file when copying it to adapt the attachment file IDs in its content. Problem is the .md file is locked when we catch the copy event. For now it's temporarily unlocked but I'm sure there is a cleaner solution :grin:
- [ ] The .md file ID is stored in each attachment link in order to use it as parameter when getting images through Text. This is kind of hacky. It would be better to find a way to get the .md file ID in the ImageView component context.
- [ ] As mentioned earlier, it would be better to have a single public endpoint to serve images to logged in users and public share links. The session token would be used to authenticate.

eneiluj:
Rebased on master.

* @juliushaertl I tried to implement authentication via edition session like you said. There is now only one public endpoint for each upload method. In non-public contexts, we now get the userId from the session instead of relying on NC auth.
Could you have a look at https://github.com/nextcloud/text/pull/1900/commits/06d4d3866caf7c9c9de65b31ee09e22b9721b934 ? Is this more or less what you had in mind?

* It would be nice to use the edition session when serving images too. It would allow us to avoid storing the document ID in each image link (which is...meh). But I can't find a clean way to get the session info in the ImageView component to be able to put it in the "get image" requests.

eneiluj:
I think it's finally in a good shape :tada: .

Thanks to @mejo- 's suggestion, images are now served via a single endpoint in both public and non-public contexts. Session information is now accessible via the Vuex store so the ImageView component can use it to generate the image URL.

@juliushaertl @mejo- I wonder if those changes to the Vuex store are ok: 7e0df82d076a6dcaf1c20babcbfb45ef542e9277 more specifically in `src/nodes/ImageView.vue` and `src/store.js`.

The document file ID and the image file IDs are not written in the markdown content anymore which makes it easier to handle the .md file copy/move/delete events. Only the attachment image file name is written.

Oh and I don't think the failing CI checks are blocking :grin:.

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+		if ($attachmentFolder !== null) {
+			try {
+				$imageFile = $attachmentFolder->get($imageFileName);
+			} catch (NotFoundException $e) {
+				return null;
+			}
+			if ($imageFile instanceof File) {
+				if ($this->previewManager->isMimeSupported($imageFile->getMimeType())) {
+					return $this->previewManager->getPreview($imageFile, 1024, 1024);
+				} else {
```
mejo-, line 133: From PhpStorm: [EA] Child instructions can be extracted here (clearer intention, lower complexity numbers).

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 * @param string $userId
+	 * @return array
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	public function uploadImage(int $documentId, string $newFileName, string $newFileContent, string $userId): array {
+		$textFile = $this->getTextFile($documentId, $userId);
+		if (!$textFile->isUpdateable()) {
+			throw new Exception('No write permissions');
```
mejo-, line 156: Maybe throw a `\OCP\Files\NotPermittedException` here instead of a generic `\Exception`?

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 * @param string $newFileContent
+	 * @param string $userId
+	 * @return array
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	public function uploadImage(int $documentId, string $newFileName, string $newFileContent, string $userId): array {
+		$textFile = $this->getTextFile($documentId, $userId);
+		if (!$textFile->isUpdateable()) {
```
mejo-, line 155: Better to check for $textFile to prevent null pointer exception:

```suggestion
		if (!$textFile || !$textFile->isUpdateable()) {
```

Or even better, throw a `NotFoundException` if empty in `$this->getTextFile()`?

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+		$saveDir = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+		if ($saveDir !== null) {
+			$fileName = (string) time() . '-' . $newFileName;
+			$savedFile = $saveDir->newFile($fileName, $newFileContent);
+			return [
+				'name' => $fileName,
+				'id' => $savedFile->getId(),
+				'documentId' => $textFile->getId(),
+			];
+		} else {
```
mejo-, line 167: From PhpStorm: [EA] Child instructions can be extracted here (clearer intention, lower complexity numbers).

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+
+	/**
+	 * Get or create the user-specific attachment folder
+	 *
+	 * @param string $userId
+	 * @return Folder|null
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	private function getOrCreateTextDirectory(string $userId): ?Folder {
```
mejo-, line 390: This seems like an old leftover? At lest the function is not used.

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+				}
+				$targetPath = $saveDir->getPath() . '/' . $fileName;
+				$savedFile->move($targetPath);
+				// get file type and name
+				return [
+					'name' => $fileName,
+					'id' => $savedFile->getId(),
+					'documentId' => $textFile->getId(),
+				];
+			} else {
```
mejo-, line 365: From PhpStorm: [EA] Child instructions can be extracted here (clearer intention, lower complexity numbers).

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 */
+	public function insertImageLinkPublic(?int $documentId, string $link, string $shareToken): array {
+		if (!$this->hasUpdatePermissions($shareToken)) {
+			throw new Exception('No write permissions');
+		}
+		$textFile = $this->getTextFilePublic($documentId, $shareToken);
+		$saveDir = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+		if ($saveDir !== null) {
+			return $this->downloadLink($saveDir, $link, $textFile);
+		} else {
```
mejo-, line 303: From PhpStorm: [EA] Child instructions can be extracted here (clearer intention, lower complexity numbers).

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 */
+	public function insertImageLink(int $documentId, string $link, string $userId): array {
+		$textFile = $this->getTextFile($documentId, $userId);
+		if (!$textFile->isUpdateable()) {
+			throw new Exception('No write permissions');
+		}
+		$saveDir = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+		if ($saveDir !== null) {
+			return $this->downloadLink($saveDir, $link, $textFile);
+		} else {
```
mejo-, line 279: From PhpStorm: [EA] Child instructions can be extracted here (clearer intention, lower complexity numbers).

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 * @param string $link
+	 * @param string $userId
+	 * @return array
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	public function insertImageLink(int $documentId, string $link, string $userId): array {
+		$textFile = $this->getTextFile($documentId, $userId);
+		if (!$textFile->isUpdateable()) {
```
mejo-, line 273: Same as above: either check to prevent null pointer exception, or throw in `$this->getTextFile()`.

mejo- (changes requested):
In general, I'm really happy with the implemention and think it's almost done. All my comments are rather minor compared to the huge PR.

But how about adding some cypress end-to-end tests for image uploads?

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	/**
+	 * Get attachment file names listed in the markdown file content
+	 *
+	 * @param string $content
+	 * @return array
+	 */
+	private function getAttachmentNamesFromContent(string $content): array {
+		$matches = [];
+		preg_match_all(
+			'/\!\[[^\[\]]+\]\(text:\/\/image\?[^)]*imageFileName=([^)&]+)\)/',
```
mejo-, line 599: The regex doesn't match image links with square brackets in the description. I had a quick look at [the relevant part of the CommonMark spec](https://spec.commonmark.org/0.30/#links) and indeed, square brackets are allowed (under limited circumstances).

php-markdown uses the following regex syntax for the link/image name/description part between squares:

```regex
(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[\])*\])*\])*\])*\])*\])*
```

See [`$this->nested_brackets_re`](https://github.com/michelf/php-markdown/blob/f0911e737ddaad081803424203407a0f01189eca/Michelf/Markdown.php#L106) and [`doImages()`](https://github.com/michelf/php-markdown/blob/f0911e737ddaad081803424203407a0f01189eca/Michelf/Markdown.php#L693)

lib/Service/ImageService.php
eneiluj: :+1:

lib/Service/ImageService.php
eneiluj: Yep. `ImageController::getImage()` now catches the exceptions.
`ImageService::getTextFile()` and `ImageService::getTextFilePublic()` now throw a `NotFoundException`.

lib/Service/ImageService.php
eneiluj: :+1:

lib/Service/ImageService.php
eneiluj: I don't get this on my side. What does it mean? :grin:

src/components/MenuBar.vue
```diff
@@ -34,6 +41,43 @@
...
+						:default-icon="'icon-image'"
+						@close="onImageActionClose">
+						<button slot="icon"
+							:class="{ 'icon-image': true, 'loading-small': uploadingImage }"
+							:title="icon.label"
+							:aria-label="icon.label"
+							:aria-haspopup="true" />
+						<ActionButton
+							icon="icon-upload"
+							:close-after-click="true"
```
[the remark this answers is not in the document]
eneiluj: :+1:

src/components/MenuBar.vue
```diff
@@ -34,6 +41,43 @@
...
+							@click="onUploadImage(commands.image)">
+							{{ t('text', 'Upload from computer') }}
+						</ActionButton>
+						<ActionButton v-if="!isPublic"
+							icon="icon-folder"
+							:close-after-click="true"
+							@click="showImagePrompt(commands.image)">
+							{{ t('text', 'Insert from Files') }}
+						</ActionButton>
+						<ActionButton v-show="!showImageLinkPrompt"
```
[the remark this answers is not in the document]
eneiluj: Yes, right. I remember having some issues with v-if in <Action*> where an action is not visible if it is rendered after <Actions>. But here it seems to work fine.

lib/AppInfo/Application.php
```diff
@@ -53,6 +59,10 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(LoadViewer::class, LoadViewerListener::class);
 		$context->registerEventListener(LoadAdditionalScriptsEvent::class, FilesLoadAdditionalScriptsListener::class);
 		$context->registerEventListener(BeforeTemplateRenderedEvent::class, FilesSharingLoadAdditionalScriptsListener::class);
+		// for image attachments
+		$context->registerEventListener(NodeCopiedEvent::class, NodeCopiedListener::class);
+		$context->registerEventListener(BeforeNodeRenamedEvent::class, BeforeNodeRenamedListener::class);
+		$context->registerEventListener(BeforeNodeDeletedEvent::class, BeforeNodeDeletedListener::class);
```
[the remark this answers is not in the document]
eneiluj: It's not possible to get the attachment dir after a file has been deleted or moved because we use it (the file) to get its owner's storage.
I agree it would be nicer to use NodeRenamedEvent and BeforeNodeDeletedEvent but I think it's not possible.

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	/**
+	 * Get or create file-specific attachment folder
+	 *
+	 * @param File $textFile
+	 * @return Folder|null
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	private function getOrCreateAttachmentDirectoryForFile(File $textFile): ?Folder {
```
mejo-, line 392: Since `getOrCreateAttachmentDirectoryForFile()` and `getAttachmentDirectoryForFile()` share a lot of code, wouldn't it be better to merge them and instead e.g. add a boolean parameter `$create`?

eneiluj:
@mejo- Thanks a lot for the review.

mejo- (changes requested):

lib/Service/ImageService.php
eneiluj: :+1:

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 */
+	private function getAttachmentNamesFromContent(string $content): array {
+		$matches = [];
+		preg_match_all(
+			'/\!\[[^\[\]]+\]\(text:\/\/image\?[^)]*imageFileName=([^)&]+)\)/',
+			$content,
+			$matches,
+			PREG_SET_ORDER
+		);
+		return array_map(function (array $match) {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 */
+	public function cleanupAttachments(int $fileId): int {
+		$textFile = $this->rootFolder->getById($fileId);
+		if (count($textFile) > 0 && $textFile[0] instanceof File) {
+			$textFile = $textFile[0];
+			if ($textFile->getMimeType() === 'text/markdown') {
+				// get IDs of the files inside the attachment dir
+				$attachmentDir = $this->getOrCreateAttachmentDirectoryForFile($textFile);
+				$attachmentsByName = [];
+				foreach ($attachmentDir->getDirectoryListing() as $attNode) {
```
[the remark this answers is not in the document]
eneiluj: Yep. Moreover, we don't need to create the attachment dir there if it does not exist.

lib/Service/ImageService.php
mejo-: It basically means that you can remove the `else {` statement and instead put everything inside `else` **behind** the `if` statement.

It's about changing

```php
if {
    ...
    return x;
} else {
    do_something();
}
```

to

```php
if {
    ...
    return x;
}
do_something();
```

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 * @throws \OCP\Lock\LockedException
+	 */
+	public function moveAttachments(File $source, File $target): void {
+		// if the parent directory has changed
+		if ($source->getParent()->getPath() !== $target->getParent()->getPath()) {
+			$sourceAttachmentDir = $this->getAttachmentDirectoryForFile($source);
+			// if there is an attachment dir for this file
+			// and it is in the same directory as the source file
+			if ($sourceAttachmentDir !== null
+				&& $source->getParent()->getId() === $sourceAttachmentDir->getParent()->getId()
```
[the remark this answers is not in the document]
eneiluj: I improved the comment. This is tricky.
* move:
    * if done by the .md file's owner, the attachment dir is next to the .md file => we should move the attachment dir
    * if someone else having access does it, as the file won't move in the owner's storage => we don't move the attachment dir
* copy is different because we duplicate the attachments, this is done regardless if the file is copied by its owner or someone else.

Does it make sense? :grin:

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+	 */
+	public function deleteAttachments(File $source): void {
+		// if there is an attachment dir for this file
+		$sourceAttachmentDir = $this->getAttachmentDirectoryForFile($source);
+		if ($sourceAttachmentDir !== null) {
+			$sourceAttachmentDir->delete();
+		}
+	}
+
+	public function copyAttachments(File $source, File $target): void {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Service/ImageService.php
```diff
@@ -0,0 +1,676 @@
...
+		try {
+			$share = $this->shareManager->getShareByToken($shareToken);
+			if ($share->getShareType() === IShare::TYPE_LINK) {
+				// shared file or folder?
+				if ($share->getNodeType() === 'file') {
+					$textFile = $share->getNode();
+					if ($textFile instanceof File) {
+						return $textFile;
+					}
+				} elseif ($share->getNodeType() === 'folder' && $documentId !== null) {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,224 @@
...
+	 * @param string $sessionToken
+	 * @param string|null $shareToken
+	 * @param string $imageFileName
+	 * @return DataDisplayResponse
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\NotFoundException
+	 * @throws \OCP\Files\NotPermittedException
+	 * @throws \OCP\Lock\LockedException
+	 */
+	public function getImage(int $documentId, int $sessionId, string $sessionToken, ?string $shareToken = null, string $imageFileName): DataDisplayResponse {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Service/ImageService.php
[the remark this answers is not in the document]
mejo-: Ah, makes sense, thanks for the explanation! Maybe just use the last comment?

```php
		// Move attachement dir only if .md file is owned by it's owner.
```

lib/Service/ImageService.php
mejo-: See https://github.com/kalessil/phpinspectionsea/blob/master/docs/control-flow.md#redundant-else-keyword for details :blush:

I guess it's the setting in Editor -> Inspections -> PHP -> Control flow -> Unnecessary 'else' branch.

lib/Service/ImageService.php
```diff
@@ -0,0 +1,658 @@
...
+			if ($ownerParentFolder->nodeExists($attachmentFolderName)) {
+				$attachmentFolder = $ownerParentFolder->get($attachmentFolderName);
+				if ($attachmentFolder instanceof Folder) {
+					return $attachmentFolder;
+				}
+			} elseif ($create) {
+				return $ownerParentFolder->newFolder($attachmentFolderName);
+			}
+		}
+		return null;
```
mejo-, line 416: I wonder whether it would be cleaner to throw a `NotFoundException` instead of returning null in low-level functions like this one, `getFileFromPath()` and `getImagePreview()`. That way you don't have to check for the returned value being null in the higher level functions and don't have to care about null pointer exceptions.

lib/Service/ImageService.php
```diff
@@ -0,0 +1,658 @@
...
+	}
+
+	/**
+	 * Download a file and write it to a resource
+	 * @param string $url
+	 * @param $resource
+	 * @param array $params
+	 * @return array
+	 */
+	private function simpleDownload(string $url, $resource, array $params = []): array {
```
mejo-, line 500: The parameter `$params` is never passed to the private function `simpleDownload()` and probably can be dropped, no?

mejo- (changes requested):

lib/Service/ImageService.php
eneiluj: :+1: thanks

lib/Service/ImageService.php
eneiluj: :+1:

lib/Service/ImageService.php
eneiluj: Much better, simpler and more readable.
Exceptions are now caught in the controller (except for the delete/copy/move methods in the service).
Done in c204d904ab3a1bba6c58d32d8a615ebab04808b4

lib/Service/ImageService.php
```diff
@@ -0,0 +1,658 @@
...
+	 * @param File $imageFile
+	 * @param Folder $saveDir
+	 * @param File $textFile
+	 * @return array
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 */
+	private function copyImageFile(File $imageFile, Folder $saveDir, File $textFile): array {
+		$mimeType = $imageFile->getMimeType();
+		if (in_array($mimeType, ImageController::IMAGE_MIME_TYPES)) {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,225 @@
...
+	 */
+	public function uploadImage(int $documentId, int $sessionId, string $sessionToken, ?string $shareToken = null): DataResponse {
+		if (!$this->sessionService->isValidSession($documentId, $sessionId, $sessionToken)) {
+			return new DataResponse([], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		try {
+			$file = $this->request->getUploadedFile('image');
+			if ($file !== null && isset($file['tmp_name'], $file['name'], $file['type'])) {
+				if (!in_array($file['type'], self::IMAGE_MIME_TYPES)) {
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Service/ImageService.php
```diff
@@ -0,0 +1,658 @@
...
+		$resource = $savedFile->fopen('wb');
+		$res = $this->simpleDownload($link, $resource);
+		if (is_resource($resource)) {
+			fclose($resource);
+		}
+		$savedFile->touch();
+		if (isset($res['Content-Type'])) {
+			if (in_array($res['Content-Type'], ImageController::IMAGE_MIME_TYPES)) {
+				if ($res['Content-Type'] === 'image/jpeg') {
+					$fileName = $fileName . '.jpg';
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,210 @@
...
+	 * @return DataResponse
+	 */
+	public function uploadImage(int $documentId, int $sessionId, string $sessionToken, ?string $shareToken = null): DataResponse {
+		if (!$this->sessionService->isValidSession($documentId, $sessionId, $sessionToken)) {
+			return new DataResponse([], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		try {
+			$file = $this->request->getUploadedFile('image');
+			if ($file !== null && isset($file['tmp_name'], $file['name'], $file['type'])) {
```
juliushaertl, line 156: Some validation on the uploaded file would be nice here, to give some more feedback to the user or log, for possible inspiration: https://github.com/nextcloud/deck/blob/13dcacc3bb4149ebc7283f34f6d9af05322adea5/lib/Service/FilesAppService.php#L218-L235 but not a blocker for me.

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,210 @@
...
+				if (!in_array($file['type'], self::IMAGE_MIME_TYPES, true)) {
+					return new DataResponse(['error' => 'Image type not supported'], Http::STATUS_BAD_REQUEST);
+				}
+				$newFileContent = file_get_contents($file['tmp_name']);
```
juliushaertl, line 160: Also no blocker but would be good to stream the file to the storage instead of loading it into memory. https://github.com/nextcloud/deck/blob/13dcacc3bb4149ebc7283f34f6d9af05322adea5/lib/Service/FilesAppService.php#L194-L198

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,210 @@
...
+		}
+
+		try {
+			if ($shareToken) {
+				$imageFile = $this->imageService->getImagePublic($documentId, $imageFileName, $shareToken);
+			} else {
+				$imageFile = $this->imageService->getImage($documentId, $imageFileName, $this->userId);
+			}
+			return $imageFile !== null
+				? new DataDisplayResponse($imageFile->getContent(), Http::STATUS_OK, ['Content-Type' => $imageFile->getMimeType()])
```
juliushaertl, line 203: I'm not fuilly sure if necessary, but to be on the safe side this should use https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/public/Files/IMimeTypeDetector.php#L68 to avoid any mimetype that would allow potential XSS

lib/Service/ImageService.php
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+
+	/**
+	 * Get attachment file names listed in the markdown file content
+	 *
+	 * @param string $content
+	 * @return array
+	 */
+	private function getAttachmentNamesFromContent(string $content): array {
```
juliushaertl, line 550: Bonus points for some quick unit test ;)

lib/Service/ImageService.php
```diff
@@ -0,0 +1,631 @@
...
+	/**
+	 * Check if the shared access has write permissions
+	 *
+	 * @param string $shareToken
+	 * @return bool
+	 */
+	private function hasUpdatePermissions(string $shareToken): bool {
+		try {
+			$share = $this->shareManager->getShareByToken($shareToken);
+			return ($share->getShareType() === IShare::TYPE_LINK && $share->getPermissions() & Constants::PERMISSION_UPDATE);
```
juliushaertl, line 293: `IShare::TYPE_EMAIL` would also be a case where this should work as they behave the same like share links.

juliushaertl: `IShare::TYPE_ROOM` would also be possible if the room is shared to the public, then such shares would be used to access the file as guests

juliushaertl (changes requested):
Very nice work @eneiluj I've added a few comments inline, but that should be rather small stuff.

lib/Service/ImageService.php
eneiluj: This is tricky. The inserted `[photo.name.jpg](text://image?imageFileName=photo.name.jpg)` has to be valid from the renderer point of view. I'll first fix that and then adjust `getAttachmentNamesFromContent()`. Thanks for mentioning this as image insertion fails in many cases.

juliushaertl:
Regarding the psalm issues on CI, the following two can be added to the baseline as they are due to leaking oc namespace classes in ocp ones: `oc\hooks\emitter` and `OC\User\NoUserException`

lib/Service/ImageService.php
eneiluj: :+1: Done in fad15fa0a42010eab2905dba2e9479d1a53ca025

lib/Service/ImageService.php
eneiluj: Yep, will do after fixing smaller stuff.

lib/Controller/ImageController.php
eneiluj: :+1: Done in 7628714403f41706f5dd9465c7a4b48898a6ecbc

lib/Controller/ImageController.php
eneiluj: :+1: Done in d1cf1fb7f80b68bf3f3af0769c5e410ba831e2c2

lib/Controller/ImageController.php
eneiluj: :+1: Done in bef8698e291caab8ff5e0485e9b2b53f63aa888c (almost like in Deck).

lib/Controller/ImageController.php
```diff
@@ -0,0 +1,210 @@
...
+	 * @param string $link
+	 * @param int $documentId
+	 * @param int $sessionId
+	 * @param string $sessionToken
+	 * @param string|null $shareToken
+	 * @return DataResponse
+	 */
+	public function insertImageLink(string $link, int $documentId, int $sessionId, string $sessionToken, ?string $shareToken = null): DataResponse {
+		if (!$this->sessionService->isValidSession($documentId, $sessionId, $sessionToken)) {
+			return new DataResponse([], Http::STATUS_INTERNAL_SERVER_ERROR);
```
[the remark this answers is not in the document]
eneiluj: :+1:

lib/Controller/ImageController.php
```diff
@@ -175,6 +183,35 @@ public function uploadImage(int $documentId, int $sessionId, string $sessionToke
...
+		];
+
+		if (empty($file)) {
+			$error = $this->l10n->t('No file uploaded or file size exceeds maximum of %s', [Util::humanFileSize(Util::uploadLimit())]);
+		}
+		if (!empty($file) && array_key_exists('error', $file) && $file['error'] !== UPLOAD_ERR_OK) {
+			$error = $phpFileUploadErrors[$file['error']];
+		}
+		if ($error !== null) {
+			throw new Exception($error);
```
mejo-, line 221: Maybe again a more specific exception would be nice here? :thinking:

lib/Service/ImageService.php
eneiluj: The expression from php-markdown is now used in `getAttachmentNamesFromContent()`.
As a simple fix for the nested brackets, as we generate the link ourselves and the user is not supposed to manually edit it afterwards, brackets in image file names are removed in the link text.
There is no problem in the link itself as it is "url encoded".
This is done in c351e66607d23b1672603752b2bdd1dba83b59a6

lib/Service/ImageService.php
eneiluj: A small test was added in c351e66607d23b1672603752b2bdd1dba83b59a6
Do you have more ideas of potentially problematic names?

lib/Controller/ImageController.php
eneiluj: Yep, done in c92316183af72ee92a6a73a2b31e4b7da0255c74

eneiluj:
@juliushaertl @mejo- I think everything you mentioned is fixed.

I'll rebase, squash the commits and compile when we're done with the review.

lib/Service/ImageService.php
mejo-: Perfect, I think that's fine for now :ok_hand:

mejo-:
Super nice work :ok_hand:

As mentioned, some cypress tests would be nice - apart from that, the PR is ready to be merged from my side :tada:

eneiluj:
> Super nice work ok_hand
> 
> As mentioned, some cypress tests would be nice - apart from that, the PR is ready to be merged from my side tada

Oh yes right, I'll dive into cypress soon.
Thanks again for the solid review!

lib/Service/ImageService.php
juliushaertl: Looks good :)

juliushaertl (approved):

eneiluj:
Rebased on master, added small tests for `<ImageView>`.
We can now merge or wait until a pairing session about Cypress and some more tests. As you wish.

juliushaertl:
Good to merge right away from my side 👍

mejo-:
> We can now merge or wait until a pairing session about Cypress and some more tests. As you wish.

I don't mind to merge now and add cypress tests later (if we don't forget about it :grimacing:). But the failing psalm tests should be tackled first, no?

mejo- (approved):
Super nice work :ok_hand:

juliushaertl:
Filed https://github.com/nextcloud/text/issues/2049 for the cypress tests

tobiasKaminsky:
Is this also ready for clients via our DirectEditingMobileInterface?

eneiluj:
@tobiasKaminsky All related endpoints are public and use the session edition token and optionally the public share token to authenticate.
So it is supposed to be ok but in reality it partially works. I've tried it on our Android app:
* Inserting an image from a link :heavy_check_mark: 
* Uploading a local image :x: The `<input type=file>` does not work when rendered in the Android app.
* Serving the images :x: 2 mistakes:
    * The share token presence detection is broken in <ImageView>
    * The ImageController does not get the user ID from the session

I'll make a PR to fix the 2 image serving issues.
I don't know about the local image upload issue. Do you think we should just hide this feature on mobile apps? Or is there a way to get a system file selector when clicking on such file input?

tobiasKaminsky:
> I don't know about the local image upload issue. Do you think we should just hide this feature on mobile apps? Or is there a way to get a system file selector when clicking on such file input?

This is possible, as we do it with Collabora.

In this regard, we have this https://github.com/nextcloud/server/issues/16132, but want to wait for your side (move RichDocuments to directEditing)

PR #1904, opened 2021-10-20
Fix wrong generation of ocs url for rich workspaces

juliusknorr:
Found when testing #1860 and also encountered by @azul 

This was a regression introduced with #1565 where the ocs path for fetching rich workspace details was not properly constructed anymore.


juliushaertl:
Might be a good case to add a cypress test for showing rich workspaces

juliushaertl:
Filed https://github.com/nextcloud/text/issues/1905 for that

Issue #75, opened 2019-06-18
Support underline


jospoortvliet:
It'd be great to add underline when using `_` like this works in discord. Right now, one _ adds `<em>` tags, two add a `<b>` tag, just like * does. I suggest to let `_` add a `<u>` tag. You could let `__` add a double underline, but that's a bit much perhaps.

ryanprior:
Counterpoint: don't support underlining at all.

The argument has been made elsewhere better than I could make it, so I'll quote and link to the source:

> In a printed document, don’t underline. Ever. It’s ugly and it makes text harder to read.
> 
> Underlining is another dreary typewriter habit. Typewriters had no bold or italic styling. So the only way to emphasize text was to back up the carriage and type underscores beneath the text. It was a workaround for shortcomings in typewriter technology.
> 
> Neither your word processor nor your web browser suffers from these shortcomings. If you feel the urge to underline, use bold or italic instead. In special situations, like headings, you can also consider using all caps, small caps, or changing the point size.
> 
> Not convinced? I invite you to find a book, newspaper, or magazine that underlines text. That look is mostly associated with supermarket tabloids. If that’s the impression you want to make with your writing, by all means, use underlining. If not, don’t.

From [Butterick's Practical Typography](https://practicaltypography.com/underlining.html)

jancborchardt:
@ryanprior as a fellow typography nerd, I can appreciate that. ;) However, the points with Discord using underline and healthcare professionals commonly using underline are very good ones. So we shouldn’t restrict it on principle if there’s actally people using it, especially for such an important field as healthcare.

We still won’t underline anything (except maybe links) in Nextcloud, so there’s nothing to worry about. :)

brainchild0:
I would agree that it is unnecessary to  support underlining, largely being a tool for preparing manuscripts using pen or typewriter, to indicate that the typesetter should choose some appropriate typeface distinct from the main document text, such as is unavailable historically except in commercial printing.

In an environment where italics and bold can be used without effort, we should agree that underlining loses its historic function.



I would also agree that as users will be accustomed to single and double underscores demarcating emphasis and strong emphasis, and as documents from external sources will also use the same convention, these elements should be rendered similarly as in other environments.

Triple underscores should further be treated no differently from triple asterisks, as doing otherwise will break compatibility.


[...]

PR #1913, opened 2021-10-25
Support underline - use `__underlined__` in markdown

azul:
* Resolves: #75
* Target version: master 

### Summary

Underlines are not supported in commonmark markdown. So we need our own way of encoding them. Tripple `_` means double `_` plus single `_` - so bold AND italic in most markdown editors. So use quadrupple `_` instead.

Only available with `Ctrl + u` keyboard shortcut for now. Maybe that's enough as we don't want to encorage underline anyway.


eneiluj:
Nice!

So how do we let users know how to underline things? Why not adding a button in the menubar? In other words, why don't we encourage underlining?

Quadruple underscore won't be recognized as underline in any other markdown renderers. Are we ok with that?

azul:
@eneiluj I think one reason not to encourage it is the second point you brought up. It's not supported by other markdown things. Also most designers i talked to don't like underlining text. It's harder to read and does not add meaning unless it's in a specific context like medicine.

My reasoning for now would be if people ask for underline we can tell them the keyboard combination. If people paste underlined text it will be preserved. From my point of view that solves the main use cases. But I'd be okay with adding a button - maybe below the bold button like a menu to not make the toolbar even wider.

Is there any underline syntax that other markdown editors understand?

azul:
Looks like some use `++underline++` and some use `__underline__`. We currently convert `__hello__` to `**hello**` when loading the markdown in the editor. I would like to preserve the original markdown as much as possible.

Markdown has emphasis and strong emphasis. Most of the time emphasis is rendered as italics and strong emphasis is rendered as bold. I'm not sure about having two different ways to do strong emphasis based on what chars were used might be confusing with other editors that display both as the same.

juliushaertl:
> My reasoning for now would be if people ask for underline we can tell them the keyboard combination. If people paste underlined text it will be preserved. From my point of view that solves the main use cases. But I'd be okay with adding a button - maybe below the bold button like a menu to not make the toolbar even wider.

I'd agree on not having a button there. @jancborchardt What do you think?

> Looks like some use ++underline++ and some use `__underline__`. We currently convert __hello__ to **hello** when loading the markdown in the editor. I would like to preserve the original markdown as much as possible.
> Markdown has emphasis and strong emphasis. Most of the time emphasis is rendered as italics and strong emphasis is rendered as bold. I'm not sure about having two different ways to do strong emphasis based on what chars were used might be confusing with other editors that display both as the same.

Preserving the original markdown is actually a strong argument of just using __underline__ for this i think. That I totally missed in the previous discussion in the issue. This would mean we just render the two usages of emphasis differently depending on their source.

mejo-:
Great work, I like :blush: 

Though I'm not sure about the decision to not expose underlining in the menubar. In my opinion, hiding features and only making them available to "power users" is kind of an anti-pattern.

Also, in case you go with exposing it in the menubar, mentioning the keyboard shortcut in the tooltip would be super nice :blush:

jancborchardt:
> Though I'm not sure about the decision to not expose underlining in the menubar. In my opinion, hiding features and only making them available to "power users" is kind of an anti-pattern.

Yup, agree with that @mejo- – especially since to anyone who doesn’t know Markdown (a.k.a. majority of people ;), underlining is super normal, we should have it in the bar. The common order is: Bold, Italic, Underline, Strikethrough.

(A more likely candidate for hiding would be "Strikethrough" btw, which neither Google Docs nor Office 365 shows by default in the bar. But since we don’t have a menu bar, it’s fine to keep it.)

juliushaertl (approved):
Lets stick with the 4 underscores which in common mark would just be rendered as strong and we don't interfere with other syntax then

juliushaertl:
@azul I took the opportunity and added the menu item right away when testing it. Would be good to go from my side.

juliushaertl (changes requested):
Sorry, blocking again, tired to add a test case and it indeed is an issue with reading the file from markdown as we use markdown-it for that which converts `____test____` to `<b>test</b>`

juliushaertl:
We could probably introduce something similar to this for handling that https://github.com/outline/rich-markdown-editor/blob/f31ab5c0af38c7c1170c1c5328f5f8140673c674/src/rules/underlines.ts

mejo-:
Would be nice to get this added to the formatting help modal as well: https://github.com/nextcloud/text/blob/master/src/components/HelpModal.vue

max-nextcloud:
> Sorry, blocking again, tired to add a test case and it indeed is an issue with reading the file from markdown as we use markdown-it for that which converts `____test____` to `<b>test</b>`

I ended up going with the double `_` for underline. Seemed better than introducing a completely different syntax than everyone else. I also considered just using `<u>underlined</u>` - which would be the common mark compatible approach. But then we do not allow html in markdown and working around was less obvious - still might be an option though.

Benefits of just using `__`:
* more consistent with `**` from a users perspective
* keeps previously used markdown intact (does not change `__` to `**`).
* easier to type and more consistent as a shortcut.

So for tests i simply reused the common mark tests modifying the expected tag from `strong` to `u` if the markdown contained `__`. There were no tests mixing `__` and `**` - so for now it's working fine.

Issue #1925, opened 2021-11-02
Ensure js tests are run in CI


azul:
**Describe the bug**
Looks like we lost our js test runs here:
https://github.com/nextcloud/text/pull/1688/files#diff-e98936aa52a6dd7416e4296e9628456227d834f7245967383fd9ff80fd985dadL26-L28

**Expected behavior**
CI should run the js tests

azul:
Now Ci is showing the same failure i see locally:
https://github.com/nextcloud/text/runs/4077426188?check_suite_focus=true

PR #1926, opened 2021-11-02
ci: run js tests

azul:
* Resolves: #1925 
* Target version: master 

### Summary

Bring back js test in ci.

Will probably fail now because we broke them with updates in between.


azul:
We should probably backport this to stable23

juliushaertl (approved):

juliushaertl:
I just also double checked, stable22 and below are not affected

PR #1947, opened 2021-11-11
Update version on master

nickvergessen:


PVince81:
cypress failure related: `Cannot read properties of undefined (reading 'currentVersion')`

PR #1959, opened 2021-11-22
Hide menu and keep focus when using menubar popover items

juliushaertl:
When using any formatting action from the collapsed menubar, this change will make sure that the popover gets hidden and the focus stays within the document.


eneiluj (approved):
Code looks good. Works perfectly.

mejo- (approved):

PR #1960, opened 2021-11-23
fix issues on formatting bar

luka-nextcloud:

* Resolves: #1075
`Increase height of formatting bar to 50px, size of avatars to 44 (might fix the status bubble display issue too)`
`A little transparency and blur for more open feel`

* Target version: master 

### Summary
- Increase height of formatting bar to 50px, size of avatars to 44 (might fix the status bubble display issue too)
- A little transparency and blur for more open feel

image


juliushaertl:
Having a bit of text in it I'm no longer sure if we should actually go with the blur here, which we initially discussed with @jancborchardt. Feels a bit less out of place compared to the general Nextcloud design:

https://user-images.githubusercontent.com/3404133/143309782-2f951747-067c-4d0c-a61b-742321728f94.mov
 
@nimishavijay Maybe you also have some thoughts about that?

Other than that, there are a few further things that should be more aligned with the increase of the formatting buttons.

- The menubar itself should only be 50px in total (including the padding)

Screenshot 2021-11-24 at 21 40 11


- The avatars still seem to have a slightly different size and the user icon on the right some margin (the margin actually seems to be the reason that the menubar exceeds the 50px size)
Screenshot 2021-11-24 at 21 42 16

nimishavijay:
I think with text behind the formatting bar it is a bit confusing, because initially I couldn't tell that it was blurred text, I thought the colour of the bar changed to a weird grey. It looks really nice with images though.
I can also imagine it _might_ be an accessibility issue because of insufficient contrast.

I suggest we do something like add gradient from 100% opacity at the middle of the formatting bar to 0% at the bottom along with the blur, so it is not a hard stop but a softer gradient. What do you think?

luka-nextcloud:
@jancborchardt @juliushaertl What do you think about @nimishavijay 's suggestion?
I think gradient opacity would look better.

juliushaertl:
I don't see a way with CSS to fade out the blur effect (which itself also doesn't work in some browsers). @jancborchardt Do you have strong feeling about the blur at all? I feel the design is much cleaner and more structured without now looking at the recording another time.

jancborchardt:
> I don't see a way with CSS to fade out the blur effect (which itself also doesn't work in some browsers). @jancborchardt Do you have strong feeling about the blur at all? I feel the design is much cleaner and more structured without now looking at the recording another time.

Yeah, agree with you, don’t have a strong feeling about the blur – the current state looks better and less jarring. Similarly, as soon as the document is loaded, the header bar (where title and top right actions are in) should be fully opaque so the Nextcloud header does not shine through.

luka-nextcloud:
@juliushaertl Please check again.
1. Remove blur
2. The menubar itself should only be 50px in total (including the padding)
3. The avatars still seem to have a slightly different size and the user icon on the right some margin (the margin actually seems to be the reason that the menubar exceeds the 50px size)

max-nextcloud:
> @juliushaertl Please check again.
Since julius is out, i'll review.

Looks really goodat first sight. In particular scrolling workspaces without messing with the menubar is great.

> 1. Remove blur

:+1: 

> 2. The menubar itself should only be 50px in total (including the padding)

Confirmed in viewer. looks good.
In the workspace it seems to be 44 (3 + 38 + 3) because there's no outer `.menubar`. Seems fine to me.

> 3. The avatars still seem to have a slightly different size and the user icon on the right some margin (the margin actually seems to be the reason that the menubar exceeds the 50px size)

Confirmed. Changing the margin in SessionList.vue to `margin: 0px 6px;` centers the avatar. Reducing `top` of `.save-status` to `9px` (6 less than before) ensures the status is still centered.

I'll create a new commit on top of this to apply these changes.

src/components/SessionList.vue
```diff
@@ -177,19 +183,18 @@ export default {
 
 		.icon-more, .icon-group, .icon-settings-dark {
 			background-color: var(--color-background-dark);
-			width: 36px;
-			height: 36px;
-			margin: 6px 6px 6px 0px;
+			width: 44px;
+			height: 44px;
+			margin: 0px 6px 0px 0px;
```
eneiluj, line 188:
```suggestion
			margin: 0 6px 0 0;
```

src/components/SessionList.vue
```diff
@@ -177,19 +183,18 @@ export default {
...
 		}
 	}
 
 	.avatar-wrapper {
-		width: 32px;
-		height: 32px;
+		width: 42px;
+		height: 42px;
```
eneiluj, line 194:
```suggestion
		width: 44px;
		height: 44px;
```
Otherwise it's not vertically aligned with `.avatardiv.icon-group`.

eneiluj (changes requested):
1. The top/bottom margin of `.avatardiv.icon-group` is 0 and the one of `.avatar-wrapper` is 6px so they are not aligned:
ttt

As the menubar elements and the save-status are already aligned, we could just remove the margin of `.avatar-wrapper`:
``` css
.avatar-wrapper {
    margin: 0 -8px 0 0;
}
```

2. Why not generating avatars in 44x44px instead of 32x32px in SessionList's template?

3. Inline suggestions

jancborchardt:
> as soon as the document is loaded, the header bar (where title and top right actions are in) should be fully opaque so the Nextcloud header does not shine through.

@luka-nextcloud is this something you could adjust here too? It’s sort of separate but makes the formatting bar look off since the Nextcloud header shines through above.

max-nextcloud:
I rebased and addressed all changes proposed by @eneiluj .

max-nextcloud:
> > as soon as the document is loaded, the header bar (where title and top right actions are in) should be fully opaque so the Nextcloud header does not shine through.
> 
> @luka-nextcloud is this something you could adjust here too? It’s sort of separate but makes the formatting bar look off since the Nextcloud header shines through above.

My understanding is that the top header bar with hte nextcloud shining through is part of the viewer - not the text app. As far as i know we can only pick a light or a dark theme for the viewer.

max-nextcloud:
> 2\. Why not generating avatars in 44x44px instead of 32x32px in SessionList's template?

Yes... looks much better. Will do this.

eneiluj (approved):

PR #1961, opened 2021-11-23
Add emoji autocompletion (#987)

mejo-:
Adds emoji autocompletion by typing `:<search string>`.

It utilizes the emoji functions from nextcloud-vue[1] and the TipTap
suggestions extension.

[1] https://github.com/nextcloud/nextcloud-vue/pull/1474

## Screencasts

recording

* Resolves: #987 
* Target version: master


mejo-:
As discussed, I first looked into opportunities to outsource the emoji autocompletion into nextcloud-vue components. But unfortunately that's not really an option - at least I didn't see a good one.

The [nextcloud-vue implementation](https://github.com/nextcloud/nextcloud-vue/pull/1474) would be specific to the RichContenteditable component and our implementation utilizes the [TipTap Suggestions extension](https://v1.tiptap.dev/suggestions).

I'd be happy to learn about others' ideas on how to merge (parts of) the two approaches though :blush:

mejo-:
@nimishavijay and @jancborchardt feel free to drop some early feedback as well :blush:

src/components/EmojiSuggestions.vue
```diff
@@ -0,0 +1,176 @@
...
+		},
+	},
+	computed: {
+		hasResults() {
+			return this.filteredEmojis.length
+		},
+
+		showSuggestions() {
+			return this.emojiQuery || this.hasResults
+		},
```
eneiluj, line 79: For readability and to make sure we manipulate booleans here.
```suggestion
		hasResults() {
			return this.filteredEmojis.length > 0
		},

		hasEmojiQuery() {
			return this.emojiQuery !== null && this.emojiQuery !== ''
		},
		
		showSuggestions() {
			return this.hasEmojiQuery || this.hasResults
		},		
```

eneiluj (changes requested):
:rocket: Wow it is very nice already!

> The nextcloud-vue implementation would be specific to the RichContenteditable component and our implementation utilizes the TipTap Suggestions extension.

I guess it would require to implement something generic enough so it could be used in RichContenteditable and as a replacement of your EmojiSuggestions component.
I think it would be nicer to factorize the work on text suggestion.
I'll have a look at our Vue lib soonish.

When there is no suggestion, the TipTap emoji plugin sometimes provides an emojiRect with x==0 and y==0 which causes the suggestion box to jump to the top left corner. I didn't dig too deep but I feel this could be avoided with a few checks.

nimishavijay:
Looking super nice! Pretty much done looks-wise :D
Some questions I have: 
- does it auto complete on pressing tab or enter or both? Ideally it should be both 
- is the list of suggested emojis limited to a certain number? I'd say something around 5-6 would work
- do you think we need to impose a max width on mobile and ellipsise so it doesn't extend beyond the screen?

eneiluj:
@nimishavijay 

> * does it auto complete on pressing tab or enter or both? Ideally it should be both

The suggestion dropdown appears right after typing ":". You can then change the highlighted one with the arrows and select one with enter or by clicking it. Pressing the tab has no effect on the dropdown, it changes the focus to other stuff in the page.

> * is the list of suggested emojis limited to a certain number? I'd say something around 5-6 would work

Right now it's 5.

> * do you think we need to impose a max width on mobile and ellipsise so it doesn't extend beyond the screen?

Sounds good, yes.

src/components/EmojiSuggestions.vue
```diff
@@ -0,0 +1,176 @@
...
+					position.top = `${this.emojiRect.y}px`
+				} else {
+					// TODO: Fix y value for positioning on top
+					position.bottom = `${this.emojiRect.y}px`
+				}
+			} else {
+				position.top = '0px'
+			}
+			// TODO: Remove debugging statement
+			console.log(position) // eslint-disable-line no-console
```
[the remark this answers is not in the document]
mejo-: Ah, thanks for the hint! I always wanted to look into this - now you explained to me and I don't have to any longer <3

src/components/EmojiSuggestions.vue
mejo-: Full ack. Much cleaner that way :)

mejo-:
> When there is no suggestion, the TipTap emoji plugin sometimes provides an emojiRect with x==0 and y==0 which causes the suggestion box to jump to the top left corner. I didn't dig too deep but I feel this could be avoided with a few checks.

I digged a bit deeper here. The problem is that `getBoundingClientRect()` on the emojiSuggestion element returns positions (0, 0) for a short time in some cases. The only solution I could find so far is to cache the coordinates and to *not* update them in case of (0, 0): https://github.com/nextcloud/text/pull/1961/commits/b5feddf0500d069e77ec4740b6c94b1fabd8452a

This works well in my tests, but not sure if it's the smartest solution. Let me know what you think :grimacing:

mejo-:
>     * does it auto complete on pressing tab or enter or both? Ideally it should be both

Good point, I'll make it work on `<Tab>` as well.

>     * is the list of suggested emojis limited to a certain number? I'd say something around 5-6 would work

As @eneiluj said, it's limited to five for now (Github also shows five).

>     * do you think we need to impose a max width on mobile and ellipsise so it doesn't extend beyond the screen?

Good idea. What would be a good max-width on mobile?

juliushaertl:
> > When there is no suggestion, the TipTap emoji plugin sometimes provides an emojiRect with x==0 and y==0 which causes the suggestion box to jump to the top left corner. I didn't dig too deep but I feel this could be avoided with a few checks.
> 
> I digged a bit deeper here. The problem is that `getBoundingClientRect()` on the emojiSuggestion element returns positions (0, 0) for a short time in some cases. The only solution I could find so far is to cache the coordinates and to _not_ update them in case of (0, 0): [b5feddf](https://github.com/nextcloud/text/commit/b5feddf0500d069e77ec4740b6c94b1fabd8452a)
> 
> This works well in my tests, but not sure if it's the smartest solution. Let me know what you think grimacing

Seems good and a similar approach is used in https://github.com/ueberdosis/tiptap/blob/v1/examples/Components/Routes/Suggestions/index.vue#L230

mejo-:
>     * do you think we need to impose a max width on mobile and ellipsise so it doesn't extend beyond the screen?

I went with `max-width: 90wv` on mobile as that seemed like the most sensible value to me.

Also did some further CSS/positioning improvements, see https://github.com/nextcloud/text/pull/1961/commits/95e2583d523a936a41af0be1a09ed476ff7bede6 for details. I updated the screencast above accordingly.

mejo-:
> Supress further marks in Emoji nodes

So I looked into the problem of marks being applied within emoji search and came up with #1964, which is the proper fix here in my eyes. This also fixes the problem for inserting e.g. URLs with two underscores etc. Therefore I removed the quoted todo as it's out of scope for this PR.

azul:
I did not take a look at the code yet... but i am super excited to see this! Just wanted to share the excitement. And yes... i know there are emoji reactions for this... sometimes they are not enough though ;)

eneiluj:
> I digged a bit deeper here. The problem is that `getBoundingClientRect()` on the emojiSuggestion element returns positions (0, 0) for a short time in some cases. The only solution I could find so far is to cache the coordinates and to _not_ update them in case of (0, 0): [b5feddf](https://github.com/nextcloud/text/commit/b5feddf0500d069e77ec4740b6c94b1fabd8452a)
> 
> This works well in my tests, but not sure if it's the smartest solution. Let me know what you think grimacing

This fix makes sense. It now works very nicely :rocket:.

mejo-:
So I changed the implementation to depend on https://github.com/nextcloud/nextcloud-vue/pull/1474.

This brings unified emoji datasource, defaults and recents for EmojiPicker and the emoji autocompletion feature.

max-nextcloud:
@mejo- I've been trying to rebase this. It all worked fine until the last commit. There i get:
```
/home/max/code/nextcloud/server/apps/text/src/components/EditorWrapper.vue
  100:40  error  Unable to resolve path to module '@nextcloud/vue/dist/Functions/emoji'  import/no-unresolved
  100:40  error  "@nextcloud/vue/dist/Functions/emoji" is not found                      node/no-missing-import
```

To me this looks as if `nextcloud-vue` did not expose the `emoji` function. Looking that the node_modules folder there's also no such file:
```
max@ubuntu-21.10:~/code/nextcloud/text$ ls node_modules/@nextcloud/vue/dist/Functions/
usernameToColor.js  usernameToColor.js.map
```

I suspect this worked for you locally because of a npm link or so. Could you try building this without npm link to nextcloud-vue and with an `npm ci`?

eneiluj:
Same compilation error after bumping `@nextcloud/vue` to v4.3.0.
@mejo- Is `@nextcloud/vue/dist/Functions/emoji` supposed to be exposed in any release?

max-nextcloud:
> I'll take a look later, but in any case the emoji autocompletion in @nextcloud/vuejs didn't get released yet (will be released with 5.0.0) - so you'll have to build against a local checkout of the master branch using `npm link` for now. Did you try this?

Okay... cool. with npm link it works. I finished my rebase and pushed to keep this branch somewhat up to date.

max-nextcloud (approved):
I played with this some and it's really great.
There's one issue with combined emojis right now - but that's independent of the autocompletion and i will create a separate issue for it.

eneiluj (commented):
Awesome!
As far as I've seen, in the suggestion list, when there's no query, only the last used emoji is added in the end. It would be nice to have the 3 last used or the 3 most used. What do you think? Does it make sense? Is it even possible/easy to implement?

juliushaertl:
`@nextcloud/vue` 5.0.0 was merged https://github.com/nextcloud/text/pull/2178

mejo-:
> The only issue i ran into is that it's right now impossible to insert two emojis using autocompletion directly after each other. You need to put a space in between otherwise the dialog will not open.
> 
> For other keyboard shortcuts it makes sense not to be triggered within words. If you have `text_with_underscores` we don't want the `with` to become italic. However we're not converting `:heart:` to a heart emoji unless one presses tab or enter. So i think opening the emoji selector even when typing a word would be okay and would allow adding multiple emojis in a row or directly following a letterexclamation

An alternative would be to always append a space when adding an emoji. That's also, what others do. What do you think about that?

max-nextcloud:
> An alternative would be to always append a space when adding an emoji. That's also, what others do. What do you think about that?

That's a great idea. Avoids opening the emoji picker when ending a word with a colon like this:
Also should be fairly easy implementationwise. :D

mejo-:
I put some more work into the PR and think that it's ready to be merged now.

In particular, implementing automatic scrolling when navigating through the suggestion list was tricky and to be honest I'm a bit proud about the result :blush: https://github.com/nextcloud/text/pull/1961/commits/b05c62b863f2af022557fdf4f860b3903a11816e

Also, as discussed with @max-nextcloud, every inserted emoji now always gets a whitespace appended, to make it easy to insert several emojis in a row.

max-nextcloud (approved):
Code looks very clean. Will give it a try and then merge.

juliushaertl (approved):
Works like a charm and also nice and clean from a code perspective 👍

max-nextcloud:
Trying to fix the jest issue. I ran into something similar before.
- [ ] reproduce it localy
- [ ] figure out what its missing
- [ ] maybe add to transformIgnorePatterns?

max-nextcloud:
https://github.com/nextcloud/nextcloud-vue/pull/2521 has the upstream fix. We'll have to wait for a release of `@nextcloud/vue` to get this in.

PR #1962, opened 2021-11-23
Add remove link button

luka-nextcloud:

* Resolves: #1075
`"Remove link" action in the bar when a link is set`
* Target version: master 

### Summary
Added "Remove link" action in the bar when a link is set.

image


eneiluj (commented):
Nice and clean!

Removing a link only affects the selected text so if a sub part of the link text is selected and one hits "remove link", the rest is still a link. This makes sense but I wonder if users would expect the whole link to be removed.

What do you think?

How do you build the scripts? In your branch, when doing `npm ci; npm run build` I get 18 different files in `js/`.

luka-nextcloud:
@eneiluj @juliushaertl 
I guess, it would make more sense to remove the whole link.

I just ran `npm run build`.

luka-nextcloud:
@eneiluj @juliushaertl 
I've just checked again, it works exactly the same as the update link option. I think we need to update the behavior of the update link option as well or leave it as it is.

juliushaertl:
As discussed via chat, might need another rebase against latest master ;)

juliushaertl:
@luka-nextcloud Please make sure to add the original ticket where this was mentioned to the pull request description, so we have a back reference there. If it is an overview ticket just mention that it `contibutes to #ticketnumber` or something similar :)

Issue #1941, opened 2021-11-09
Document markdown syntax and keyboard shortcuts


mejo-:
The options to directly type markdown syntax and use keyboard shortcuts in order to format your text are currently hidden to new users. I wonder whether it would be a good thing to document these features and that way make them more accessible to users.

One option would be to describe markdown syntax and keyboard shortcuts (if applicable) in menubar icon tooltips. That's e.g. how the Github, Gitlab and Etherpad editors do it. I read in #1075 though that you decided to remove tooltips altogether from the menubar icons. What was the reasoning behind that?

Cc @jancborchardt :wave:

jancborchardt:
For keyboard shortcuts: Yep, let’s do that for sure! :) Putting it in the tooltips is a good call too, so like "Bold (Ctrl+B)" and "Underline (Ctrl+U)".

For Markdown syntax I’m less sure, what we could do is have one last entry with an `icon-info` which opens a modal explaining Markdown syntax quickly?

mejo-:
> For keyboard shortcuts: Yep, let’s do that for sure! :) Putting it in the tooltips is a good call too, so like "Bold (Ctrl+B)" and "Underline (Ctrl+U)".

I opened a PR to do this: #1963

mejo-:
So @nimishavijay and me brainstormed a bit how the formatting help modal could look like. That was great fun and @nimishavijay's input helped a lot to develop a proper concept here :ok_hand: 

Idea is basically to have
1. Some short introductionary text that explains the fact that markdown formatting syntax exists
2. A table with styles, their markdown syntax and (if not on mobile) keyboard shortcuts
3. A list of examples that show how the styles are actually rendered

For the examples we ideally reuse existing CSS so we don't have to update the examples if we ever change something in the text styling. Another reason is that custom theming will apply automatically that way.

Please excuse my ugly drawing, but that's a first mockup:

![formatting_help_mockup](https://user-images.githubusercontent.com/3582805/144244266-6bedf436-9a81-487b-9d14-204ff2b50d7d.jpeg)

mejo-:
First ugly screenshots:

![Screenshot 2021-12-07 at 15-23-43 Readme md - Nextcloud](https://user-images.githubusercontent.com/3582805/145046580-db8fcb59-863a-46e7-a686-ad86a836ee08.png)

[...]

PR #1963, opened 2021-11-25
Mention keyboard shortcuts in tooltips (#1941)

mejo-:
Doesn't show the shortcuts on mobile (they don't apply there).

recording

* Target version: master


azul:
I'm unsure about this one. I see github is doing the same and i see the point of making keyboard shortcuts discoverable. At the same time it's always there for everyone and clutters the ui a bit. I try to use the markdown shortcuts such as `_italic_` instead of ctrl+i myself.

src/mixins/menubar.js
```diff
@@ -115,6 +127,7 @@ export default [
 	},
 	{
 		label: t('text', 'Unordered list'),
+		keyboard: t('text', 'Shift') + '+' + t('text', 'Ctrl') + '+8',
```
azul, line 130: I feel like constructing these strings could be moved to a helper function. Something like:
```
keyboardShortcut('8', shift: true, ctrl: true)
```

azul (approved):
Codewise looks good to me. I have one small comment I will add in line.

mejo-:
> I'm unsure about this one. I see github is doing the same and i see the point of making keyboard shortcuts discoverable. At the same time it's always there for everyone and clutters the ui a bit. I try to use the markdown shortcuts such as `_italic_` instead of ctrl+i myself.

I started looking into this after I got the feedback that people didn't know at all about formatting shortcuts in Nextcloud Text. That's why I think it's a good thing to make it obvious.

As discussed in #1941, the idea is to mention the keyboard shortcuts in toolbar tooltips in a first step and add a motion with some explanatory text and details about the markdown shortcuts there.

I took a look at several web editors (Google Docs, Github, Gitlab, Discourse, Etherpad) and they all mention keyboard shortcuts in their tooltips :blush:

jancborchardt (changes requested):
The tooltips are good as discussed, since they only appear after some time. :)

But in the headings action menu and also for blockquote, having the shortcut directly in the menu is indeed a bit cluttered as @azul mentions. Is it possible to put the shortcut info in a tooltip there too?

mejo-:
> But in the headings action menu and also for blockquote, having the shortcut directly in the menu is indeed a bit cluttered as @azul mentions. Is it possible to put the shortcut info in a tooltip there too?

Unfortunately neither `ActionButton` nor `PopoverMenuItem` support setting the `title` property for `button` element so far. I opened a PR in nextcloud-vue to fix that: https://github.com/nextcloud/nextcloud-vue/pull/2375. With this PR being merged (and released), it's possible to set tooltips for actions in submenus:

recording

mejo-:
After some further discussion and brainstorming with @skjnldsv, I went with using `v-tooltip` instead, which is the prefered way for tooltips in Nextcloud anyways.

Also, @skjnldsv made the valid point that tooltips in the headings submenu would be confusing (basically tooltip inside a tooltip). Therefore I went with `Heading (Ctrl+Shift+1…6)` as a generic headings tooltip and no tooltips for the concrete levels. What do you think about it?

See the updated gif above.

src/mixins/menubar.js
mejo-: Did this now, thanks to pair programming with @azul  :)

nimishavijay (approved):
Looks good design wise! Agreed with using `Ctrl+Shift+1...6` because otherwise it would be repeated many times in the submenu

mejo-:
@jancborchardt could you take a quick look? Github tells me that you requested changes and you probably have to approve before Github lets us merge :)

PR #1964, opened 2021-11-25
Update mark input/paste rules to tiptap v2 regular expressions

mejo-:
The most important change is that a mark is added only if you are at
the beginning of a block or there is a space before it.

This especially fixes accidently formatting parts of a string as italic
when it contains two underscores (often true in URLs or emoji names).

This change can be reverted when we migrated to tiptap v2, but it's
a low hanging fix for an annoying bug, and the tiptap v2 migration might
take a bit longer.

See Philipp Kuehns comment at https://discuss.prosemirror.net/t/4230/3.

* Target version: master


eneiluj (approved):
This is nice!

It feels a bit weird when you want a sub part of a word to be bold. You type `un**cann**y` and it does not get formatted...
But I guess this is a very uncommon use case.

juliushaertl (approved):

mejo-:
This probably should also be backported to stable23 and stable22 at least, no? @juliushaertl

juliushaertl:
Yes, definitely.

PR #1966, opened 2021-11-26
Add a regular link if a file was inserted instead of an image

luka-nextcloud:

* Resolves: #1075
`If a file is inserted just add a regular link`

* Target version: master 

### Summary
If a file is inserted instead of an image, just add a regular link.
image


jancborchardt (approved):
Very cool, looks nice! :)

juliushaertl:
Maybe we could even not insert an image into the markdown file for non-image mimetypes but just a link `[filename](path-to-file)`

https://github.com/nextcloud/text/blob/93e3a0905f6b38a3201a0663bbb4ceff75b25936/src/components/MenuBar.vue#L290

The fileInfo should already contain the mimetype, so it would just require to not only pass the image command in https://github.com/nextcloud/text/blob/93e3a0905f6b38a3201a0663bbb4ceff75b25936/src/components/MenuBar.vue#L180 but all or at least also the link command.

eneiluj:
If the only way to insert a non-image file is via the "insert image" button then I think it's better to restrict the mimetypes in the filepicker dialog and avoid this use case.

If inserting a non-image file is a feature we want, it could be clearer to trigger it with another menu button for example (although we already have a lot of them).

I also like the solution mentioned by @juliushaertl. We could trigger a different action depending on the selected file mimetype. But then we need to make it explicit in the UI that it's possible to insert all kind of files and not just images.

mejo- (approved):
Works pretty well, thank you :)

mejo-:
> If inserting a non-image file is a feature we want, it could be clearer to trigger it with another menu button for example (although we already have a lot of them).
> 
> I also like the solution mentioned by @juliushaertl. We could trigger a different action depending on the selected file mimetype. But then we need to make it explicit in the UI that it's possible to insert all kind of files and not just images.

In general, I agree with you both. But given that inserting non-image files through the "Insert image" action right now, I'd consider it a regression if we dropped support for it. And as long as we don't have a proper "Link file" dialog, I think that @luka-nextcloud's change is an improvement over the current situation. Don't you think so?

juliushaertl:
Yes, I'm also fine to get this in and handle the inserting separately :+1: Nice work @luka-nextcloud

juliushaertl (approved):

juliushaertl:
Mind to rebase and push the updated js bundles again? :see_no_evil:

Issue #903, opened 2020-06-23
ToDo list - Add checkboxes to all selected lines


themedleb:
**Describe the bug**
As the title says, checkboxes should be added to all selected lines, not just one.

**To Reproduce**
![This is what happens](https://raw.githubusercontent.com/themedleb/Uploads/master/Actual.gif)

**Expected behavior**
![This is what should happen instead](https://raw.githubusercontent.com/themedleb/Uploads/master/Supposed.gif)

<details>
<summary>Server details</summary>

**Text app version:** 3.0.1

**Nextcloud version:** 19.0.0

</details>

susnux:
Already fixed with #1973 (master and NC24)

PR #1973, opened 2021-12-01
Fix only the first item gets tasklist-ified issue

luka-nextcloud:

* Resolves: #1075 
`If you mark a whole list and switch it to a task list, only the first item gets tasklist-ified`

* Target version: master 

### Summary
Fixed the issue: If you mark a whole list and switch it to a task list, only the first item gets tasklist-ified
image


azul:
Code looks good to me. I hope i find some time to try it out.

src/nodes/ListItem.js
```diff
@@ -128,12 +128,14 @@ export default class ListItem extends TiptapListItem {
...
+						if (node.type === schema.nodes.list_item) {
+							tr.setNodeMarkup(pos, node.type, { type: parentList.node.attrs.type === TYPES.CHECKBOX ? TYPES.BULLET : TYPES.CHECKBOX })
+						}
+					})
 					tr.scrollIntoView()
 
-					if (dispatch) {
-						dispatch(tr)
-					}
+					if (dispatch) dispatch(tr)
```
mejo-, line 138: Not sure about our coding style convention here, I thought that we prefer to always use curly brackets for better readability. But I might be wrong :)

mejo- (commented):
Apart from my minor coding style comment, looks and works good. Well done :blush:

mejo-:
This also would fix #903. Just a notice that we don't forget to close the bugeport afterwards.

src/nodes/ListItem.js
```diff
@@ -128,12 +128,14 @@ export default class ListItem extends TiptapListItem {
 						return false
 					}
 
-					tr.setNodeMarkup(parentList.pos, schema.nodes.list_item, { type: parentList.node.attrs.type === TYPES.CHECKBOX ? TYPES.BULLET : TYPES.CHECKBOX })
+					state.doc.nodesBetween(selection.from, selection.to, (node, pos) => {
```
juliushaertl, line 131:
```suggestion
					tr.doc.nodesBetween(tr.selection.from, tr.selection.to, (node, pos) => {
```

Shouldn't this operate on the transaction instead of the original document?

juliushaertl: With that it would also fix #903 right away it seems, otherwise when you select a block of paragraph nodes and trigger the checklist, they will only get converted to a regular bullet list, as the `toggleList` call from above is not yet dispatched to the original `state.doc`

src/nodes/ListItem.js
juliushaertl: Yes, I'd prefer that, though not enforced through our eslint rules currently. 

https://github.com/nextcloud/eslint-config/blob/master/index.js
https://eslint.org/docs/rules/curly

eneiluj (commented):
Works fine, nice.

As mentioned by @juliushaertl, It still feels weird that when selecting multiple lines of raw text, it requires 2 clicks on the "todo list" button to actually get checkboxes.

mejo-:
@luka-nextcloud: seems like some file called `patch) {` somehow sneaked itself into your commit :wink:

mejo- (approved):
Works like a charm :ok_hand:

mejo-:
TODO: backports to stable21 and stable22 need to be done manually.

Issue #1906, opened 2021-10-21
Migrate to tiptap v2


juliusknorr:
Might need some additional evaluation to see what we actually need to change, especially with the custom marks/blocks and the collaboration/annotation plugins, but i think we should at some point consider an upgrade.

https://tiptap.dev/overview/upgrade-guide

azul:
As a start i will summarize considerations from the last time we talked about this.

## Why upgrade:
* stay up to date with our dependencies
* new, cleaner codebase
* benefit from new features and plugins
* longer term compatibility for things we develop ourselves
* Collaborative editing is fairly slow right now. Faster sync times of doc status would be nice.

## Challanges with the Collaboration plugin
* We need a way to sync clients that can be deployed on a plain LAMP stack.
* WebRTC might not work because of lack of turn server in certain network setups
* We need to save the state of the document to the file storage
* Optionally improve performance with non LAMP addition

## Possible approaches

### Forward port the old collaboration plugin to tiptap2

We have a working collaboration plugin for tiptap1 - forward port it to tiptap2.
Would be the closest to what we have so far.

pros:
* no need to change server code
* should be pretty straight forward
* maybe reusible for other projects in a similar situation
* known to work on LAMP stack
* Would work in all deployments
* Could be used as a fallback

cons:
* cannot use tiptap2 plugin as is
* not possible to combine with webRTC
* no performance improvements
* no Y.js merges

### PHP implementation of Y.js server side

pros:
* easy to integrate with nextcloud file storage
* Would work in all deployments

cons:
* not sure if this is possible at all
* Might be cumbersome to write performant async code in php

### Rust implementation of Y.js as part of the notify_push app

pros:
* Rust (type safety, high performance, async primitives)
* Rust implementation of Y.js is on the way
* push_notify is already written in rust and promoted as "high performance backend"

cons:
* Requires fallback for LAMP deployment
* Maybe harder to integrate with storage (but notify_push already connects to the storage)

### Deploy Y.js server

pros:
* No need to (partly) develop a Y.js server.
* easy to deploy optionally and containerize (i guess)

cons:
* Requires fallback for LAMP deployment
* We'd still need a way to store state to the files backend

PR #1942, opened 2021-11-09
Bump highlight.js from 10.7.3 to 11.3.1

azul:
Use path to the node_modules folder to work around
webpack/webpack#13865

Bumps [highlight.js](https://github.com/highlightjs/highlight.js) from 10.7.3 to 11.3.1.
- [Release notes](https://github.com/highlightjs/highlight.js/releases)
- [Changelog](https://github.com/highlightjs/highlight.js/blob/main/CHANGES.md)
- [Commits](https://github.com/highlightjs/highlight.js/compare/10.7.3...11.3.1)

---
updated-dependencies:
- dependency-name: highlight.js
  dependency-type: direct:production
  update-type: version-update:semver-major
...

Signed-off-by: dependabot[bot] <support@github.com>

* Resolves: # 
* Target version: master

PR #1981, opened 2021-12-02
WIP: Upgrade/tiptap2

azul:
* Resolves: #1906 
* Target version: master 

### Summary

This is still work in progress.

- [ ] update dependencies
- [ ] update imports
- [ ] migrate marks
- [ ] port collaboration extension
- [ ] integrate ported collaboration extension
- [ ] migrate nodes
- [ ] migrate extensions


juliushaertl:
FYI I added  plain text editing / syntax highlight to the checklist as that is something that has likely changed but we might be able to tackle https://github.com/nextcloud/text/pull/1942#issuecomment-965010720 with that.

eneiluj:
I think the test failure (while testing images features) is caused by a session sync happening after the user has been deleted. It does not happen all the time.
My bad. I'm on it.

eneiluj:
#2167 fixes the Cypress image test.

eneiluj (approved):
Tried all aspects of the description task list, everything works fine and looks the same as before.
I didn't deeply check the code but lgtm.

It's incredible we don't loose anything with this tiptap upgrade.
Impressive work!

mejo-:
The keyboard shortcuts for strikethrough changed upstream (<kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>X</kbd> instead of <kbd>Ctrl</kbd>-<kbd>D</kbd>). We either have to overwrite it or (imho better) keep it the upstream way and update our documentation.

max-nextcloud:
> I think the test failure (while testing images features) is caused by a session sync happening after the user has been deleted. It does not happen all the time. My bad. I'm on it.

Hehe... Actually this caught an issue in the error handling. Before we were catching errors in the PollingBackend like this:
```
axios.post(...)
  .then(handleResponse)
  .catch(handleErrors)
```
This caused exceptions in handleResponse to be caught by the catch that was meant for network issues.
I had changed this to:
```
axios.post(...)
  .catch(handleErrors)
  .then(handleResponse)
```

Now when errors happened they would be handled by `handleErrors` and then `handleResponse` was triggered with the return value `handleErrors` (undefined). So that caused the failing cypress test we saw.

So in the end i changed this to:
```
axios.post(...)
  .then(handleResponse, handleErrors)
```
This will either handle the response or the errors - but not both.

mejo-:
> The keyboard shortcuts for strikethrough changed upstream (Ctrl-Shift-X instead of Ctrl-D). We either have to overwrite it or (imho better) keep it the upstream way and update our documentation.

Same is true for shortcuts for headings (<kbd>Ctrl</kbd>-<kbd>Alt</kbd>-<kbd>[1-6]</kbd> instead of <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>[1-6]</kbd>).

max-nextcloud:
> The keyboard shortcuts for strikethrough changed upstream (Ctrl-Shift-X instead of Ctrl-D). We either have to overwrite it or (imho better) keep it the upstream way and update our documentation.

Ctrl-D is creating a bookmark in firefox by default. I don't think we should overwrite that. I don't use it - but seems to be a very core browser functionality.

In libreoffice ctrl-D is double underline. Strike through has no keyboard shortcut by default.
So i'm tempted to either go with ctrl+shift+X or no keyboard short cut at all.

mejo-:
> Ctrl-D is creating a bookmark in firefox by default. I don't think we should overwrite that. I don't use it - but seems to be a very core browser functionality.
> 
> In libreoffice ctrl-D is double underline. Strike through has no keyboard shortcut by default. So i'm tempted to either go with ctrl+shift+X or no keyboard short cut at all.

Thanks for further looking into this. I can imagine that's the reason why TipTap switched from Ctrl-D to Ctrl-Shift-X and I think that we should follow. So only the shortcut documentation in `src/components/MenuBar.vue` and `src/components/HelpModal.vue` needs to be updated.

juliushaertl (approved):
Awesome work 👍 Lets get this in

PR #1993, opened 2021-12-06
[stable22] Fix scss deprecated syntax

eneiluj:
This makes the compilation fail in some contexts.


juliushaertl:
Pushed a commit as the package.json allowed using node 16 and then there is no prebuilt libsass available for that node version and the compile command fails with that. Remaining question is why the local build is different from the one in the Node CI / build action, but we can try to compare when the compile action commited the expected result. Let's see

juliushaertl:
At least CI is stable with that but the local build is still different even with the exact same node/npm versions.  Not sure why though, the diff is rather odd with just some mapping hash changes it seems, using `git diff --text --word-diff=porcelain --word-diff-regex=.`

mejo- (approved):

PR #2001, opened 2021-12-07
Add formatting help modal (Fixes #1941)

### LINKED ISSUE #1941: Document markdown syntax and keyboard shortcuts

mejo-:
## Desktop

Full screen:
Screenshot 2021-12-14 at 16-00-27 Keyboard Shortcuts md - Nextcloud

Toolbar on wide screen:
2021-12-14T13:42:15,175191977+01:00

Toolbar on narrow screen:
Screenshot 2021-12-14 at 16-01-13 Keyboard Shortcuts md - Nextcloud

## Mobile

Screenshot 2021-12-14 at 16-00-08 Keyboard Shortcuts md - Nextcloud

Toolbar:
Screenshot 2021-12-14 at 15-59-31 Keyboard Shortcuts md - Nextcloud

* Target version: master


mejo-:
> How does it look like on very small display sizes? E.g. 360px wide?

Good question :) See the updated screenshot in the PR description above. It's not super smooth there, you have to scroll from left to right to read the full syntax.

Do you have a good idea how to improve the mobile experience?

szaimen:
> Do you have a good idea how to improve the mobile experience?

Maybe hide the keyboard shortcuts column and make the modal full-screen on such small screens?

mejo-:
> Maybe hide the keyboard shortcuts column and make the modal full-screen on such small screens?

Keyboard shortcuts are [already hidden on mobile](https://github.com/nextcloud/text/pull/2001/files#diff-e49d874caf205cd10782af6d3215331accd0e17b9df6068bfdaa72f6d48e7413R16) :blush: 

And [modals alread have full width on mobile per default](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L758-L770). The non-used space on both sides seems to [come from the invisible prev/next elements](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L680). If I override their width to 0, the button to close the modal is no longer visible :disappointed:

szaimen:
> The non-used space on both sides seems to [come from the invisible prev/next elements](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L680). If I override their width to 0, the button to close the modal is no longer visible 😞

Strange!
Would maybe the following work?
```
	.prev,
	.next { display:none !important }
```

jancborchardt:
Looks very nice now! :) Only thing: Would be better to put the button as last item of the fornatting bar instead of next to sharing. That way on narrower viewports it doesnt take a space but gets ellipsized away, as it also isnt always needed. :)

azul:
Just a minor note: The description on mobile still talks about keyboard shortcuts.
I'd also remove the 'experience'. 'Speed up your writing with simple shortcuts' is enough i think.

How about using the icons instead of words for the style in mobile? That would also serve as an explaination for the icons. Something like this:
|style| syntax|
|------|------|
| **B** | `**Bold**` |
| *I* | `*Italic*` |
| ~~S~~ | `~~Strikethrough~~` |
| H1 | `# largest heading` |
| H6 | `###### small heading` |

This way people would also know it's the same thing as that button.

mejo-:
> Looks very nice now! :) Only thing: Would be better to put the button as last item of the fornatting bar instead of next to sharing. That way on narrower viewports it doesnt take a space but gets ellipsized away, as it also isnt always needed. :)

Agreed, done so now:

2021-12-13T18:42:43,344139281+01:00

szaimen:
> Would maybe the following work?
> 
> ```
> 	.prev,
> 	.next { display:none !important }
> ```

So I suppose this doesn't work either?

mejo-:
> > The non-used space on both sides seems to [come from the invisible prev/next elements](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L680). If I override their width to 0, the button to close the modal is no longer visible disappointed
> 
> Strange! Would maybe the following work?
> 
> ```
> 	.prev,
> 	.next { display:none !important }
> ```

Yep, that indeed works, but as written the button to close the modal is hidden in this case (it's not really hidden, but it's white on white). Also, being able to click into the prev/next areas for closing the modal is no longer possible.

Do you have a good idea how to tackle this? :thinking:

szaimen:
> Do you have a good idea how to tackle this? 🤔

I think it will be already resolved with vue 5.0 due to my changes here: https://github.com/nextcloud/nextcloud-vue/pull/2126/files#diff-907877e92ada6bf9340562928a4046f5a937f4985041a4f2e3981313538f8f5dR679-R697
Maybe this could be of some inspiration?

mejo-:
> > Do you have a good idea how to tackle this? thinking
> 
> I think it will be already resolved with vue 5.0 due to my changes here: https://github.com/nextcloud/nextcloud-vue/pull/2126/files#diff-907877e92ada6bf9340562928a4046f5a937f4985041a4f2e3981313538f8f5dR679-R697 Maybe this could be of some inspiration?

Uh nice, that helped a lot! Thanks @szaimen :blush:

I'm now setting `position: absolute` and `top: 50px` on mobile, see my last commit: https://github.com/nextcloud/text/pull/2001/commits/10488842f93acc5aa9cc68df5d0d75d71c4bc69a

I updated the screenshots above.

mejo-:
> This looks great! I particularly like the i next to the save indicator as the trigger.
> 
> Just a minor note: The description on mobile still talks about keyboard shortcuts. I'd also remove the 'experience'. 'Speed up your writing with simple shortcuts' is enough i think.
> 
> How about using the icons instead of words for the style in mobile? That would also serve as an explaination for the icons. Something like this (with the icons from the menu bar on the left obviously):
> style 	syntax
> **B** 	`**Bold**`
> _I_ 	`*Italic*`
> ~S~ 	`~~Strikethrough~~`
> H1 	`# largest heading`
> H6 	`###### small heading`
> 
> This way people would also know it's the same thing as that button.

Nice idea! I gave it a try, but found the written style ("Bold", ...) to be much clearer. Now that all content fits on mobile screens (at least for english), I'd prefer to keep it that way.

src/components/HelpModal.vue
```diff
@@ -196,15 +196,28 @@ export default {
...
+			padding: 30px 40px 20px;
+			user-select: text;
+		}
+
+		// TODO: can be removed once migrated to @nextcloud-vue >= 5.0
+		// Make modal full screen on mobile
+		@media only screen and (max-width: 512px) {
+			.modal-container {
+				position: absolute;
+				top: 50px;
```
szaimen, line 211: if you do this, I think you will also need to modify the heigth to substract the 50px at the top from the modal size. E.g.:
```css
heigth: calc(100vh - 50px);
```

mejo-: Indeed, good catch :blush: Fixed now (I used `height: calc(100% - var(--header-height));` just like in nextcloud-vue upstream).

jancborchardt (approved):
Except the "Show formatting help" → "Formatting help" as mentioned @mejo- this is super nice work design-wise! :)

juliushaertl (approved):
Very nice :+1: Good to go from my side

PR #2002, opened 2021-12-07
Fix landscape image issue

luka-nextcloud:

* Resolves: #1075
`Possibly remove max-height for landscape pictures (ref Jos’ screenshot issue when zooming)`

* Target version: master 

### Summary
Fix landscape pictures position issue.
image
image


jancborchardt:
The image still has a rather small max-height though, as seen from the padding on the left and right?

The images could get a `max-height: 100vh;` that way they would always fit the screen/viewport height.

luka-nextcloud:
@jancborchardt 
I've checked again, the light green color is the author's background that we need to leave as it is.

src/nodes/ImageView.vue
```diff
@@ -260,12 +260,15 @@ export default {
 
 	.image__view {
 		text-align: center;
-
-		.image__main {
-			max-height: 40vh;
+		img {
+			width: 100%;
```
juliushaertl, line 264:
```suggestion
			max-width: 100%;
```

Should still be max-width, otherwise portrait images would be stretched.

juliushaertl:
@jancborchardt The max-height was picked deliberately iirc also in discussion with you at some point to always make the image fit into the visible area. I'd be fine to drop, just want to double check that this is fine with you then. 100vh would not take into account that there is also a menubar visible, so ideally it should be `calc(100vh - 100px)` to take the size without header + menubar

jancborchardt:
> @jancborchardt The max-height was picked deliberately iirc also in discussion with you at some point to always make the image fit into the visible area. I'd be fine to drop, just want to double check that this is fine with you then. 100vh would not take into account that there is also a menubar visible, so ideally it should be `calc(100vh - 100px)` to take the size without header + menubar

No you’re right, the max-height should be kept, ideally with the value you propose. :)

src/nodes/ImageView.vue
```diff
@@ -260,12 +260,15 @@ export default {
...
 		}
 	}
 
+	.image__main {
+		max-height: 80vh;
```
juliushaertl, line 269:
```suggestion
		max-height: calc(100vh - 50px - 50px);
```

juliushaertl:
@luka-nextcloud THen with my two code suggestions this would be good :+1:

luka-nextcloud:
@juliushaertl Fixed :)

juliushaertl (approved):

PR #1903, opened 2021-10-20
Position menububble relative to content-wrapper

azul:
* Resolves: #1903 
* Target version: master 

### Summary

Add a `position: relative` to `content-wrapper`
to fix misalignment of the menububble.

The menububble has an absolute position
which is calculated by tiptap
based on the position of the highlighted text.

It is styled with `left` and `bottom` attributes.
These are relative to the position of the `.content-wrapper`.

PR #2019, opened 2021-12-20
Azul/rebase stable22 fix 1677 menububble position

azul:
* Backport of #1903 
* Target version: stable22

This backports the fix as is in stable23 and master.

This fix broke the display of the heading menu (it's not showing above the files in rich workspaces anymore).
We'll furthermore always want to display the menu below the selected line on mobile so it's not overlaying with copy and paste menus.
Will need to fix this in another pr... but for now get stable22 into the state of current development.


azul:
#2020 to track the regression.

danxuliu (approved):
Tested and works* :+1: 

* Fixes what it should fix, breaks what it was expected to break :-)

azul:
The build failure is unrelated. Will fix that elsewhere.

Issue #2020, opened 2021-12-20
Heading menu is getting cut off again


azul:
**Describe the bug**
in the rich workspace the menu for headings is getting cut off again.

This was fixed in #1821 and broken again by #1903.

We'll want to have the menu bubble positioned properly AND the heading menu overflow the files list if the doc is small.
Or we just expand the editor so the menu always fits.

In addition we might want to always display the link menu below the selected text on mobile to make sure it does not overlap with copy and paste native menus.

PR #2021, opened 2021-12-21
fix: use stable22 branch for cypress tests

azul:
* Resolves: CI failures in stable22 branch.
* Target version: stable22

### Summary

The server image will use the BRANCH environment variable
to determine which branch of the server to checkout:
https://github.com/nextcloud/docker-ci/blob/master/server/initnc.sh#L3-L12


azul:
We were effectively using the latest master which led to test failures as it has a rewritten API for injecting app scripts.
For a failing test see #2020

mejo- (approved):

PR #2022, opened 2021-12-21
fix: use stable21 branch for cypress tests

azul:
* Resolves: ci build failures on stable21 branch
* Target version: stable21

### Summary

The server image will use the BRANCH environment variable
to determine which branch of the server to checkout:
https://github.com/nextcloud/docker-ci/blob/master/server/initnc.sh#L3-L12

So we were effectively using the latest master which led to test failures as it has a rewritten API for injecting app scripts.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -18,9 +18,17 @@ jobs:
         # containers: [1, 2, 3]
         php-versions: [ '7.4' ]
         databases: [ 'sqlite' ]
-        server-versions: [ 'master' ]
+        server-versions: [ 'stable21' ]
 
     steps:
+      - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


azul:
see also #2021

mejo- (approved):

azul:
Also include a commit from stable22 that makes sure we actually use the specified node version.

azul:
@juliushaertl I am not entirely sure if including the change to build the viewer in testing mode should be backported. I used the original commit because i felt it would increase consistency accross branches.

azul:
> @juliushaertl I am not entirely sure if including the change to build the viewer in testing mode should be backported. I used the original commit because i felt it would increase consistency accross branches.

Okay... it's causing the build to fail... So i guess i will just make sure we use the propper node version and remain consistent with what we had with stable21.

PR #2026, opened 2021-12-21
Re-enable cypress records

juliushaertl:
From previous discussions having the record key in the public org is fine since all that this would allow is to push test results.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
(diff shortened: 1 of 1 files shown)


juliushaertl:
https://dashboard.cypress.io/projects/hx9gqy/runs/2427/specs?utm_source=github

mejo- (approved):

skjnldsv (approved):

skjnldsv:
Btw, I updated the cypress workflow if you want to adapt from it :wink: https://github.com/nextcloud/viewer/blob/master/.github/workflows/cypress.yml

PR #2027, opened 2021-12-22
update offline state message when saving

luka-nextcloud:

* Resolves: # 
`Offline state: "Offline, changes will be saved when online"`

* Target version: master 

### Summary
Add a status message for offline state.


src/components/EditorWrapper.vue
```diff
@@ -182,7 +182,7 @@ export default {
 			return this.$store.state.showAuthorAnnotations
 		},
 		lastSavedStatus() {
-			return this.dirtyStateIndicator ? t('text', 'Saving …') : t('text', 'Saved')
+			return this.hasConnectionIssue ? t('text', 'Offline, changes will be saved when online') : this.dirtyStateIndicator ? t('text', 'Saving …') : t('text', 'Saved')
```
mejo-, line 185: Great improvement! Maybe the wording could be tweaked a bit:

```suggestion
			return this.hasConnectionIssue ? t('text', 'Connection lost. Will be saved once connection is restored.') : this.dirtyStateIndicator ? t('text', 'Saving …') : t('text', 'Saved')
```

mejo- (changes requested):

src/components/EditorWrapper.vue
luka-nextcloud: @jancborchardt What do you think about the message?

juliushaertl, line 185: Small suggest to improve readability of the code:

```suggestion
			if (this.hasConnectionIssue) {
				return t('text', 'Connection lost. Will be saved once connection is restored.')
			}
			return this.dirtyStateIndicator ? t('text', 'Saving …') : t('text', 'Saved')
```

Otherwise good to go with the new wording from my side.

jancborchardt: Hmm, I think offline and online are a bit less technical and more commonly understood terms, so I'd keep them?

mejo-: Ok, no hard feelings here :blush: Funny sidefact: I thought that 'Connection lost' would be less technical and more accurate at the same time, that's why I suggested it :grimacing:

mejo-: Now the logic to check for `this.hasConnectionIssue` is doubled, looks like that was accidently @luka-nextcloud  :grimacing:

luka-nextcloud:
@mejo- Fixed :)

src/components/EditorWrapper.vue
```diff
@@ -755,4 +758,14 @@ export default {
...
 		height: 100%;
 	}
+
+	@media only screen and (max-width: 600px) {
+		.connection-error {
+			visibility: hidden;
+		}
+		.connection-error:after {
+			visibility: visible;
+			content: 'Offline';
```
juliushaertl, line 768: That won't work with translations unfortunately, but you can just return the translated text of Offline within the lastSavedStatus method by using the isMobile mixin https://github.com/nextcloud/text/blob/5eb7b205b3aed1f1e6fd5bbec41c947df46d4f39/src/mixins/isMobile.js#L23

mejo- (approved):

eneiluj:
@luka-nextcloud You might have launched `npm install` instead of `npm ci` and then you included `package-lock.json`. Could you remove those changes and force push to this branch?

IMO a simple way to do this would be:
``` bash
# while being in your branch with a clean status
# this will cancel the last commit but keep the changes
git reset HEAD~
# then this will get rid of the changes to package-lock.json
git checkout package-lock.json
git commit -a -m "update offline state message when saving"
git push origin fix/offline-state-issue -f
```

luka-nextcloud:
Hmm, it looks strange - let me check and try again.

Issue #1974, opened 2021-12-01
(un)ordered list button not effective when the content is empty


eneiluj:
Hitting the ordered list or unordered list button does not add a first bullet point (or number) while there is no content and this is the first action.
Once there has been some edition, it works fine, even efter deleting all the content.

This happens in the browser and in the Android client.

**To Reproduce**
* Create a new markdown file
* Hit the (un)ordered list button => nothing happens
* Type some text and delete it to get an empty file again
* Hit the (un)ordered list => it works

mejo-:
Actually this seems to be the case with any action from the toolbar. Checklist, blockquote, code block, etc. all don't work on a new document without any content. Using the markdown formatting syntax on the other hand works.

Seems like something is not fully initiallized before the document has content :thinking: 

It also happens when opening an empty but existing markdown file. So it's not related to creating a new file, but more to the file not having content yet.

Nothing relevant is logged to the JS console.

mejo-:
Or probably there's no prosemirror node yet to align the formatting to/with :thinking:

mejo-:
I did some further testing but didn't find the reason for this bug yet.

* reproducible if Text gets loaded by Viewer app
* reproducible for public view of single-file shares (not loaded by Viewer app)
* not reproducible in workspaces
* not reproducible in Collectives

`editor.getHTML()` returns the same content for new empty documents in all cases. Also, a quick manual comparison of the editor object didn't reveal anything obvious.

Also interesting: in scenarios where the issue is reproducible, keyboard shortcuts don't get caught by the editor at all. If they have a meaning in the browser, that one applies (e.g. for Ctrl+Shift+1). So apparently something is not registered yet properly? :thinking:

juliushaertl:
For the keyboard shortcuts maybe there is some conflicting event handler that stops propagation in some regards, but I don't see how that would influence clicking the buttons in the same way then

juliushaertl:
Stepped through it a bit and seems that getting the range between $from=0 and $to=0 fails:

![image](https://user-images.githubusercontent.com/3404133/144991026-f11fd02e-057c-4e3a-b2b5-7650a1a8ef68.png)

juliushaertl:
For comparison, this is after typing one char and removing it again: 
![image](https://user-images.githubusercontent.com/3404133/144991277-4e5cabef-fa5f-4a9f-abc1-b12cfaa250a6.png)

juliushaertl:
| Failing | Working |
|---|---|
[...]

PR #2034, opened 2021-12-28
Fix autofocus on empty documents without a node (Fixes: #1974)

mejo-:

* Resolves: #1974
* Target version: master


mejo-:
Tested manually with new empty file, existing empty file and existing file with content.

juliushaertl (approved):

PR #2039, opened 2021-12-30
fix: cypress login with new session feature

azul:
Use the new session feature to create and reuse sessions:
https://docs.cypress.io/api/commands/session

The first time `login` with a given name is called
the steps described in the login function are performed
and cookies and localstorage are cached.

The next time `login` is called with the same name
cookies and localStorage are restored and used again.

This allows us to keep fast test runs
while still separating the tests more cleanly.

The old logout command was broken because of the way
we used `Cypress.Cookies.defaults({ preserve })` before:

Cypress runs all the `cy.*` commands during initialization
and builds a list of commands that are then executed during the tests.
However `Cypress.Cookies.defaults` is evaluated when preparing the list
not while performing the actual steps.

* Target version: master

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -2,5 +2,6 @@
   "baseUrl": "https://localhost:8081/index.php/",
   "projectId": "hx9gqy",
   "viewportWidth": 1280,
-  "viewportHeight": 720
+  "viewportHeight": 720,
+  "experimentalSessionSupport": true
 }
--- a/cypress/integration/files.spec.js
+++ b/cypress/integration/files.spec.js
@@ -21,11 +21,8 @@
  */
 
 describe('Files default view', function() {
-	before(function() {
+	beforeEach(function() {
 		cy.login('admin', 'admin')
-	})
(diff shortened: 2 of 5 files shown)


azul:
I think this is the cleanest way and would be a good basis for implementing more complex tests with cypress.

However since it's only available since 8.2.0 we can only backport it to `stable23` (cypress 8.6.0) but not to `stable22` (cypress 7.5.0) - unless we update cypress there.

I'm currently working on #2020 and i want to prevent regressions like this in the future. Cypress tests seem to be the best way to ensure we don't break things over and over again to me. However if they are very destinct for old branches and the current one backporting the tests adds yet more maintainance overhead.

juliushaertl:
Very nice, that is a lot cleaner :clap: 

I think we could also just bump cypress on the stable branches

juliushaertl (approved):

mejo- (approved):
Nice, indeed much cleaner than before :)

PR #2052, opened 2022-01-04
fix: 2020 let heading menu overflow workspace

azul:
* roll back parts of #1903 that broke it again.
* Calculate top of menububble based on scrollHeight of content-wrapper.
* Always display menububble below selected line.
  This will make it less likely to conflict with the menubar
  or mobile copy and paste toolbars.
* Add cypress tests for the workspace.

* Resolves: #2020
* Target version: master


azul:
I would like to backport this to stable22 and stable23.
Stable22 requires an update of cypress and a backport of #2039.

src/components/MenuBubble.vue
```diff
@@ -159,15 +174,14 @@ export default {
 			command({ href: null })
 		},
 		bubblePosition(menu) {
-			// below the first line, above all others
-			const vertical = menu.top < 45
-				? { top: `${menu.top}px` }
-				: { bottom: `${menu.bottom}px` }
+			const wrapper = this.$parent.$refs.wrapper
```
eneiluj, line 177: Let's keep in mind this breaks if the component structure changes.

eneiluj (commented):
Nice!

src/components/MenuBubble.vue
azul: I could not pass scrollTop in because it changes whenever the text is scrolled up and down - but it does not seem to be reactive.
I'm now passing in the content wrapper explicitely. That should allow using the Menu Bubble in other contexts, and is less magical than requiring the parent to define a `ref` called `wrapper`

eneiluj: :+1:

eneiluj:
@azul I hope you don't mind, I edited your commits to be able to rebase and compiled again in your last commit.

eneiluj (approved):
Nice!

max-nextcloud:
Will need to backport this next week.

PR #2059, opened 2022-01-07
Fix image serving in direct editing

eneiluj:
This fixes 2 mistakes:
* Avoid passing empty shareToken when not logged in
* In the controller, use the edition session to get the user ID instead of relying on NC authentication

Image serving then works in our mobile clients (tested NC Android app).

refs #1900


lib/Controller/ImageController.php
```diff
@@ -246,7 +246,9 @@ public function getImage(int $documentId, int $sessionId, string $sessionToken,
 			if ($shareToken) {
 				$imageFile = $this->imageService->getImagePublic($documentId, $imageFileName, $shareToken);
 			} else {
-				$imageFile = $this->imageService->getImage($documentId, $imageFileName, $this->userId);
+				$session = $this->sessionService->getSession($documentId, $sessionId, $sessionToken);
+				$userId = $session->getUserId();
+				$imageFile = $this->imageService->getImage($documentId, $imageFileName, $userId);
```
mejo-, line 251: Mh, `$userId` seems to be needed in each public function of the controller. It gets injected into the constructor by DI, but in all functions but this one (`getImage()`), it's fetched again using `$session->getUiserId()`. So does this mean that `$userId` as injected by DI is not reliable? Then probably `$this->userId` should be removed from the class altogether?

Maybe a function like the following could work?

```php
private function getUserId(int $documentId, int $sessionId, string $sessionToken): string {
        if ($this->userId === null) {
                $this->userId = $this->sessionService->getSession($documentId, $sessionId, $sessionToken)->getUserId();
        }

        return $this->userId;
}
```

mejo- (changes requested):

lib/Controller/ImageController.php
eneiluj: It's the same in all methods of `ImageController`. Whether the user is authenticated or not, we always choose to rely on the edition session rather than the "classic" authentication. This way it works with the mobile clients which are making unauthenticated requests but our UI passes the edition session ID and token.

This was suggested by @juliushaertl in https://github.com/nextcloud/text/pull/1900#discussion_r756144161

The unused `$userId` attribute has been removed.

mejo-: Looks good to me now :blush: You could still move the logic to get `$userId` into a dedicated private function to lower code-duplication, but that's really just nitpicking :wink:

mejo- (approved):

eneiluj:
@mejo- I completely agree. Factorization done.

eneiluj:
Cypress failure is unrelated. It's the "share" test.

mejo- (approved):

PR #2069, opened 2022-01-10
Fix Content-type of served images

eneiluj:
`image/svg+xml` is replaced by `text/plain` by MimeTypeDetector which prevents displaying SVG images even though they are supported.

IMO we can "whitelist" all the mimetypes we support and still keep using MimeTypeDetector.

### DIFF
--- a/lib/Controller/ImageController.php
+++ b/lib/Controller/ImageController.php
@@ -31,7 +31,7 @@ use OCA\Text\UploadException;
 use OCP\AppFramework\Http;
 use OCA\Text\Service\ImageService;
 use OCP\AppFramework\Controller;
-use OCP\AppFramework\Http\DataDisplayResponse;
+use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\Files\IMimeTypeDetector;
 use OCP\IL10N;
@@ -225,11 +225,11 @@ class ImageController extends Controller {
 	 * @param string $sessionToken
 	 * @param string $imageFileName
 	 * @param string|null $shareToken
-	 * @return DataDisplayResponse
+	 * @return DataDownloadResponse|DataResponse
 	 */
-	public function getImage(int $documentId, int $sessionId, string $sessionToken, string $imageFileName, ?string $shareToken = null): DataDisplayResponse {
+	public function getImage(int $documentId, int $sessionId, string $sessionToken, string $imageFileName, ?string $shareToken = null) {
 		if (!$this->sessionService->isValidSession($documentId, $sessionId, $sessionToken)) {
-			return new DataDisplayResponse('', Http::STATUS_FORBIDDEN);
+			return new DataResponse('', Http::STATUS_FORBIDDEN);
 		}
 
 		try {
@@ -240,15 +240,15 @@ class ImageController extends Controller {
 				$imageFile = $this->imageService->getImage($documentId, $imageFileName, $userId);
 			}
 			return $imageFile !== null
-				? new DataDisplayResponse(
+				? new DataDownloadResponse(
 					$imageFile->getContent(),
 					Http::STATUS_OK,
-					['Content-Type' => $this->mimeTypeDetector->getSecureMimeType($imageFile->getMimeType())]
+					$this->getSecureMimeType($imageFile->getMimeType())
 				)
-				: new DataDisplayResponse('', Http::STATUS_NOT_FOUND);
+				: new DataResponse('', Http::STATUS_NOT_FOUND);
 		} catch (Exception $e) {
 			$this->logger->error('getImage error', ['exception' => $e]);
-			return new DataDisplayResponse('', Http::STATUS_NOT_FOUND);
+			return new DataResponse('', Http::STATUS_NOT_FOUND);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


juliushaertl:
We probably need to workaround that in a similar way as the viewer does, as otherwise we expose a potential way for XSS attacks: https://github.com/nextcloud/viewer/pull/16/files

eneiluj:
Ok I think I get it. You mean we should serve SVGs as base64 encoded strings which would be directly put in the data attribute. I'll check this out.

eneiluj:
No idea why we get
```
TypeError: (0 , _auth.getRequestToken) is not a function
```
when importing `@nextcloud/axios` in the jest tests...

src/nodes/ImageView.vue
```diff
@@ -295,6 +301,15 @@ export default {
...
 		},
+		async setImageSrc(imageUrl) {
+			const file = await axios.get(imageUrl)
+			const mime = file.headers['content-type']
+			if (mime === 'image/svg+xml') {
+				this.imgSrc = `data:${mime};base64,${btoa(file.data)}`
+			} else {
+				this.imgSrc = imageUrl
+			}
+		},
```
max-nextcloud, line 312: This does not seem ideal since it needs to download the image to determine the content-type and then again to display it if it's not an svg. In case of large images this might increase traffic / slow things down.
I don't know if we can avoid it though. I see two options:
* using a `HEAD` request instead that only sends metadata like the mimetype
* using the image in the src as it used to be, responding with some specific error response if it's an svg and handling that response somehow. I have no idea if the browser supports handling failing image request responses.

max-nextcloud: Okay... so looks like there is [onerror](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement#errors) handling for images but i don't see any way to inspect the cause of the error. So the only option would be to handle all errors the same way - which might be:
1. load image with src
2. server rejects it because its an svg
3. javascript catches error tries to initiate alternative image with data property by fetching data from other url
4. server sees other url request and hands out svg
5. browser renders the svg.

Downside being that the second request would also trigger for images not loading for other reasons.

max-nextcloud: Oh... and we would need to make sure that the request in 4. cannot originate from an image src as people can modify that. Including a token that cannot be determined by attackers or a non GET request seems like an option. The first seems cleaner the latter seems easier.

max-nextcloud: Thought about this some more. The best of both worlds to me seems to be to:
1. attempt to load the image via src
2. trigger a second request at the same time that fetches image metadata and where the server includes the data if it's an svg image (might be a `PROPFIND`).
3. have an onerror handler for 1. that will check if 2. returned an svg mimetype and some data. If it did fill the data into the `` replacing the src property.

Since only 1. or 2. will return the actual data traffic should stay roughly the same.
Since both 1. and 2. get started early there should not be additional delays.
Since 2. returns metadata and a data blob in it it cannot be rendered as an image as is. So that prevents using its url in an image src to get the svg to load via the src property.

src/nodes/ImageView.vue
```diff
@@ -295,6 +301,15 @@ export default {
 			tr.delete(pos, pos + this.node.nodeSize)
 			this.view.dispatch(tr)
 		},
+		async setImageSrc(imageUrl) {
+			const file = await axios.get(imageUrl)
```
juliushaertl, line 305: To avoid getting the whole image data before knowing if the mime type can actually be handled, we could use a HEAD request which also returns the `Content-Type` header.

eneiluj: Is this what you have in mind? 5ea058b94bc6a3404a292244f81abce9690b0a51

Now I feel like the `imageUrl2` trick could be removed and image loading could be clarified. How about doing this: 5ffccc411e8838cbc52ac8c2d025b3c8375bed68 ? (pushed in another branch for the moment). @max-nextcloud I think it's close to what you described.

The logic is to always load the image with an `Image()` object. If it's successful, we set the `` src attribute. As the image has already been loaded, the browser does not download it again.

There are 3 major cases:
* (1) data URL: we directly load it in the `Image()`
* not a data URL: we get the mime type with a HEAD request
    * (2) SVG: we download it, generate a base64 string and load it in the `Image()`
    * (3) other mime types: we directly use the URL in the `Image()`

juliushaertl: I've been talking about this with Max earlier today, and we could probably even avoid the HEAD request and use the GET, if browsers are able to properly cache that and then use the cached image when setting the src of the img-tag

src/nodes/ImageView.vue
```diff
@@ -149,6 +151,10 @@ export default {
 			}
 		},
 		imageUrl() {
+			this.setImageSrc(this.imageUrl2)
```
vinicius73, line 154: Why do we have a set inside a computed?
This can potentially create an infinity loop.

src/nodes/ImageView.vue
```diff
@@ -149,6 +151,10 @@ export default {
...
+			return this.imageUrl2
+		},
+		imageUrl2() {
```
vinicius73, line 157: This computed can have a more expressive name?

src/nodes/ImageView.vue
eneiluj: What do you think of https://github.com/nextcloud/text/commit/5ffccc411e8838cbc52ac8c2d025b3c8375bed68 ?

vinicius73: It looks better to me.

eneiluj:
So according to Lukas and Julius, it's fine to give an URL to a SVG file as the src attribute of a `` tag if we use out `DataDownloadResponse` instead of `DataDisplayResponse`. So we can keep the frontend unchanged :grin:.

juliushaertl (approved):

PR #2070, opened 2022-01-10
Trigger cypress on push/pull_request

juliushaertl:
Cypress workflow runs seemed to fail since the merge of #2026 on dependabot pull requests, the only noticable difference was the events that trigger them compared to https://github.com/nextcloud/viewer/blob/master/.github/workflows/cypress.yml, so this should hopefully let github actions pass the record key again also on dependabot pull request.


azul (approved):
Let's give it a try.

PR #2071, opened 2022-01-10
Factorize image upload cypress tests

eneiluj:
Just some more factorization to only leave specific stuff in the `it()` functions: How to trigger the insertion once the action has been clicked.


juliushaertl (approved):

juliushaertl:
Retriggered cypress to see if https://github.com/nextcloud/server/pull/30571 fixes the runs

PR #1916, opened 2021-10-28
Only register trailing node for rich text editing

juliushaertl:
Follow up to #1771, otherwise loading plain text files will fail hard

PR #2075, opened 2022-01-11
[stable22] allow to insert text after trailing codeblock

azul:
* Backports #1771 
* Target version: stable22


azul:
We need to make sure to also backport #1916

juliushaertl (approved):

PR #2077, opened 2022-01-12
ci: dependabot PRs against stable23 rather than stable20

max-nextcloud:
stable20 has reached end of live.

* Target version: master


juliushaertl:
Added to https://github.com/nextcloud/office/issues/5#issue-1091030902 to also have that for the next major bump and then as a template for future releases.

juliushaertl (approved):

PR #2090, opened 2022-01-12
Ci: github action for php unit tests

max-nextcloud:
Also runs composer validate and install.

* Target version: master 

The idea here is to replace drone as it's often overloaded throughout the day.


max-nextcloud:
oh... nice... we already have php unit tests!

Issue #1679, opened 2021-06-24
Replace drone with github actions 


nickvergessen:
Please migrate your drone usage to github actions

PR #2091, opened 2022-01-12
ci: github action for composer

max-nextcloud:
Validate composer and ensure composer install actually works.

* Target version: master 

This is meant to replace the drone compatibility check


eneiluj:
Can we implement all drone checks in a GitHub action? That would be awesome.

max-nextcloud:
> Can we implement all drone checks in a GitHub action? That would be awesome.
I think this is the only missing part. We already have php unit tests run by github actions.

max-nextcloud:
Removed the drone config... but i think it's just listening to what's in master currently. On the other hand this means that if we merge this other branches will also stop running drone.

mejo-:
Probably this is obsolete now that we have #1679 in, right?

max-nextcloud:
> Probably this is obsolete now that we have #1679 in, right?

I don't know. Does `composer validate` do anything that the php setup routine does not? I'll look it up.

max-nextcloud:
Looks like composer validate performs some checks - but some of them we don't care about such as checking if it could be published. Also we still ignore warnings in this test. Not sure if it would catch anything that our other tests don't catch. So i'll just close this and we can add it once we notice we would have needed it.

PR #2093, opened 2022-01-13
composer: Specify php version so dependabot respects it

max-nextcloud:
Dependabot just came up with an invalid composer.lock for php 7.3.
It picks the lowest version requirement from composer.json
and selects the dependencies based upon that.

Specify our php versions so Dependabot can take them into account.

* Resolves: # 
* Target version: master

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,9 @@
             "email": "jus@bitgrid.net"
         }
     ],
-    "require": {},
+    "require": {
+      "php": "^7.3|^8.0"
+    },
     "scripts": {
       "lint": "find . -name \\*.php -not -path './vendor/*' -print0 | xargs -0 -n1 php -l",
       "cs:check": "php-cs-fixer fix --dry-run --diff",


juliushaertl:
Hm, i guess you are referring to psr/container (1.0.0 => 1.1.2) which is still bumped to the incompatible version 

In deck we added the plaform requirement https://github.com/nextcloud/deck/pull/3498/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R30 as this seems to still pull in the newer version on a `composer update`

max-nextcloud:
> Hm, i guess you are referring to psr/container (1.0.0 => 1.1.2) which is still bumped to the incompatible version

yes. It broke the build in #1954 

> In deck we added the plaform requirement https://github.com/nextcloud/deck/pull/3498/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R30 as the generic requirement here seems to still pull in the newer version on a `composer update`

I based this on [a comment in a dependabot issue](https://github.com/dependabot/feedback/issues/761#issuecomment-555161692):
> Before running composer, dependabot will get your lowest supported PHP version in your composer.json (at key require.php) and put it in the key config.platform.php so that composer will resolve the updates of your dependencies on your lowest supported PHP version no matter what PHP version composer is actually running on.

So my hope was this would fix dependabot. Might not fix all uses of `composer update` - but i'd be fine with that. Specifying the range that we support seems more correct to me than locking everything to the lowest version number.

But i am also happy to change this to pinning config.platform.php - as long as dependabot creates correct dependency updates i'm happy.

juliushaertl:
> I based this on [a comment in a dependabot issue](https://github.com/dependabot/feedback/issues/761#issuecomment-555161692):
> 
> > Before running composer, dependabot will get your lowest supported PHP version in your composer.json (at key require.php) and put it in the key config.platform.php so that composer will resolve the updates of your dependencies on your lowest supported PHP version no matter what PHP version composer is actually running on.
> 
> So my hope was this would fix dependabot. Might not fix all uses of `composer update` - but i'd be fine with that. Specifying the range that we support seems more correct to me than locking everything to the lowest version number.
> 
> But i am also happy to change this to pinning config.platform.php - as long as dependabot creates correct dependency updates i'm happy.

Thanks for the reference. Makes sense 👍

juliushaertl (approved):

PR #2102, opened 2022-01-17
[stable23] fix: use BRANCH when checking out viewer

max-nextcloud:
Use the viewer that fits the nextcloud release.

Currently cypress tests are failing for all branches
even though the cause seems to be a recent commit to viewers stable branch.

Cypress runs for the stable23 branch of text
should use the stable23 branch of server
and the stable23 branch of viewer
so they match the actual releases that will go out.

* Resolves: # 
* Target version: master


max-nextcloud:
okay... looks like it is NOT a viewer issue. Respecting `$BRANCH` still makes sense but i'll add that later.

PR #2104, opened 2022-01-17
Fix closing the editor modal in public folder view

eneiluj:
Apparently the `close` event emitted by the `<PublicFilesEditor>` component is not caught anywhere. The `active` property of `<PublicFilesEditor>` never changes and the modal is not properly closed.

This destroys the Vue instance rendering  `<PublicFilesEditor>` when the `close` event is emitted.

This needs to be backported at least in stable23.


mejo- (approved):
Yay :+1:

max-nextcloud (approved):
Looks good to me. Will trigger a recompilation of the js assets. Then this should be ready to go.

PR #2107, opened 2022-01-18
test: closing the non-viewer modal in shared folder

max-nextcloud:
This is a regression tests for #2104.

* Target version: master

### DIFF
--- a/cypress/integration/share.spec.js
+++ b/cypress/integration/share.spec.js
@@ -138,6 +138,8 @@ describe('Open test.md in viewer', function() {
 					cy.get('#editor-container').should('be.visible')
 					cy.get('#editor .ProseMirror').should('contain', 'Hello world')
 					cy.get('#editor .ProseMirror h2').should('contain', 'Hello world')
+					cy.get('.modal-header button.header-close').click()
+					cy.get('.modal-mask').should('not.exist')
 				})
 			})
 	})


max-nextcloud:
interesting failure... *sigh*

eneiluj:
The failure happens there: https://github.com/nextcloud/text/blob/9f59d9e1fa32c86bf9d7532109c8a5cada9b94e2/src/services/PollingBackend.js#L232

The response when requesting `apps/text/public/session/push` with valid parameters is `[]`.
Here is an example of passed parameters:
```
{
    "documentId": 26558,
    "sessionId": 1395,
    "sessionToken": "i7RyQF6rPyhJwmYZLrEi7BKeNBf9/OGVKHwun3cUVykoyDvZn3b+MNclmj1jwlCL",
    "steps": [
        {
            "stepType": "replace",
            "from": 13,
            "to": 13,
            "slice": {
                "content": [
                    {
                        "type": "paragraph"
                    }
                ]
            }
        }
    ],
    "version": 0,
    "token": "JgAkDesDfxG2Wtk",
    "filePath": "//test.md"
}
```

I'll dig deeper on the server side code to see if it's expected to get an empty array in the response or if it's a bug.

max-nextcloud:
https://github.com/nextcloud/text/pull/2193/commits/0f59c2019746a4acbca72ee6b4ad93186d3d1e87 fixes the issue in question. It also has an explanation in the commit message:

> Cypress tests were failing because they triggered editor.focus on a read only doc.
> Focus seems to cause sendable steps - maybe because the current cursor changes.
> 
> Prevent this on various levels:
> * Do not autofocus read only docs.
> * Do not send steps to read only docs.
> * Handle server response with 403 without content gracefully when sending steps.
> 
> The last response originated here:
> https://github.com/nextcloud/text/blob/master/lib/Service/ApiService.php#L158
> So wither the session was not valid or the document read only.
> 
> Not entirely sure what best to do in this situation.
> Logging to console.error for now.

mejo- (approved):

PR #2108, opened 2022-01-18
fix: vue warning about invalid prop type

### LINKED ISSUE #2103: Vue prop type check warnings in MenuBubble
When compiling Text's frontend code in development mode, there are some Vue warnings about the contentWrapper prop being an `HTMLDivElement` instead of an Object.

@max-nextcloud It's related with https://github.com/nextcloud/text/pull/2052 . Is it possible to make the prop type check pass?

max-nextcloud:
* Fixes #2103.
* Targets master.


eneiluj (approved):
:rocket:

PR #2113, opened 2022-01-19
Backport/stable22/2020

max-nextcloud:
* Resolves: #2052
* Target version: stable22

### Summary

* roll back parts of #1903 that broke it again.
* Calculate top of menububble based on `scrollHeight` of `content-wrapper`.
* Always display menububble below selected line.
  This will make it less likely to conflict with the menubar
  or mobile copy and paste toolbars.
* Add cypress tests for the workspace.
* Include the fix for #2108.


max-nextcloud:
cypress is still failing because of the regression in server.
Need to wait for https://github.com/nextcloud/server/pull/30752 and then run the tests again.

mejo- (approved):

PR #2128, opened 2022-01-24
Updating dependabot-approve-merge.yml workflow from template

nextcloud-command:
Automated update of the dependabot-approve-merge.yml workflow from https://github.com/nextcloud/.github


max-nextcloud (approved):
Looks good to me.
Just a rename, pulling up the if condition and some formatting.
Oh... and it won't acto on `dependabot-preview[bot]` anymore... which we don't use anyway as far as i can tell.

PR #2129, opened 2022-01-24
ci: increase cypress default command timeout

max-nextcloud:
Prevent timeouts due to slow CI runs by waiting longer for normal commands.

Some of our commands open the editor which has to load quite a bit of additional javascript.
This can be slow on CI.
Instead of failing after 4 seconds - wait 10 seconds when run in CI.

* Target version: master


eneiluj:
Maybe a syntax problem? It's not running with your changes.

max-nextcloud:
> Maybe a syntax problem? It's not running with your changes.

Well.... i need to specify the new timeout i guess m). Now it's passing.

eneiluj (approved):
:+1:

PR #2137, opened 2022-01-28
Update master php testing versions

nickvergessen:


mejo-:
This also completely removes `.drone.yml`. Is this intended?

nickvergessen:
the drone file only contained php7.3 tests. feel free to restore and adjust to 7.4 images. Prefered solution would be to do it via github actions

mejo- (approved):

PR #2143, opened 2022-02-02
Add index for last_contact in text_sessions table

eneiluj:
This will most likely improve this query:
https://github.com/nextcloud/text/blob/427b262e51efecbbaffb24639c23424fcdd0c360/lib/Db/SessionMapper.php#L84-L92
which is involved in the session cleanup job.

### DIFF
--- a/lib/Migration/Version030501Date20220202101853.php
+++ b/lib/Migration/Version030501Date20220202101853.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\Text\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
+ */
+class Version030501Date20220202101853 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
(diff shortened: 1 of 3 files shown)


lib/Migration/Version030501Date20220202101853.php
```diff
@@ -0,0 +1,32 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('text_sessions');
+		$table->addIndex(['last_contact'], 'ts_lastcontact');
+
+		return $schema;
```
juliushaertl, line 30: Could you add that to https://github.com/nextcloud/text/blob/f256eafe22234a6eb4586c03ace9ba480c50309c/lib/Migration/Version010000Date20190617184535.php#L103 and wrap it with a `hasIndex` call here?

```suggestion
		$table = $schema->getTable('text_sessions');
		if (!$table->hasIndex('ts_lastcontact')) {
			$table->addIndex(['last_contact'], 'ts_lastcontact');
			return $schema;
		}

		return null;
```

That way we safe an extra migration step which might have a performance impact on new installations, especially on oracle.

juliushaertl (changes requested):

lib/Migration/Version030501Date20220202101853.php
eneiluj: Sure, done.

mejo- (approved):

juliushaertl (approved):

eneiluj:
Should we backport this to stable23, stable22 and stable21? If so I'll make them manually.

PR #2144, opened 2022-02-03
Use php 7.4 and 8.0 in drone CI

eneiluj:
Since the server does not support php7.3 anymore.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,7 +2,7 @@ kind: pipeline
 name: compatibility
 steps:
   - name: compatibility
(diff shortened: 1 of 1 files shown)


eneiluj:
I don't know why it fails with Php 8.0. How about merging this to make drone CI work again and fix the Php 8.0 issue later?

mejo-:
Sounds like a good plan - especially since we should move these checks to Github actions anyway :wink:

mejo- (approved):

Issue #2050, opened 2022-01-04
Enabling capitelized letters in keyboard shortcuts in markdown editor 


Chnapik2:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
I was working on a markdown document the other day and I was curently in caps lock mode and noticed that ctrl + B and other keyboard shortcuts do not work. When I turned caps lock off, everything was ok.

**Describe the solution you'd like**
Enabling keyboard shortcuts for capitalized letters, e. g. ctrl+b = ctrl+B. I am shure that it should be an easy fix and would help with being a bit more user freindly.

mejo-:
I checked other online editors (Notion and Google Docs) and there indeed shortcuts with capitalized characters work - but shortcuts with Shift don't (e.g. `<Ctrl>+<Shift>+<B>` doesn't toggle bold). I guess we should do the same.

juliushaertl:
It hink its not about regular shift, but having caps lock activated, there both notion and google docs seem to also toggle.

mejo-:
I reported the issue upstream: https://github.com/ueberdosis/tiptap/issues/2426

Also did some local testing, and adding the uppercase key to the keymap for a mark works as expected:

```js
	keys({ type }) {
		return {
			'Mod-b': toggleMark(type),
			'Mod-B': toggleMark(type),
		}
	}
```
Only downside is, that key combinations with shift (<kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>b</kbd> in this example) will be caught as well. But as written in the upstream issue, I think that's an acceptable tradeoff.

mejo-:
Opened a PR upstream: https://github.com/ueberdosis/tiptap/pull/2478

PR #2146, opened 2022-02-03
Add key bindings for uppercase letters for bold, italic underline

mejo-:
This way, key bindings 'Mod-B', 'Mod-I' and 'Mod-U' with active caps
lock have the same effect as their lowercase siblings.

This is supposed to be fixed in Tiptap v2 upstream:
https://github.com/ueberdosis/tiptap/issues/2426
https://github.com/ueberdosis/tiptap/pull/2478

Fixes: #2050

* Resolves: #2050
* Target version: master


eneiluj (approved):
Wow it works!

PR #2145, opened 2022-02-03
Use file.path to track EditorWrapper instances more accurately

artonge:
This fixes rich workspace loading when moving a `Readme.md` into a folder and navigating into that folder.

The `file.id` stays the same, so no new components are created and RichWorkspace stays in a `ready=false` state because EditorWrapper never emits a new ready event.

PR #2152, opened 2022-02-03
[stable21] Use file.path to track more accurately EditorWrapper instances

eneiluj:
Backport of #2145


mejo- (approved):

artonge (approved):

max-nextcloud (dismissed):
Blocking since this may have unintended consequences.

Say i edit a file and you rename it in the meantime... What should happen? I think it's a feature and not a bug, that the editing keeps working and i now edit the file with the new name.

Issue #1827, opened 2021-08-17
"data:" URLs handled wrong for images


alx-tuilmenau:
**Describe the bug**
If Markdown-Files (especially Readme.md) contain `data:` URLs, a http(s) request for the preview (including data) is executed, trying to get the the image preview. 
`https://example.com/core/preview.png?file=%2Fdata%3Aimage%2Fpng%3Bbase64%2CiVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAAAl21bKAAAAA1BMVEUAAACnej3aAAAAAXRSTlMAQObYZgAAAApJREFUCNdjYAAAAAIAAeIhvDMAAAAASUVORK5CYII%3D&x=1024&y=1024&a=true`
This triggers (sometimes, if data is long enough) a "OCP\Files\InvalidPathException: Path length(7565) exceeds max path length(4000)" in the log. 
Because there is no such image on the http(s) server, there is a 404 result and it will replaced with a placeholder, linked to the correct original data: - url. Clicking the placeholder opens the image in new tab.

**To Reproduce**
1. Create a Readme.md file with the content `![](data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAAAl21bKAAAAA1BMVEUAAACnej3aAAAAAXRSTlMAQObYZgAAAApJREFUCNdjYAAAAAIAAeIhvDMAAAAASUVORK5CYII=)`
2. Close and Reopen the file in text

**Expected behavior**
The image is displayed correctly (or at least not requested via http(s)).

**Client details:**
 - OS: Win 10
 - Browser: Firefox
 - Version: 91
 - Device: PC

<details>
<summary>Server details</summary>

**Text app version:** 3.1.0

**Operating system:** Debian 10

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud version:** 20.0.11

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
if data is long enough, a OCP\Files\InvalidPathException: Path length(7565) exceeds max path length(4000) - Exception (but this is correct)

#### Browser log
```
The javascript console log:

[...]

PR #2153, opened 2022-02-04
[stable23] Render data URL images as-is

DerpgonCz:
### Summary
When the image URL is a data URL, render the image as-is with `src="dataUrl"`. No need to fetch anything.

Also prevents a race condition while loading images.

Resolves: #1827


DerpgonCz:
Well, what to do with this? Wait until the NC main repo updates to not be `RC`?

src/nodes/ImageView.vue
```diff
@@ -227,6 +229,7 @@ export default {
 			this.imageLoaded = false
 			this.loaded = true
 		}
+		img.src = this.imageUrl
```
eneiluj, line 232: Why did you move this?
Is it to make sure `onload` is defined before the image is actually loaded?
I agree it's better but it works fine without moving it.
@mejo- @juliushaertl @max-nextcloud What do you think about that?

eneiluj (approved):
:+1: Well done, thanks a lot.
Could you make another PR for the master branch? Those changes apply perfectly to master as well.

@juliushaertl What do you think about potential XSS with SVG? Should we simply not render SVG data?

eneiluj:
@juliushaertl Ignore my comment about XSS. It's actually safe to load base64 content in the src attribute.
Sorry for the noise.

eneiluj:
> Well, what to do with this? Wait until the NC main repo updates to not be `RC`?

Exactly, we'll merge the approved PRs after the minor releases have been produced.

src/nodes/ImageView.vue
DerpgonCz: This is due to a race condition, detailed in the issue #1827 

tldr: Assigning Image.src before Image.onload does not fire the event if the actual `load` event on the Image object happens before assigning the event handler function. (so between `onload=` and `src=`).

I understand that this is extremely unlikely to happen, maybe even impossible or really a tiny edge-case. However, better safe than pull out our hair when something like that happens.

DerpgonCz:
> +1 Well done, thanks a lot. Could you make another PR for the master branch? Those changes apply perfectly to master as well.
> 
> @juliushaertl What do you think about potential XSS with SVG? Should we simply not render SVG data?

We could possibly check for the mimetype. Also described in the issue #1827.

src/nodes/ImageView.vue
juliushaertl: Makes sense 👍

juliushaertl:
>  Ignore my comment about XSS. It's actually safe to load base64 content in the src attribute.

While a quick test at least seems to not execute any inline JS from base64 encoded svgs, I'd rather be careful about this. Any reference mentioning that this is not an issue? In general I'd say we should avoid rendering user provided SVG especially if they might be presented to other users. 

Example payload I used:
```
<img src="data:image/svg+xml;base64,CjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB3aWR0aD0iMTAwIiBoZWlnaHQ9IjEwMCI+CiAgPGNpcmNsZSByPSIxMCIgY3g9IjEwIiBjeT0iMTAiIGZpbGw9ImdyZWVuIi8+CiAgPGltYWdlIGhyZWY9IngiIG9uZXJyb3I9ImphdmFzY3JpcHQ6YWxlcnQoJ1hTUycpIiAvPgo8L3N2Zz4=" />
```

juliushaertl:
I couldn't find any good resource how different browsers process files in img tags, but if they are not parsed in terms of HTML/JS rendering it should be safe.

DerpgonCz:
If possible, it would be a major vulnerability. Most conversations online point to it not being possible while in an `img` tag. However, opening the source directly in a browser (opening DevTools and clicking the `src` attribute and `Open link in a new tab` actually executes the code.

I am not sure how big of a vulnerability this is.

juliushaertl:
Sounds reasonable then 👍  Also according to https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image#restrictions this should be restricted by the browsers.

max-nextcloud:
The freeze is over. The CI is green again.. time to get this in alongside the other pending stable23 PRs. I'm going through them one by one and hope i get to it today.

PR #2163, opened 2022-02-07
Work around OCA.Viewer.file not set when editing in richworkspace mode.

rotdrop:
* Resolves: #2164
* Target version: master 

### Summary
See linked issue.

### DIFF
--- a/src/components/EditorWrapper.vue
+++ b/src/components/EditorWrapper.vue
@@ -291,6 +291,7 @@ export default {
 				this.$parent.$emit('error', 'No valid file provided')
 				return
 			}
+			OCA.Text.RichWorkspaceFilePath = this.relativePath
 			const guestName = localStorage.getItem('nick') ? localStorage.getItem('nick') : getRandomGuestName()
 			this.syncService = new SyncService({
 				shareToken: this.shareToken,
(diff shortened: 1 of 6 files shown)


rotdrop:
The second commit removes the extra "context=BASEFILE.md" from the generated links which would modify the Readme.md markdown in an irritating manner.

max-nextcloud:
Hey @rotdrop,
Thanks a lot for the analysis of the proplem and proposed fix. Your fix is perfectly in line with what we did there so far.
However I'd prefer to avoid introducing another *global* variable that magically switches behavior inside the helper. Using `OCA.Viewer` inside that helper was a mistake in the first place. 

I think we could hand the `currentDir` to the `domHref` function in the helper. My understanding is that it's only called from the link mark which should have access to the editor state. We could inject `currentDir` the way we do it for the `Image` node.

max-nextcloud:
Actually it would probably be better to get this fixed and backported first and clean it up after. Would be good to have a test for the broken behavior to make sure this fixes it and also to prevent it from braking during the refactor afterwards.

mejo-:
Thanks again for the PR @rotdrop. A similar fix got implemented in https://github.com/nextcloud/text/pull/3269

PR #2168, opened 2022-02-07
[master] Render data URL images as-is

DerpgonCz:
Summary
When the image URL is a data URL, render the image as-is with src="dataUrl". No need to fetch anything.

Also prevents a race condition while loading images.

Related for stable23: #2153

Resolves: https://github.com/nextcloud/text/issues/1827


max-nextcloud (approved):
Same as the PR for stable23. Looks good to me.

PR #2169, opened 2022-02-08
Migrate phpunit tests from drone to github actions (Fixes: #1679)

mejo-:
Only test with sqlite for now. There's no added value to run our
basic phpunit test on mysql and postgresql environments.

* Resolves: #1679
* Target version: master


max-nextcloud:
I thought we were already running php unit tests in ci... Ahh... i see what you are doing. I guess we only ran them with OCI backend. I guess this does not really matter as they are hardly testing anything but makes sense to run them in the default setup.

I ran into the same issue with CI still wanting to run drone in https://github.com/nextcloud/text/pull/2091.
I think our two PRs compliment each other. Mine adds a composer validate run and yours adds the unit tests. Pr do we only need one job instead?

With these two jobs i really think we could turn of drone. @juliushaertl I think you have the powers to do so, right?

juliushaertl (approved):

juliushaertl:
Adjusted the required checks accordingly

PR #2182, opened 2022-02-14
fix: bring back horizontal rule

max-nextcloud:
Also drop @tiptap/starter-kit as we are not using it anymore.

* Target version: master 

### Summary

We were missing the horizontal rule extension since the migration to tiptap2. This brings it back and also adds a unit test for it.


package.json
```diff
@@ -55,7 +56,6 @@
 		"@tiptap/extension-strike": "^2.0.0-beta.26",
 		"@tiptap/extension-text": "^2.0.0-beta.15",
 		"@tiptap/extension-underline": "^2.0.0-beta.22",
-		"@tiptap/starter-kit": "^2.0.0-beta.183",
```
juliushaertl, line 58: Should not be removed I guess?

juliushaertl (approved):

Issue #2185, opened 2022-02-17
Ordered lists nested in unordered lists do not show numbers


max-nextcloud:
**Describe the bug**
When nesting an ordered list in another list type it has bullets rather than numbers.

**To Reproduce**
1. Open the editor
2. Create a bullet list
3. when on a new item press tab to add a nested item
4. toggle to ordered list

**Expected behavior**
Nested list should start with 1.

**Screenshots**
![Bildschirmfoto von 2022-02-17 10-50-13](https://user-images.githubusercontent.com/97337118/154450449-d29c3c1c-1117-4190-bf09-5cab5cdb878c.png)

**Nextcloud version:** 23

PR #2186, opened 2022-02-17
fix: only apply bullet style to ul > li

max-nextcloud:
* Make list items in ordered lists still show numbers.
* Use the same bullets on second and third level when first level is ol.

* Resolves: #2185
* Target version: master, 23, 22, 21


juliushaertl (approved):
Makes sense 👍 

Restarted cypress as failures seemed unrelated.

Issue #2171, opened 2022-02-09
Use PROPFIND instead of custom workspace URL in files app


juliushaertl:
This can save one additional request that is just used for checking if a README file is part of the file list. This should speed up loading of the rich workspace area when switching folders. 

Should work in a similar way as we do it in files_lock https://github.com/nextcloud/files_lock/blob/master/js/files.js#L3-L22

PR #2193, opened 2022-02-21
use filelist PROPFIND for loading workspace

max-nextcloud:
Resolves: #2171.
Target Branch: master

Even though this is meant to improve load times
I did not observe any improvements locally.

I suspect that is because the workspace was loaded in parallel with the PROPFIND anyway.
Experience may differ in production environments obviously.

It is still one request less and the code also seems more clean.


max-nextcloud:
It seems to work for me. However i have not looked into error handling and the like yet.
Also fixes the autofocus when creating a new readme.

src/helpers/files.js
```diff
@@ -149,6 +150,21 @@ const FilesWorkspacePlugin = {
...
+			const dir = (data.mimetype === 'httpd/unix-directory')
+				? data.path + (data.path.endsWith('/') ? '' : '/') + data.name
+				: data.path
+			if (dir !== fileList.getCurrentDirectory()) {
+				return
+			}
+			if (data.mimetype === 'httpd/unix-directory') {
+				this.vm.folder = { ...data }
+			}
+			if (data.mimetype === 'text/markdown' && data.name === 'Readme.md') {
```
juliushaertl, line 164: This should also handle other filenames that are taken into consideration:

https://github.com/nextcloud/text/blob/adf5a760a04b50ec4e538f7b2c9a7d8549ef4cae/lib/Service/WorkspaceService.php#L30-L46

A different approach might also be to extend the DAV plugin in order to return the filename of the rich workspace file as a dedicated property, similar to https://github.com/nextcloud/text/blob/master/lib/DAV/WorkspacePlugin.php#L99-L114 which would already return the content.

juliushaertl: The second approach would have the benefit that we only maintain one codepath on how the workspace file is determined.

max-nextcloud: Oh wow... that looks good. I think we could actually make use of the content.

max-nextcloud:
Funny.... that cypress tests passed for me locally.

juliushaertl:
I think we force the language in cypress tests when running on CI:

```
cypress/server.sh
4:php occ config:system:set force_language --value en

.github/workflows/cypress.yml
89:          php occ config:system:set force_language --value en
```

That would explain the Bad Request on trying to set it:
https://github.com/nextcloud/server/blob/309354852f12ae88d5eef05d311d6ebcba8ee762/apps/provisioning_api/lib/Controller/UsersController.php#L757-L762

So either doing the language update as an admin or changing the cypress setup to rather just set the default language for tests on user creation would work both.

cypress/support/commands.js
```diff
@@ -62,6 +62,22 @@ Cypress.Commands.add('nextcloudCreateUser', (user, password) => {
 	})
 })
 
+Cypress.Commands.add('nextcloudUpdateUser', (user, password, key, value) => {
```
juliushaertl, line 65: Nothing related to this PR, but maybe we could consider wrapping all those nice helpers into a small js library, so that they could be reused across apps when writing cypress tests.

juliushaertl (changes requested):
Figured out when testing that we actually need to handle public share links of folders in a special way, as there the DAV property currently does not work as in throwing an error when requested. 

I've been playing with this patch to make the property work, but unsure yet if that is a proper way to get the node for the public share:

```diff
diff --git a/lib/DAV/WorkspacePlugin.php b/lib/DAV/WorkspacePlugin.php
index 764aa1f3..63b2062c 100644
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -29,6 +29,7 @@ use OC\Files\Node\File;
 use OC\Files\Node\Folder;
 use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Files\FilesHome;
+use OCA\DAV\Storage\PublicOwnerWrapper;
 use OCA\Text\AppInfo\Application;
 use OCA\Text\Service\WorkspaceService;
 use OCP\Files\IRootFolder;
@@ -98,9 +99,9 @@ class WorkspacePlugin extends ServerPlugin {
                // Only return the property for the parent node and ignore it for further in depth nodes
                if ($propFind->getDepth() === $this->server->getHTTPDepth()) {
                        $propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
+                               $owner = $this->userId ?? $node->getFileInfo()->getStorage()->getOwner('');
                                /** @var Folder[] $nodes */
-                               $nodes = $this->rootFolder->getUserFolder($this->userId)->getById($node->getId());
-                               if (count($nodes) > 0) {
+                               $nodes = $this->rootFolder->getUserFolder($owner)->getById($node->getId());
                             if (count($nodes) > 0) {
                                        /** @var File $file */
                                        try {
                                                $file = $this->workspaceService->getFile($nodes[0]);
@@ -114,8 +115,9 @@ class WorkspacePlugin extends ServerPlugin {
                                return '';
                        });
                        $propFind->handle(self::WORKSPACE_FILE_PROPERTY, function () use ($node) {
+                               $owner = !$node->getFileInfo()->getStorage()->instanceOfStorage(PublicOwnerWrapper::class) ? $this->userId : $node->getFileInfo()->getStorage()->getOwner('');
                                /** @var Folder[] $nodes */
-                               $nodes = $this->rootFolder->getUserFolder($this->userId)->getById($node->getId());
+                               $nodes = $this->rootFolder->getUserFolder($owner)->getById($node->getId());
                                if (count($nodes) > 0) {
                                        /** @var File $file */
                                        try {
```

However the frontend parsing in src/helper/files.js also needs some adjustments for that then as well.

juliushaertl:
Might be nice to extend the cypress tests also with one case for the public share link.

max-nextcloud:
> Figured out when testing that we actually need to handle public share links of folders in a special way, as there the DAV property currently does not work as in throwing an error when requested.
> 
> I've been playing with this patch to make the property work, but unsure yet if that is a proper way to get the node for the public share
I gave it a try and it seems to be working. No idea if i am missing any scenario though.

> However the frontend parsing in src/helper/files.js also needs some adjustments for that then as well.
Yes. I think i fixed it... looks like for public shares the data object did not include a name. So now we find it based on id.

cypress/support/commands.js
max-nextcloud: Created #2201 to address this.

max-nextcloud:
If we decide to backport this we need to make sure to also backport #2216

PR #2197, opened 2022-02-22
[stable22] fix: only apply bullet style to ul > li

max-nextcloud:
Make list items in ordered lists still show numbers.
Use the same bullets on second and third level when first level is ol.

* Backports: #2186
* Resolves: #2185
* Target version: stable22


juliushaertl:
@max-nextcloud Would be cool to always have reviewers requested and labels about the state (read for review/in development) and purpose (bugfix/enhancement) for new PRs 😉

juliushaertl (approved):

PR #2208, opened 2022-02-26
[stable23] fix: update psalm baseline to account for changes in server

max-nextcloud:
https://github.com/nextcloud/server/commit/003f2b59c92fa5a9455ddacea11ee0d66c828f5b changed the types psalm expects and thus triggered an error not yet recorded in the baseline.


max-nextcloud:
https://github.com/nextcloud/text/runs/5343100357?check_suite_focus=true is the failure this is trying to address.

mejo- (approved):

PR #2223, opened 2022-03-05
Updating node.yml workflow from template

nextcloud-command:
Automated update of the node.yml workflow from https://github.com/nextcloud/.github

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -9,6 +9,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
       - master
       - stable*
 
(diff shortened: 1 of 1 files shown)


.github/workflows/node.yml
```diff
@@ -38,14 +39,9 @@ jobs:
...
       - name: Install dependencies & build
         run: |
-          make clean
           npm ci
           npm run build --if-present
 
-      - name: npm test
-        run: |
-          npm test
-
```
max-nextcloud, line 48: So this would not run the tests anymore?
Then we should at least have another workflow to run the tests.

max-nextcloud: turns out the original template never ran tests. Will update the PR to keep our changes and only update the dependencies.

max-nextcloud (approved):

Issue #2172, opened 2022-02-09
Embed initial file fetching into the create session endpoint


juliushaertl:
Currently when starting a document editing session we always perform two requests:
- create: initialize the sesion
- fetch: fetch the base document version
Followed by a sync call to get any intermediate steps

It should be possible to embed the file content directly inside of the create response to save an extra API call there.

In addition the initial sync call could also be saved if we sent those along with the create call and there are no further steps to be fetched from the server.

PR #2225, opened 2022-03-07
Performance/create session with content

max-nextcloud:
* Resolves: #2172 
* Target version: master 

### Summary

Fetch the content of the file when creating the session. Saves the `fetch` request whenever no `initialSession` is given. On my dev instance this saves 80ms - but it's hard to tell how much it will save in production. Well it will save one roundtrip that is currently blocking the rendering of the content.


lib/Service/ApiService.php
```diff
@@ -106,10 +106,17 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
 		}
 
 		$session = $this->sessionService->initSession($document->getId(), $guestName);
+		try {
+			$baseFile = $this->documentService->getBaseFile($document->getId());
+			$content = $baseFile->getContent();
+		} catch (NotFoundException $e) {
+			$content = null;
+		}
```
max-nextcloud, line 114: This really should not happen as the file was just created. We should probably log it if we end up here.

juliushaertl (approved):
Nice, thanks for looking into that 👍

Issue #2183, opened 2022-02-15
Improve read only performance when just displaying rich workspace


juliushaertl:
In most cases the rich workspace area is probably not edited, so we should think about having a more sane process for opening up a read only editor and just kicking of the heavy sync parts and session setup when starting to edit.

max-nextcloud:
I profiled the workspace loading some and it turns out rendering the menubar is surprisingly expensive.
I experimented some with it and we can postpone that until the content has been rendered. In theory this would reduce the time till the content is displayed by 600-800ms. In practice it's still hard to measure.

ImaCrea:
Great idea :) It's very disturbing when I'm ready to click on a file and then everything scrolls down because finally the editor is loaded.

tcitworld:
The PR from @max-nextcloud doesn't really answer the original issue, may this be reopened to avoid loading sync & session until the workspace is actually opened?

juliushaertl:
https://github.com/nextcloud/text/pull/4776 also contributed to that by showing the content already before the session is opened but what would be missing is a read only and just open the session once clicked option

elzody:
Notes from meeting with @marcoambrosini:

- [ ] Suggestion buttons row should not be displayed there 
- [ ] When we navigate to a folder, the rich workspaces section should not be in "edit" mode, and thus the file should not be locked
- [ ] Cursor becomes pointer when hovering the inactive text input. Clicking on it activate the text area and locks the file
- [ ] Clicking outside saves and disables edit mode again
- [ ] Little scrolling arrows should not be shown
- [ ] Plus button should not be cut off

PR #2226, opened 2022-03-07
display content first and then load menus

max-nextcloud:
Rendering the menubar and menububble takes a significant amount of time.
First display the content and then render the menus.

* Resolves: #2183 
* Target version: master 

Renders the content >500ms faster on my local dev setup.

Now the content is drawn (LCP) before the large computation load of rendering the menubar:
Bildschirmfoto von 2022-03-07 10-19-02


max-nextcloud:
This should also speed up opening files but the content seems to render at the same time as the menubar there. Will need to investigate if it actually speeds up loading there as well or what is blocking it.

vinicius73:
We also can use [window.requestAnimationFrame()](https://developer.mozilla.org/pt-BR/docs/Web/API/Window/requestAnimationFrame) to "delay" some stuff to create a more smooth experience.

src/components/EditorWrapper.vue
```diff
@@ -34,8 +34,8 @@
 			</p>
 		</div>
 		<div v-if="displayed" id="editor-wrapper" :class="{'has-conflicts': hasSyncCollission, 'icon-loading': !initialLoading && !hasConnectionIssue, 'richEditor': isRichEditor, 'show-color-annotations': showAuthorAnnotations}">
-			<div id="editor">
-				<MenuBar v-if="tiptap && !syncError && !readOnly"
+			<div v-if="tiptap" id="editor">
+				<MenuBar v-if="initialLoading && !syncError && !readOnly"
```
vinicius73, line 38: Should it be better to use a computed here?

max-nextcloud:
> We also can use [window.requestAnimationFrame()](https://developer.mozilla.org/pt-BR/docs/Web/API/Window/requestAnimationFrame) to "delay" some stuff to create a more smooth experience.

For now i am mostly trying to render the content as early as possible. In the rich workspaces in files the menubar will not show anyway. So we can easily speed up the loading of the text by postponing the menubar loading.
When opening a file to edit it would be nice if the menubar was snappy... however the content still seems more important.

I did read upon `requestAnimationFrame` but i am not sure how it would apply here. What's the effect you would want to achieve?

vinicius73:
> I did read upon `requestAnimationFrame` but i am not sure how it would apply here. What's the effect you would want to achieve?

Nothing for now @max-nextcloud, maybe in new interactions about performance improvements.

vinicius73 (approved):

juliushaertl (approved):

PR #2231, opened 2022-03-10
More image upload cypress tests

eneiluj:
* check uploaded images are in the attachment folder
* check attachment folder is moved when a markdown file is moved, check images are still in there
* check attachment folder is copied when a markdown file is copied, check images are there and are different than the ones from the original markdown file
* check attachment folder is deleted when a markdown file is deleted


juliushaertl:
Cool 🚀 

One thing that would be generally nice is to get rid of all the different cy.wait calls since this is an anitpattern with cypress. https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting might give some good hints on how to rather wait for the elements to appear or triggered requests to be finished before continuing.

cypress/integration/images.spec.js
```diff
@@ -161,6 +172,86 @@ describe('Test all image insertion methods', () => {
...
+						.should('exist')
+						.should('have.attr', 'data-id')
+						.should('eq', String(attachmentFileNameToId[name]))
+				}
+			})
+	})
+
+	it('test if attachment folder is moved with the markdown file', () => {
+		cy.createFolder('subFolder')
+		cy.wait(2000)
```
vinicius73, line 196: There is a specific reason to use wait here?

eneiluj: The reason was: to make sure the folder is created before going on. Tests were passing locally but failing in a slow context like the CI run.

eneiluj:
@juliushaertl Very true. No more `cy.wait(seconds)`. It's now waiting for expected related requests.

We could wait in the command functions but as it was not done in `deleteFile` for example, I didn't want to risk breaking other tests or having heterogeneous commands logic.

juliushaertl (approved):

vinicius73 (approved):

PR #2236, opened 2022-03-14
Support PHP 8.1

juliushaertl:
PHP unit tests are almost non-existent until now, but let's see if psalm has any complaints

Local tests so far seem good and no deprecation reporting from text itself during regular editing.


vinicius73 (approved):

max-nextcloud (commented):
Great to get this in. Thanks for looking into it.
Code changes look legit.
Have not tested it locally.

PR #2244, opened 2022-03-16
Fix participant popover design

juliushaertl:
Contributes to #1075 

- Always show all participants in dropdown (including yourself so you know your own color)
- Do not show own entry for guests (as there we have the name picker)
- Change tooltip to just say "Participants"
- Align avatars of the list with the guest name editor

Screenshot 2022-03-16 at 16 00 29


src/components/SessionList.vue
```diff
@@ -110,18 +110,18 @@ export default {
...
+		participantsPopover() {
+			if (this.currentSession.guestName) {
+				return this.participantsWithoutCurrent
 			}
-			return tooltipPrefix + this.activeSessions.slice(0, 3).map((session) => session.guestName ? session.guestName : session.displayName).join(', ')
+			return this.participants
 		},
-		activeSessions() {
+		participantsWithoutCurrent() {
+			return this.participants.filter((session) => !session.isCurrent).sort((a, b) => a.lastContact < b.lastContact)
```
vinicius73, line 120: Why this function is sorting the nbUploadingImages again?

juliushaertl: Right, pushed an update as it is already sorted in the participants computed prop.

vinicius73 (approved):

juliushaertl:
Switched to the latest 14.x node version, lets see if that makes the build asset ci happy

PR #2248, opened 2022-03-23
Implement collaborative locking

juliushaertl:
- [x] Lock/unlock a file when opening/closing
- [x] Open files in read only mode when locked
- [x] Indicate other lock in the UI
- [x] Safeguard unlocking through the cleanup cronjob if no close request was sent
- [x] Changing ETag on lock/unlock causes a conflict, so we either need a way to still propagate the change to the root without changing the actual file tag for shares as mentioned in https://github.com/nextcloud/files_lock/pull/54 or adjust the change detection in text to maybe also compare the mtime in addition

To be tested together with https://github.com/nextcloud/files_lock/pull/54


lib/Service/ApiService.php
```diff
@@ -113,11 +109,23 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
 			$this->logger->logException($e, ['level' => ILogger::INFO]);
 			$content = null;
 		}
+
+		$lockInfo = $this->documentService->getLockInfo($file);
+		if ($lockInfo && $lockInfo->getType() === ILock::TYPE_APP && $lockInfo->getOwner() === Application::APP_NAME) {
```
eneiluj, line 114:
```suggestion
		if ($lockInfo !== null && $lockInfo->getType() === ILock::TYPE_APP && $lockInfo->getOwner() === Application::APP_NAME) {
```
If I may suggest.

lib/Service/DocumentService.php
```diff
@@ -453,7 +440,26 @@ public function isReadOnly($file, $token) {
 		} else {
 			$readOnly = !$file->isUpdateable();
 		}
-		return $readOnly;
+
+		$lockInfo = $this->getLockInfo($file);
+		$isTextLock = (
+			$lockInfo && $lockInfo->getType() === ILock::TYPE_APP && $lockInfo->getOwner() === Application::APP_NAME
```
eneiluj, line 446:
```suggestion
			$lockInfo !== null && $lockInfo->getType() === ILock::TYPE_APP && $lockInfo->getOwner() === Application::APP_NAME
```

lib/Service/DocumentService.php
```diff
@@ -490,4 +496,41 @@ private function ensureDocumentsFolder(): bool {
 
 		return true;
 	}
+
+	public function lock(int $fileId): bool {
```
eneiluj, line 500: Missing phpdoc blocks.

lib/Service/DocumentService.php
```diff
@@ -490,4 +496,41 @@ private function ensureDocumentsFolder(): bool {
...
+		if (!$this->lockManager->isLockProviderAvailable()) {
+			return true;
+		}
+
+		$file = $this->getFileById($fileId);
+		try {
+			$this->lockManager->lock(new LockContext(
+				$file,
+				ILock::TYPE_APP,
+				Application::APP_NAME
```
eneiluj, line 510: Out of topic but this could become `Application::APP_ID` at some point.

lib/Service/DocumentService.php
```diff
@@ -490,4 +496,41 @@ private function ensureDocumentsFolder(): bool {
...
+
+		$file = $this->getFileById($fileId);
+		try {
+			$this->lockManager->unlock(new LockContext(
+				$file,
+				ILock::TYPE_APP,
+				Application::APP_NAME
+			));
+		} catch (NoLockProviderException $e) {
+		} catch (PreConditionNotMetException $e) {
```
eneiluj, line 533: No warning logs here?

eneiluj (approved):
LGTM. A few minor remarks.

lib/Service/DocumentService.php
juliushaertl: I'd let the actual lock provider handle that, from the text app side there is no special interest if we unlock a file that has already been unlocked.

lib/Service/DocumentService.php
juliushaertl: Yes, something for a separate PR.

lib/Service/ApiService.php
juliushaertl: You may ;) Adjusted

PR #2249, opened 2022-03-23
debug: ci failures with server errors in cypress

max-nextcloud:
Cypress tests are failing even on master right now.

Since they are failing with a server error... let's try to get some more debug info.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -112,3 +112,11 @@ jobs:
           name: Upload screenshots
           path: apps/${{ env.APP_NAME }}/cypress/screenshots/
           retention-days: 5
+
+      - name: Upload nextcloud logs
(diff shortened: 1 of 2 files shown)


max-nextcloud:
This was intended for ad-hoc debugging - but it seems generally useful to be able to see server errors in case of 500 responses and read some more of the trace. So I think we should just keep this.

juliushaertl:
In addition we could also dump the Nextcloud log after running the cypress tests, so that they are then available in the GitHub action output.

max-nextcloud:
> In addition we could also dump the Nextcloud log after running the cypress tests, so that they are then available in the GitHub action output.

I'm adding them to the uploaded files on failure. That way we can download and open them in an editor of our choice.

vinicius73 (approved):

PR #2264, opened 2022-03-29
Add image upload via drag'n'drop

eneiluj:
This includes those changes:

* add image upload via drag'n'drop from file explorers
* remove image upload by link (because it triggers a network request on the server side which potentially exposes stuff that shouldn't be, the feature is not essential)
* move image insertion and upload methods from MenuBar to EditorWrapper
* allow multiple files selection in upload file input

We imitate what's done style-wise in Files when dragging stuff over: set background color to `--color-primary-light`.


src/components/EditorWrapper.vue
```diff
@@ -99,6 +108,18 @@ import { Step } from 'prosemirror-transform'
 
 const EDITOR_PUSH_DEBOUNCE = 200
 
+const imageMimes = [
```
vinicius73, line 111:
```suggestion
const IMAGE_MIMES = [
```

src/components/EditorWrapper.vue
```diff
@@ -179,6 +200,8 @@ export default {
 			readOnly: true,
 			forceRecreate: false,
 			menubarLoaded: false,
+			nbUploadingImages: 0,
```
vinicius73, line 203:
```suggestion
			uploadingImagesCount: 0,
```

vinicius73 (approved):

mejo- (approved):

PR #2268, opened 2022-04-01
Fix broken file links

eneiluj:
Currently, when opening Text from the files app and clicking a file link (created with the menu bubble), the target URL is `/apps/files/RELATIVE_PATH?fileId=FILE_ID` which does not lead anywhere so we get redirected to the default app.

This simple fix checks if the link points to NC and contains the `fileID` GET param. If so, an internal link is generated and opened.

We could also change what's written in the file's content. Currently it's something like
`[LINK_TEXT](RELATIVE_PATH?fileId=FILE_ID)`
Any thoughts on that?


src/marks/Link.js
```diff
@@ -81,6 +82,11 @@ const Link = TipTapLink.extend({
 									OCA.Viewer.open({ path })
 									return
 								}
+								if (query.fileId) {
+									const directFileLink = generateUrl('/f/' + query.fileId)
+									window.open(directFileLink)
+									return
```
vinicius73, line 88:
```suggestion
									window.open(
									  generateUrl(`/f/${query.fileId}`)
									)
									return
```

vinicius73 (approved):

PR #2278, opened 2022-04-05
WIP: split into src and package

max-nextcloud:
* Target version: master  (after the Nextcloud 24 fork)

## Summary

In order to publish an npm package
and reuse parts of the text code in collectives
split the js source into
* src - for all things text specific, in particular webpack entries
* package - for the package, that is yet to be defined in package.json.


package/index.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import store from './store'
+
+export {
+	store
```
vinicius73, line 26: @max-nextcloud do you have the intention to export more packages here?

src/helpers/files.js
```diff
@@ -23,25 +23,10 @@
 import { openMimetypes } from './mime'
 import RichWorkspace from '../views/RichWorkspace'
 import { imagePath } from '@nextcloud/router'
-import store from '../store'
+import store from '@nextcloud/text-editor/store'
```
vinicius73, line 26:
```suggestion
import { store } from '@nextcloud/text-editor'
```

src/helpers/files.js
```diff
@@ -92,7 +77,7 @@ const registerFileActionFallback = () => {
 				const file = window.FileList.findFile(filename)
 				Promise.all([
 					import('vue'),
-					import(/* webpackChunkName: "files-modal" */'./../components/PublicFilesEditor'),
+					import(/* webpackChunkName: "files-modal" */'@nextcloud/text-editor/components/PublicFilesEditor'),
```
vinicius73, line 80: I guess we will change this approach.

src/main.js
```diff
@@ -1,4 +1,4 @@
-import store from './store'
+import store from '@nextcloud/text-editor/store'
```
vinicius73, line 1:
```suggestion
import { store } from '@nextcloud/text-editor'
```

src/public.js
```diff
@@ -6,7 +6,7 @@ import {
 } from './helpers/files'
 import { openMimetypes } from './helpers/mime'
 import { loadState } from '@nextcloud/initial-state'
-import store from './store'
+import store from '@nextcloud/text-editor/store'
```
vinicius73, line 9:
```suggestion
import { store } from '@nextcloud/text-editor'
```

vinicius73 (commented):
There are some eslint false positives.

package/index.js
max-nextcloud: @vinicius73 Yes - I think it would be nice to make use of this to export everything we use from the app and collectives. This way we have an overview of what is the external api and what is internal.

However for now this was part of debugging the eslint failures.

max-nextcloud:
> There are some eslint false positives.

Yes... i am wondering how to tell eslint that the files are available in the `package` dir.

src/helpers/files.js
max-nextcloud: But how... this is really one of the parts i have no idea of.

I think we will still want to have async imports. But we don't want to import from a nested path inside the package but from the package itself i think.

src/components/ViewerComponent.vue
```diff
@@ -33,7 +33,7 @@
 export default {
 	name: 'ViewerComponent',
 	components: {
-		EditorWrapper: () => import(/* webpackChunkName: "editor" */'./EditorWrapper'),
+		EditorWrapper: () => import(/* webpackChunkName: "editor" */'@nextcloud/text-editor'),
```
vinicius73, line 36: We can move it to a local file and re-export it as a local component.
It will turn easier to adjust lazy load and other possible adjustments.

```js
const EditorWrapper = () => import(/* webpackChunkName: "editor" */'@nextcloud/text-editor')

export default EditorWrapper
```

vinicius73: Also, we can access any exported value inside `@nextcloud/text-editor`.

```js
const EditorWrapper = () => import(/* webpackChunkName: "editor" */'@nextcloud/text-editor')
  .then(package => package.default)
```

It is important because we are using named exports like `store`

```js
import { store } from '@nextcloud/text-editor'
```

Lazy loading the default export and do not use lazy loading in named exports are incompatibles.

To make it, we must use lazy loading in named imports too.

```js
const Store = () => import(/* webpackChunkName: "editor" */'@nextcloud/text-editor')
  .then(package => package.store)
```

But it does not make a lot of sense.
In our architecture, we can't use the async store without changing some parts of code and flow.

To make it possible we need to split the store and components into different parts/files of the `@nextcloud/text-editor` package.
It is possible, but I don't know if webpack will handle it well.

src/helpers/files.js
vinicius73: This is hard do to in actual versions of webpack 😞 
But we can try to create local files to make references to direct components on `'@nextcloud/text-editor'` and lazy load those files.


```js 
// src/components/editor.js

const Editor = () => import(/* webpackChunkName: "editor" */'@nextcloud/text-editor') 

export const PublicFilesEditor = () => Editor().then(package => package.PublicFilesEditor)
```

That is not the best solution, and we need to test how webpack will handle it, but I guess we can try.

max-nextcloud: We also need to take into account the bundle size of the various webpack chunks. I tried making the `EditorWrapper` the default export of the package - but that made all chunks that used the `{ state }` import from the same package grow by more than 1 MB. So right now i think i'd go with just using file references from inside the alias and try and separate the code into the src and package dirs in a meaningful way.

src/components/ViewerComponent.vue
max-nextcloud: Well... it's possible - the store for example already is in a file of its own and importing just that file worked fine. I also noticed that highlight.js also uses paths into the package as more fine grained imports. So that seems to be an somewhat established option. Not pretty but maybe enough to get a first working iteration.

I might just use the file paths into the alias in src and then index.js to define the exports of the package that we can then use in collectives. I think there we just want the tiptap extensions for displaying the tables etc. nicely.

We might also want the editor wrapper to get rid of the magic we use to extract the editor component from the viewer. But that could be a separate step or we could just reference it by path to avoid bloating the read only editor in collectives if we include the `EditorWrapper` in index.js.

max-nextcloud:
Open questions:
* how do we call the package / alias? `@nextcloud/text-editor` for now.

Next steps:
- [ ] create npm package (at least locally)
- [ ] include css
- [ ] allow disabling the `Link` mark in `RichText` extension
- [ ] use `RichText` extension to build the preview
- [ ] include css from text in collectives rather than having a copy
- [ ] use `EditorWrapper` component to build editor in collectives rather than taking it from the view handlers
- [ ] publish 0.1.0 of the package.

package.json
```diff
@@ -11,14 +11,19 @@
 	],
 	"license": "agpl",
 	"private": true,
+	"files": [
```
vinicius73, line 14: also, we need to use [`.npmignore`](https://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html) to prevent non-intentional exports like `/js` folder

max-nextcloud: I think for now we are fine as we explicitly list the dirs we want to include in the files list. I think we'd need .npmignore if we wanted to ignore files from within those dirs.

Issue #2267, opened 2022-03-31
Table tweaks


max-nextcloud:
Some of these were _originally posted by @juliushaertl in https://github.com/nextcloud/text/pull/2235#issuecomment-1077682558_ and _by @max-nextcloud in https://github.com/nextcloud/text/issues/2235#issuecomment-1075993639_

In order of priority:
- [x] make text after a table start below - not next to it on the right #2289 
- [x] hide table editing buttons in read only editor #2289 
- [ ] check if rows and columns can be added / removed and disable buttons accordingly https://github.com/nextcloud/text/issues/2267
- [x] ~~refactor test layout, put all tables fixtures in one folder~~
- [ ] accessibility for adding columns and rows (currently tab moves past the buttons in the cells)
- [ ] drop colspan=1, rowspan=1, etc from rendered html - first attempt at this failed. https://github.com/nextcloud/text/issues/2267
- [x] #3304 (allow left, right and center align of columns)
- [ ] https://github.com/nextcloud/text/issues/2266

luka-nextcloud:
@max-nextcloud 
> refactor test layout, put all tables fixtures in one folder

I don't really understand this point. Currently, the table fixtures is only one file `cypress/fixtures/Table.md`. Do you mean to split this file into many smaller files and put them into a folder?

> drop colspan=1, rowspan=1, etc from rendered html - first attempt at this failed.

I don't really understand this point.
![Image](https://github.com/user-attachments/assets/1b161412-d8f8-4e12-8b63-4a216c143729)

max-nextcloud:
> I don't really understand this point. Currently, the table fixtures is only one file `cypress/fixtures/Table.md`. Do you mean to split this file into many smaller files and put them into a folder?

I don't remember either.
I think i was referring to the fixtures in `src/tests/fixtures/tables` - maybe to combine them with the ones in the cypress fixture, or to change the folder structure. Now there are two subfolders for two different scenarios - which makes sense from my point of view. So I crossed out this point and checked it off.

max-nextcloud:
> I don't really understand this point.

https://github.com/nextcloud/text/blob/6453780dffd25630c2ccec6c123ca137875769a3/src/tests/fixtures/tables/handbook/handbook.out.html#L6 has the output of `tiptap.getHTML()` from an editor with tables from the old handbook source.

It has `colspan="1"` and `rowspan="1"` all over the place - even though that's not needed - as it's the default.

[...]

PR #2289, opened 2022-04-11
Fix/2267 table tweaks

max-nextcloud:
* See: #2267 
* Target version: master 

### Summary

Fixes the two main ui issues:
* make sure text after the table starts on a new line
* hide actions in read only view


juliushaertl:
@max-nextcloud Feel free to adjust the label to "to review" once ready :)

max-nextcloud:
@juliushaertl Good point... let's get this in as it affects the ui and would be nice to have for screencasts etc. The other remaining todo items in the issue are less important for now.

PR #2292, opened 2022-04-12
Changes menu entry.

pierreozoux:
We realized with our users that when the apps Text and ONLYOFFICE are installed in parallel, this is a bit confusing.

You can see a screenshot here, in french:
https://forge.liiib.re/indiehost/tech/plateforme/-/issues/217#note_9733

So we propose this change.

In relation to this change in [onlyoffice app](https://github.com/ONLYOFFICE/onlyoffice-nextcloud/pull/617).


max-nextcloud:
Hi @pierreozoux 

Thanks a lot for your proposal. I'm a bit unsure about using 'Note' instead as there is a Notes app - so this might be confusing if you create a note somewhere and then it does not show in the notes app (because that app only works on one folder).

I agree that `New text document` is confusing. For me it's mostly the document part. That sounds like something formal on a page with layout and all that.

My personal preference would be just `New text` - do you think that would make it easier to distinguish?

I'd be really curious what @jancborchardt and @nimishavijay think.

pierreozoux:
@imacrea what do you think?

ImaCrea:
> @ImaCrea what do you think?

What about `New pad` @max-nextcloud ?

juliushaertl:
The text app is advertised and documented as a text editor so from my perspective either "New text document" or "New text file" are the most fitting terms. Pad is not really a term that is understandable for non-technical users from my perspective and note is something that may add a wrong connection to the notes app, but I would appreciate feedback from @jancborchardt and @nimishavijay here as well :)

ImaCrea:
> The text app is advertised and documented as a text editor so from my perspective either "New text document" or "New text file" are the most fitting terms. Pad is not really a term that is understandable for non-technical users from my perspective and note is something that may add a wrong connection to the notes app, but I would appreciate feedback from @jancborchardt and @nimishavijay here as well :)

Understood. `New text file` sounds good to me then. It's the 'document' part of the current wording that can be misleading when office suite like collabora or onlyoffice are integrated.

nimishavijay:
I can understand how it could be confusing when there is a "New document" right underneath. 
"New text" seems a bit incomplete, and "New pad" won't be understood by many as I've seen it being used only with EtherPad, so I would vote for "New text file" :)

jancborchardt:
Yep, agree with the points made by @max-nextcloud @juliushaertl and @nimishavijay. Let’s go for "New text file". :)

src/helpers/files.js
```diff
@@ -55,8 +55,8 @@ const registerFileCreate = () => {
 			// register the new menu entry
 			menu.addMenuEntry({
 				id: 'file',
-				displayName: t('text', 'New text document'),
-				templateName: t('text', 'New text document') + '.md',
+				displayName: t('text', 'New note'),
+				templateName: t('text', 'New note') + '.md',
```
juliushaertl, line 59:
```suggestion
				displayName: t('text', 'New text file'),
				templateName: t('text', 'New text file') + '.md',
```

juliushaertl (changes requested):
See inline comment for the wording as per discssion, in addition the wording in https://github.com/nextcloud/text/blob/551729e08d7c044ae34d8972129caa448725f7a9/lib/AppInfo/Application.php#L71 will also need to be adjusted.

juliushaertl:
Looks good for the frontend part, could you also adjust that accordingly in https://github.com/nextcloud/text/blob/551729e08d7c044ae34d8972129caa448725f7a9/lib/AppInfo/Application.php#L71 ?

pierreozoux:
@juliushaertl I think I did all the changes, looks goot to :ship: to me :)

max-nextcloud:
I rebased on top of the current master and rebuild the js. Hope that tests pass now and we can merge after.
Thanks a lot for your contribution @pierreozoux

max-nextcloud:
cypress failure is not related.Will look into that separately.

pierreozoux:
@max-nextcloud do you think it is possible to backport this little change? Or is it a "big" change, and will have to wait 25th?

max-nextcloud:
I'd be okay with backporting this - I'll open a backport pr to have the discussion there.

PR #2294, opened 2022-04-13
fix: keep menubar in one line - prioritize entries

max-nextcloud:
target version: master, NC24

Menubar entries with submenus cannot be moved to the more-button.
Prioritize them when picking the icons for the toolbar.
If they still do not fit - do not add them to the more button either
they do not work there.

This makes sure that the menubar fits in one line and does not overflow.

Still prioritizing `undo` and `bold` over `emoji` and `callout`.
The former ones are need fairly frequently while the latter not so often.
In addition mobile phones have an emoji picker in their text entry.


max-nextcloud:
Prioritized callout and emoji so they are available even on small screens as recommended by @jancborchardt in chat:
> It’s better to have all things available but a bit moved, than to have some not available

src/components/MenuBar.vue
```diff
@@ -68,14 +69,14 @@
...
-					v-show="$index < iconCount"
+					v-show="icon.priority <= iconCount"
 					:key="icon.label"
 					v-tooltip="getLabelAndKeys(icon)"
 					:class="getIconClasses(icon)"
 					:disabled="disabled(icon)"
 					@click="clickIcon(icon)" />
 				<template v-else>
-					<div v-show="$index < iconCount || !icon.class"
+					<div v-show="icon.priority <= iconCount"
```
vinicius73, line 79: **Question:** can we use `v-if` instead of `v-show`?

```suggestion
					<div v-if="icon.priority <= iconCount"
```

src/mixins/menubar.js
```diff
@@ -220,12 +233,25 @@ export default [
...
+	},
+	{
+		label: t('text', 'Insert image'),
+		class: 'icon-image',
+		priority: 2,
+	},
+	{
+		label: t('text', 'Formatting help'),
+		class: 'icon-help',
+		click: (view) => view.$emit('show-help'),
```
vinicius73, line 254: **Suggestion:** Just a small comment
```suggestion
		// receive component instance to handle custom actions
		click: (view) => view.$emit('show-help'),
```

vinicius73 (approved):

mejo- (approved):

PR #2303, opened 2022-04-18
📌 explicit latest versions

vinicius73:
* Target version: master 

### Summary

Just make explicit we are using the updated versions of dependencies.


juliushaertl (approved):
Sure, I guess dependabot would bump them anyways with future bumps then to be in sync...

mejo- (approved):

vinicius73:
> Sure, I guess dependabot would bump them anyways with future bumps then to be in sync...

We can use `versioning-strategy: increase`, looks like it will do what we are talking about.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

PR #2306, opened 2022-04-19
Fix matrix variable reference in Github static-analysis action

mejo-:

* Resolves: # 
* Target version: master


max-nextcloud:
Awesome! Thanks for fixing this! :heart_eyes:

max-nextcloud (approved):

PR #2307, opened 2022-04-19
Fix position of dropped images

eneiluj:
The cursor position was not set to the drop position so the images were always inserted at the current cursor position.

This required to handle the drop event in the Image node, like we do with the Paste event. The position is then passed to the editor wrapper via a custom event.


mejo- (approved):
Yay, nice :heart_eyes: Works like a charm.

PR #2308, opened 2022-04-19
Fix matrix variable reference in Github static-analysis action

mejo-:
Backport of #2306.


mejo-:
The failure of "Static analysis" job is due to `christophwurst/nextcloud` missing a `dev-stable24` release. The advice is to ignore this for now and just go ahead with the merge :elephant:

PR #2315, opened 2022-04-19
fix: adjust menubar on sidebar toggling

max-nextcloud:
* Target version: master, stable24

### Summary
When the sidebar is opened or closed the width of the menubar changes. Adjust the entries so they still fit it nicely.

#### Before:

https://user-images.githubusercontent.com/97337118/164085267-8b5957af-8755-4b35-9bac-170d5bb05a61.mp4

#### After:

https://user-images.githubusercontent.com/97337118/164085264-aa5c72d8-a20a-417c-ad97-9880ea3d0d56.mp4


src/components/MenuBar.vue
```diff
@@ -284,8 +287,14 @@ export default {
 	},
 	beforeDestroy() {
 		window.removeEventListener('resize', this.getWindowWidth)
+		unsubscribe('files:sidebar:opened', this.redrawAfterTransition)
+		unsubscribe('files:sidebar:closed', this.redrawAfterTransition)
 	},
 	methods: {
+		redrawAfterTransition() {
+			// wait for transition to complete (100ms)
+			setTimeout(this.redrawMenuBar, 110)
```
max-nextcloud, line 296: It's somewhat funny that this is necessary. Files sidebar is based on nc vue and that says 'opened' and 'closed' are emitted AFTER the transition:
https://github.com/nextcloud/nextcloud-vue/blob/e6bb18756ebf2c00d4f7e37fe7faaa4340ef8c13/src/components/AppSidebar/AppSidebar.vue#L439
But if i don't put a delay redraw menu bar is called when the menubar still has the old size.

mejo-:
Mh, I failed to reproduce the described issue with either Firefox (99.0.1) or Chromium (100).

But for me, resizing the window results in the first (or the first two) tools becoming invisible in some cases - both on master branch and on this PR branch - and both on Firefox and on Chromium:

## Screencast with Text from this branch/PR

https://user-images.githubusercontent.com/3582805/164241847-d173f1ff-d267-456f-b035-5e1ef23bb48d.mp4

max-nextcloud:
> Mh, I failed to reproduce the described issue with either Firefox (99.0.1) or Chromium (100).

Interesting... so even with current master you don't see menu items overflowing into the next row when opening the sidebar?

> But for me, resizing the window results in the first (or the first two) tools becoming invisible in some cases - both on master branch and on this PR branch - and both on Firefox and on Chromium:

I suspect the menubar is exceeding the visible area on the left. The Calculation is based on the actual size of the menubar - even if not all of it is visible. I'll take another look at this.

mejo-:
> > Mh, I failed to reproduce the described issue with either Firefox (99.0.1) or Chromium (100).
> 
> Interesting... so even with current master you don't see menu items overflowing into the next row when opening the sidebar?

Exactly.

mejo- (approved):
With @max-nextcloud help I was able to reproduce :zany_face:

Issue #2097, opened 2022-01-14
Add support for autolink (tiptap v2)


ya-d:
[Tiptap v2 adds support for autolink](https://github.com/ueberdosis/tiptap/issues/2087) and creates hyperlinks automatically when URLs are inserted by typing oder pasting. 

It would be nice if Text would support that as well. Thanks for considering!

Related to #1906 #1981

mejo-:
Interesting, this should be enabled per default with tiptap v2 (see https://tiptap.dev/api/marks/link#autolink), but for some reason it doesn't work in latest Text :thinking:

Seems like our custom `addProseMirrorPlugins()` in https://github.com/nextcloud/text/blob/master/src/marks/Link.js#L55 overwrites the default plugins autolink, onOpenClick and linkOnPaste. For linkOnPaste, this probably even is a regression in Nextcloud 24.

vinicius73:
Yes @mejo- , we must fix it.
I will work on it.

PR #2316, opened 2022-04-19
Bugfix: add support to auto link

vinicius73:
* Resolves: #2097
* Target version: master, stable24

### Summary

Auto-link is included in the original tiptap extension-link.

We should keep the original behavior as much as possible.


src/plugins/link.js
```diff
@@ -0,0 +1,55 @@
...
+	return new Plugin({
+		props: {
+			handleClick: (view, pos, event) => {
+				const linkElement = event.target.parentElement instanceof HTMLAnchorElement
+					? event.target.parentElement
+					: event.target
+
+				const isLink = linkElement && linkElement instanceof HTMLAnchorElement
+
+				const htmlHref = linkElement?.href
```
vinicius73, line 15: I've changed this part of the original code to make it possible to run tests inside Cypress.

Originally we used `editor.getAttributes('link')` but it not works in cypress

When we interact natively/manually with the editor, the original events "bubble" from the `span` tag, but in cypress tests, the event comes from `a` tag, and this tag doesn't have attributes in the editor.
It makes the automated test impossible to work, because of this I've changed the code to use the native properties of the link.

cypress/integration/links.spec.js
```diff
@@ -0,0 +1,71 @@
...
+					getEditor()
+						.type(link)
+
+					getEditor()
+						.get(`a[href*="${Cypress.env('baseUrl')}`)
+						.click({ force: true })
+
+					cy.get('@winOpen')
+						.should('have.been.calledOnce')
+						.should('have.been.calledWith', `/index.php/f/${id}`)
```
eneiluj, line 54: This fails if NC is accessible with a sub path as the call will be like `/sub/path/index.php/f/${id}`. Not a big deal but this test may fail in the future if we change how NC is deployed in the CI job.

We could either check that the target of `@winOpen` ends with `/index.php/f/${id}` or get the sub path and include it in the expected value.

eneiluj (approved):
Works like a charm!

juliushaertl:
@vinicius73 Mind to take care of the manual backport then? :)

Issue #2317, opened 2022-04-20
Drag and drop of anything except images is broken


mejo-:
Unfortunately, #2307 broke dragging and dropping anything but images into the editor. E.g. it's no longer possible to drag and drop a hyperlink into the editor.

Cc @eneiluj - if you don't have time to look into it, I could take a look tomorrow or on Monday :relaxed:

PR #2329, opened 2022-04-25
Fix drag&drop of non-image objects

mejo-:
Only catch drop with custom image-drop event if it contains files.

Fixes: #2317

* Resolves: #2317
* Target version: master


eneiluj (approved):
:see_no_evil: Sorry for this mistake.
Nice fix :heart:

PR #2356, opened 2022-05-01
ci: only run php unit tests when php files changed

max-nextcloud:
They take 10 minutes to setup nextcloud
and block available CI runners.

This is particularly wasteful with javascript dependency updates.
Dependabot will rebase the pending updates
upon every push to the underlying branch.

* Resolves: # 
* Target version: master

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -2,10 +2,26 @@ name: PHPUnit
 
 on:
   pull_request:
+    paths:
+      - .github/workflows/phpunit.yml
+      - appinfo/**
+      - composer.*
+      - lib/**
+      - psalm.xml
+      - templates/**
+      - tests/**
   push:
     branches:
       - master
       - stable*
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
Good point 👍

max-nextcloud:
@juliushaertl frankly I think the backports do not matter so much. Most PR's run against master and master is getting updated frequently so that dependabot rebases also happen quite often. So having this in master will already reduce ci times significantly.
It might be good for consistency to apply it to older branches - but i don't think it's really needed. I'm thinking about doing the same for nextcloud/server - i think that would save more CI minutes with a comparable effort to backporting it in text.

max-nextcloud:
Looks like i've been wrong about stable24. Tiptap has a lot of small packages in a monorepo that all get bumped at once. And stable24 is still up to date with them. So we get a lot of stable24 updates for tiptap. I'll start a backport of this to stable24.

PR #2359, opened 2022-05-04
fix: only show delete image button when editing

max-nextcloud:

* Target version: master 

### Summary
We used to show the delete image button even in read only views - such as a public share without editing permission.
Only show it when people are actually allowed to remove the image.


src/tests/nodes/ImageView.spec.js
```diff
@@ -13,6 +13,7 @@ describe('Image View src attribute based on markdown', () => {
 	const factory = attrs => {
 		const propsData = {
 			extension: { options: {currentDirectory: '/current'} },
+			editor: { isEditable: true },
```
vinicius73, line 16: Do we have a test case for when this value is false?

eneiluj (approved):
Fine by me!

Issue #2338, opened 2022-04-26
Don't prepend timestamp to image filename


mejo-:
When a new image gets added to the document, we shouldn't prepend filename with a timestamp. Instead, filename conflicts should be resolved by appending ` (2)` etc. when a file already exists.

As discussed in the design review call, see https://gitlab.com/collectivecloud/collectives/-/issues/322 for details.

susnux:
Looks like this is fixed with https://github.com/nextcloud/text/pull/2376 ?

mejo-:
Yes, indeed. Thanks for checking @susnux :heart:

PR #2376, opened 2022-05-09
Use '(n)' suffix instead of timestamp prefix for uploaded image names

eneiluj:
refs #2338

If `foo.png` is uploaded but already exists in the attachment folder, file is named `foo (2).png`. If uploaded again, `foo (3).png`. And so on.

### DIFF
--- a/lib/Service/ImageService.php
+++ b/lib/Service/ImageService.php
@@ -152,7 +152,7 @@ class ImageService {
 			throw new NotPermittedException('No write permissions');
 		}
 		$saveDir = $this->getAttachmentDirectoryForFile($textFile, true);
-		$fileName = (string) time() . '-' . $newFileName;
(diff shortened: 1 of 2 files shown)


vinicius73:
@eneiluj it's possible to create a test for it to prevent regressions?

mejo- (approved):
Nice, works like a charm. Thanks a lot @eneiluj for looking into it :purple_heart:

One day we should add such a `getUniqueFilename()` function to Nextcloud itself - I already know of at least four independent implementations for that :zany_face:

eneiluj:
@vinicius73 Yes I'll create a test, thanks for the suggestion.

vinicius73 (approved):

mejo-:
We probably should backport this to stable24, no?

eneiluj:
@mejo- Yep!
And thanks for the review!

Issue #2339, opened 2022-04-26
Allow to upload non-image files


mejo-:
We want to extend the functionality to upload images (either via toolbar or via drag&drop) to allow uploading non-image attachments.

When uploading a non-image attachment, a link to it should be added to the document.

As discussed in the design review call, see https://gitlab.com/collectivecloud/collectives/-/issues/322 for details.

eneiluj:
Hey, so how do we want to make that happen exactly?

* How should other file types be displayed in Text? Generic type icon or preview if available?
* Do we want to open the viewer on click for compatible file types like PDFs or videos?

More questions will come during the implementation :grin:.

mejo-:
In a private discussion with @juliushaertl, we came to the conclusion that it might be rather confusing to allow non-image attachments in the standalone text editor. The most common usecase for standalone text is probably that it's launched from the Files app, and there people expect to have "control" over their files.

So it might not be the best UX to expose a feature that allows to upload files which then disappear in a hidden folder and are removed automatically once their reference is removed from the markdown file. Especially as there's no interface to list the attachment files in standalone text.

The situation is different if the Collectives app is the primary user interface. There, users will have an attachments list to browse the attachments, and the attachments are tied to a page UI-wise.

IIRC, our idea implementation-wise was to alllow non-image attachment uploads in the Text backend, but only expose the feature in the UI when inside a collective.

@juliushaertl please add if I missed something from our brainstorming :wink:

This is all not set in stone and I'd like to hear other opinions on it :speech_balloon:

mejo-:
This already got implemented in #2378 thanks to gorgeous @julien-nc :blush:

PR #2378, opened 2022-05-09
Non-image media attachments

eneiluj:
refs #2339

Allow non-image media upload. Some details needs to be discussed before moving forward.


eneiluj:
@mejo- @max-nextcloud @vinicius73 @juliushaertl 

This is now based on latest master which includes the material icon PR :grin:.
:warning: compiled code is not included.

Before moving forward, do we agree on the following aspects?

1. We could just rename image related entries in the menubar `s/image/media/` as we want to let users insert any kind of file via: the menubar, the paste event, the drop event
2. Images and Medias (any other mimetype) can be handled by the same controller/service on the server side?
    * We could rename `ImageController` => `MediaController` 
    * We could rename `ImageService` => `MediaService`
3. We have to choose how to write media files in the markdown content (and then make the magic happen in Tiptap's parsing mechanism)
4. We will handle the media display in a dedicated node (pretty much like ImageView). It could look like the file list items in Files.

What's been implemented for now:
* It's possible to upload any kind of file
     * If it's an image, it's done like before
     * If not, then it's almost inserted like an image except the link target ends with `#media`
* ImageView calls different endpoints when the link target ends with `#media`
* We now have 3 endpoints to serve image/media stuff:
    * one to get an image (it existed before), it provides a preview if possible, the real image content if not
    * one to get a media file, it provides the file content
    * one to get a media file preview, it provides a file preview if possible, the mimetype icon URL if not

What's weird but not definitive :stuck_out_tongue_winking_eye: :
* There's no styling when ImageView display a media file as it will be done somewhere else later
* The attachment cleanup does not work anymore because the trailing `#media` breaks it. It will be adjusted when we decide how to write media files in the markdown content

TODO
* [x] Implement server side logic
* [ ] Implement the parsing magic for media attachments
* [ ] Implement a MediaFileView node to display media attachments
* [ ] Adjust the attachment cleanup methods

Let's go!

jancborchardt:
@eneiluj the delete icon looks a bit off, it seems to be a community one from materialdesignicons.com (please try not to use that, only when the official site has no icon) – use this instead :) https://fonts.google.com/icons?selected=Material%20Icons%3Adelete%3A

Regarding the layout of the element itself, it would be a bit nicer if the element doesn’t go full width. The size info could go into the subline of the element, using `color: var(--color-text-maxcontrast);`, and the delete icon on the right of it immediately. Since the delete icon is only shown on hover/focus it is fine to have it at different positions.

Example mockup like it is in Mail:
image

eneiluj:
@jancborchardt Thanks for the feedback.
* Delete icon is now the official one
* We can move it in another PR, there were no changes made to the icon position
* It now looks like that (2 files and one image):
image

What about alignment? It's currently centered so the icons are not vertically aligned. Would it make sense to left-align the media attachments? But then do we leave the images centered?

@vinicius73 I think I addressed all your remarks, thanks! Rebased on master, tests are finally passing.

src/services/ImageResolver.js
```diff
@@ -50,35 +51,70 @@ export default class ImageResolver {
...
+				{
+					type: 'image',
+					url: this.#davUrl(src),
+				},
+				{
+					type: 'image',
+					url: this.#getImageAttachmentUrl(imageFileName),
+				},
+				{
+					type: 'media',
```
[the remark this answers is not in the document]
eneiluj: They are now attributes of the AttachmentResolver class.

src/nodes/ImageView.vue
```diff
@@ -212,24 +233,31 @@ export default {
...
+		async loadImage(imageUrl, attachmentType, name = null) {
 			return new Promise((resolve, reject) => {
 				const img = new Image()
 				img.onload = () => {
 					this.imageUrl = imageUrl
 					this.imageLoaded = true
 					this.loaded = true
+					this.attachmentType = attachmentType
+					console.debug('SUCCESS type', attachmentType)
+					if (attachmentType === 'media') {
```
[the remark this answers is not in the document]
eneiluj: I'm open to commits on top of mines :grin:. It is not as simple as it seems because we need the file name to load the metadata.

src/components/Editor.provider.js
```diff
@@ -78,10 +78,10 @@ export const useFileMixin = {
...
-export const useImageResolver = {
+export const useAttachmentResolver = {
 	inject: {
-		$imageResolver: {
-			from: IMAGE_RESOLVER,
+		$attachmentResolver: {
+			from: ATTACHMENT_RESOLVER,
 			default: {
 				resolve(src) {
 					console.warn('No image resolver provided. Some image sources cannot be resolved.')
```
vinicius73, line 87:
```suggestion
					console.warn('No attachment resolver provided. Some attachment sources cannot be resolved.')
```

src/components/Menu/entries.js
```diff
@@ -322,7 +322,7 @@ export default [
 		key: 'insert-image',
 		label: t('text', 'Insert image'),
 		icon: Images,
```
vinicius73, line 324: Should we change it to attachments?

src/nodes/Image.js
```diff
@@ -53,7 +53,7 @@ const Image = TiptapImage.extend({
 						// only catch the drop if it contains files
 						if (event.dataTransfer.files && event.dataTransfer.files.length > 0) {
 							const coordinates = view.posAtCoords({ left: event.clientX, top: event.clientY })
-							const customEvent = new CustomEvent('image-drop', {
```
vinicius73, line 56: Should we move this to a more specific plugin?

src/nodes/ImageView.vue
```diff
@@ -124,15 +142,15 @@ export default {
 	name: 'ImageView',
```
vinicius73, line 142: What do you think about `MediaView` or `AttachmentView`?

src/nodes/ImageView.vue
```diff
@@ -238,6 +267,11 @@ export default {
 				img.src = imageUrl
 			})
 		},
+		loadMediaMetadata(name) {
+			this.$attachmentResolver.getMediaMetadata(name).then((response) => {
+				this.attachmentMetadata = response.data
+			})
```
vinicius73, line 273:
```suggestion
			return this.$attachmentResolver.getMediaMetadata(name).then((response) => {
				this.attachmentMetadata = response.data
			})
```

src/nodes/ImageView.vue
```diff
@@ -212,24 +235,30 @@ export default {
...
 			return new Promise((resolve, reject) => {
 				const img = new Image()
 				img.onload = () => {
 					this.imageUrl = imageUrl
 					this.imageLoaded = true
 					this.loaded = true
+					this.attachmentType = attachmentType
+					if (attachmentType === this.$attachmentResolver.ATTACHMENT_TYPE_MEDIA) {
+						this.loadMediaMetadata(name)
+					}
```
vinicius73, line 261: Wait for `loadMediaMetadata`  is not relevant?

vinicius73 (commented):
LGTM;

I did not test yet.
Despite of my comments, everything seems good.
Thanks!

jancborchardt:
> Would it make sense to left-align the media attachments? But then do we leave the images centered?

@eneiluj yes to both. :) Would look better to left-align non-images, and better to keep images centered.

(Btw, image filenames can be styled as color-text-maxcontrast since the image is more relevant than the text.)

src/components/Editor.provider.js
eneiluj: :+1:

src/nodes/Image.js
eneiluj: I don't know because it still will be written with the image syntax in the markdown content.

src/nodes/ImageView.vue
eneiluj: Same here, the ImageView node still renders the markdown image elements.

src/nodes/ImageView.vue
eneiluj: :+1:

src/nodes/ImageView.vue
eneiluj: Yes, better indeed. `loadImage` is now resolved once we got the metadata.

eneiluj:
@jancborchardt Here is how it looks now:

https://user-images.githubusercontent.com/11291457/186871688-e2f8866c-3206-49e9-8e6b-9fcef8af737b.mp4

* Medias are left aligned
* The delete icon is now in NcButton
* The image name color is now color-text-maxcontrast (but not in the video).

eneiluj:
Failin Cypress test is not related (sharing).

src/nodes/ImageView.vue
```diff
@@ -212,24 +251,30 @@ export default {
...
-				if (fallback) {
-					return this.loadImage(fallback)
+			const candidates = this.$attachmentResolver.resolve(this.src)
+			return this.load(candidates)
+		},
+		async load(candidates) {
+			const candidate = candidates.shift()
+			return this.loadImage(candidate.url, candidate.type, candidate.name).catch((e) => {
+				if (candidates.length > 0) {
+					return this.load(candidates)
```
max-nextcloud, line 261: nice use of recursion and much cleaner than before. :)

src/services/AttachmentResolver.js
```diff
@@ -43,42 +47,77 @@ export default class ImageResolver {
 	/*
 	 * Resolve a given src.
 	 * @param { string } the original src in the node.
-	 * @returns { Array<string> } - resolved urls to try.
+	 * @returns { Array<Array> } - resolved candidates to try.
```
max-nextcloud, line 50: This returns an Array of Objects, right?

src/services/AttachmentResolver.js
```diff
@@ -99,6 +138,40 @@ export default class ImageResolver {
...
+		return generateUrl('/apps/text/mediaMetadata?documentId={documentId}&sessionId={sessionId}&sessionToken={sessionToken}&mediaFileName={mediaFileName}&shareToken={shareToken}', {
+			...this.#textApiParams(),
+			mediaFileName,
+			shareToken: this.#shareToken,
+		})
+	}
+
+	getMediaMetadata(mediaFileName) {
+		return axios.get(this.#getMediaMetadataUrl(mediaFileName))
+	}
```
max-nextcloud, line 173: Could we maybe move the call to `axios` to the calling function? So far this class just encapsulates the logic of generating the urls. Without the axios call it has no side effects - which makes it very easy to test and reason about.

That might also make it easy to add a test for the logic generating the url.

max-nextcloud (commented):
Looks great!
So far I just have some minor comments from a first pass over the code.

I still want to:
* [ ] give this a try locally
* [ ] take a closer look at the changes to the backend
* [ ] think through possible attack vectors on the file copying functions

Submitting this now never the less for some initial feedback.

src/nodes/ImageView.vue
eneiluj: Thanks!

src/services/AttachmentResolver.js
eneiluj: Oops yes, fixed.

src/services/AttachmentResolver.js
eneiluj: I completely agree. Done.

Metadata url generation is now tested.

max-nextcloud: thanks

src/nodes/ImageView.vue
```diff
@@ -34,27 +34,59 @@
...
 							type="text"
 							class="image__caption__input"
 							:value="alt"
 							@keyup.enter="updateAlt">
-						<div v-if="editor.isEditable && showIcons"
-							class="image__caption__delete"
-							title="Delete this image"
-							@click="deleteNode">
-							<TrashCan />
+						<div v-if="editor.isEditable && showIcons && !isMediaAttachment"
```
vinicius73, line 81: We can turn it into a computed.

src/nodes/ImageView.vue
```diff
@@ -212,24 +252,30 @@ export default {
...
 		async init() {
-			const [url, fallback] = this.$imageResolver.resolve(this.src)
-			return this.loadImage(url).catch((e) => {
-				if (fallback) {
-					return this.loadImage(fallback)
+			const candidates = this.$attachmentResolver.resolve(this.src)
+			return this.load(candidates)
+		},
+		async load(candidates) {
+			const candidate = candidates.shift()
```
vinicius73, line 259: I liked, but we are mutating the `candidates` argument.
It can generate unexpected behaviors in the future.

```suggestion
		async load(candidates) {
		    candidates = [...candidates] // prevent argument mutation
			const candidate = candidates.shift()
```

What do you think?

src/nodes/ImageView.vue
```diff
@@ -238,6 +284,16 @@ export default {
 				img.src = imageUrl
 			})
 		},
+		loadMediaMetadata(name) {
+			const metadataUrl = this.$attachmentResolver.getMediaMetadataUrl(name)
+			console.debug('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa', metadataUrl)
```
vinicius73, line 289:
```suggestion
```

src/nodes/ImageView.vue
max-nextcloud: while we are at it...how about:
```js
const [candidate, ...rest] = candidates
```
So the meaning of candidates does not change within the function.

Or even

```js
...
  load(...candidates)
...
async load(candidate, ...fallbacks) {
  ...
  load(...fallbacks)
}
```

jancborchardt:
cc @juliushaertl we could slightly adjust the styles here and in the OpenGraph embedding PR.

@eneiluj I think the only things I would still change are:
- On files, a `border: 2px solid var(--color-border)` on the files for a nice framing (not extending the whole width, just nicely framing them)
- On the image, the delete button could be directly next to the image filename, currently it’s a bit floating in the air
- Said before, image filenames can be styled as color-text-maxcontrast since the image is more relevant than the text. :)

juliushaertl:
Images are special as the filename field is editable and is used as the alt text of the image on markdown.

src/nodes/ImageView.vue
eneiluj: Right, done.

src/nodes/ImageView.vue
eneiluj: @vinicius73 :sweat: Nice catch.
@max-nextcloud I prefer your first one which is slightly more explicit.
Done.

eneiluj:
@jancborchardt Thanks.
* Border is set for medias, is it fine to leave the delete button outside the border? I think it looks better than if it's inside because we can keep the border fixed when the button appears.
* Image delete button: This was harder than I thought. I had to add a parent element with `position: relative` to be then able to use `position: absolute` for the button so it does not "push the caption away" when it appears.
* Caption color (it was already done but not effective in the previous video). I don't have any opinion on that one but I agree with the remark from @juliushaertl, this is a special case... As you wish.


https://user-images.githubusercontent.com/11291457/187412429-17f1d399-e4e4-4133-993f-e2a44d232dc7.mp4

jancborchardt:
Looks great @eneiluj! The image (filename) doesn't need the border. :) Otherwise good to go design-wise!

jancborchardt:
> Images are special as the filename field is editable and is used as the alt text of the image on markdown.

Ah it's an input field? Can we only show the border on hover & focus @eneiluj?

eneiluj:
@jancborchardt hover&focus of the input or of the whole image wrapper (so the input border appears even when hovering the image itself)?

eneiluj:
@vinicius73 @max-nextcloud Some news, there's a new Cypress attachment test. I did a bit of renaming there.

jancborchardt:
@eneiluj it's the input field border, so I'd say should just appear on hover/focus of the input. :)

eneiluj:
@jancborchardt It's kind of nice that it appears at the same time the button does. What do you think?

https://user-images.githubusercontent.com/11291457/187433368-e09f6f3a-04f3-4a46-9a45-6f680e797c47.mp4

jancborchardt:
@eneiluj you're absolutely right! Let's do it like that :)

jancborchardt (approved):

eneiluj:
Cypress failure seems unrelated (share.spec).
@vinicius73 @max-nextcloud Ready IMO.

max-nextcloud:
@eneiluj I really need to improve my skills on creating screen casts with audio. Those are great fun!

max-nextcloud:
Had another look at the code and also failed to exploit the file copy. It's checking the path so one cannot escape the own home dir. So all that looks good.

Ran into one issue when playing with it locally though:
https://user-images.githubusercontent.com/97337118/188385405-8bd115dc-fd8e-4c6b-bd93-0fdf541f9b7b.mp4

Worst part.... i could not even insert an emoji there :wink: . Would be okay with merging this anyway and fixing afterwards.

eneiluj:
> I really need to improve my skills on creating screen casts with audio

@max-nextcloud The song you chose perfectly matches the screencast purpose :joy: Thank you for this epic moment!
I can't reproduce though. I don't know why inline emoji insertion is never working on my setup...
Anyway this is a more general issue with the ImageView node. Let's fix that later.

I pushed one more commit (https://github.com/nextcloud/text/pull/2378/commits/f58afa1dbcced49cf46f76657508b4caac3778ef) which improves the situation in some edge cases: When PHP's post_max_size is exceeded, the request is truncated and the authentication params (documentId, sessionToken...) might not be included in it so the controller complains about the session not being found when it's just about the POST content being too big. This can be solved by passing the auth params as GET-like ones which is perfectly handled by our middleware.

Any objection on using GET-like params in a POST request? If not, we can merge :tada: .

max-nextcloud (approved):

max-nextcloud:
I wonder if the cypress failures are due to this... maybe because the urls changed now that the parameters are in there...

PR #2382, opened 2022-05-10
Fix attachment cleanup when file names contain parenthesis

### LINKED ISSUE #2338: Don't prepend timestamp to image filename

eneiluj:
To reproduce the issue (and check it works fine this fix :grin:):
* create a new markdown document and open it
* upload the same image twice so the second one is named `image (2).png`
* close the document
* check if `image (2).png` has been deleted in the attachment folder

The regex we were using did not match file names with parenthesis. The code is now a bit more complex but I couldn't find a way to make it work with one big single regex. So it first gets all the link targets and then filter to keep the `text://image?imageFileName=blabla.png` ones.

This should be backported to stable24 as it fixes #2338 .

### DIFF
--- a/lib/Service/ImageService.php
+++ b/lib/Service/ImageService.php
@@ -478,18 +478,29 @@ class ImageService {
 	 */
 	public static function getAttachmentNamesFromContent(string $content): array {
 		$matches = [];
+		// get first level of parenthesis after ![.*]
+		$regex = '/!\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[\])*\])*\])*\])*\])*\])*\](\(((?>[^()]+)|(?-2))*\))/';
+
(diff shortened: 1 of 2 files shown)


eneiluj:
I forgot to mention the limitations of this regex: If the nested parenthesis are not coherent, this might not work.

By the way, I discovered that Php's `urlencode()` actually encodes parenthesis but Javascript's `encodeURIComponent()` does not.

If we manage to encode the parenthesis when generating the file name to be embedded in the MD content, we can stick with the old cleanup process. To be continued.

mejo-:
Thanks a lot @eneiluj for fixing this in #2383 <3

So this one can be closed then, right?

eneiluj:
Thank you for spotting the issue in the first place.

Yep yep closing...now.

PR #2383, opened 2022-05-10
Fix attachment cleanup when file names contain parenthesis

eneiluj:
This is an alternative of #2382.

We can keep the basic regex to look for attachment links if we make sure they are properly encoded by the frontend.
`encodeURIComponent` does not actually encode those chars `!'()*`. Let's make sure we do.


eneiluj:
Btw, tests were working because we encode links with Php but in real life, Javascript does it.

cypress/integration/images.spec.js
```diff
@@ -45,50 +55,81 @@ const clickOnImageAction = (actionIndex, callback) => {
...
+				if (index > 0) {
+					expect(imageName).include(`(${index + 1})`)
+				}
+
+				cy.wrap($el)
+					.should('be.visible')
+					.find('img')
+					.should('have.attr', 'src')
+					.should('contain', 'apps/text/image?documentId=' + documentId)
+					.should('contain', 'imageFileName=' + encodeURIComponent(imageName))
```
vinicius73, line 104: @eneiluj I don't know exactly the reason for this behavior, but looks like the element appears different when we read the attribute.

vinicius73 (approved):

PR #2389, opened 2022-05-13
[stable24] Fix attachment cleanup when file names contain parenthesis

eneiluj:
Manual backport of #2383


tests/.phpunit.result.cache
```diff
@@ -0,0 +1 @@
+{"version":1,"defects":[],"times":{"OCA\\Text\\Tests\\TextTest::testDummy":0.001,"OCA\\Text\\Tests\\TextTest::testGetAttachmentNamesFromContent":0.002}}
```
mejo-, line 1: Is this file pushed on purpose?

tests/clover.xml
```diff
@@ -0,0 +1,1461 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mejo-, line 1: Is this file pushed on purpose? Looks like a coverage report that is probably subject to change with each code change, no?

eneiluj:
@mejo- Nice catch, thanks. We need to add some entries to the gitignore file :grin:

mejo- (approved):

Issue #2411, opened 2022-05-14
Embedded images generate links that are not markdown compatible / interoperable. Also unnecessary copys of image are created.


lars-becker:
**Describe the bug**

If I embed an image in a markdown file with Nextcloud Text *an image link is created which is not compatible with other markdown apps* **which defeats the purpose of using markdown in the first place**. The usage of Markdown was is one major reason why we started using text documents. 

**Because of this change Text effectively  becomes a product which uses a **proprietary format** that is not interoperable and can't be rendered properly outside of Nextcloud.** Cumbersome office documents are now again much more interoperable than Text in this regard. This is a highly annoying regression that renders Text almost unusable in our use case (in which people also use Markdown files on their desktops).

Also unnecessary copies of the image will be created. This *unwanted* **redundancy** increases the likelyhood that there will be confusing among users and that in later edits wrong images might be edited. This kind of magic is, from my point of view, neither user friendly nor helpful in keeping a consistent and error free documentation.
 

**To Reproduce**
1. Open a markdown file.
2. Click on Insert image and select an existing image
3. The image is inserted with a link that is not standard conform like this: `![Raumübersicht](text://image?imageFileName=1652529375-nc_bbb_raumanzeige.png`
4. Also a copy of the image is created and moved to a attachment subdirectory; which might be fine in case of a new file, but not in case of an existing file. Redundancy is bad for many reasons.

**Expected behavior**

1. Produce CommonMark compatible links.
2. Don't copy existing images.

**Text app version:** 3.5.1
**Nextcloud version:** 24.0.0

lars-becker:
As a workaround a switch to disable the possibility for users to upload images would be nice. A downgrade to an old text version is not feasible and this feature breaks things. With such a setting only our users who are using desktop apps can create images and others can't break things unknowingly.

allesblinkt:
I second that oppinion. Up until now we had established a very nice workflow for working on technical documents (manuals, documentation, etc.) using Text. The images for those documents already exist in a good directory structure. Then we would use *pandoc* to generate PDF output.
Text now has a couple of very nice additional features (thank you for callouts and tables), but the image handling breaks our workflow. Having a configuration switch to disable upload would be a welcome option. Alternatively, there could be a distinction between **uploaded** images and **linked** images.

eneiluj:
Hi,

### Issues with the old image links

* In Text, an image could only be seen by users having access to the image file (accessed by file ID).
[...]

PR #2426, opened 2022-05-20
Improve image attachment management

eneiluj:
refs #2411

Implementation of the behaviour suggested by @max-nextcloud in https://github.com/nextcloud/text/issues/2411#issuecomment-1128776961 with which I completely agree :heart: :

* [x] 1. If the image url matches `.attachments.MD_FILE_ID/FILENAME` use FILENAME to get the image with the attachment API.
* [ ] 2. If the image url matches `.attachments.OTHER_FILE_ID/FILENAME`
  * [x] a. try to retrieve the image with the WebDav API.
  * [x] b. If that fails try to retrieve FILENAME with the attachment API.
  * [ ] c. Maybe rewrite the url with the own file id if b. succeeds.
* [x] 3. If not, use the path with the WebDav API.
* [x] 4. If that fails check if the url has a fileID param and try to get the file from that. (So we keep the old behavior for old images)

So this PR brings back the relative path in the markdown content's image target. We now write:
```
![IMAGE_FILE_NAME](.attachments.MARKDOWN_FILE_ID/ENCODED_IMAGE_FILE_NAME)
```

The problem with 2.a. and 2.b. is that `imageUrl` is a computed prop in which we would need a non-async (sync?) image loading check. But we can only make async network calls in a computed prop :grin:.
There already is a check which depends on `imageUrl`.
In short, we would need to make deeper changes to ImageView to allow 2.a. and 2.b.

We could discuss that and in the meantime I might come up with a first solution.


eneiluj:
@max-nextcloud I found a reasonable solution for 2.a. and 2.b. : move the content of the `imageUrl` computed prop in a method (called `init`) which is called in `beforeMount` (which is the only time and place we loaded the image anyway). This way we can happily call async stuff.

`imageUrl` is now a data attribute. It is set if the image has been successfully loaded.

The image loading has been moved in a method as well (`loadImage`). It has an optional error callback that we use for 2.

The `init` method now handles all the different cases and is more flexible than the old `imageUrl` computed prop.

src/nodes/ImageView.vue
```diff
@@ -301,6 +310,29 @@ export default {
...
+			} else {
+				return generateUrl('/apps/text/image?documentId={documentId}&sessionId={sessionId}&sessionToken={sessionToken}&imageFileName={imageFileName}&shareToken={shareToken}',
+					{
+						documentId,
+						sessionId,
+						sessionToken,
+						imageFileName,
+						shareToken: this.token,
+					})
+			}
```
vinicius73, line 334:
```suggestion
			}
		return generateUrl('/apps/text/image?documentId={documentId}&sessionId={sessionId}&sessionToken={sessionToken}&imageFileName={imageFileName}&shareToken={shareToken}',
			{
				documentId,
				sessionId,
				sessionToken,
				imageFileName,
				shareToken: this.token,
			})

```

src/nodes/ImageView.vue
```diff
@@ -280,18 +248,59 @@ export default {
...
+				const imageFileName = getQueryVariable(this.src, 'imageFileName')
+				this.loadImage(this.getTextApiUrl(imageFileName))
+				return
+			}
+			if (this.src.startsWith(`.attachments.${this.currentSession?.documentId}/`)) {
+				const imageFileName = decodeURIComponent(this.src.replace(`.attachments.${this.currentSession?.documentId}/`, '').split('?')[0])
+				this.loadImage(this.getTextApiUrl(imageFileName))
+				return
+			}
+			if (this.isRemoteUrl || this.isPreviewUrl || this.isDataUrl) {
```
vinicius73, line 265: Can we turn it into a computed?
```suggestion
			if (this.isDirectUrl) {
```

src/nodes/ImageView.vue
eneiluj: Why not.

src/nodes/ImageView.vue
```diff
@@ -280,18 +248,59 @@ export default {
...
+				if (failureCallback !== null) {
+					failureCallback()
+				} else {
+					this.failed = true
+					this.imageLoaded = false
+					this.loaded = true
+				}
+			}
+			img.src = imageUrl
+		},
```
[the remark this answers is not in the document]
eneiluj: Thank you for showing me the way to the promises. Much better indeed :heart:

src/nodes/ImageView.vue
eneiluj: Yep!

eneiluj:
@vinicius73 Thanks for the review. If you have any idea about how to fix the Jest tests. It seems the checks are done before the async stuff is finished.

eneiluj:
Let's commit the compiled code in the end to make our lives easier.

So don't forget to compile locally :wink:.

src/nodes/ImageView.vue
```diff
@@ -280,18 +251,72 @@ export default {
...
-		img.onerror = () => {
+		this.init()
+	},
+	methods: {
+		init() {
+			if (this.src.startsWith('text://')) {
+				const imageFileName = getQueryVariable(this.src, 'imageFileName')
+				this.loadImage(this.getTextApiUrl(imageFileName)).catch((e) => {
+					this.onImageLoadFailure()
+				})
```
vinicius73, line 262: You can simplify this code.

```suggestion
				this.loadImage(this.getTextApiUrl(imageFileName))
				  .catch(this.onImageLoadFailure)
```

vinicius73: Also, we can use async/await and wrap `this.init()` and handle error from `this.init()`

```suggestion
				await this.loadImage(this.getTextApiUrl(imageFileName))
```

or

```suggestion
				return this.loadImage(this.getTextApiUrl(imageFileName))
```

and, change `this.init()` call.

```js
this.init().catch(this.onImageLoadFailure)
```

max-nextcloud:
Thanks a lot for picking up the idea. :heart_eyes: 
I only had a quick look at the code and checked if the `text://` urls are still parsed and will still display nicely. Looks good to me in that regard.

eneiluj:
@vinicius73 @max-nextcloud Ready to merge!

vinicius73 (approved):

tests/TextTest.php
```diff
@@ -7,36 +7,54 @@
...
-			'aaa (2.png',
-			'aaa ((2.png',
-			'aaa 2)).png',
-			'a[a]a.png',
-			'a(a)a.png',
-			'a](a.png',
-			',;:!?.§-_a_',
-			'a`a`.png',
-		];
+	public function testGetOldAttachmentNamesFromContent() {
```
juliushaertl, line 29: Something for a separate PR but we should move those tests into tests/unit/Service/ImageServiceTest.php

lib/Service/ImageService.php
```diff
@@ -476,20 +479,33 @@ public function cleanupAttachments(int $fileId): int {
...
+			$oldMatches,
+			PREG_SET_ORDER
+		);
+		$oldNames = array_map(static function (array $match) {
+			return urldecode($match[1]);
+		}, $oldMatches);
+
+		$matches = [];
+		preg_match_all(
+			'/\!\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[\])*\])*\])*\])*\])*\])*\]\(\.attachments\.'.$fileId.'\/([^)&]+)\)/',
```
juliushaertl, line 499: Would be nice to have a comment on what the regex should match to make is easier to catch its impact on the surrounding logic

juliushaertl (approved):
Minor comments, but feel free to address them separately :)

eneiluj:
Should we backport this to stable24? I think we should.

Issue #2440, opened 2022-05-23
Improve Header bar


luka-nextcloud:
This issue is for the header improvement mentioned on https://github.com/nextcloud/text/issues/2211

- [x] Make backgound solid var(--color-main-background)
- [x] Add small blue flavor bar on the top like in Nextcloud Office
- [x] Bold the filename / title

PR #2441, opened 2022-05-24
Improve Header bar

luka-nextcloud:

* Resolves: #2440
* Target version: master 

### Summary
Some improvements on the header bar.

- [ ] Requires https://github.com/nextcloud/viewer/pull/1281


css/style.scss
```diff
@@ -52,3 +52,12 @@ li.ProseMirror-selectednode:after {
     visibility: hidden;
   }
 }
+
+#viewer > div.modal-header > div.modal-title {
```
vinicius73, line 56: This selector will impact all modals inside viewer

I don't know if it will be good.

But if is our intention, maybe change it directly in viewer app will be better.

eneiluj: It looks much nicer!
But yeah it seems better not to make changes to the viewer style in apps. This might break if conflicting changes are then made to the viewer.
@luka-nextcloud @vinicius73 This change was suggested for Text but maybe @jancborchardt meant it would be good to make it in the viewer itself. Did he? :grin:

vinicius73: Yes, we must track the root cause of this error and fix it properly.
Apps like `files_pdfviewer` do not have this issue in my actual setup.
I don't know if it is related to vue components version, but `files_pdfviewer` do not use Nextcloud vue components.

juliushaertl: Lets wait for input from @jancborchardt or @nimishavijay if we should add the full background color without opacity on the viewer header and the theme color hint with the small 3px border on top to any viewer apps (with light theme). 

If we should only do this for text, then maybe we need to expose the handler of viewer somehow in the attributes so we can apply a selector only if viewer displays a text file.

luka-nextcloud:
> Before/after screenshots would be great for quick design review @luka-nextcloud @vinicius73 :) thanks!

@jancborchardt Please check the following screenshots.

Before:
image

After:
image

jancborchardt (approved):
Looks great! :)

css/style.scss
vinicius73: Will we keep the global selectors @luka-nextcloud ?

jancborchardt: > Lets wait for input from @jancborchardt or @nimishavijay if we should add the full background color without opacity on the viewer header and the theme color hint with the small 3px border on top to any viewer apps (with light theme).

@juliushaertl @luka-nextcloud @vinicius73 what would "any viewer apps (with light theme)" mean? Are there any except Text and Office which use (or look like they use) the Viewer in light theme?

But yeah, what is for sure:
- The style for the normal file viewer with the dark slightly transparent overlay is fine as is
- Text should absolutely have the full main background color without opacity in the header, and the theme color hint border is nice to have it consistent with Office

juliushaertl: > Are there any except Text and Office which use (or look like they use) the Viewer in light theme?

Not yet, but I could imagine for pdf viewer it would also look nice to have a white header instead of the dark one.

> Text should absolutely have the full main background color without opacity in the header, and the theme color hint border is nice to have it consistent with Office

@luka-nextcloud We can achieve that with a pull request in the viewer app to avoid the global style override in here.

Relevant code parts would be 
https://github.com/nextcloud/viewer/blob/master/src/views/Viewer.vue#L833
https://github.com/nextcloud/viewer/blob/master/src/views/Viewer.vue#L845

luka-nextcloud: Okay, great!

skjnldsv: Pdf cannot have theming as it's using an external library entirely and load an iframe.
So there is no css selector inheritance here.

Since this is only about text then, I don't see why we should have custom app-related css inside a standard app like Viewer.
I'll suggest to go for that for code quality

> If we should only do this for text, then maybe we need to expose the handler of viewer somehow in the attributes so we can apply a selector only if viewer displays a text file.

juliushaertl (changes requested):
@vinicius73 Let's wait for the merge of the viewer PR, as this one then needs to be adjusted to use the data attribute.

juliushaertl (approved):

Issue #652, opened 2020-02-03
Do not auto-assign random vegetable names to guests


wiswedel:
It seems to have been plotted conjointly to auto-assign veggie names to guest editors (https://github.com/nextcloud/text/issues/52), but from an enterprise-grade software perspective, this is silly. Who wants to share a protocol with their bosses who eventually appear then as _Chinese Broccoli_ and _Spinach_?

Please take a minute to consider rolling back to the input field.

juliushaertl:
cc @jancborchardt 

We basically have two options then, either we don't show the editor at all until the user has entered a name (this is how it is currently done with Collabora) or we just use something like "Anonymous user" as the default guest name.

wiswedel:
+1 for enforced name entering:
1. It's usual across the web
2. you can still enter a pseudonym
3. Self-renaming of auto-anonymous-users is very likely to not happen in real-life

kesselb:
Seems to be enterprise-grade enough for Google: https://support.google.com/docs/answer/2494888

jancborchardt:
Yes – as @kesselb said Google does just the same, except with animal names.

Forced name entering doesn’t seem like much, but it’s an annoying hurdle when you just want to get editing. We literally had this before and it was bad: https://github.com/nextcloud/text/issues/52

Yes we can show the name input field more presently, but let’s not _require_ it. People just want to edit documents, not fill their names (or come up with pseudonyms).

juliushaertl:
It might help a bit that google prefixes the animal names with "Anonymous". I also can imagine that a lot of people can easier get what animal names are about as they might apply some personification to those which is harder for vegetables. :wink:

schiessle:
> +1 for enforced name entering:
> 
>     1. It's usual across the web
> 
>     2. you can still enter a pseudonym
> 
>     3. Self-renaming of auto-anonymous-users is very likely to not happen in real-life

FWIW, I agree with @wiswedel.

Looking at the initial issue which lead to the vegetable names I think there can be some middle ground.

1. If it is a read-only document we should skip the input field, as @jancborchardt said in one of the other issues, this classifies clearly as a bug.
2. It would be fine for me if the input field would be pre-filled with a fantasy name, so people who don't want to share their real name and don't want to be creative can just press enter.

[...]

PR #2447, opened 2022-05-25
Avoid random guest names and make guest state more visible

juliushaertl:
* Resolves: https://github.com/nextcloud/text/issues/652
* Target version: master 

### Summary

- Make sure that guest users use "Guest" as a default name instead of random vegetable names
- Allow guests to reset their name to the default after changing it initially
- Adjust the avatar style for guests so that their avatar just uses a grey background with the first character of the name
  - Refactor the avatar handling in the session menu a bit to have a shared AvatarWrapper that takes care of the user/guest rendering and can be reused

This is the same behaviour as it is already in place in the Talk app.

First 3 commits can be backported separately to address the confusing default names
| Before | After |
| --- | --- |
| Screenshot 2022-05-25 at 14 38 30 | Screenshot 2022-05-25 at 14 38 02 |


juliushaertl:
Rebased and added screenshots for comparison

src/components/AvatarWrapper.vue
```diff
@@ -0,0 +1,68 @@
...
+		sessionColorBorder() {
+			return {
+				...this.sessionBackground,
+				'border-color': this.session.color,
+				'border-width': '2px',
+				'border-style': 'solid',
+				'--size': this.size + 'px',
+				'--font-size': this.size / 2 + 'px',
+			}
+		},
```
vinicius73, line 43: This is not only for border.

juliushaertl: Yes, renamed :)

luka-nextcloud (approved):

juliushaertl:
Test succeeded after restart. Might still be worth to monitor if that is flaky and needs some improvements:

https://dashboard.cypress.io/projects/hx9gqy/runs/4000/test-results/9c8707e6-4ab8-4124-a226-e360db96653c

juliushaertl:
Let's skip this backport since this is more an enhancement and would bring a large changeset to the stable branches.

PR #2448, opened 2022-05-25
Do not register text/csv mimetype when office can handle it

juliushaertl:
Make sure to only register CSV handling for the viewer when there are no other options as Collabora and ONLYOFFICE have better capabilities for showing and editing those. 

We could have this as an optional file opener through the 3 dots menu, but the viewer currently does not support that and only allows registering for the default file action.


max-nextcloud (approved):
Looks good to me. Thanks for tackling this.

PR #2455, opened 2022-05-27
Bugfix/hard break

juliushaertl:
Seems like the hard break plugin got lost during migration to tiptap v2 (https://github.com/nextcloud/text/blob/stable23/src/EditorFactory.js#L73)

Now the first commit restores the previous behaviour to support hard breaks in text. The second transforms the output markdown to use double space instead of a backslash at the line end to indicate the hard break, which is less intrusive in the plain markdown. Both are allowed according to https://spec.commonmark.org/0.30/#hard-line-breaks

The backslash is the default format provided by prosemirror-markdown:

https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/to_markdown.ts#L107-L113


max-nextcloud:
Thanks for catching this. Looks good.

src/extensions/Markdown.js
```diff
@@ -79,6 +79,7 @@ const createMarkdownSerializer = ({ nodes, marks }) => {
 		),
 		serialize(content, options) {
 			return this.serializer.serialize(content, { ...options, tightLists: true })
+				.split('\\\n').join('  \n')
```
max-nextcloud, line 82: what will happen to a line that ends with a literal `\` in the editor?
I suspect it will be transformed into a line ending with `\\` by the serialization and then turned into `\  ` as the end by this regexp.

Maybe better overwrite the `to_markdown` property of the `HardBreak` node?

max-nextcloud (changes requested):

src/extensions/Markdown.js
max-nextcloud: https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/to_markdown.ts#L107-L113 could serve as a starting point.

juliushaertl: Move to an extension that implements toMarkdown and added test cases

max-nextcloud (approved):

eneiluj (approved):
lgtm

juliushaertl:
Manual backport to 23/22 might need a bit more work to make it work with tiptap v1

juliushaertl:
Leaving for backport for the next maintenance releases.

Issue #2173, opened 2022-02-10
Feature-request: inter-page links


rotdrop:
**Is your feature request related to a problem? Please describe.**
Trying to convert pages from DokuWIki: inter-document links do not seem to be supported by NC text app?

**Describe the solution you'd like**
Link to a header in the same document  like
```
See below [LinkText](#heading-text)
...
#### Heading Text
Blah blah.
```
The link should be followed by scrolling down to the linked header. This is BTW supported on Github, I think.
**Describe alternatives you've considered**
Do not use Nextcloud collectives and text-app, stick to DokuWiki

juliushaertl:
Related to https://github.com/nextcloud/text/issues/107 as we'd need the anchors then anyways which would easily allow to link to them.

PR #2464, opened 2022-05-30
Feature: Add page anchors and inter page links

susnux:
* Resolves: #2173
* Target version: master 

### Summary
Implements the anchors needed for inter-page links

This requests consists of three commits, the first part adds the required anchors.
IDs are generated using a similar slugify function github and
[gitlab use](https://docs.gitlab.com/ee/user/markdown.html#header-ids-and-links).
For initial content markdown-it-anchor is used.

The second part implements visible links, using ProseMirror view decorations.

The last part is handling opening inter-page links, as just adding an onClick listener
using `Element.scrollIntoView()` would not handle user defined links within the document.
And the current default handler would open the document again in a new tab.


susnux:
@vinicius73 
> Despite small CSS bugs (master branch) seem it works.

Sorry, CSS is not one of my strengths, I get it work but it might not that clean (help one the CSS part is highly welcome)

> But I must check it close, decorators looks a _heavy_ approach.

I thought it might be the easiest way, as the links should only be visible on the frontend and not serialized.
Using decorations will not interfere with the editor state and so not interfere with any collaboration stuff.
It should no lead to performance losses, as the decorations are cached and only re-rendered if the ID changed (user changed the heading or removed a heading with the same content and type above) or the position changed and the decoration has to be rendered new anyways. 

> It is the only way to do it? @susnux

No, two other option I can imagine:
* Using a NodeView for headings
  * pro: No Decorations (this would be the tiptap way) and still no unneeded nodes in the editor state
  * con: Requires to override the `renderHTML` of the Heading class, so it might get hard to inherite stuff from the core heading (rewrite of existing code)
* Insert a custom `AnchorLink` Node into the editor state, the same way the IDs are updated
  * pro: At least no decorations needed
  * con: Part of the editor state, as such they are handled by transactions (collab?) and need to serialized (or dropped)

So from my point decorations are a cleaner way, but `NodeView` could be also a option.

juliushaertl:
Extending the Heading node sounds reasonable to me. We already extend from the tiptap one in https://github.com/nextcloud/text/blob/87aa0695d7bb49fb65c486de3758d48abeee7925/src/nodes/Heading.js#L3 so that should give a starter point for adjusting the renderHTML. It may only require the onUpdate event https://tiptap.dev/guide/custom-extensions#events  in case the anchor should be updated on a content change.

susnux:
@vinicius73 @juliushaertl How about doing it this way: https://github.com/nextcloud/text/compare/master...susnux:feature/lightweight-inter-page

I rewrote it to be similar to the already existing custom nodes.

juliushaertl:
Yes, looks reasonable to me from a first look. @vinicius73 What do you think?

susnux:
created a new PR from the modified (lightweight) version: https://github.com/nextcloud/text/pull/2520

susnux:
created a new PR from the modified (lightweight) version: https://github.com/nextcloud/text/pull/2520

Issue #2462, opened 2022-05-30
Trigger event on content change in syncService


mejo-:
For Collectives, it would be very nice if Text could trigger an event when the content changed in syncService.

This way, we could update the page content that we pass into the [RichText component][https://gitlab.com/collectivecloud/collectives/-/blob/main/src/components/Page/RichText.vue] in realtime when the [Editor component](https://gitlab.com/collectivecloud/collectives/-/blob/main/src/components/Page/Editor.vue) is loaded (and has the up-to-date content from syncService anyway).

This would make life easier and we could avoid a `setInterval` loop in https://gitlab.com/collectivecloud/collectives/-/merge_requests/533/diffs#4ce2ff67d66d8af017006277e1e4338890fccb1b_192_192

vinicius73:
Now, Editor triggers 4 events (without payload).

```
sync-service:idle
sync-service:save
sync-service:error
sync-service:sync
```

PR #2465, opened 2022-05-31
Trigger event on content change in syncservice

vinicius73:
* Resolves: #2462 
* Target version: master 

### Summary

Fires 4 events from `SyncService`

- `sync-service:idle`
- `sync-service:save`
- `sync-service:error`
- `sync-service:sync`

#### Prevention of Memory Leak

All `SyncService` event handlers are storage in arrays, and never removed from these arrays.
Although the `SyncService` whore removed from Vue component context, we had a circular reference here.
This is a very common scenario for Memory Leak in JavaScript.

Now using [`mitt`](https://github.com/developit/mitt) library is possible to use `.off` to remove an event handler.

To do it, I've change part of event contracts (error event) and I've split the event handlers in component methods.

There are more potential problems to solve, but we can do it later.


juliushaertl (approved):
Much cleaner 👍

max-nextcloud:
Not sure if we really want to backport this - but since it prevents memory leaks and cleans things up seems like a good idea.

mejo-:
Manual backported to stable24 in #2679

Issue #2463, opened 2022-05-30
Load fallback images when image could not be loaded


mejo-:
It would be nice to load a fallback image in [ImageView](https://github.com/nextcloud/text/blob/master/src/nodes/ImageView.vue) if the requested image could not be loaded.

This would also fix a bug with the print view in Collectives in case of missing images: https://gitlab.com/collectivecloud/collectives/-/issues/333

In the long run, we could make Text emitr an event when images failed to load, but that's out of scope for this issue :wink:

PR #2466, opened 2022-06-01
🩹 (#2463): show a fallback image when image can't be loaded

vinicius73:
* Resolves: #2463
* Target version: master 

### Summary

Display a fallback image when image loading fails.

image


jancborchardt:
@vinicius73 some design feedback:
- The icon can be the general image icon, does not need to be a specific "broken" one: https://fonts.google.com/icons?selected=Material%20Icons%3Aimage%3A
- Color should be grey, also for the text, color-text-maxcontrast. It’s not a "red error" type big deal. :)
- The message "Failed to load image" can show below together with the filename, as to not introduce another text block above. Just: "Failed to load [filename]" like "Failed to load image2.jpg"
- The initial `/` should be removed, it’s too technical

jancborchardt (changes requested):
@vinicius73 some design feedback:
- [x] The icon can be the general image icon, does not need to be a specific "broken" one: https://fonts.google.com/icons?selected=Material%20Icons%3Aimage%3A
- [x] Color should be grey, also for the text, color-text-maxcontrast. It’s not a "red error" type big deal. :)
- [x] The message "Failed to load image" can show below together with the filename, as to not introduce another text block above. Just: "Failed to load [filename]" like "Failed to load image2.jpg"
- [ ] The initial `/` should be removed, it’s too technical

vinicius73:
> The initial / should be removed, it’s too technical

It's the description of the image, I've put it as an example.

eneiluj (commented):
I agree with @jancborchardt comments.

Code-wise, nothing to add. I guess the `ErrorLoadImage` is part of the solution we figured out with @mejo- about knowing when all images are loaded in Collectives, right?

The error could be renamed to `LoadImageError` but that's nitpicking :grin:.

vinicius73:
@jancborchardt what is the best alternative?

Original/raw src parameter. (for now I am using this)
image

With full URL used to load image (parsed by ImageView component)
image

mejo-:
> @jancborchardt what is the best alternative?
> 
> Original/raw src parameter. (for now I am using this) image

I would vouch for `Failed to load invalid.png`, without square brackets and without any path elements - just the basename of the image. That's the least technical and easiest to understand for users in my opinion.

src/components/icons.js
```diff
@@ -46,6 +46,8 @@ import MDI_FormatQuote from 'vue-material-design-icons/FormatQuoteClose'
 import MDI_FormatStrikethrough from 'vue-material-design-icons/FormatStrikethrough'
 import MDI_FormatUnderline from 'vue-material-design-icons/FormatUnderline'
 import MDI_Help from 'vue-material-design-icons/HelpCircle'
+import MDI_Image from 'vue-material-design-icons/ImageOutline'
+import MDI_ImageBroken from 'vue-material-design-icons/ImageBrokenVariant'
```
mejo-, line 50: This can probably be removed now that you use `ImageOutline`, right?

src/components/icons.js
```diff
@@ -110,6 +113,8 @@ export const FormatQuote = makeIcon(MDI_FormatQuote)
 export const FormatStrikethrough = makeIcon(MDI_FormatStrikethrough)
 export const FormatUnderline = makeIcon(MDI_FormatUnderline)
 export const Help = makeIcon(MDI_Help)
+export const Image = makeIcon(MDI_Image)
+export const ImageBroken = makeIcon(MDI_ImageBroken)
```
mejo-, line 117: Same here :relaxed:

mejo- (commented):
Thanks a lot for implementing this, @vinicius73 <3

vinicius73:
@juliushaertl I've reverted this commit: a70841ac98cdfdbd376f2406eebd3ce30ce5fc32

It was generating error 500 from API.

```
POST /index.php/apps/text/session/fetch HTTP/1.1" 500
````

juliushaertl:
Rebased to retrigger tests after #2495 was merged

juliushaertl (approved):

PR #2457, opened 2022-05-27
Feature: Allow linking to directories

susnux:
* Resolves: #2162
* Target version: master 

### Summary

Add the ability to create links to directories from within the
link menu bubble.

PR #2481, opened 2022-06-05
Fix cypress scripts for local run + Add documentation

susnux:
* Resolves: https://github.com/nextcloud/text/pull/2457#issuecomment-1143764123 (comment)
* Target version: master 

### Summary

Fix the scripts for local cypress runs,
they need to be run within the `npm run` environment,
as the nextcloud webpack config requires the `npm_package_name`
variable to be set.

Also a up-to-date nextcloud instance is required for testing
text at the current git head, so updated the dockerfile
and docker-compose files to create a nextcloud image from current git head.
(To get it work I modified the docker part to create a similar image as the one used by the github CI)

Additionally updated the tests to use cypress 10.x and
added missing documentation on how to run the cypress tests.


vinicius73:
Thanks a lot for this @susnux 
I've adjusted some parts of your PR 

- The speed of build test image build (using [`--depth`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt))
- Apply [ShellCheck](https://www.shellcheck.net/) in `runLocal.sh`
- Fix apps folder permission

vinicius73 (approved):

susnux:
> Thanks a lot for this @susnux I've adjusted some parts of your PR
> 
>     * The speed of build test image build (using [`--depth`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt))
> 
>     * Apply [ShellCheck](https://www.shellcheck.net/) in `runLocal.sh`
> 
>     * Fix apps folder permission

Looks good, it's only important to mount into custom apps, because docker will otherwise remove your current directory (or more precisely the nextcloud entry point)

juliushaertl (approved):
Very nice, thanks for documenting those parts and the improvements :)

susnux:
> Thanks a lot for this @susnux I've adjusted some parts of your PR

Have you tried it? For me it works the first time but then stops working

vinicius73:
Very strange @susnux 
Works every time for me, I've spent a long time adjusting it :/

susnux:
Not sure why but the `custom_apps` directory gets created with `root` as the owner, see #2515 which works for me by forcefully `chown` the directory.
I tried to debug it but not sure if it is caused by docker while bind-mount the directory or the nexcloud entrypoint script.

PR #2488, opened 2022-06-07
Build app before running cypress tests

juliushaertl:

* Resolves: # 
* Target version: master


max-nextcloud (approved):
Looks good to me.

max-nextcloud:
ahh... never mind - no need to backport this as the behavior is determined by the file in master i guess.

PR #2490, opened 2022-06-08
[stable22] upgrade: all of prosemirror

max-nextcloud:
There have been some fixes to Android behavior.

* Target version: stable22


max-nextcloud:
For now i just created the branch for testing... but we might want to get this in anyway... What do you think @juliushaertl ?

juliushaertl:
Yeah probably makes sense to regularly do that for the stable branches manually as dependabot would only bump patch releases.

max-nextcloud:
@juliushaertl any idea what's up with the static analysis?

juliushaertl:
Yes, seems there is a conflict between out composer lock file version and what the OCP package pulls in ... pushed an update in #2491

max-nextcloud:
@juliushaertl just confirmed with @tobiasKaminsky that this fixes some of the Android issues in text. So we should definitely get this in.

juliushaertl (approved):

juliushaertl:
Failure unrelated

Issue #2442, opened 2022-05-24
Re-add inline code (missing since NC24)


Pilzinsel64:
**Is your feature request related to a problem? Please describe.**
Since NC24 the inline code feature is missing. Probably an unknown bug, but I guess that's just missing now due a depency update e.g.
Also, it get's actively forced removed. That breaks existing files.

**Describe the solution you'd like**
I'd like to have a inline code feature as it was there before NC24. It can be used to highlight some words in a line without generating a whole code block.
It could be a hidden feature (as it was if you typed a ` at beginning and one at the end) or done by clicking the "Add Code" button that desices if you selected the whole line / multiple lines or if you selected only a part of the line. Test it yourself while writing a comment here on GitHub.

Formatted it looks like this:
![grafik](https://user-images.githubusercontent.com/23138465/170004202-fed7445b-48e3-4447-b791-d9669408be1b.png)

In Plain text it looks like this:
![grafik](https://user-images.githubusercontent.com/23138465/170004250-6d373a5c-d6c9-490c-beba-f422145440a1.png)

**Describe alternatives you've considered**
The only alternative is to leave it as it is now - without inline code.

**Additional context**
_None_

juliushaertl:
Yes, seems we also missed that. Pulling in https://tiptap.dev/api/marks/code is probably enough. I'm wondering a bit why the jest tests haven't caught that, same as with the line breaks, but then we need to extend those as well.


@max-nextcloud Could you maybe have a look?

juliushaertl:
@max-nextcloud I was so free to push a fix since I was looking into code blocks anyways ;)

PR #2502, opened 2022-06-09
Bring back inline code support

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/2442


max-nextcloud (approved):
Thanks heaps for taking care of this.

PR #2513, opened 2022-06-11
Updating node.yml workflow from template

nextcloud-command:
Automated update of the node.yml workflow from https://github.com/nextcloud/.github

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -13,6 +13,9 @@ on:
       - master
       - stable*
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest


.github/workflows/node.yml
```diff
@@ -39,14 +42,8 @@ jobs:
 
       - name: Install dependencies & build
         run: |
-          make clean
           npm ci
           npm run build --if-present
-          npm run build:package
```
max-nextcloud, line 45: I'd like to ensure the package build does not break.
We could use a different workflow for that though.

.github/workflows/node.yml
```diff
@@ -39,14 +42,8 @@ jobs:
 
       - name: Install dependencies & build
         run: |
-          make clean
```
max-nextcloud, line 42: If we remove the make clean the js dir will contain the artifacts from the repo, right? So if `npm run build` did nothing the comparison to the repo would still match in the next step.

I'm not sure which of these are changes from the template and which are just cleanup.

.github/workflows/node.yml
```diff
@@ -39,14 +42,8 @@ jobs:
...
           npm ci
           npm run build --if-present
-          npm run build:package
-
-      - name: npm test
-        run: |
-          npm test
```
max-nextcloud, line 49: Do we run these tests elsewhere?

max-nextcloud (commented):
I don't quite understand the goal here. @vinicius73 are you aiming to make this match the template more closely? Shall we move the text specific checks to other workflows?

vinicius73:
> I don't quite understand the goal here. @vinicius73 are you aiming to make this match the template more closely? Shall we move the text specific checks to other workflows?

No @max-nextcloud 
It is an automatic PR, and I know it will impact the generation of text package.
Because of that, I've called for you.

max-nextcloud:
> > I don't quite understand the goal here. @vinicius73 are you aiming to make this match the template more closely? Shall we move the text specific checks to other workflows?
> 
> No @max-nextcloud It is an automatic PR, and I know it will impact the generation of text package. Because of that, I've called for you.

Ahh... now i get it. Thanks. then we should only add the permissions i think. I'll fix this.

max-nextcloud (approved):
Okay... let's just add what was added upstream.

PR #2514, opened 2022-06-14
package: expose ImageResolver so it can be provided to RichTextReader

max-nextcloud:
* Target version: master 

### Summary

In order to resolve images correctly in various contexts the RichTextReader needs an ImageResolver. The ImageResolver implements a resolve function that turns a `src` attribute into a full url that can be retrieved from the server.

Texts `ImageResolver` is quite powerful and flexible. So let's expose it so it can be used by other apps such as collectives.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,17 +9,25 @@
   </head>
   <body>
     <div id="app"></div>
+	<script>
+	OC = { webroot: '/OC_WEBROOT' }
+	</script>
     <script type="module">
-	    import { RichTextReader } from './src/package.js'
+	    import { RichTextReader, ImageResolver, IMAGE_RESOLVER } from './src/package.js'
 		import Vue from 'vue'
+		import content from './src/tests/fixtures/basic.md?raw'
 		const app = new Vue({
 			data() {
 				return {
-					content: '**Hello There**',
+					content
 				}
 			},
-			provide: {
-				currentDirectory: '/'
+			provide() {
+				const val = {}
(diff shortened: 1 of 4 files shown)


max-nextcloud:
@vinicius73 I'm curious to hear what you think about this approach. I'm quite unsure myself. I don't like to export such a complex API - but i have not been able to figure out something more simple.

vinicius73:
> @vinicius73 I'm curious to hear what you think about this approach. I'm quite unsure myself. I don't like to export such a complex API - but i have not been able to figure out something more simple.

Since yesterday, I am thinking about this change.
Though, I share your fear... I did not think any technical issue related about this change.

If we provide a mixin, the main issue still the same:

How expose/provide: `currentDirectory`, `shareToken`, `user` and `fileId` to build `ImageResolver`

The pain point of this change is exposing the contract of `ImageResolver` and maybe turn hard to change it over the time.

Can we have problems if we pass it as props to `RichTextReader`?
For now is the only other way what I am able to think.

max-nextcloud:
> Can we have problems if we pass it as props to `RichTextReader`?
> For now is the only other way what I am able to think.
I might give this a try. Seems simpler indeed as props are kind of the default api of components and we could also handle reactivity in one place.

The only downside i see is not being able to implement a different ImageResolver. I was thinking about providing one in `index.html` that allows using images with vite - but that's just a convenience thing for development.

mejo- (approved):

PR #2516, opened 2022-06-15
Fix formatting help not triggered if button is in submenu

susnux:
* Target version: master 

### Summary
When the formatting help button is pushed into the hidden
actions, meaning the sub-menu for low resolution displays,
the help modal is not shown as the click event is intercepted.

This is fixed by checking the menu entry for a click listener
before intercepting the event.


vinicius73 (approved):

vinicius73:
I've create another PR just to compile files @susnux 
Thanks

PR #2584, opened 2022-06-20
test: retry failed cypress tests twice

max-nextcloud:
This will hopefully make the CI more stable
while allowing us to also see which tests are flaky
in the cypress ui.

Some tests may not be retryable yet.
So they will always fail on a second attempt because of remains of a first attempt.
We can tackle that on a test by test basis later.

* Target version: master


max-nextcloud:
https://dashboard.cypress.io/projects/hx9gqy/runs/4319/test-results/b476bae3-44e3-4184-9d68-81c968f8875a image test retry failing because on the second and third run it has a file conflict.

max-nextcloud:
https://dashboard.cypress.io/projects/hx9gqy/runs/4319/test-results/b476bae3-44e3-4184-9d68-81c968f8875a with the file conflict gone it now fails because the file name counter continues to go up even if the images were removed from the file content itself. This makes sense as the editor was not closed in between.

vinicius73 (approved):

PR #2603, opened 2022-06-20
text: fix cypress selector to find the right image

max-nextcloud:
`get().filter()` might not take into account elements
that did not exist in the dom yet when the `get` part was fired.

`.filter()` would then wait for the elements that already existed
during `.get()` to contain the newly created image.

* Target version: master 
* fixes failures like https://github.com/nextcloud/text/runs/6966514212?check_suite_focus=true


eneiluj (approved):
Much better indeed!

PR #2608, opened 2022-06-21
Fix/dav

CarlSchwan:
* Resolves: https://github.com/nextcloud/server/issues/32514
* Target version: master and stable24

### Summary

Clients expects this behavior:
If rich-workspaces is enabled and readme.md is empty or does not exist, it must return "".
If disabled -> null.
If readme.md has content -> show content

This revert to it and also make it possible in a separate commit to not have to add  'Depth: 1' in the curl request to get the workspace information

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -98,8 +98,9 @@ class WorkspacePlugin extends ServerPlugin {
 		// Only return the property for the parent node and ignore it for further in depth nodes
 		if ($propFind->getDepth() === $this->server->getHTTPDepth(1)) {
 			$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
+				$owner = $this->userId ?? $node->getFileInfo()->getStorage()->getOwner('');
 				/** @var Folder[] $nodes */
-				$nodes = $this->rootFolder->getUserFolder($this->userId)->getById($node->getId());
+				$nodes = $this->rootFolder->getUserFolder($owner)->getById($node->getId());
 				if (count($nodes) > 0) {
 					/** @var File $file */
 					try {
@@ -121,7 +122,7 @@ class WorkspacePlugin extends ServerPlugin {
 					try {
 						$file = $this->workspaceService->getFile($nodes[0]);
 						if ($file instanceof File) {
-							return $file->getFileInfo()->getName();
+							return $file->getFileInfo()->getId();
 						}
 					} catch (StorageNotAvailableException $e) {
 						// If a storage is not available we can for the propfind response assume that there is no rich workspace present
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -96,7 +96,7 @@ class WorkspacePlugin extends ServerPlugin {
 		}
 
 		// Only return the property for the parent node and ignore it for further in depth nodes
-		if ($propFind->getDepth() === $this->server->getHTTPDepth()) {
+		if ($propFind->getDepth() === $this->server->getHTTPDepth(1)) {
 			$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
 				/** @var Folder[] $nodes */
 				$nodes = $this->rootFolder->getUserFolder($this->userId)->getById($node->getId());
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -96,40 +96,38 @@ class WorkspacePlugin extends ServerPlugin {
 		}
 
 		// Only return the property for the parent node and ignore it for further in depth nodes
-		if ($propFind->getDepth() === $this->server->getHTTPDepth(1)) {
-			$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
-				$owner = $this->userId ?? $node->getFileInfo()->getStorage()->getOwner('');
-				/** @var Folder[] $nodes */
-				$nodes = $this->rootFolder->getUserFolder($owner)->getById($node->getId());
-				if (count($nodes) > 0) {
-					/** @var File $file */
-					try {
-						$file = $this->workspaceService->getFile($nodes[0]);
-						if ($file instanceof File) {
-							return $file->getContent();
-						}
-					} catch (StorageNotAvailableException $e) {
-						// If a storage is not available we can for the propfind response assume that there is no rich workspace present
+		$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
+			$owner = $this->userId ?? $node->getFileInfo()->getStorage()->getOwner('');
+			/** @var Folder[] $nodes */
+			$nodes = $this->rootFolder->getUserFolder($owner)->getById($node->getId());
+			if (count($nodes) > 0) {
+				/** @var File $file */
+				try {
+					$file = $this->workspaceService->getFile($nodes[0]);
+					if ($file instanceof File) {
+						return $file->getContent();
 					}
+				} catch (StorageNotAvailableException $e) {
+					// If a storage is not available we can for the propfind response assume that there is no rich workspace present
 				}
-				return '';
-			});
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
@max-nextcloud can we then also have a test, so this regression cannot happen again? :)

CarlSchwan:
@vinicius73 why was master merged in this branch?

vinicius73:
> @vinicius73 why was master merged in this branch?

This branch has master as target, we can't merge it properly without update it.
Last commits in master update some dependencies, it can affect tests.

CarlSchwan:
we should do rebasing instead then instead of creating merge commits, always a bit cleaner

tobiasKaminsky:
👍 "add folder info" works again.
(though i now cannot use text within Android app, but either this is a problem on my side or a different bug)

tobiasKaminsky:
One problem I now see is:
- have folder 1, inside subfolder 2
- go into folder 1, retrieve content of it: subfolder 2 does not show rich workspaces
- go into folder 2, see that it has content -> is updated and shown in Android
- go back to folder 1, refresh -> subfolder 2 has no content thus it is deleted

--> if requested, the folder and its children should always have the rich workspace property properly set

max-nextcloud:
The CI failures are indicating that names other than 'Readme.md' - i.e. 'README.md' or 'Anleitung.md' do not work anymore for the rich workspaces.

This used to be server side logic in the `/workspace` text endpoint. We stopped using that endpoint and rely purely on PROPFIND results since NC 24 to speed up rendering of rich workspaces in the files view.

I'm not a big fan of supporting translated or all caps readme file names - but it's what we did in the past so breaking it does not seem like a good idea either.

max-nextcloud:
> also make it possible in a separate commit to not have to add 'Depth: 1' in the curl request to get the workspace information

@CarlSchwan Sounds like you've been using curl to test this... Could you share your curl commands so i can base the tests on it?

max-nextcloud:
Added tests and fixed issues that the reverted commit had fixed.

The workspace property is still not set for nested folders (see https://github.com/nextcloud/text/pull/2608#issuecomment-1162715522). I think in order to bring that back we need to revert https://github.com/nextcloud/text/commit/be562a41329fccb7c6bdc8d1a5e2e1e187a10869.

CarlSchwan:
```
curl 'http://nextcloud.local/remote.php/dav/files/admin/' -v -X PROPFIND -u admin:admin --data '<?xml version="1.0"?>
<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns" xmlns:ocs="http://open-collaboration-services.org/ns">
  <d:prop>
    <d:getlastmodified />
    <d:getetag />
    <d:resourcetype />
    <oc:fileid />
    <nc:file-metadata-size /><oc:permissions />
    <oc:size />
    <d:getcontentlength />
    <d:quota-available-bytes />
    <nc:rich-workspace /> <nc:rich-workspace-file /> <nc:has-preview />
    <nc:mount-type />
    <nc:is-encrypted />
    <ocs:share-permissions />
    <oc:tags />
    <oc:favorite />
    <oc:owner-id />
    <oc:owner-display-name />
    <oc:share-types />
    <oc:comments-unread />
  </d:prop>
</d:propfind>' | xmllint --format -
```

max-nextcloud:
@tobiasKaminsky I think this should fix both issues now. Also added regression tests

lib/DAV/WorkspacePlugin.php
```diff
@@ -96,26 +96,38 @@ public function propFind(PropFind $propFind, INode $node) {
 		}
 
 		// Only return the property for the parent node and ignore it for further in depth nodes
-		if ($propFind->getDepth() === $this->server->getHTTPDepth()) {
+		$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($node) {
```
max-nextcloud, line 99: I'm not entirely sure about removing this condition. https://github.com/nextcloud/text/pull/2608#issuecomment-1162715522 sounds like workspace needs to be set for nested folders and we rely on it being set in our new tests for depth=0.

My understanding is that this will only be evaluated anyway if the property is requested. Would like to hear @juliushaertl  opinion on this before a merge though.

tobiasKaminsky:
> @tobiasKaminsky I think this should fix both issues now. Also added regression tests

Looks good 👍👍👍

tobiasKaminsky:
I found a strange bug?
On Android:
- create subfolder 1
- go into it
- add folder content / rich workspaces

Use Insomnia/cURL to show rich-workspaces property of /1/ folder -> see nothing
Force refresh folder in Android -> still nothings
Go to web in /1/ folder, see rich workspaces -> then it is also in Insomnia / Android

max-nextcloud:
This PR reverts a change that was introduces as a performance fix. We need to evaluate the impact of undoing that fix.

szaimen:
> This PR reverts a change that was introduces as a performance fix. We need to evaluate the impact of undoing that fix.

👍 I hope this will be possible before the next RC for 24.0.3...

max-nextcloud:
So i profiled this some.
<details>
<summary>Profiling Command</summary>

```bash

for n in {1..10} ; do curl -w "%{time_total}\n" -o /dev/null -s 'http://nextcloud.local/remote.php/dav/files/admin/' -X PROPFIND -u admin:admin --data '<?xml version="1.0"?>
<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns" xmlns:ocs="http://open-collaboration-services.org/ns">
  <d:prop>
    <d:getlastmodified />
    <d:getetag />
    <d:resourcetype />
    <oc:fileid />
    <nc:file-metadata-size /><oc:permissions />
    <oc:size />
    <d:getcontentlength />
    <d:quota-available-bytes />
    <nc:rich-workspace /> <nc:rich-workspace-file /> <nc:has-preview />
    <nc:mount-type />
    <nc:is-encrypted />
    <ocs:share-permissions />
    <oc:tags />
    <oc:favorite />
    <oc:owner-id />
    <oc:owner-display-name />
    <oc:share-types />
    <oc:comments-unread />
  </d:prop>
</d:propfind>'; done | LC_ALL=C awk '{ total += $1; count++ } END { print total/count }'
```
</details>

## 22 Folders inside root folder of admin:

* `fix/dav`: 0.542629
* `master`: 0.414488

## 222 Folders inside root folder of admin:

* `fix/dav`: 2.3157
* `master`: 0.763323

juliushaertl:
@tobiasKaminsky Can you maybe clarify on how the android client is requesting the workspaces especially in regards to the Depth value? From https://github.com/nextcloud/text/pull/994 we had issues in the past with large folder structures that is why the response was limited to only return the workspace on the requested root node.

tobiasKaminsky:
@juliushaertl initial bug fix works, what we discussed yesterday!
Also works on iOS.

But this one still happens:

> I found a strange bug? On Android:
> 
>     * create subfolder 1
> 
>     * go into it
> 
>     * add folder content / rich workspaces
> 
> 
> Use Insomnia/cURL to show rich-workspaces property of /1/ folder -> see nothing Force refresh folder in Android -> still nothings Go to web in /1/ folder, see rich workspaces -> then it is also in Insomnia / Android

juliushaertl:
> But this one still happens:

@tobiasKaminsky I could reproduce but seems related to the files_lock app. Do you have that one enabled in your test instance? I'd consider that a separate bug then.

tobiasKaminsky:
> @tobiasKaminsky I could reproduce but seems related to the files_lock app. Do you have that one enabled in your test instance? I'd consider that a separate bug then.

I have, yes.
Disabling "fixes" it.
Shall I create a new bug in files_lock, or will you directly create a fix 🤡?

juliushaertl:
> Shall I create a new bug in files_lock, or will you directly create a fix 🤡?

Should be fixed on master already with https://github.com/nextcloud/files_lock/pull/75 but I have a follow up patch in the pipeline that I worked on with @max-nextcloud now.

PR #2612, opened 2022-06-23
Fix conflict between @nextcloud/stylelint-config and manually added peerDependencies

susnux:
* See comment https://github.com/nextcloud/text/pull/2519#issuecomment-1164400486
* Target version: master 

### Summary
1. Remove manually added [peerDependencies](https://github.com/nextcloud/stylelint-config/blob/e15bae69defa0745da9056f87feeb70f947f0a95/package.json#L10) of @nextcloud/stylelint-config as they are installed, using the correct version, automatically by npm.
2. Updated package-lock.json to NPMv7 version (lockfileVersion 2), prevent warnings when using npm v7 which is the minimum supported version on the project anyways. Dependabot supports it as well, see:\
https://github.blog/changelog/2021-02-08-dependabot-npm-v7-support/

PR #2614, opened 2022-06-24
Remove unnecessary dependencies

skjnldsv:
They are not needed as they are automatically handled by npm7 and the associated configs packages

Closes #2612
cc @susnux

Btw, you seems to have a lot of dependencies in dev that should be in the `dependencies` section ? Please double-check, they might be here because of the secondary vite building package, :shrug:


susnux:
For the app package it should make not difference if you put the dependencies into dev- or dependencies. But for the @nextcloud/text package that would make a difference.

skjnldsv:
> For the app package it should make not difference if you put the dependencies into dev- or dependencies. But for the @nextcloud/text package that would make a difference.

Well, I'd say since they are runtime deps, they should be in `dependencies`, since we're exclusing those anyway from the final bundles: https://github.com/nextcloud/text/blob/96f0cdd92400454b1d9956f28dd284978f68c5f1/vite.config.js#L18

skjnldsv:
Well, according to the components used in the `@nextcloud/text` package, I see absolutely no usage of those lib. They should be in `dependencies`....
https://github.com/nextcloud/text/blob/96f0cdd92400454b1d9956f28dd284978f68c5f1/src/components/RichTextReader.vue#L28-L30
https://github.com/nextcloud/text/blob/96f0cdd92400454b1d9956f28dd284978f68c5f1/src/extensions/RichText.js#L23-L46
https://github.com/nextcloud/text/blob/96f0cdd92400454b1d9956f28dd284978f68c5f1/src/components/BaseReader.vue#L28-L29

vinicius73 (approved):

package.json
```diff
@@ -93,65 +100,28 @@
 		"npm": "^7.0.0"
```
vinicius73, line 100: Can we just remove it?
It not affects the host's npm.

max-nextcloud:
> Well, according to the components used in the `@nextcloud/text` package, I see absolutely no usage of those lib. They should be in `dependencies`....

My understanding is that if they are not used at all vite will not package them. But if they are in dependencies apps that use `@nextcloud/text` will start packaging them, right? Is that what we want? For example should collectives start packaging them?

package.json
```diff
@@ -93,65 +100,28 @@
...
 		"@vue/test-utils": "^1.3.0",
 		"@vue/vue2-jest": "^27.0.0",
-		"babel-loader": "^8.2.5",
-		"core-js": "^3.22.7",
-		"css-loader": "^6.7.1",
 		"cypress": "^10.1.0",
 		"cypress-file-upload": "^5.0.8",
 		"debounce": "^1.2.1",
 		"escape-html": "^1.0.3",
-		"eslint": "^8.18.0",
```
max-nextcloud, line 122: This is a peer dependency of `@nextcloud/eslint`. My understanding is that peerDependencies of dependencies should be added as devDependencies to ensure they are installed and keep track of their versions even though npm will auto install peer dependencies nowadays.

max-nextcloud:
So far my thinking was this:
1. if it's a runtime dependency of the package -> add to dependencies
2. if it's not a runtime dependency of the package but a dependency for building the app -> add to devDependencies
3. if it's a peerDependency of one of our dependencies or devDependencies -> add to devDependencies.

The goal of 1. and 2. is to depend only on the things the package actually requires. If we forget something here it will be bundled with the package and the package will still work - but it will grow in size. I was thinking about adding a CI check for that as it would be hard to notice otherwise.

About 3. I was mostly following what i heard from @vinicius73 and thinking this makes conflicts explicit and defines a working set of dependencies. I'm fine with fully relying on npm magic here. I think we should pick one strategy and stick to it - maybe even document it and standardize it?

vinicius73:
> About 3. I was mostly following what i heard from @vinicius73 and thinking this makes conflicts explicit and defines a working set of dependencies. I'm fine with fully relying on npm magic here. I think we should pick one strategy and stick to it - maybe even document it and standardize it?

We have alt least 2 problems.

- NPM and Node inconsistency: Node v14 is our development target, but we use NPM v7, a legacy NPM version who is not bundled with Node.

- We are using the same `package.json` to be used in an _application_ and a _package_.


From the perspective of an application, every dependency matters. Those dependencies can be declared as `dependencies` (who will be bundled in production) or `devDependencies` (who will not be bundled in production).
From the perspective of a package, we must avoid bundle dependencies with the release package.

We can choose between `dependencies`, `devDependencies` and 'peerDependencies`, but it only makes sense in a _package_, not in an _application_.

susnux:
> 3\. if it's a peerDependency of one of our dependencies or devDependencies -> add to devDependencies.
>
> ...
> 
> About 3. I was mostly following what i heard from @vinicius73 and thinking this makes conflicts explicit and defines a working set of dependencies. I'm fine with fully relying on npm magic here. I think we should pick one strategy and stick to it - maybe even document it and standardize it?

3. was the way to go with npm 3-6, but led to version conflicts as you have to manually check that the peer dependencies and your dev dependencies match.
And monitor any bot output as a dependency update might break the manually pinned peer dependencies

> * NPM and Node inconsistency: Node v14 is our development target, but we use NPM v7, a legacy NPM version who is not bundled with Node.

Node 14 bundles npm 6, I guess 7 is used because of it's huge improvements(?). In 10 months node 14 will be EOL and the current LTS is node 16 bundling npm 8.
So there should be no problem with that versions.

package.json
skjnldsv: > This is a peer dependency of `@nextcloud/eslint`. My understanding is that peerDependencies of dependencies should be added as devDependencies to ensure they are installed and keep track of their versions even though npm will auto install peer dependencies nowadays.

It's the other way, we made sure each config dependency handle their required versions themselves.
All the other apps are doing it that way, it actually helped us a lot to reduce noise and ensure further compatibilities

skjnldsv:
> About 3. I was mostly following what i heard from @vinicius73 and thinking this makes conflicts explicit and defines a working set of dependencies. I'm fine with fully relying on npm magic here. I think we should pick one strategy and stick to it - maybe even document it and standardize it?

This is already standardize with those global packages.We did a lot of work over the last year and a half to go that direction.
With npm 7 came the fix for peer dependencies. We took advantage of that and migrated to global packages.

skjnldsv:
> * NPM and Node inconsistency: Node v14 is our development target, but we use NPM v7, a legacy NPM version who is not bundled with Node.

No, this is not an issue
https://github.com/nextcloud/text/blob/96f0cdd92400454b1d9956f28dd284978f68c5f1/package.json#L91-L94
You just have to be compliant about what your app requires

package.json
skjnldsv: No, this is what we requires at Nextcloud for now.
Min npm7 and node14. We move alltogether when npm7 got out and communicated about it already

max-nextcloud:
Okay... so removing the peer dependencies of our dependencies for package.json seems to be the agreed upon way to go within Nextcloud. I'm fine with that. And that will fix the issue this PR is addressing, right?

I think dependencies that are not runtime dependencies of the package should stay in devDependencies and not become dependencies. There's a number of them where it's hard to make a clear decision. I moved everything into devDependencies that is not imported by the package - as that was an easy criterion to apply. But one could obviously argue that a package that exposes a vue component depends on vue.

I'm okay with moving on with this PR but i'd probably move for example `@nextcloud/initial-state` back into dev dependencies as there's no reason for an app that uses `@nextcloud/text` to also include `@nextcloud/initial-state` if it's not using it.

max-nextcloud (approved):

skjnldsv:
> I'm okay with moving on with this PR but i'd probably move for example `@nextcloud/initial-state` back into dev dependencies as there's no reason for an app that uses `@nextcloud/text` to also include `@nextcloud/initial-state` if it's not using it.

The dependency might be installed yes, but it will not be part of the final bundle of the app then
In any case, for me this is an extra point against https://github.com/nextcloud/standards/issues/3. If you want to do that, please use a differet package.json. Having unrelated depdndencies seems like a bad idea :/

Issue #2607, opened 2022-06-21
Button style


luka-nextcloud:
- [x]  use primary for active icons, primary-light for the background
- [x]  raise opacity for all buttons by default
- [x]  current opacity for the disabled buttons
- [x]  Remove blue circle below active icon

PR #2627, opened 2022-06-28
Update toolbar button style

luka-nextcloud:

* Resolves: #2607 
* Target version: master 

### Summary

- [x]  use primary for active icons, primary-light for the background
- [x]  raise opacity for all buttons by default
- [x]  current opacity for the disabled buttons
- [x]  Remove blue circle below active icon


luka-nextcloud:
Here are videos on desktop and mobile:

[Screencast from 30-06-2022 23_32_23.webm](https://user-images.githubusercontent.com/89908051/176732548-e2fde24e-8620-43a7-a677-3ec548330a69.webm)

[Screencast from 30-06-2022 23_34_56.webm](https://user-images.githubusercontent.com/89908051/176732571-0accfefa-7f76-4a9f-b17a-e484a3066a00.webm)

jancborchardt:
Looks nice! Just what's the blue border around the textarea, that shouldn't be there. :)

vinicius73:
> Looks nice! Just what's the blue border around the textarea, that shouldn't be there. :)

I guess is related to https://github.com/nextcloud/server/pull/31584

vinicius73 (approved):

src/components/Menu/ActionEntry.scss
```diff
@@ -1,16 +1,9 @@
...
-		background: var(--color-primary-element);
-		bottom: 0;
-		height: 6px;
-		width: 6px;
-		content: '';
-		left: 50%;
+	background-color: var(--color-primary-light);
+	border-radius: 50%;
+	svg {
+		fill: var(--color-primary) !important;
```
vinicius73, line 6: Is it possible to avoid `!important` here and in line 75?

luka-nextcloud: I tried to get rid of it but we need it in those lines.

juliushaertl: Maybe setting `color: var(--color-primary);` on the parent could work as the svg should use currentColor?

Issue #470, opened 2019-12-11
Add Word Count


TobiasPreiss:
It would be nice if the app would have a function to count the number of words in the document text.

juliushaertl:
We had that in the notes app as well at some point iirc but it is not present anymore. @jancborchardt is that something we want to have or should we keep the UI clean from additional details like that?

jancborchardt:
Word count is indeed useful for writers. How it could be displayed is as the last item in the 3-dot-menu list, simply saying "14 words" or whichever number it is.

The entry doesn’t need an icon and would be inactive. As such the text should be grey via `var(--color-text-maxcontrast)`.

If there would be no 3-dot-icon, it’s fine if the word count is the only entry in the 3-dot-menu – so there would always be a menu. (We could also think about moving the modified date into there, but that’s a separate discussion.)

elpraga:
Are there any news about this feature?

IMHO it could be displayed in the bottom or top corner in a similar way as the "last saved counter" - there is usually an abundance of space that is unused. (in case it is needed, it could disappear - the same way the counter does)

What do you think about this idea?

elpraga:
What do you think about the idea @jancborchardt ? I am not a writer, I am an English teacher. I'd like to use **Text** app to emulate the computer based exams, hence my interest in the word count.

jancborchardt:
No news and my design proposal in https://github.com/nextcloud/text/issues/470#issuecomment-577464003 still stands. If there is news, this issue will be updated. :)

PR #2630, opened 2022-07-01
Feature: Show word count in RichText

susnux:
* Resolves: #470
* Target version: master 

### Summary
Show word count as last element of the overflow menu,
which now will be always shown. The word count
is updated when the menu is openend as the
tiptap word count is not reactive.

Modified ActionList: Added a named slot for adding
elements that should always shown as the last elements.

---
What should be mentioned: This solution refreshes on menu open, but it is also possible to refresh it on content change.
So the current solution do not introduce many unneeded calculations, but (and I am not sure if this is even possible) does not update the label if typing while the menu is open.


juliushaertl:
I'm a bit hesitant to add even more to the menu, especially on mobile. Also it is not fully related to the other actions which are more formatting actions that are targeting the current position/selection. To me it would be more suitable on the right hand side. Maybe we need to adapt the "collaborator" menu a bit to cover more details about the document in general.

For reference this is how it currently looks:
Screenshot 2022-07-04 at 14 09 37

Screenshot 2022-07-04 at 14 07 58

@jancborchardt What do you think?

Dropbox Paper for example also has a separate context menu for the document itself: Screenshot 2022-07-04 at 14 09 37

PR #2660, opened 2022-07-03
cleanup: wait statements - rely on promises instead

max-nextcloud:
If a command returns a promise cypress will wait for that promise
before proceeding with the next commands.

Since 0162bb016c we return the promises from most commands.
The explicit calls to cy.intercept and cy.wait are not needed anymore.

* Target version: master


max-nextcloud:
sigh... looks like opening a file in a share does not trigger the same propfind - probably because the sidebar is not shown or so.

mejo- (approved):

PR #2662, opened 2022-07-04
Fix/base reader injection

mejo-:
This PR replaces #2628 by @susnux in order to allow our compile bot to push to the PR branch.

* Resolves: https://gitlab.com/collectivecloud/collectives/-/issues/349
* Target version: master

### Summary

BaseReader: Use a factory pattern for creating the extension list

Provide functions are neither called on the providing object nor
on the injecting object, so using this within provided functions
will refer to the global this for which $emit is not defined.
This uses a factory pattern so that this is bound to the injecting
object.

Also forwarding the click-link event is needed.


susnux:
Thank you!

max-nextcloud:
there's no point in backporting this to stable24. The code in question did not exist there as far as i can tell.

Issue #107, opened 2019-06-27
Outline view of headings for quick navigation (table of contents)


jancborchardt:
Here’s some examples of how other apps use headings to automatically create an outline for quick navigation. This is always shown on the left and we could do that too.

## Dropbox Paper
Very nicely designed, non-obtrusive by default, with h1 and h2 levels highlighted on the left:
https://blog.dropbox.com/topics/product-tips/paper-hidden-gems
![dropbox](https://user-images.githubusercontent.com/925062/60253644-ca856180-98cc-11e9-8b5d-48bb86029fb9.gif)

--------------------

## Google Drive
h1 and h2 are shown normally, and it seems starting with h3 they are indented:
![google docs outline](https://user-images.githubusercontent.com/925062/60253643-ca856180-98cc-11e9-9778-0df7fb147367.jpg)

--------------------

## [FoldingText](http://www.foldingtext.com/)
![foldingtext outline](https://user-images.githubusercontent.com/925062/60253645-ca856180-98cc-11e9-8f49-536f1553763c.png)

thorsten-panknin:
I was just looking around for an issue mentioning a better overview of headlines and stumbled over this. Great idea!

feutl:
like this too!

b3:
Any idea if this still considered and how one can try to help?

biva:
This idea would be great and would make the Collectives app even better!

As a complement (or maybe it would be a workaround?), it would be great to make sections collapsible (https://github.com/nextcloud/text/issues/1841)

rinurinu:
I would love seeing this feature implemented. I use nextcloud text a lot and this would help declutter long notes.

rotdrop:
We can actually add any mature Wiki-Software to the list of apps which have this feature, including Gitlab markdown-pages and perhaps Github. Comparison with Wiki-Software is perhaps a bit unfair if we look at the text-app alone. However, as the Nextcloud-Wiki app, which calls itself "collectives", is using "text"-app as backend it seems that many features yet missing from the collectives can only be implemented if they are also present in the text app. Besides TOC and general inter-page links (links to headings in the same document and to headings in other documents) also foot-notes come into my mind.

b3:
Comparison with wiki may be unfair, but comparison with other collaborative editing tools may not be that unfair.
[...]

PR #2664, opened 2022-07-04
Feature: Show word count in RichText

susnux:
* Resolves: #470, closes #2630
* Target version: master 

### Summary
Show word count as last element of the overflow menu,
which now will be always shown. The word count
is updated when the menu is openend as the
tiptap word count is not reactive.

Modified ActionList: Added a named slot for adding
elements that should always shown as the last elements.

---
What should be mentioned: This solution refreshes on menu open, but it is also possible to refresh it on content change.
So the current solution do not introduce many unneeded calculations, but (and I am not sure if this is even possible) does not update the label if typing while the menu is open.


susnux:
from https://github.com/nextcloud/text/pull/2630#issuecomment-1173746059 @juliushaertl 

---

I'm a bit hesitant to add even more to the menu, especially on mobile. Also it is not fully related to the other actions which are more formatting actions that are targeting the current position/selection. To me it would be more suitable on the right hand side. Maybe we need to adapt the "collaborator" menu a bit to cover more details about the document in general.

For reference this is how it currently looks:
Screenshot 2022-07-04 at 14 09 37

Screenshot 2022-07-04 at 14 07 58

@jancborchardt What do you think?

Dropbox Paper for example also has a separate context menu for the document itself:
Screenshot 2022-07-04 at 14 10 37

vinicius73:
@susnux I'm working in a "sidebar" (ouline) for text editor to wrap the Table Of Contents #107 
Maybe we can use it to show word counter as well.

I have some visual issues, but I will open a PR and get some feedback from the team.

susnux:
@jancborchardt  affirmed this should go into the "..." menu: https://github.com/nextcloud/text/pull/2688#issuecomment-1184356837

So I think this is ready to go.

max-nextcloud (approved):
Code looks good. Thanks for including a test.
Will give it a try and then work towards getting this merged.

max-nextcloud:
UI also looks okay to me. It looks somewhat funny when it's the only entry in the three dots menu. But i guess that's okay. We'll redesign the menubar anyway.

mejo-:
I rebased the PR to latest master and pushed a few additional changes. Hope you don't mind, @susnux.

After discussing with @jancborchardt we decided (and I implemented) to:
* Add an icon to the word count
* Alsow add character count
* Always move the formatting help under the three-dot-menu. That way the counts are never the only entry

Screenshot:

grafik

susnux:
:+1: 
The reason why I did not use `ActionText` but `ActionCaption` is that the latter one does not show a pointing cursor when hovering it. As there is no action behind the the text, so the action text is a bit confusing.

mejo-:
>  The reason why I did not use ActionText but ActionCaption is that the latter one does not show a pointing cursor when hovering it. As there is no action behind the the text, so the action text is a bit confusing.

I agree. At the same time, `ActionCaption` doesn't support an icon slot, so that's why I went with `ActionText`. I think, it's an acceptable tradeoff, don't you think so? Also, there's still no background change on hover, so it's different from the clickable menu entries.

susnux:
Yes I think this is ok, just wanted to mention this as it might be confusing.

jancborchardt (approved):
Looks great, thanks a lot @susnux and also @mejo- :)

mejo- (approved):

vinicius73 (approved):

adrienneteicher:
Hi all,
This is just what I need but I don't see it in my version of the app. I am on 3.5.1
Is it that I should just be patient for updates?
Thank you!

mejo-:
Hey @adrienneteicher. Which Nextcloud version do you use? The feature is only available starting with Nextcloud 25.

susnux:
`3.5.1` is the version used on Nextcloud 24. So you would have to upgrade to Nextcloud 25 to get this feature.

adrienneteicher:
Thanks Patrizio. Yes, I am on Nextcloud 24. I have to wait for my 
provider to upgrade to 25.

PR #2666, opened 2022-07-04
Backport/stable24/2326 update eslint config

max-nextcloud:
* Backports: #2326 
* Target version: stable24


max-nextcloud:
This is not a bugfix - but i think it will make our lives much easier in the future as it reduces the diff between stable24 and master significantly and means we apply the same style in both.

juliushaertl (approved):

PR #2685, opened 2022-07-06
Update to Vue 2.7

vinicius73:
* Resolves: #2670 #2744 #2689 
* Target version: master 

## Summary

Update to Vue 2.7

https://blog.vuejs.org/posts/vue-2-7-naruto.html


max-nextcloud:
Thanks a lot for looking into this :heart_eyes:

juliushaertl:
🥳

css/prosemirror.scss
```diff
@@ -2,7 +2,7 @@
 
 /* Document rendering styles */
 div.ProseMirror {
-	margin-top: 44px;
+	// margin-top: 44px;
```
juliushaertl, line 5: Looks unrelated

vinicius73:
It is working now 🎉

css/prosemirror.scss
juliushaertl: @vinicius73 Can you check that, then we could get this in so we can get the PR list down a bit :)

vinicius73: Yes, it's not related, but I need to send another PR to fix it.

juliushaertl (approved):

juliushaertl:
Vue packages version mismatch:
- vue@2.7.8 (/home/runner/work/text/text/node_modules/vue/dist/vue.runtime.common.js)
- vue-template-compiler@2.7.4 (/home/runner/work/text/text/node_modules/vue-template-compiler/package.json)

package.json
```diff
@@ -89,12 +89,12 @@
 		"prosemirror-tables-contently": "^1.1.2",
 		"prosemirror-transform": "^1.6.0",
 		"prosemirror-utils": "^1.0.0-0",
-		"prosemirror-view": "^1.26.4",
+		"prosemirror-view": "^1.26.5",
 		"proxy-polyfill": "^0.3.2",
 		"tippy.js": "^6.3.7",
-		"vue": "^2.6.14",
+		"vue": "^2.7.8",
```
juliushaertl, line 95: Let's maybe add vue-template-compiler as a dependency directly then dependabot can also take care of bumping them in sync

vinicius73: Done.

marcelklehr:
> Vue packages version mismatch

The migration docs say template compiler isn't necessary anymore as a separate dependency in v2.7, I think

vinicius73:
> The migration docs say template compiler isn't necessary anymore as a separate dependency in v2.7, I think

Yes, it is true.

But I guess we must fix it on https://github.com/nextcloud/webpack-vue-config/blob/b3100850cc78c51fa6258d53b2116484f7fff64d/package.json#L38

package.json
```diff
@@ -108,23 +108,25 @@
...
+		"jest": "^28.1.3",
+		"jest-environment-jsdom": "^28.1.3",
 		"jest-serializer-vue": "^2.0.2",
 		"lowlight": "^1.20.0",
 		"mitt": "^3.0.0",
-		"vite": "^2.9.9",
+		"vite": "^3.0.2",
 		"vite-plugin-commonjs": "^0.5.0",
-		"vite-plugin-vue2": "^2.0.0"
+		"vite-plugin-vue2": "^2.0.2",
```
marcelklehr, line 127: > 2.7 support for Vite is provided via a new plugin: [@vitejs/plugin-vue2](https://github.com/vitejs/vite-plugin-vue2). This new plugin requires Vue 2.7 or above and supersedes the existing [vite-plugin-vue2](https://github.com/underfin/vite-plugin-vue2).

from https://github.com/vuejs/vue/blob/main/CHANGELOG.md#vite

marcelklehr:
> Note: if you are using @vue/test-utils, you may need to keep it in the dependencies for now, but this requirement will also be lifted in a new release of test utils.

Ah, when using test-utils it might still be needed

package.json
```diff
@@ -108,23 +108,25 @@
 		"@nextcloud/browserslist-config": "^2.2.0",
 		"@nextcloud/eslint-config": "^8.0.0",
```
marcelklehr, line 109: > Update eslint-plugin-vue to latest version (9+) if you run into unused variable lint errors when using <script setup>.

For good measure uping eslint-config to 8.1 pulls in eslint-plugin-vue v9 instead of v8

vinicius73: It is not release yet.
https://www.npmjs.com/package/@nextcloud/eslint-config

marcelklehr: Ah, nevermind then. We'll probably get the new version when it's out anyway.

marcelklehr (approved):

Issue #2337, opened 2022-04-25
File open always triggers save and overwrites last modified date


relikd:
**Describe the bug**
I have multiple files which are saved as soon as they are opened. I am not sure why these files are updated. I tried to reproduce the bug with a new file but the new file was not affected. Maybe something with version history.

However, the network log shows a needless push `"stepType":"replace","from":35,"to":35`

**Expected behavior**
The file should not be written to disk if it has no changes. The unnecessary write destroys the "last changed" date.

**Screenshots**

https://user-images.githubusercontent.com/1321720/165187173-0049478b-7d4d-4c6a-852c-0a623c6001a0.mp4

**Client details:**
 - OS: macOS
 - Browser: Firefox

<details>
<summary>Server details</summary>

**Text app version:** 3.4.0

**Operating system:** Arch

**Web server:** nginx

**Database:** mysql 8

**PHP version:** 8

**Nextcloud version:** 23.0.2 and 23.0.3

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
N/A

#### Browser log
```
$ is deprecated: The global jQuery is deprecated. It will be removed in a later versions without another warning. Please ship your own. 9 main.js:1358:3634
Opening viewer for file  /Briefkasten/Protokoll-2022-01-13.md viewer-main.js:2:993487
oc_defaults is deprecated: use OC.theme instead, this will be removed in Nextcloud 20 main.js:1358:3634
Fetching additional files... viewer-main.js:2:994064
receivedSteps newVersion 1 editor.js:2:132079
receivedSteps newVersion 2 editor.js:2:132079
Saved document 
Object { id: <redacted>, currentVersion: 2, lastSavedVersion: 2, lastSavedVersionTime: 1650926125, baseVersionEtag: "<redacted>", initialVersion: 0 }
editor.js:2:124054
No new notification data received notifications-main.js:2:1120986
Polling interval updated to 30000 notifications-main.js:2:1121750
OC.Util.relativeModifiedDate is deprecated and will be removed in Nextcloud 21. See @nextcloud/moment main.js:116:2282
```

```
POST https://<server>/index.php/apps/text/session/push

[...]

PR #2686, opened 2022-07-08
Avoid saving without changes

juliushaertl:
First attempt towards fixing #2337 

- [x] Test case for trailing node transaction triggering the step and save
- [x] Hotfix by appending a trailing paragraph before creating the tiptap document
  - [ ] Hopefully figure out a nicer way to solve this


src/EditorFactory.js
```diff
@@ -122,7 +122,7 @@ const createEditor = ({ content, onCreate, onUpdate, extensions, enableRichEditi
 	}
 	extensions = extensions || []
 	return new Editor({
-		content,
+		content: content + '<p/>',
```
juliushaertl, line 125: I fail to come up with a nicer approach for this, any ideas or objections on using this @nextcloud/text ?

vinicius73: I can't think about better solution than this.

vinicius73:
the expects now have a `p` tag 🙈 

image

juliushaertl:
@vinicius73 Which would be expected so that we always end up with a new line. But I didn't see a failing CI check. Anyways rebased, so let's see.

max-nextcloud:
Ahh.... i see. Tests still need to be updated.

max-nextcloud (approved):
Not ideal but seems to do the trick.

juliushaertl:
/backport cdfcd37e39067fb60b515d306251b0ec24776f70,2fa358f8e7bfdc74158fa1e3a8b352db81806f25,5c56006d74a95396eb8492c45ca17d544f7d7a5e to stable24

juliushaertl:
/backport cdfcd37e39067fb60b515d306251b0ec24776f70,2fa358f8e7bfdc74158fa1e3a8b352db81806f25,5c56006d74a95396eb8492c45ca17d544f7d7a5e to stable23

PR #2687, opened 2022-07-08
Refactor: Split Editor Wrapper into multiple components

max-nextcloud:
`EditorWrapper` is fairly large and hard to understand. It handles almost all editor related things - Sessions, Sync, Document, EditorLayout, Menubars etc.

In order to rework the collaboration it might be helpful to split it into smaller pieces so it becomes easier to understand and handle.

`Editor/Status` handles the status display in the menubar.


max-nextcloud:
This is just some preparation - but i think it's helpful in its own right and changing quite a lot of file structures - so maybe good to get it in already.

juliushaertl:
@max-nextcloud Setting labels is highly appreciated ;)

juliushaertl (approved):

src/components/Editor/Main.vue
```diff
@@ -0,0 +1,54 @@
...
+	name: 'Main',
+	components: {
+		MediaHandler,
+	},
+
+}
+</script>
+
+<style scoped lang="scss">
+	.text-editor__main, .editor {
```
vinicius73, line 43: We do not have a `.editor` element in this component.

vinicius73 (approved):

Issue #299, opened 2019-09-18
Links to sections?


jans23:
**Describe the bug**
My Markdown file contains several sections and I want to link to those. Is this possible somehow?

**To Reproduce**
1. Add a section, say `# section1`
2. Add a link to this section: `[link](#section1)`

**Expected behavior**
When clicking the link, section 1 should be displayed.

**Actual behavior**
A new browser window opens showing the folder in which the Markdown file is stored.

**Client details:**
 - OS: Linux
 - Browser: latest Firefox

sunjam:
Are you referring to Categories, which are literally defined as folders in [Nextcloud Notes](https://github.com/nextcloud/notes/issues).

PR #2520, opened 2022-06-17
Feature: Add page anchors and inter page links

susnux:
* Resolves: #2173, and resolves #299
* Supersedes #2464
* Target version: master 

### Summary

Lightweight version of #2464, see that request for full discussion.

This implements the anchors needed for inter-page links as well as clickable
anchor links.

Issue #2142, opened 2022-02-01
Footnote support would be nice


rotdrop:
**Is your feature request related to a problem? Please describe.**
Trying to convert some Dokuwiki pages to NC collectives I'm missing the support for Footnotes.

**Describe the solution you'd like**
Support the footnote-syntax as described here:
https://www.markdownguide.org/extended-syntax#footnotes

There is an addon package for markdown-it:
https://github.com/markdown-it/markdown-it-footnote

but just adding this would of course not be enough. The editor and also the Collectives app would need to add some display and UI sugar to make the footnotes usable.

**Describe alternatives you've considered**
Don't use NC Collectives

kabajau:
Any update on this? I would like to use markdown created with hedgedoc in nextcloud, but while the former already supports footnotes, the latter doesn't.

juliusknorr:
You will see updates here in the ticket as soon as this is picked up or started. This is currently not scheduled for development, but of course we are happy to receive contributions for features like this.

mejo-:
This is being worked on in #8840.

Issue #145, opened 2019-07-08
Support Frontmatter in MD files


paulhibbitts:
**Is your feature request related to a problem? Please describe.**
I'd love to use the Collaborative editor to share Page files for the Grav CMS (a flat-file CMS). Right now any Frontmatter (https://jekyllrb.com/docs/front-matter/) originally in the file seems to be re-saved in a different format. Frontmatter is used quite widely in both static site generators and flat-file CMSs.

Here is a simple example of Frontmatter for testing purposes:

```
---
title: 'Classic Modern Architecture'
date: '17:34 06/27/2017'
---

Here is the main content area of the MD file.

```

When saved, this is the result:

```
---

title: 'Classic Modern Architecture' date: '17:34 06/27/2017'

Here is the main content area of the MD file.
```

**Describe the solution you'd like**
Any frontmatter contained in a MD should be visually distinct from the main content of the MD file, and also be editable (like a Code Block perhaps?)

**Describe alternatives you've considered**
I've not been able to find any so far.

**Additional context**

Thanks very much! The new collaborative editor is awesome to see in Nextcloud!

juliushaertl:
Thanks for your suggestion. While front matter is not really specified in commonmark, I still agree that we should make sure that it is not changed. So we should make sure to strip it from the markdown parsing and add a simple code block for editing at the beginning of the document.

paulhibbitts:
That sounds great @juliushaertl , thank you for considering this suggestion.

brainchild0:
You could also consider rendering the displayable content, such as title, in appropriate styles.  But somehow the user would need to able to add or to edit the other fields.

Also, not sure, but if the block is appearing disassembled after changes are saved, then maybe there is a separate issue with the way documents are processed.  A MarkDown editor ideally should, I think, preserve as much literal content as possible from the original document, not attempt to produce an entirely new document from the logical features of the AST.

hugoroy:
Hi, I think this feature is definitely important :-)

But I don't think it should be rendered as text like the rest. These blocks are supposed to be metadata (see, e.g. https://pandoc.org/MANUAL.html#extension-yaml_metadata_block which specifies it cleary, or https://jekyllrb.com/docs/front-matter/ which specifies it implicitly with the special rules IMO).

So it probably requires a lot of work to get right from the UI/UX perspectives (the simple solution could be to treat it like a codeblock - but instead of inserting the codeblocks delimiter, use "---" and "---"; but this does not feel very satisfactory…)

[...]

PR #2688, opened 2022-07-08
Outline view

vinicius73:
* Resolves: #107
* Target version: master 

### Summary

Add Outline view to display Table of Contents

image

image

image

https://user-images.githubusercontent.com/1561347/184963989-e19ee919-6fb1-4c07-bac9-60d0be05eab0.mp4


src/components/EditorOuline.vue
```diff
@@ -0,0 +1,33 @@
...
+		</div>
+		<TableOfContents />
+	</div>
+</template>
+
+<script>
+import TableOfContents from './TableOfContents.vue'
+
+export default {
+	name: 'EditorOuline',
```
max-nextcloud, line 14: small typo: EditorOu**t**line

max-nextcloud:
This looks really great! :clap:

mejo-:
Wow, this looks super nice @vinicius73 :heart_eyes: 

How does it look like on mobile, did you find a good solution there already? Also, what happens if your window is not wide enough to display the outline? Text width is 670px, so e.g. when you're on a tablet with 800px.

susnux:
This looks awesome! :heart_eyes: 


One note: I think the ID generation is better suited within the Heading node, like introduced in #2520, so maybe first get the anchors?

---

> Also, what happens if your window is not wide enough to display the outline? Text width is 670px, so e.g. when you're on a tablet with 800px.

at 800px

It is possible to use it, but not that nice. It works perfect starting with ~900 - 1000px.

On some other application this is solved by providing a button, something like `➤`, floating on the left side that enables some kind of drawer functionality. So the toc is faded in from the left side.

jancborchardt (changes requested):
Nice nice! :) Some first feedback:
- [x] The line below "Outline" is not necessary
- [x] The text "Outline" is a bit too present. You can reduce the size to be the same size
- [x] The text "Outline" should be left-aligned with the h1 titles below, currently it is a bit off to the left
- [ ] Currently the outline section doesn’t seem to be vertically aligned with anything. Could you align it so that when the Text app contains normal text in the first line, the text "Outline" and that text in the first line vertically aligns?
- [ ] When clicking to navigate to a section, currently it immediately jumps. Could this be done with an ever-so-quick animation? Just like 100-250ms, so really not long, but an animation to give some feedback would be great.
- [ ] When there are no headings, I assume the whole outline box just vanishes? (Didn’t test yet)
- [ ] Possibly unrelated, but that blue border around the text content is not going to stay, right?


> On some other application this is solved by providing a button, something like ➤, floating on the left side that enables some kind of drawer functionality. So the toc is faded in from the left side.

Yep, agree that would be nice. :)

vinicius73:
@jancborchardt 

> When there are no headings, I assume the whole outline box just vanishes? (Didn’t test yet)

Not now.
The title (outline) and the content (table of contents) are two different components.

The idea is to use outline to wrap "metadata" like table of contents and word counter https://github.com/nextcloud/text/pull/2664#issuecomment-1173773284 and other possible tools.

vinicius73:
@jancborchardt 

> When clicking to navigate to a section, currently it immediately jumps. Could this be done with an ever-so-quick animation? Just like 100-250ms, so really not long, but an animation to give some feedback would be great.

I agree, but unfortunately it will be a hard task for now.
The scroll is handled by [`prosemirror.scrollIntoView()`](https://prosemirror.net/docs/ref/#state.Transaction.scrollIntoView) this method is not able do it.
We must find another way.

Can we postpone it to other issue?

mejo-:
@vinicius73 this will also fix #299, right?

susnux:
> The idea is to use outline to wrap "metadata" like table of contents and word counter [#2664 (comment)](https://github.com/nextcloud/text/pull/2664#issuecomment-1173773284) and other possible tools.

And maybe also for https://github.com/nextcloud/text/issues/145 (?)

jancborchardt:
> The idea is to use outline to wrap "metadata" like table of contents and word counter https://github.com/nextcloud/text/pull/2664#issuecomment-1173773284 and other possible tools.

Word count can go in the 3-dot menu as a last entry, that’s a much cleaner place for it. Let’s keep outline for outline. :)

> The scroll is handled by [prosemirror.scrollIntoView()](https://prosemirror.net/docs/ref/#state.Transaction.scrollIntoView) this method is not able do it. We must find another way.
> Can we postpone it to other issue?

Yes absolutely, this is not critical to this issue.

vinicius73:
@susnux 
> And maybe also for #145 (?)

Yes, it is a possibility.

vinicius73:
@mejo- 
> @vinicius73 this will also fix #299, right?

It is related, but this PR no dot resolver this issues yet.
@susnux's PR is directly related #2520, although we need to discuss more about the better way to do it.

juliushaertl:
> @jancborchardt
> 
> > Word count can go in the 3-dot menu as a last entry, that’s a much cleaner place for it. Let’s keep outline for outline. :)
> 
> And about other stuffs like #145 #3 #2142 ?

As discussed lets keep those discussions separate and start off with the outline view only.

juliushaertl:
Very nice, just a few additional comments from my side:

- The animation is nice but feels a bit slow, also it does slide in every time you change the headline, maybe we can only do that on the initial adding and speed it up a bit
- I think the outline should probably be hidden by default and then we can have a button to show/hide it
  - Not sure where to position it, but maybe we can put it to the top left as the app navigation toggle would be on other apps
  - We could store the state if open or closed in the browser local storage to remember the user preference at least locally
  - On mobile we could hide it in the right popover menu next to the author colors

What do you think?

mejo-:
Thanks @vinicius73, that looks really awesome :heart_eyes:

On ultra-wide screens, the outline can be a bit too separated from the text container though. What do you think about moving the `editor-outline` container closer to the `editor__content` container? In other words, aligning `editor-outline` to the right instead of the left of `text-editor__content-wrapper__left`?

> I think the outline should probably be hidden by default and then we can have a button to show/hide it

I agree that we should have a button to show/hide the outline for sure. And it should be possible to display the outline on narrow screens (e.g. mobile) as well, then possibly overlapping the text content.

>     * Not sure where to position it, but maybe we can put it to the top left as the app navigation toggle would be on other apps

That unfortunately would not work with Collectives (and other apps that might inject Text in the future), so I don't think it's a good idea.

>     * We could store the state if open or closed in the browser local storage to remember the user preference at least locally

Nice idea!

jancborchardt:
> I think the outline should probably be hidden by default and then we can have a button to show/hide it 

2 options I’d favor:

**Simple:** Since it is related to headings, I’m thinking it could be the last item in the headings menu "Show outline"/"Hide outline". Yes it is a bit buried, but it’s also not super important in all cases? It should be shown / hidden by default either based on screen size (definitely hidden by default on mobile).

**Fancy:** Like Dropbox does, with the slight line indicators. If you hover/focus, the text shows. This also works on mobile, if it opens on tap and the element has a background.

jancborchardt:
> On ultra-wide screens, the outline can be a bit too separated from the text container though. What do you think about moving the `editor-outline` container closer to the `editor__content` container? In other words, aligning `editor-outline` to the right instead of the left of `text-editor__content-wrapper__left`?

What would be important is that the text keeps left-aligned so it stays nice to scan. But right-aligning the container of the outline and giving it a max-width + ellipsizing the text on overflow sounds good.

> We could store the state if open or closed in the browser local storage to remember the user preference at least locally

Yesss, the preference should be remembered. :)

mkrecek234:
Can you also provide the outline view to have levels collapsible/or hide sub-chapters? Cool work.
Eventually you can also show the carets/" > " signs next to the header in the text view on the right, so also you can hide sub-chapters in the running text. This is particularly helpful for large documents.

juliushaertl:
@vinicius73 I took the liberty to push a few commits during review. I changed quite a bit to enable the usage of it also on mobile. Other than that I've added some extra handling to only update the changed props of the headings list, so that DOM nodes can be preserved and the animation will only be used when the node gets added initially.

Screen share is here:

https://user-images.githubusercontent.com/3404133/184545053-7bbdb5dd-8431-498c-9028-73548d83564d.mov

juliushaertl (approved):

juliushaertl:
> On ultra-wide screens, the outline can be a bit too separated from the text container though. What do you think about moving the editor-outline container closer to the editor__content container? In other words, aligning editor-outline to the right instead of the left of text-editor__content-wrapper__left?

I'd say this is something for a follow up then.

vinicius73:
Ready to review

juliushaertl (approved):

juliushaertl:
@vinicius73 Can you make sure every commit message has a signed off message and rebase to latest master? :)

Cypress failure looked unrelated so I restarted that.

mirisbowring:
When will this be available in Nextcloud? Everything is up to date on my installation.

susnux:
> When will this be available in Nextcloud? Everything is up to date on my installation.

With nextcloud 25, scheduled for end of October (afaik)

Issue #2703, opened 2022-07-10
Link text escaping of backslash is removed


susnux:
### Describe the bug
If a link text contains a backslash the escaping of it is removed after the serialization, this leads to invalid markdown after the file is opened the next time.

### To Reproduce
Consider this markdown file (from [example 549](https://spec.commonmark.org/0.30/#example-549) of the commonmark specs):
```markdown
[bar\\]: /uri

[bar\\]
```

Open is using the `text` app, save it and have a look at the content using a plain text editor.

**Expected behavior**
Best case would be no modification, but at least the link should stay the same, so following output is expected:
```markdown
[bar\\](/uri)
```

But instead this is generated by the text app:
```markdown
[bar\](/uri)
```

Which is not a link as now the backslash escapes the `]`.

### Details
This was introduced as a fix for #325 with 3c350b2d41ea352e9d5131ff1def87182402a6e2.
There is this testcase for it, which I am not sure why it is there as this is not valid markdown: https://github.com/nextcloud/text/blob/a1a3b8720702d6fdeb6bdd7b966cbb7cb0edc4c6/src/tests/markdown.spec.js#L100

```markdown
- [ [asd](sdf)
```

The specification forbids (github) forbids more than one character inside task items and CommonMark forbids unescaped brackets within link texts, so this should be interpreted as this instead:
```markdown
- \[ [asd](sdf)
```

Issue #2577, opened 2022-06-18
Messed up entire markdown syntax


vittorio-c:
**Describe the bug**

The app messed up the entire markdown syntax of my file.

**To Reproduce**

1. Create a markdown file in your regular markdown editor.
2. Add normal stuff like lists, break lines, code fences, math expressions.
3. Then open up your file in the "nextcloud/text" app and save the file.
4. Open back your file in your regular markdown editor, and see errors : all your makrdown syntax is entirely messed up :
- Backslash are added on every single math anotations. 
- Break lines are removed. 
- Lists '-' are replaced with '*'. 
- '<' and '>' in URL are simply... removed. 
- Langage indices (ex: 'python') in code fences are removed.

Is this normal behavior ? It sounds like the app is trying to avoid langage injection, or something like this. But I don't see why, since it's just regular file. There is no "excecution" capability in markdown file, isn't it ?

**Expected behavior**

None of my markdown syntax is touched.

**Client details:**
 - OS: Linux Manjaro 5
 - Browser: Firefox
 - Device: Desktop

**Text app version:** 3.5.1

**Web server:** Nextcloud 24

susnux:
Just as a user:

> * Backslash are added on every single math anotations.
> 
> * Break lines are removed.

Sounds bad, probably a bug?

> * Lists '-' are replaced with '*'.

Text uses markdown-it to convert it convert markdown to html and later tiptap to edit the html and later on recreate markdown.
So this is normal behavior as markdown-it parses it as a list and tiptap also recreates a list, the previous style is simply not saved and the common mark specs say both are equal.

> * '<' and '>' in URL are simply... removed.

But both characters are not even allowed within an URL.
So I would guess some of the used libraries simply validates the URL. See also [rfc3986](https://www.rfc-editor.org/rfc/rfc3986/)

> * Langage indices (ex: 'python') in code fences are removed.

That's also sound's like a bug, as markdown-it and tiptap should support those (convert by commonmark specs)

XueSheng-GIT:
I assume this is related to https://github.com/nextcloud/text/issues/593

juliushaertl:
Since this ticket contains different separate topics, I'll close this and would ask you to open individual issues for each syntax problem together with a simple markdown example to reproduce.

> Langage indices (ex: 'python') in code fences are removed.

I could reproduce this and opened https://github.com/nextcloud/text/issues/3328

> '<' and '>' in URL are simply... removed.

[...]

Issue #593, opened 2020-01-19
Markdown files lose formatting on save


inthreedee:
**Summary**
Markdown files that are created in external markdown and text apps will have their formatting completely overridden and/or lost when edited through the Nextcloud web interface.

**Details**
I use external text editors for my markdown files and then sync them to my Nexctloud.  I primarly use Atom for its markdown support but will also use basic text editors like Gedit on Linux or Notepad on Windows.  That's the great thing about markdown files, after all, they're just plain text.  When I edit these files through my Nextcloud web interface, the formatting of the entire file is destroyed.  Newlines, tabs, and spaces are thrown out everywhere in the document--not just on the line I edited--and lists will be changed to asterisks instead of dashes.  See screenshots below.

Related: https://github.com/nextcloud/text/issues/390

**To Reproduce**
1. Create a markdown file in an external editor such as Atom/Gedit or a basic text editor like Notepad
2. Add some spaces, tabs, newlines, lists, etc.
3. Open the markdown file in the Nextcloud web editor and notice the formatting differences
4. Make a minor edit to the file in the Nextcloud web interface
5. View the file again in your external editor and notice the formatted has been destroyed.

**Expected behavior**
I expect the Text app to not completely destroy any formatting created by external apps.  If any special formatting is unsupported, it should be left alone for maximum compatibility with external apps.  If any formatting changes must be made based on the current edits, those changes should apply ONLY to the current edits and not the entire file.

**Screenshots**

The original file as created and edited in Atom:
![Screenshot from 2020-01-19 17-01-26](https://user-images.githubusercontent.com/54287723/72689244-73034900-3add-11ea-9ae2-669c866e503b.png)

How it looks in Gedit under Linux:
![Screenshot from 2020-01-19 17-02-41](https://user-images.githubusercontent.com/54287723/72689252-93cb9e80-3add-11ea-890d-7956b3a0f8f6.png)

How it looks when opened in my Nextcloud web interface:
![Screenshot from 2020-01-19 17-11-06](https://user-images.githubusercontent.com/54287723/72689405-1bfe7380-3adf-11ea-950e-feb34799132b.png)

Make an edit through the web interface and now the file's formatting has been destroyed when viewed in Gedit:
![Screenshot from 2020-01-19 17-12-49](https://user-images.githubusercontent.com/54287723/72689413-26207200-3adf-11ea-9b86-5b40b706b0fd.png)

**Client details:**
 - OS: Arch Linux
 - Browser: Firefox
[...]

Issue #325, opened 2019-10-04
Bracket escaping issue


juliushaertl:
I'm experiencing this issue on 1.1.0, though only with certain lines.

The following lines will all trigger the bug:

```
(Asdf [asdf asdf](asdf asdf) asdf asdf asdf asdf asdf asdf asdf asdf asdf)

* [asdf asdf asdf/asdf](Asdf Asdf)

* asdf asdf asdf [a--f asdf asdf](a--f Asdf Asdf)

* [Asdf asdf asdf asdf asdf asdf](Asdf asdf)
```

I saved this locally on my machine (vimwiki), opened it up in Text, changed a char, closed, then reloaded in vim. All brackets had escape chars.

_Originally posted by @ainola in https://github.com/nextcloud/text/issues/103#issuecomment-538218183_

joshp23:
I came here to report the same issue
`[1]` becomes `\[1\]` and so on.

PR #2704, opened 2022-07-10
Do not remove escaping from brackets

susnux:
* Resolves: #2703
* Target version: master 

### Summary
See #2703 for full description

Removed the erroneous replacement of escaped brackets with plain brackets as this breaks valid markdown.
Added a test case for this.

Added markdownit and tiptap extensions to tag special characters (characters which might have a special meaning for markdown) that are not parsed by markdownit, e.g. because they are extended markdown syntax. The tags are only stripped if the user edits it that part of the file.
This way we can ensure they do not get escaped by `prosemirror-markdown`.


susnux:
Do not merge, still some work to do (WIP).... implementing some more markdown fixes.

susnux:
Added test cases, all tests pass I think this can be reviewed :smiley:

vinicius73:
This PR fix one or two problems @susnux ?

susnux:
> This PR fix one or two problems @susnux ?

Well actually three problems

1. I fixed https://github.com/nextcloud/text/issues/2703 which is quite simple but solves the problem: https://github.com/nextcloud/text/pull/2704/files#diff-6591e0135928d4bf8e3856079a58f0453ddfd4a62a71d923c30cbd2913973cb2
2. That fixes that problem but reintroduces https://github.com/nextcloud/text/issues/325 so thats why I added the `keep syntax` extensions, those extension simply add marks for parts of the original markdown file that would be escaped when serializing the editor and if present when serializing they will prevent escaping of that region. Those marks stay until the user manually edits that part.\
So as an benefit this also fixes parts of https://github.com/nextcloud/text/issues/593 and https://github.com/nextcloud/text/issues/2577

susnux:
@vinicius73 if preferred I can split this PR into two, one for fixing the issue and one for switching the dependency to fix the other issue.

susnux:
I decided to split the PR as the switch of the dependency was not really needed to fix this issue, but another.

So this in now only fixing the issues described in the comment above ( https://github.com/nextcloud/text/pull/2704#issuecomment-1184269724 ), for the other part see: #2720

vinicius73:
Thanks for it @susnux

juliushaertl (approved):

Issue #2387, opened 2022-05-12
Focus order issues with viewer modal


juliushaertl:
When opening a text file in viewer, focus remains in the background and the user needs to tab through all the contents in the background without seeing what they’re doing to get to the modal.

vinicius73:
Scheduled to v5.3.2 https://github.com/nextcloud/nextcloud-vue/milestone/77

juliushaertl:
Done with https://github.com/nextcloud/text/pull/2794

PR #2713, opened 2022-07-12
Prevent focus to go back to file list

vinicius73:
* Resolves: #2387
* Target version: master 

### Summary

Add `role="dialog"` in to `EditorWrapper` component when it is wrapped with `ViewerComponent`


juliushaertl:
Should we rather fix this in the viewer or even the `@nextcloud/vue` Modal component?

juliushaertl:
@vinicius73 I'd say lets skip this one and open a PR in viewer. Or would there be a reason to have it in both?

vinicius73:
@juliushaertl we can wait this PR
https://github.com/nextcloud/nextcloud-vue/pull/2696

Issue #16, opened 2019-05-14
Pasting markdown does not apply styling


jancborchardt:
This is what happens when you paste some markdown text, e.g. from [Lorem Markdownum](https://jaspervdj.be/lorem-markdownum/):
![markdown paste](https://user-images.githubusercontent.com/925062/57704845-04730f00-7663-11e9-8fbe-5e8a3bd503f4.png)

Should probably be highlighted as markdown properly, or is that difficult?

Prosemirror.net has the same issue, so I guess it’s upstream? Or maybe a setting?

juliushaertl:
@jancborchardt How should we distinguish if the user wants to paste markdown or plain text?

jancborchardt:
Hmmmm good question. Does Prosemirror or tiptap not have a default or mechanism for that?

meonkeys:
> @jancborchardt How should we distinguish if the user wants to paste markdown or plain text?

Other text entry UIs appear to commonly use Ctrl-v to paste directly from the clipboard (including markup and images) and Ctrl-Shift-v to attempt to paste the clipboard as plain text. Some right-click menus provide the choice of how to paste what's on the clipboard.

ranomier:
I had this problem too. IMHO this should be higher priority :) Pretty Please :)

juliushaertl:
First implementation draft at https://github.com/nextcloud/text/pull/2716

stellarpower:
Just thinking, ahead of time - there are times when I paste some formatted text, and incorrect markdown formatting is applied (I think it may e.g. detect monospace and apply code formatting here, when I don't necessarily want it to).

I'm gonna guess after this feature, someone may come back and say that they're end-users rather than programmers, and prefer to paste in symbols from their text as literal, without applying them as markdown syntax. Such users may be more comfortable with the WYSIWYG-style formatting options present in the editor than using markdown syntax itself.

Therefore, there may be a case for having a paste menu with options on which action to perform (and a default to apply when hitting Ctrl+V). I reckon this would be useful for other use-cases too, and overall am thinking a bit how LO lets you paste as plain text, formatted text, etc. A menu allowing to "paste and interpret as markdown", "paste as (literal) plain text", "paste formatted text", etc. for example, time permitting.

juliushaertl:
> there are times when I paste some formatted text

The implementation only acts on plain text paste. Formatted text e.g. from a web page or external editor would take precedence and not apply the markdown parsing.

[...]

Issue #938, opened 2020-07-21
Pasted code blocks and inline code should preserve all text characters


dbolton:
**Describe the bug**
Nextcloud Text mangles pasted code blocks and inline code. It should preserve the original text instead of dropping characters or trying to parse it. 

**To Reproduce**
1. Select and copy the following text: 

Code block: 
```
2x = 2*x, 3(4) = 3*(4)
**Bold**
_Underline_
~~Strike through~~
`code`
https://example.com
```
Inline code: ``2x = 2*x, 3(4) = 3*(4) **Bold** _Underline_ ~~Strike through~~ `code` https://example.com ``

2. Open a markdown file in Nextcloud
3. Paste

**Actual behavior**
Asterisks (\*), underscores (\_), tildes (\~), and grave accents (\`) are mistakenly dropped or parsed. URLs become clickable.

**Expected behavior**
The text within the code blocks should not be parsed at all. It should match the text display above. URLs should remain plain text.

**Client details:**
 - OS: Windows
 - Browser: Firefox (developer edition)
 - Version: 79
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 2.0.0

**Operating system:** Ubuntu

**Web server:** Nginx

**Database:** mysql 

**PHP version:** 7.2.31

**Nextcloud version:** 18.0.7

</details>

luka-nextcloud:
I can reproduce the bug.

Issue #55, opened 2019-06-13
Links can not be inserted using Markdown syntax


jancborchardt:
This doesn’t work in [tiptap](https://tiptap.scrumpy.io/) either, but there’s a discussion at https://github.com/scrumpy/tiptap/issues/337
![writing link via markdown syntax does not work yet](https://user-images.githubusercontent.com/925062/59426564-c4758800-8dd8-11e9-9473-f057d20b1a89.png)

RobinDev:
And maybe we could add the ability to use jump link ?

juliushaertl:
> And maybe we could add the ability to use jump link ?

Not sure what you mean by that, can you clarify that a bit?

RobinDev:
jump link or fragment link, eg : 

```
[item 1](#item1)

### Item 1 <a name=item1></a>
```

jancborchardt:
@RobinDev could you open a separate issue for this? Also related: Outline view of headings #107

juliushaertl:
https://github.com/ueberdosis/tiptap/discussions/1865 might be some good input for implementing this.

PR #2716, opened 2022-07-13
Add paste event handler to allow pasting markdown

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/16

Marking as draft as this still needs a bit of testing if this is feasible for common use cases of pasting random text


juliushaertl:
While I have the use case very often, I'd still be curious if anyone could think of scenarios where just always parsing pasted text as markdown could be unexpected to users. Any objections @nextcloud/office ?

mejo-:
Nice, I would love to be able to paste markdown and get it formatted direclty!

This should also fix #55.

susnux:
> This should also fix #55.

No that one is about typing not pasting, but this would fix #938

max-nextcloud:
> I'd still be curious if anyone could think of scenarios where just always parsing pasted text as markdown could be unexpected to users.

When pasting markdown into a codeblock i would expect it to show up as source code not interpreted as markup.

marcelklehr:
Maybe this would be better as a context menu option similar to how Word processors have different options on how to paste

susnux:
Or to follow common practice:

Use this for all normal paste and add a handler for `ctrl + shift + v` to paste without formatting.

juliushaertl:
Changed the approach a bit by using `clipboardTextParser` and skipping the markdown parsing when `Ctrl+Shift+V` is used.

juliushaertl:
🏓 For review ;)

susnux (approved):

PR #2719, opened 2022-07-14
Some backend cleanup

juliushaertl:
- Add type hints
- Move exceptions to dedicated namespace
- Get rid of ILogger in favour of the PSR LoggerInterface


juliushaertl:
Thanks for taking care of the interface adjustment

Issue #886, opened 2020-06-18
Better styling for ToDo list


themedleb:
**Is your feature request related to a problem? Please describe.**
The ToDo list is functional and great, but lacking a little bit of some beauty and aesthetics that the note-taking platforms are using to make the user experience even better.

**Describe the solution you'd like**
The solution is to make the checked/done line easy on the eyes to recognize.

Add the following CSS code:
> .checkbox-item.checked {
>     color: gray;
>     text-decoration: line-through;
> }

And change the border radius to 50%:
> #editor-wrapper div.ProseMirror .checkbox-item::before {
>     border-radius: 50%;
> }

That's all.

**Additional context**
I took screenshots to show how it was and how it will be.

Before:
![ToDo before screenshot](https://raw.githubusercontent.com/themedleb/My-Nextcloud/master/Screenshot_20200618_132829.png)

After:
![ToDo after screenshot](https://raw.githubusercontent.com/themedleb/My-Nextcloud/master/Screenshot_20200618_132731.png)

juliushaertl:
I don't think we should use different checkboxes than anywhere else in Nextcloud. Also round empty boxes actually indicate a radio button so i think that might give the wrong indication, since this is a very common pattern in UI elements. cc @nextcloud/designers

stefan-niedermann:
Absolutely agree with @juliushaertl here: Round checks have a semantically different meaning.

jancborchardt:
Agree on not using round checkboxes for the confusion with radio buttons – but the grey text and striked-out text proposed by @themedleb is actually very nice!

@themedleb do you want to open a pull request for that part? Please use `var(--color-text-maxcontrast)` instead of grey so it works in both light and dark theme. :)

stefan-niedermann:
> but the grey text and striked-out text proposed by @themedleb is actually very nice! 

This is right and would also partially enhance the situation for https://github.com/nextcloud/text/issues/819

themedleb:
> I don't think we should use different checkboxes than anywhere else in Nextcloud
I noticed that after I saw files listing has squared checkboxes, I was like "Oh nooo!!, that's inconsistent!!!" but then I was like "But the ToDo list looks nice!" lol So I ignored it.

> Also round empty boxes actually indicate a radio button so i think that might give the wrong indication
Yes, you are right, I didn't think of when it's empty, all I thought about is the white dot indicates that the radio button is checked and check sign/icon indicates that the checkbox is checked.

[...]

PR #2724, opened 2022-07-16
Task lists: Show label of done tasks crossed out

susnux:
* Resolves: #886
* Target version: master 

### Summary
As recommended in #886 this changes the styling of
checked task list items to `line-through` and changes
the color to `var(--color-text-maxcontrast)`.

Here is an example of how it looks like (also with dark color mode):
Example of crossed out task itels


juliushaertl:
My personal preference it the non-strikes through one as it keeps the items a bit more readable, but let's see what @jancborchardt and @nimishavijay think.

susnux:
That somewhat true, but personally I agree with @jancborchardt on his comment:
>  It gives immediate feedback and more of a sense of "done!"

juliushaertl:
Yep, not a blocker from my side, just wanted to double check :)

nimishavijay (approved):
Really nice work! Very satisfying when it's crossed off :)

Gobytego:
need a easy option to turn off the strike thru on checked items. do not want to mess with CSS to turn this off.

Issue #2211, opened 2022-03-01
✨📱 Text app review: Mobile focus


jancborchardt:
(Notes from first review on 2020-10-06 in issue at https://github.com/nextcloud/text/issues/1075 )

## Header bar https://github.com/nextcloud/text/issues/2440 https://github.com/nextcloud/text/pull/2441

* [x] Make backgound solid var(--color-main-background)
* [x] Add small blue flavor bar on the top like in Nextcloud Office
* [x] Bold the filename / title

## Toolbar

* [x] The 3-dot menu wraps into the next line, but all toolbar entries should always be on the same row (with "Saved" indicator" and people menu) https://github.com/nextcloud/text/issues/1186 https://github.com/nextcloud/text/pull/2294
* [ ] #2606 https://github.com/nextcloud/text/pull/2611
   * In the long run, we could group the items into few menus: Formatting, headings, etc – however we might not have enough space
   * Possibly think about moving the formatting bar to the bottom on mobile
* [x] #2607 https://github.com/nextcloud/text/pull/2627
   * [x] use primary for active icons, primary-light for the background
   * [x] raise opacity for all buttons by default
   * [x] current opacity for the disabled buttons
   * [x] Remove blue circle below active icon

## Collaborators https://github.com/nextcloud/text/issues/2449 https://github.com/nextcloud/text/pull/2450

* [x] "Saved" text is too wide
  * [x] Could be combined with participants, showing a circle indicator on unsaved changes, and in the participants menu we can show extra text for "Last saved x minutes ago" on top of the participant list → **let’s do this one**
    * [x] Adjust behavior on desktop, but keep "Saved" text as additional info
  * Could collapse into a checkmark icon on mobile, could even be actionable to save
  * Could show asterisk next to filename, but filename is not present on mobile
  * Could be in a subline below the title, but filename is not present on mobile
* [x] People avatars could be a bit smaller → from 44px to e.g. 36px
* [x] Avatars should move a bit closer together to save space
* [x] Author color borders could be var(--color-main-background) when author colors are not enabled so there is separation, and when author colors are not enabled the colors don’t make sense yet anyway.
* [x] Always show all participants in dropdown (including yourself so you know your own color)

## Content

* [ ] If possible, cursor of other collaborators could show (there is a tiptap component) #90

juliushaertl:
@jancborchardt If I remembered correctly we also talked about getting rid of the menu bubble for links and move that to the menu bar, right? Then I'll add that as well

[...]

PR #2734, opened 2022-07-20
Improved styling of saving indicator

susnux:
* Target version: master 

### Summary
Improved the styling of the saving indicator on small screen sizes,
previously the contributors menu was overlapped by the saving indicator.
Also on dark color mode the saving indicator was inconveniently sticking out
by its bright (white) background color, it is now set to the main background color.

Note the overlapping of the indicator and menu and the background color of the saving indicator on lower screen width:
**Before**

https://user-images.githubusercontent.com/1855448/179947888-e0177b7b-0b9d-456e-a6c0-74e9ef11898b.mp4

**After**

https://user-images.githubusercontent.com/1855448/179947973-d149c688-447d-43bc-a362-a575a2142507.mp4


susnux:
If your browser does not support mp4, this is the important part (screenshots):
**Before:**
before


**After:**
after

juliushaertl:
@jancborchardt I'm wondering if we should fine-tune the circle indicator a bit. It doesn't seem to obvious what it is about in its current state. The original design spec is from https://github.com/nextcloud/text/issues/2211

>  Could be combined with participants, showing a circle indicator on unsaved changes, and in the participants menu we can show extra text for "Last saved x minutes ago" on top of the participant list → let’s do this one

susnux:
If understand that correctly there should be only one button / icon (the participants menu) with the saved indicator as the outline?

Or a check mark icon within the saved indicator?

mejo-:
> If understand that correctly there should be only one button / icon (the participants menu) with the saved indicator as the outline?
> 
> Or a check mark icon within the saved indicator?

From https://github.com/nextcloud/text/issues/2211 my understanding would be that we want an outline around the avatar as a saved indicator. Additionally, a text "Last saved x minutes ago" on the top of the participants list.

Not sure whether all the stacked avatars should have the outline though. I guess it looks weird if only one has it.

susnux:
To visualize:
> Or a check mark icon within the saved indicator?

Saved indicator using a check mark
Saved indicator when not saved

> If understand that correctly there should be only one button / icon (the participants menu) with the saved indicator as the outline?

Using a outline on the collaborators menu as saving indicator
Using a outline on the collaborators menu as saving indicator


> Additionally, a text "Last saved x minutes ago" on the top of the participants list.

This is already implemented

mejo-:
> > If understand that correctly there should be only one button / icon (the participants menu) with the saved indicator as the outline?
> 
> Using a outline on the collaborators menu as saving indicator Using a outline on the collaborators menu as saving indicator


Thanks for the screenshots @susnux! How would this look like with several participants being active in the document?

susnux:
> How would this look like with several participants being active in the document?

I would say like this:
Screenshot_20220726_140405
Screenshot_20220726_140356

juliushaertl:
Looks good to me, @jancborchardt what do you think?

jancborchardt:
I'd say for consistency it would be nice to use the same icon as we use in Office. :)

- Saved indicator: https://fonts.google.com/icons?selected=Material%20Icons%3Acheck%3A
- For unsaved changes: https://gitlab.collabora.com/productivity/online-branding/-/blob/master/themes/nextcloud/images/lc_savemodified_branding.svg

```
<svg width="24px" height="24px" fill="#000000" version="1.1" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="M0 0h24v24H0z" fill="none"/><path d="M9 16.17L4.83 12l-1.42 1.41L9 19 21 7l-1.41-1.41z"/><circle cx="18" cy="18" r="3" fill="#0082c9" style="paint-order:markers fill stroke"/></svg> 
```

mejo-:
Let's get this merged as it fixes the ugly white filled circle in dark mode. It doesn't replace the "saved" indicator with an icon (as suggested in https://github.com/nextcloud/text/pull/2734#issuecomment-1195291454), though. Let's tackle this in a follow-up PR - it's also tracked in https://github.com/nextcloud/text/issues/2836.

mejo- (approved):

mejo-:
/backport 6bef7e69d559a3be961d353d1e8686402bcbea97 to stable25

Issue #92, opened 2019-06-22
Mention users


juliushaertl:
It should be possible to mention users

juliushaertl:
As discussed with @mejo- this would be relatively straight forward to implement on the frontend side now with the picker from #1961 

Further steps would be:
- Clarify on markdown representation (e.g. could use the profile url as a link `[@User display name](https://cloud.example.com/u/userid)`)
- Clarify individual implementation steps in detail
  1. Plain mentions in the document
  2. Notify users about mentions (with proper cleanup once a mention gets removed again)
  3. Enhanced UX (e.g. suggest to share if a mentioned user doesn't have access to the document yet)

Should also be aligned with the ongoing effort on the Collabora integration at https://github.com/nextcloud/richdocuments/issues/1380

juliushaertl:
- [ ] Find a fitting markdown it plugin that allows us to store both the mention "@userid" and "Display name" of the user (could be https://github.com/quartzy/markdown-it-mentions)
- [ ] Integrate with https://tiptap.dev/api/nodes/mention
  - [ ] The markdown-it-mentions generates more details `<span class="mention" data-type="user" data-id="48249279">Freddie Mercury</span>` than tiptap can handle/parse
  - Therefore we need to extend the tiptap mentions plugin to handle that html https://github.com/ueberdosis/tiptap/blob/main/packages/extension-mention/src/mention.ts#L107-L124
  - In addition we need a [toMarkdown method](https://github.com/nextcloud/text/blob/46ad51019984a08864a1f5a9500710418f17fe4b/src/nodes/Callouts.js#L92-L98) to handle the last conversion step back to markdown 

## Reference on how the formats are converted

- markdown document => markdown-it => html (markdown-it-mentions)
- html => tiptap => rendering editor (tiptap needs to understand the html output from the above step)
- tiptap => prosemirror-markdown => markdown (tiptap needs to generate the same markdown format that is used in the first step)

PR #2748, opened 2022-07-23
Add user mentions feature

luka-nextcloud:

* Resolves: #92 
* Target version: master 

### Summary

It should be possible to mention users.

### ToDo

- [x] Frontend: Make sure to properly display and insert the display name
- [x] Backend: Use proper autocompletion API https://github.com/nextcloud/text/pull/2748#discussion_r940120609
- [x] Backend: Fix php cs @luka-nextcloud 
- [x] Frontend: Fix es lint errors @luka-nextcloud 
- [x] Backend: Emit notification to the users @juliushaertl 

### Follow up tickets
- [ ] Mentions for public guest names
- [ ] Highlight own user id in menu bubbles
- [ ] Pass context to notification (e.g. current line or sentence)
- [ ] On clicking the navigation, jump to the mention directly

Dependencies added (in case needed during rebase):

```
npm install --save @quartzy/markdown-it-mentions @tiptap/extension-mention
```


luka-nextcloud:
@juliushaertl 
I wanted to know if we need to put the @mention in the registered file?
Because I think the main goal of mentioning is just to notify another user not to put the mentioned users in the content right?

image

susnux:
From a user perspective:

If this is about the markdown file, I would expect everything I type to be saved in the markdown file.
E.g. if I write meeting minutes and mention users (for example assign a task) I want that user to be informed so they get remembered but I also do not want a `hole` in the minutes.

```markdown
@John volunteered to take the night shift
```

Or maybe with linking like
```markdown
[@John](/u/johns_username) volunteered to take the night shift
```

juliushaertl:
> I wanted to know if we need to put the @mention in the registered file?

Yes, it should be put into the file. As mentioned we need to store the user id, but would also want the user display name as a readable representation so I tend to use the format that https://github.com/quartzy/markdown-it-mentions#quartzymarkdown-it-mentions parses already: `@[Freddie Mercury](mention://user/48249279)
`

Notifications would be a second step then, which we could however handle in the backend when the file is actually saved.

juliushaertl:
Regarding the UI, we should stick to how Talk and comments do it. See https://nextcloud-vue-components.netlify.app/#/Components/RichContenteditable for a reference on the UI, we cannot reuse that component however.

luka-nextcloud:
@juliushaertl 
UI was updated :)

image

image

lib/Controller/UserApiController.php
```diff
@@ -0,0 +1,33 @@
...
+
+    protected IUserManager $userManager;
+
+    public function __construct($appName, IRequest $request, IUserManager $userManager) {
+        parent::__construct($appName, $request);
+
+        $this->userManager = $userManager;
+    }
+
+    /**
```
juliushaertl, line 19: Please make sure to use tab indent :)

lib/Controller/UserApiController.php
```diff
@@ -0,0 +1,33 @@
...
+     * @param string $filter
+     */
+    public function index(string $filter) {
+		$result = [];
+
+        $users = $this->userManager->search($filter, 5);
```
juliushaertl, line 25: A dedicated endpoint makes sense, since then we can later on sort the active users first. However for search ing users would be better to use https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/public/Collaboration/Collaborators/ISearch.php#L40 as this will respect the privacy settings that may be in place on the instance to only display users that are allowed to be mentioned.

shareTypes would be `[0]` to only search for users.

See also https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/apps/files_sharing/lib/Controller/ShareesAPIController.php#L228 how this is used on the API for file sharing

src/extensions/Mention.js
```diff
@@ -0,0 +1,70 @@
...
+					&& null),
+				priority: 100,
+			},
+		]
+	},
+    
+	renderHTML({ node, HTMLAttributes }) {
+		const avatarUrl = getAvatarUrl(node.attrs.id, 44)
+
+		return [
```
juliushaertl, line 28: Is there any way we could use the UserBubble component directly here?

juliushaertl (commented):
Looks very nice on the UI already, I've left some comments inline.

juliushaertl:
As a next step we may want to discuss how we issue notifications to the mentioned users. I'd say the best might be to do this on the save of the document (which might be a bit delayed), but then we can easily parse the full content and ensure that only notifications which are present are still active.

src/extensions/Mention.js
```diff
@@ -0,0 +1,70 @@
...
+						},
+					],
+				],
+			],
+		]
+	},
+
+	toMarkdown(state, node) {
+		state.write(' ')
+		state.write(`@[${node.attrs.id}](mention://user/${node.attrs.id})`)
```
juliushaertl, line 67: Should use the display name link text.

juliushaertl: Should already be available through https://github.com/nextcloud/text/pull/2748/files#diff-8431ea37f0731d83b77b1cd22d5b5213fd4fa6ff4669fdefebeb1e2220693bb9R88

lib/Controller/UserApiController.php
luka-nextcloud: To inject `UserBubble` we need to make sure that the `UserBubble` will be handled as a VueJs component.

src/extensions/Mention.js
luka-nextcloud: To inject `UserBubble` we need to make sure that the `UserBubble` will be handled as a VueJs component.

lib/Controller/UserApiController.php
luka-nextcloud: Okay, I see what you mean. But there is a concern tough: how can we access the user's id because we need it?

juliushaertl: The format is a bit different but it should return you a list of users, from the following screenshot you should be able to see the format - llabel is the display name and `value['shareWith']` the user id:

Screenshot 2022-08-11 at 18 58 06

src/extensions/Mention.js
juliushaertl: Found a way and pushed it with [fd38108](https://github.com/nextcloud/text/pull/2748/commits/fd3810853a5525b6f2701d82f98829fdd1eec65d)

juliushaertl:
@luka-nextcloud Pushed a few commits split by what they do when working on using the vue component for rendering. Let's move this to a mergeable state tomorrow. Missing points would be:

- [ ] Frontend: Make sure to properly display and insert the display name
- [ ] Backend: Use proper autocompletion API https://github.com/nextcloud/text/pull/2748#discussion_r940120609
- [ ] Backend: Fix php cs
- [ ] Backend: Emit notification to the users

luka-nextcloud:
@juliushaertl With the integration of the `VueNodeViewRenderer`, it seems that the content can not be rendered once we read the file after saving it. Do you have any idea why is behaving like that?

luka-nextcloud:
@juliushaertl Updates here:
https://user-images.githubusercontent.com/89908051/186239301-97db6a34-2d50-491e-9d47-1bbc96961ebd.mov

lib/Controller/UserApiController.php
```diff
@@ -0,0 +1,37 @@
...
+	protected ISearch $collaboratorSearch;
+
+    public function __construct($appName, IRequest $request, ISearch $ISearch) {
+        parent::__construct($appName, $request);
+
+        $this->collaboratorSearch = $ISearch;
+    }
+
+    /**
+     * @param string $filter
```
juliushaertl, line 21:
```suggestion
     * @NoAdminRequired
     * @param string $filter
```

src/extensions/Mention.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+	<NodeViewWrapper as="span" class="mention" contenteditable="false">
+		<NcUserBubble :user="node.attrs.id" :display-name="username" class="mention-user-bubble">
+			@{{ username }}
+		</NcUserBubble>
+	</NodeViewWrapper>
+</template>
+
+<script>
+import NcUserBubble from '@nextcloud/vue/dist/Components/NcUserBubble'
```
juliushaertl, line 10:
```suggestion
import NcUserBubble from '@nextcloud/vue/dist/Components/NcUserBubble.vue'
```

juliushaertl (approved):

max-nextcloud:
whoohoo... so cool this landed!

PR #2751, opened 2022-07-24
Add support to render and edit FrontMatter

susnux:
* Resolves: #145
* Target version: master 

### Summary
This adds support to show and edit a FrontMatter block as a simple code block.

This does not introduce special parsing and rendering like suggested here: https://github.com/nextcloud/text/issues/145#issuecomment-581409252
Because that requires to assume a language, while mostly YAML is used, I already saw a lot projects using TOML or even JSON in the FrontMatter. So using a simple code block allows us to not make any farfetched assumptions about the content while still gives the user a convenient way to edit.

I only added a `hr` element to visually separate FrontMatter from content.

You can even remove the FrontMatter if you like, and copy it, pasted FrontMatter will result in a simple code block as the FrontMatter is only allowed as the first element of a document.

What is not implemented: Adding a new FrontMatter to a document, while implementing this would be straightforward it would add more UI elements and so I would like to have some feedback if I should add such functionality / interface elements.

---

#### Sourcecode
````markdown
---
some: value
other: 2.3
---
# Heading
```
Code block
```
````

#### After / With this PR
Screenshot rendering with FrontMatter

#### Before / Without this PR
Screenshot rendering before

### CODE REVIEW

**Review by max-nextcloud — CHANGES_REQUESTED** [review]
This is great! I think it covers the main use case of being able to handle markdown files with front matter without adding much complexity. I'd personally stick to not exposing the front matter functionality in the ui - but if someone has a clever idea of how to handle this i'm all ears.

Two questions come to my mind:
* What happens with front matter that is affected by the escape html? It would be nice if it would be preserved when opening the file in text. Maybe we could even add a test for this.
* What happens if the front matter block is copied and pasted further down in the document? Does it add a frontmatter node?

The latter question makes me think it might be good to just define a new node type instead of inheriting code block. This would avoid inheriting the input rules, keyboard shortcuts and all that. The new node type could also not be part of the 'block' group and would then not be allowed in the document. Instead we could prefix the document content with a `frontmatter?` expression so front matter can only occur at the start of the document and only once.

**Comment by juliushaertl** [comment]
Very nice :) I agree that we should probably think about how if and how we actually present this to the user. 

- Maybe we can collapse it by default and explicitly add a front matter label, however that is quite technical
  - On the other hand we would only expose this is someone already has frontmatter in their document
- It could be separated visually a bit more to not look exactly like a code block, maybe a lighter background

Maybe @nimishavijay or @jancborchardt also have some input on this.

**Comment by jancborchardt** [comment]
It seems reasonable to show it like a code block indeed. If Frontmatter is used, then it's likely understood and can be shown directly?

We can still do any improvements if issues come up, but don't need to overengineer at this stage. :)

**Comment by susnux** [comment]
Thank you for your feedback!

> * What happens with front matter that is affected by the escape html

Not much as it is reverted when serialized to markdown, it only sanitizes for the HTML output. (see below)

> * What happens if the front matter block is copied and pasted further down in the document?

The pasted one is transformed to a normal code block. (see below)

> I'd personally stick to not exposing the front matter functionality in the ui - but if someone has a clever idea of how to handle this i'm all ears

I have some local changes which would introduce an input rule:
If you enter three dashes **at the beginning** of the document they would transform into a front matter node instead of a `hr`.
Does this solution sound reasonable to include?

> It could be separated visually a bit more to not look exactly like a code block

Maybe this way? This are my local changes including slightly different styling and the input rule.

https://user-images.githubusercontent.com/1855448/180823363-b87443ba-7a66-45d3-95b3-6201a0e1cb37.mp4

**Comment by juliushaertl** [comment]
> > I'd personally stick to not exposing the front matter functionality in the ui - but if someone has a clever idea of how to handle this i'm all ears
> 
> I have some local changes which would introduce an input rule: If you enter three dashes **at the beginning** of the document they would transform into a front matter node instead of a `hr`. Does this solution sound reasonable to include?

Yes, I think that is a nice feature to be able to add it even if we don't expose a button.

> 
> > It could be separated visually a bit more to not look exactly like a code block
> 
> Maybe this way? This are my local changes including slightly different styling and the input rule.

Looks nice 👍

**Review by max-nextcloud — APPROVED** [review]
Thanks for overwriting the tiptap code block rules to make sure they don't apply later in the doc.

I think this is ready to go! :rocket:

File: css/prosemirror.scss [hunk 1]
```diff
@@ -163,6 +163,18 @@ div.ProseMirror {
...
 	}
 
+	pre.frontmatter {
+		margin-bottom: 2em;
+		border-bottom: 2px solid var(--color-primary-element);
+	}
+
+	pre.frontmatter::before {
+		display: block;
+		content: 'Front Matter';
```
juliushaertl, line 173: Should be moved out of the styling as this would probably be bad for accessibility and is not translatable that way

juliushaertl: Ah, and it should be `Front matter` (Sentence case) as per https://docs.nextcloud.com/server/latest/developer_manual/design/foundations.html#wording

juliushaertl (changes requested):
Tested and works like a charm, apart from the small comment about the "Front matter" label this is good to go from my side, so LGTM after that is addressed.

susnux:
Added some test, unfortunately cypress does not support copy paste (somewhat related to https://github.com/cypress-io/cypress/issues/2386 ).
Also fixed serializing of a front matter containing a sequence of three or more dashes, this is valid but the fences must be longer than the sequence within the content.

css/prosemirror.scss
susnux: That will make it a bit more complex, as this requires a NodeView.
For accessibility I do not see a problem as screen readers would announce that code block as 'front matter' which is fine.

But yes translation is a problem (even if I do not know if it should be translated, for example I think German users would be confused by *Titelei* instead of the technical term *front matter*).

I will try to find a good solution :)

max-nextcloud: > That will make it a bit more complex, as this requires a NodeView.

I'm not sure... a [`renderHTML` function](https://tiptap.dev/guide/custom-extensions#render-html) might be enough.

```javascript
  renderHTML({ node, HTMLAttributes }) {
    return ['div', {},
      [ 'h3', {}, node.translatedHeading],
      this.parent?.(),
    ]
  }
```

I have not tested this. No idea if `this.parent` can be used in `renderHTML`. I took it from https://tiptap.dev/guide/custom-extensions#settings . If it does not work one could also repeat the `renderHTML` code from tiptaps code block there.

I'm also not sure if the translated heading needs to be in the node... You can probably also just translate it right in the `renderHTML` function.

juliushaertl (approved):
Awesome contribution @susnux Thanks a lot for that :)

juliushaertl:
Ah failure seems related to the recent server login page change ... waiting for https://github.com/nextcloud/text/pull/2756 then for the merge.

jancborchardt (approved):
Looks great now, nice work @susnux! :)

The blue line on the bottom we could instead do on the left like for the infoboxes – but that is not blocking and can be done in a follow-up.

susnux:
> Looks great now, nice work @susnux! :)

Thank you!

> The blue line on the bottom we could instead do on the left like for the infoboxes – but that is not blocking and can be done in a follow-up.

I did this on purpose, to create some (visual) vertical separation (front matter vs file content).
But I am fine with changing it to match the infobox design :)

susnux:
> The proposed one (with just the blue bar on the left, no icon) is definitely the best. :) And it still looks consistent with the rest.

Ok changed the design accordingly :)

Issue #2749, opened 2022-07-23
Asterisk not changed to unordered list


snowphone:
**Describe the bug**
When I type CJK letters, asterisk are not changed into unordered lists.

**To Reproduce**
1. Open any markdown file.
2. Switch into non-ascii keyboard (e.g. CJK).
3. Press `asterisk` `space` `CJK letters` consecutively.

**Expected behavior**
An asterisk should be changed into a bullet point.

**Client details:**
 - OS: [Windows 11]
 - Browser: [ms edge]
 - Version: [103.0.1264.62 (Official build) (64-bit)]
 - Device: [Thinkpad X1 carbon 6th]

**Text app version:** (see Nextcloud apps page)
3.5.1
**Operating system:**
Host: Ubuntu 20.04, Docker instance: alpine I guess (linuxserver/nextcloud:24.0.3)
**Web server:**

**Database:**
mariadb

**PHP version:** 7

**Nextcloud version:** (see Nextcloud admin page)
24.0.3

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)

[nextcloud.log](https://github.com/nextcloud/text/files/9174575/nextcloud.log)

</details>

susnux:
I can not reproduce this, works on 24.0.3 and on master for me (Chrome and Firefox both on Linux):
![input CJK letter](https://user-images.githubusercontent.com/1855448/180650814-fec2d50b-f026-4c7e-b2ab-2875856ffe97.gif)

The input rule is quite simple `\s*([-+*])\s([^\s[])` meaning you must enter:
1. `*` (or `-` or `+`)
2. exactly **one** space
3. Any non whitespace **and** no `[` or `]`

So I guess either multiple spaces are used or Windows CJK input uses a paste event (text must be typed not pasted to be recognized) or something weird is going on.

snowphone:
First, thanks for the quick reply!

https://user-images.githubusercontent.com/23140464/180652435-dcfea248-864d-4fc6-80c8-8e4edfcddc35.mp4

I recorded the screen and it seems quite weird.
I'll let you know if find out anything.

snowphone:
I think the bug is related the Windows IME.

When I type `asterisk`, `space`, `CJK letter`, _and_ **`change input methods key`**, it does work.

susnux:
> I think the bug is related the Windows IME.

I was also able to reproduce this with ibus / mozc on Linux, can you reproduce this bug on this sandbox?
[...]

PR #2752, opened 2022-07-24
Fix bullet list input role for CJK input

susnux:
* Resolves: #2749
* Target version: master 

### Summary
When using IME for CJK character input it is called after the composing
is finished, at this point there might be more than one character.

#### Example
Before entering the same work (`sakura`) as Japanese Kanji and Hiragana, Kanji is one symbol so it worked, Hiragana are three syllables symbols so it does not:
Input using IME does not work

After this patch using IME works as expected:
Input using IME is correctly formatted


max-nextcloud (approved):
Good catch, minimal fix to the point -- thanks a lot!
I wonder if there would be an easy way to test this. But i don't think we're testing input rules yet.

juliushaertl (approved):

susnux:
> I wonder if there would be an easy way to test this. But i don't think we're testing input rules yet.

I wanted to add a cypress test, but I did not find a way to test IME input

susnux:
I think this should also be backported to stable 24

Issue #2723, opened 2022-07-16
Feature request: Autolink feature should be optional


ZID-TU-Graz:
**Is your feature request related to a problem? Please describe.**
Our users reports, that the autolink feature is annoying, e.g. if the destination link does not exist or if the user accidentally did not put a space after a dot/period (see screenshot).

![2fadc8c7-0855-402f-9783-e367f2943f0f](https://user-images.githubusercontent.com/100593683/179351026-158fbf5e-ae6a-41dc-8359-e2c4cb9d073f.PNG)

**Describe the solution you'd like**
Give the user the choice to enable/disable the autolink feature and/or introduce a setting to enable/disable autolink feature systemwide.

susnux:
I guess this can be easily implemented after https://github.com/nextcloud/text/pull/2718 is merged

juliushaertl:
I'd say we should just finetune the autolinker to not convert things like example.com for the false positive cases instead of another config option. Or we try to only match links if they are surrounded by spaces.

susnux:
> I'd say we should just finetune the autolinker to not convert things like example.com for the false positive cases instead of another config option.

The links created in the report are all fine and they are all valid, as `.md` is a valid TLD (it the TLD of Moldova).

> Or we try to only match links if they are surrounded by spaces.

That would also result in false positives: `... please create a readme.md file` will be automatically linked to `https://changelog.md/`.

I guess it is only safe to auto-link if a protocol is given (`http`, `https`...)

juliushaertl:
Right, then I'd say it makes sense to only do this with for links that actually have a protocol set.

Should be fairly straight forward with https://tiptap.dev/api/marks/link#validate

PR #2754, opened 2022-07-25
Only autolink URLs containing protocol

Raudius:
* Resolves: https://github.com/nextcloud/text/issues/2723
* Target version: master


juliushaertl (approved):

vinicius73 (changes requested):
LGTM;
But tests should be updated.
https://github.com/nextcloud/text/runs/7501209373?check_suite_focus=true

https://github.com/nextcloud/text/blob/0b8750d158aeac8168322a9e2a60c1507d33f961/cypress/e2e/links.spec.js#L43-L55

susnux:
I guess restricting autolink to http(s) is fine, so lgtm after fixing the tests.

cypress/e2e/links.spec.js
```diff
@@ -45,8 +45,8 @@ describe('test link marks', function() {
 				.type('google.com')
 
 			cy.getContent()
-				.find('a[href*="google.com"]')
-				.click({ force: true })
+				.get('a[href*="google.com"]')
```
susnux, line 48: You should stick with `find` as `get` will not only search the editor content but the whole document.
Probably something like `getContent().find(...).should('not.exist')` would work

cypress/e2e/links.spec.js
```diff
@@ -45,8 +45,8 @@ describe('test link marks', function() {
...
+				.expect([])
 
 			cy.get('@winOpen')
```
susnux, line 51: This part is not needed anymore as there is no link, thus no windows is opened.

susnux (changes requested):

cypress/e2e/links.spec.js
Raudius: Thanks for the tip, I am still a cypress noobie :)

cypress/e2e/links.spec.js
Raudius: Removed!

vinicius73 (approved):

susnux (approved):

Issue #2625, opened 2022-06-25
Doc saved indicator and user color controls are missing for .txt files


max-nextcloud:
**Describe the bug**
In Nextcloud 24 when opening a .txt file there are no controls anymore.
This used to work in Nextcloud 22

**Expected behavior**
![nc22](https://user-images.githubusercontent.com/97337118/175763863-76fde23f-3acf-41d8-a893-b3e53a629074.png)

**Current behavior**
![nc24](https://user-images.githubusercontent.com/97337118/175763871-74f8c6c1-11be-4b27-bb01-934d29271d3d.png)

max-nextcloud:
Took the second screenshot after reopening the file. So user colors are gone... But that's not the way it's supposed to work. ;)

PR #2757, opened 2022-07-27
Bring back menubar without formatting on plain text mode

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/2625

- Ensure that the session list on the right of the menubar is always visible
  - For plain text files
  - For read only shares of markdown and plain text files


juliushaertl:
Rebased and ready for review :)

max-nextcloud (dismissed):
All in all looks good. I think we do not need to render the menu bubble unless we are editing a richtext document.
With the change to `renderMenus` it would now be rendered for text. Since it even has it's own dynamic import might be worthwhile having two computed conditions `renderMenubar` and `renderMenububble` or so.

juliushaertl:
Rebased and ready for review again :)

src/components/Editor.vue
```diff
@@ -262,12 +269,16 @@ export default {
 		displayed() {
 			return this.currentSession && this.active
 		},
-		renderMenus() {
+		renderRichEditorMenus() {
 			return this.contentLoaded
 				&& this.isRichEditor
 				&& !this.syncError
 				&& !this.readOnly
 		},
```
max-nextcloud, line 277: Are we using this anywhere? Or are you just keeping it in case we need it later?
```suggestion
```

max-nextcloud (approved):
Looks good. Just one function that i think is not needed anymore. But that should not block a merge.

max-nextcloud:
cypress was failing earlier - curious to see if this still happens.

max-nextcloud:
Not sure if a backport to 23 is needed. We need the one to stable 24 for sure as a customer complained about this (in ios but i think it's this issue.)

src/components/Editor.vue
max-nextcloud: I guess we still need this in earlier versions that still have the menu bubble. So good to keep it for now for the backports and clean it up separately.

max-nextcloud:
Looks like for the keyboard shortcut tests retries run into file version conflicts

max-nextcloud:
/backport 665764231b461f392b03da93fa949133975b5be7 to stable24

Issue #2735, opened 2022-07-20
[BITV/A11Y] Currently keyboard accessible using keyboard shortcuts common in text editors, however there is a WAI-ARIA best practice


AndyScherzinger:
Moved over from https://github.com/nextcloud/server/issues/21443

Details see [wcagFixes.xlsx](https://github.com/nextcloud/server/files/4788866/wcagFixes.xlsx) as well as discussions in https://github.com/nextcloud/server/issues/21443#issuecomment-645389427

c&op from the content:

> 4 | Keyboard | Currently keyboard accessible using keyboard shortcuts common in text editors, however there is a WAI-ARIA best practice | None | 2.1.1 - Text Editor |   | Text Editor is not configurable due to it being another third party application
> -- | -- | -- | -- | -- | -- | --

Is the missing `role="menubar"` on formatting bar as well as the missing `role="menuitem"` on each of the items what you are referring to? That we should of course fix.

Ref @juliushaertl for the Text editor:
- https://www.w3.org/TR/wai-aria/#menubar
- https://www.w3.org/TR/wai-aria/#menuitem

## ToDo

- [x] role menuitem/menubar https://github.com/nextcloud/text/pull/2715
- [ ] Expose shortcuts through [aria-keyshortcuts](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts) #2762

AndyScherzinger:
@juliushaertl assigned you just as a "ping"

juliushaertl:
role menubar and menuitem were already addressed in https://github.com/nextcloud/text/pull/2715

juliushaertl:
> Currently keyboard accessible using keyboard shortcuts common in text editors, however there is a WAI-ARIA best practice

As far as I get that this is about adding keyboard shortcuts through the aria-key shortcuts attribute: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts

About the other topics I'm not sure I understand the report:

> Text Editor is not configurable due to it being another third party application

What should be configurable here? Is there a way we can clarify that?

jancborchardt:
Tbh I also don't understand their remark @juliushaertl quoted.

Normally standard keyboard shortcuts should be ok as long as they are not single letters which might interfere with typing. And we don't use any of the like in Text, instead having standard shortcuts like Ctrl+B for bold, so that seems fine?

juliushaertl:
[...]

Issue #2445, opened 2022-05-25
macOS specific keyboard modifiers


juliushaertl:
On macOS keyboard shortcuts work with the `Command` key instead of [Ctrl](url). While this works with the current implementation [as tiptap covers both](https://tiptap.dev/api/keyboard-shortcuts) the tooltip and the keyboard shortcut help modal currently show `Ctrl` and should detect and rather show `Command` or `Cmd` on macOS.

PR #2762, opened 2022-07-29
[a11y] add aria-keyshortcuts

vinicius73:
* Resolves: #2735
* Resolves: https://github.com/nextcloud/text/issues/2445
* Target version: master 

### Summary

- add missing [`aria-keyshortcuts`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts)
- fix some shortcuts
  - strikethrough
  - ordered-list
  - blockquote
- add missing shortcuts
  - code-block
- improve mac compatibility
  - use command instead of control in Apple devices
- add shortcuts tests


juliushaertl:
Also fixes https://github.com/nextcloud/text/issues/2445

juliushaertl:
Also fixes https://github.com/nextcloud/text/issues/2445

juliushaertl (approved):

src/components/Menu/keys.js
```diff
@@ -0,0 +1,24 @@
+const isMac = (navigator.userAgent.includes('Mac'))
```
susnux, line 1: I would recommend to not use the `navigator.userAgent` which can be easily spoofed, but the more reliable `navigator.platform`.
So something like:
```js
/(Mac|iPhone|iPod|iPad)/i.test(navigator.platform)
```

vinicius73: Unfortunately `platform` is deprecated 😥
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

src/components/Menu/keys.js
```diff
@@ -0,0 +1,24 @@
...
+	Shift: 'Shift',
+
+	// unused
+	// AltGraph: 'AltGraph',
+	// Meta: 'Meta', // Command key on Apple computers
+}
+
+const TRANSLATIONS = {
+	[MODIFIERS.Mod]: t('text', isMac ? 'Command' : 'Control'),
+	[MODIFIERS.Control]: t('text', 'Ctrl'),
```
mejo-, line 16: Is it on purpose that `MODIFIERS.Control` is translated to 'Ctrl' while `MODIFIERS.Mod` is translated to `Control`? Also, if I'm not mistaken, then `MODIFIERS.Control` is not used at all in the code, only `MODIFIERS.Mod`, right?

src/components/Menu/keys.js
susnux: Oh I did not know, guess I will have to rewrite some code.
I guess the future way to go is using the `userAgentData`, as at least [chrome plans migrate](https://developer.chrome.com/docs/privacy-sandbox/user-agent/) from the user agent string to the user agent client hints.

Of cause not every browser has implemented the client hints by now...
Meaning until Firefox and Safari implemented them a fallback like this must be used:

```js
const platform = navigator?.userAgentData?.platform || navigator?.platform || 'unknown'
```

src/components/Menu/keys.js
susnux: Additionally when using `ctrl` for the key, which is fine to make it clear the key is meant, why not also using the abbrev for the command key too? (`cmd`)

src/components/Menu/keys.js
```diff
@@ -0,0 +1,24 @@
...
+	Control: 'Control',
+	Shift: 'Shift',
+
+	// unused
+	// AltGraph: 'AltGraph',
+	// Meta: 'Meta', // Command key on Apple computers
+}
+
+const TRANSLATIONS = {
+	[MODIFIERS.Mod]: t('text', isMac ? 'Command' : 'Control'),
```
juliushaertl, line 15:
```suggestion
	[MODIFIERS.Mod]: isMac ? t('text', 'Command') : t('text', 'Control'),
```

Ah, I missed that, we need two separate `t()` calls so that the translation script is picking this up properly picking that up.

src/components/Menu/keys.js
vinicius73: `Control` key exist in Mac's keyboard.
So we can directly address it.

But, tiptap change `Ctrl` to `Cmd` natively.

https://tiptap.dev/api/keyboard-shortcuts

> You can use Mod as a shorthand for Cmd on Mac and Control on other platforms.

src/components/Menu/keys.js
vinicius73: This is a good idea, unfortunately I don't have a Mac to proper test it.
Can you open an issue for tracking this in the future?

susnux: > This is a good idea, unfortunately I don't have a Mac to proper test it. Can you open an issue for tracking this in the future?

Done, see #2770 :)

Issue #2733, opened 2022-07-20
Bad focus order for collaborator menu


juliushaertl:
The collaborator menu is not accessible directly after opening as it is put at the very end of the DOM tree, so there user basically needs to tab through the rest of the page in order to reach it

![Screenshot 2022-07-20 at 08 31 50](https://user-images.githubusercontent.com/3404133/179912936-214cff5c-91f4-4759-aba8-e7ace13b94f9.png)
![Screenshot 2022-07-20 at 08 34 21](https://user-images.githubusercontent.com/3404133/179913120-6d9d39af-c0a0-4c59-ad76-34aa123070df.png)

vinicius73:
Waiting for next release of `@nextcloud/vue`

https://github.com/nextcloud/nextcloud-vue/pull/2941

juliushaertl:
Done with https://github.com/nextcloud/text/pull/2794

PR #2771, opened 2022-08-01
[a11y] Enable tab navigation over SessionList

vinicius73:
* Resolves: #2733
* Target version: master 

### Summary

Enable tab usage on session list popover.

- Add visual feedback about focus
- restore focus on editor after close popover

https://user-images.githubusercontent.com/1561347/182258054-60e7216b-06c4-42b1-ab5e-900fe471e6b9.mp4

https://user-images.githubusercontent.com/1561347/182258071-2f35f429-037b-4500-9a17-4755e7c2ecb6.mp4

### DIFF
--- a/src/components/Editor/SessionList.vue
+++ b/src/components/Editor/SessionList.vue
@@ -143,7 +143,7 @@ export default {
 			this.$refs.form.focus()
 		},
 		revertFocus() {
-			this.$editor.chain().focus().run()
+			this.$editor.commands.focus()
 		},
 	},
 }
--- a/src/components/Editor/SessionList.vue
+++ b/src/components/Editor/SessionList.vue
@@ -21,7 +21,10 @@
   -->
 
 <template>
-	<Popover class="session-list" placement="bottom">
+	<Popover class="session-list"
+		placement="bottom"
+		@after-show="applyFocus"
+		@after-hide="revertFocus">
 		<button slot="trigger"
 			v-tooltip.bottom="label"
 			:title="label"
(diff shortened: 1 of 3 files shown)


src/components/Editor/SessionList.vue
```diff
@@ -132,6 +138,14 @@ export default {
 			return this.participantsWithoutCurrent.slice(0, 3)
 		},
 	},
+	methods: {
+		applyFocus() {
+			this.$refs.form.focus()
+		},
+		revertFocus() {
+			this.$editor.chain().focus().run()
```
max-nextcloud, line 146: since there is only a single command in your chain i think you could do
```suggestion
			this.$editor.commands.focus()
```

max-nextcloud (approved):
Codewise looks good to me... just one minor comment. I still approve so you can fix if you want and then go ahead without another review.

I did not try this myself - your screen capture seems to make sense.

src/components/Editor/SessionList.vue
max-nextcloud: https://tiptap.dev/api/commands#execute-a-command

src/components/Editor/SessionList.vue
```diff
@@ -48,14 +51,16 @@
...
 				</ul>
-				<input id="toggle-color-annotations"
-					v-model="showAuthorAnnotations"
-					type="checkbox"
-					class="checkbox">
-				<label for="toggle-color-annotations">{{ t('text', 'Show author colors') }}</label>
-				<p class="hint">
-					{{ t('text', 'Author colors are only shown until everyone has closed the document.') }}
-				</p>
+				<form ref="form" tabindex="0" @submit.prevent>
```
juliushaertl, line 54: I think we should rather put the initial focus on https://github.com/nextcloud/text/pull/2771/files#diff-9e0a2a342f91d6277778b1bb6b17da4f37cdab38ef53c98a7c97a84deaf488a5R40 so that the content gets read and maybe give this a descriptive label then

src/components/Editor/SessionList.vue
```diff
@@ -21,7 +21,10 @@
   -->
 
 <template>
-	<Popover class="session-list" placement="bottom">
+	<Popover class="session-list"
+		placement="bottom"
+		@after-show="applyFocus"
+		@after-hide="revertFocus">
```
juliushaertl, line 27: Maybe we can additionally catch tab/escape and prevent the default event handling and instead just revert the focus.

src/components/Editor/SessionList.vue
juliushaertl, line 146: Should this rather go back to focus on the button after closing?

juliushaertl:
@vinicius73 Any thoughts about using something like in https://github.com/nextcloud/nextcloud-vue/pull/2654 ?

Could also be fixed upstream globally then https://github.com/nextcloud/nextcloud-vue/issues/2860

src/components/Editor/SessionList.vue
vinicius73: Thanks!

vinicius73: I don't know. But I am open for ideas.

vinicius73:
> @vinicius73 Any thoughts about using something like in [nextcloud/nextcloud-vue#2654](https://github.com/nextcloud/nextcloud-vue/pull/2654) ?

Looks interesting, I will play around this a bit.

vinicius73:
Works like a charm @juliushaertl 
I will do some upstream of this, maybe we can drop this PR.

vinicius73:
Waiting for next release of `@nextcloud/vue`

https://github.com/nextcloud/nextcloud-vue/pull/2941

PR #2775, opened 2022-08-02
Allow creating hard line breaks on mobile devices

### LINKED ISSUE #2761: Do not force line break spacing on the user
Two of the greatest advantages of markdown editors are 1) simplicity and 2) control. Nextcloud Text misses Point 2 when it FORCES line break spacing on the user. There are many times I don't want spacing in line breaks when taking notes -- such as if I am just jotting loose ideas, or I want to cluster concepts together. But with forced line break spacing the file becomes long and spaced out, with no discernible idea clusters. Furthermore, when downloading to markdown, those spaces turn into full line breaks of their own, which makes the document even more spaced out. Most of the major rich-text notetaking apps are not forcing line break spacing, such as Evernote, OneNote, Google Keep, etc, and I think that's for good reason -- they recognize the user doesn't always want this. 

_Option A:_ Get rid of line break spacing entirely. If the user wants space after a line break, they can do it manually with their keyboard (just hit 'enter' one more time). A tried and true method that dates back to the typewriter :wink:.

_Option B:_ If we must keep some iteration of this feature, then give the user the OPTION to turn this on or off with a button in the feature tray. Line break spacing would be attributed to a block of text (i.e. the text in between line breaks). If it is turned on for the block, then there is spacing underneath. If it is turned off then there is no spacing. 

**Additional context**
1. If Nextcloud will approve of the concept, I can contribute to the development of this. 
2. Below is a video from Google Docs, exemplifying how it could be turned on/off for the text. 
[Google Docs Line Break Spacing Example.webm](https://user-images.githubusercontent.com/32799575/181697331-d8fbac6e-5146-486b-850d-8d4cf4c8e6e6.webm)
3. Below is a rendering of what the icon could look like in the feature tray.
Line Spacing Icon

### LINKED ISSUE #2809: 2 line breaks = new paragraph
In a previous feature (#2761) I requested removing the auto-double space feature when selecting the `Enter` key. This was particularly problematic for mobile devices, since a single line break with `Shift + Enter` was not possible. Ultimately, it was decided not to change this behavior for web, but for mobile it was modified such that two spaces would equate to a new paragraph (#2775). 

I'm wondering if we can at least make a modification in web where two single spaces (`Shift + Enter` twice) will convert to a new paragraph. Or put differently, `Shift + Enter` twice is treated the same as `Enter` once.  

There are three main reasons for this.
1. It flows with the [traditional markdown logic](https://www.jetbrains.com/help/hub/Markdown-Syntax.html#quick-notes-markdown-paragraphs), whereby a line break between two lines of texts signifies a new paragraph. 
2. This logic will be consistent with the new behavior in mobile (#2775). 
3. It creates less mental burden on the user. There are times when I would like to create two spaces in between paragraphs, to signify a new section/chapter. Currently, to do this properly I will have to hit `Shift + Enter` once, and then `Enter` once -- two separate actions. Otherwise, it won't be treated as a new paragraph and formatting will carry over. This is an extra mental step that seems unnecessary. 

Note: for users that like the auto-double spacing with the `Enter` key, this won't change anything.

susnux:
* Resolves: #2761 
* Resolves #2809 (with latest changes, see comment below)
* Target version: master 

### Summary
This is a proposal to resolve #2761, let me know what you think about this solution.

This allows to create hard line breaks on mobile, currently a hard
line break could only created using `shift + enter`, but most mobile
keyboards do not support the `shift` modifier.
So on mobile devices pressing the enter key will trigger a hard line break
and pressing it again will convert it to a new paragraph.

On desktop devices the behavior is unchanged, but documentation on how to
add hard line breaks is added.

**Edit**\
This now also unifies the behavior on mobile and on desktop, pressing enter once will result in a hard line break, pressing enter again will convert the hard line break to a paragraph break (see comment below).


juliushaertl:
I'm generally fine with the approach, thanks for the proposal. To me it seems still way more common to use paragraph breaks, but that might also include a bit of my personal usage behaviour of text. Would love to get some more input from @nextcloud/text and @nextcloud/designers

max-nextcloud:
I think this is a good fix for the issue in question. I like it because it makes the hard breaks easy to discover. On the other hand it might make the paragraph breaks harder to discover / maybe people think they are just empty lines.

From my point of view the situation is kind of difficult to handle because a lot of people still think of an text editor like a drawing program... they type text and apply styling to it. Enter moves the text below down, spaces move it to the right, headings increase the font size etc. The more useful mental model is a semantic one (headings, paragraphs, hard breaks).

What i like about this fix is that it works for both mental models.

I don't like so much that it encourages hard breaks.
I like to write markdown one sentence per line, 
or even one phrase per line.
Since this relies on different rendering in editing and viewing mode
it cannot be done in a meaningful way in a wysiwyg editor.

...and i just learned that github will render newlines as hard breaks in comments. :disappointed:

Anyway. I'm all for applying this fix. :+1:

mejo-:
@jancborchardt and @nimishavijay what do you think about this topic?

jancborchardt:
I replied in the original issue at https://github.com/nextcloud/text/issues/2761#issuecomment-1293862285

One thing I’d say specifically is that the experience should be the same on desktop and mobile when pressing enter, otherwise it’s highly confusing. :)

susnux:
@jancborchardt I pushed a commit to unify the user experience on mobile and desktop.
So you will have to press enter once for a hard line break "newline" and twice for a new paragraph.

I think this way is cleaner than the other way round, as then the cursor would jump up again.

We might want to think of a keyboard short cut for "new paragraph" on desktop (using shift+enter is not an option as this would confuse anyone I think.)

src/components/HelpModal.vue
```diff
@@ -23,6 +23,25 @@
...
+						<kbd>{{ t('text', 'Enter') }}</kbd>
+					</td>
+					<td v-if="!isMobile" />
+				</tr>
+				<tr>
+					<td>{{ t('text', 'Hard line break') }}</td>
+					<td>
+						<kbd>{{ t('text', 'Enter') }}</kbd>
+					</td>
+					<td>
```
max-nextcloud, line 39: I think this also needs a `v-if="!isMobile"` : 

```suggestion
					<td v-if="!isMobile">
```

src/helpers/platform.js
```diff
@@ -0,0 +1,24 @@
...
+		/iPad/i,
+		/iPod/i,
+		/playbook/i,
+		/silk/i,
+		/BlackBerry/i,
+		/Windows Phone/i,
+	]
+
+	return mobileDevices.some(regex => navigator.userAgent.match(regex))
+}
```
max-nextcloud, line 24: I think adding an alternative definition of `isMobile` is confusing.

My understanding is that `isMobile` is being used in a number of different nextcloud apps and usually it means has a small screen. Maybe we should rename the original. But I'd be tempted to leave the definition as is for now.

Is this adjustment of `isMobile` still needed now that mobile and desktop behave almost the same?

max-nextcloud (approved):
Looks good to me. I added two smaller comments.

src/helpers/platform.js
susnux: There are no keyboard shortcuts on real mobile devices (IPhone, Android...) as there are no meta keys, but "small screen size" could also be a desktop device with small sized browser window.

max-nextcloud: The `Editor` component provides is mobile and then it's also getting [used](https://github.com/nextcloud/text/blob/879eaec610f310b3bc4e695f662eb88fe56563eb/src/components/Editor/Status.vue#L25) by the `Editor/Status` component.

In that context it hides the status and i think that is so it fits on a smaller screen.

All other uses that i found hide keyboard shortcuts - which as you explained makes sense on mobile devices.

How about calling the function here something like `hasKeyboardShortcuts` or `displayShortcuts` or `hasKeyboard`?
I think the `isMobile` name is choosen badly and `hasSmallScreen` would be better - but I would not touch taht for now.

max-nextcloud: Thanks for following up on this. I think this is resolved now.

max-nextcloud (changes requested):
I was about to merge this but noticed some issues with headings when i tried it again.

Headings can get into a strange state with this:
* create a new file.
* write a first line
* press enter twice
* write another line
* turn that line into a heading by clicking the corresponding button

If you hove the heading now the anchor `#` symbol will show below. If you press enter when at the end of the heading a new heading is created below. This does not happen if you turn the first line into a heading.

max-nextcloud:
Hey @susnux 
Thanks a lot for tackling this and following up on it. I'm sorry for the bad news above and curious to hear if you can reproduce it. I can also try and write a cypress test for it if that helps.

Here's what the funny state of the heading looks like for me:
Bildschirmfoto vom 2022-12-07 08-33-53
Bildschirmfoto vom 2022-12-07 08-34-02

I noticed this does not happen if i start with a file with two paragraphs that was created previously. So my suspicion is that after pressing enter twice the document somehow differs from what it did in the past after pressing enter once.

susnux:
I can reproduce this, there is a hardbreak within the heading (which is by the way not allowed following the commonmark rules).

But I noticed also some other usability issues, which make me think @juliushaertl was right about making paragraph breaks the default. Because after some testing I noticed you mostly need paragraph breaks. It is quite annoying to always have to press enter twice while typing markdown
E.g. headlines:
* type test
* press enter once
* type `# head`
* **-->** does not work

By the way I discovered an other, unrelated, bug with newlines inside headings, I will file an issue.

max-nextcloud:
> I can reproduce this, there is a hardbreak within the heading

Yes. I think the hardbreak is already there before creating the heading and then the entire paragraph is converted so the heading also contains a hardbreak.

max-nextcloud:
#3568 superseeded this pr.

PR #2794, opened 2022-08-11
Bump @nextcloud/vue to 6.0.0-beta.1

juliushaertl:
Still throwing some errors in the viewer as https://github.com/nextcloud/viewer/pull/1319 / https://github.com/nextcloud/viewer/pull/1300 are missing and the Vue instance has a different version there.

Ref https://github.com/vuejs/vue/issues/12627

At some point we should probably rethink our way of embedding text in viewer to avoid version dependencies like that.


PVince81:
obsoleted by https://github.com/nextcloud/server/pull/33508

vinicius73:
viewer has been updated to vue 2.7

https://github.com/nextcloud/viewer/pull/1300

There no more errors like https://github.com/vuejs/vue/issues/12627

vinicius73:
I just found a style error, it seems to work correctly.

The main problem is the new popover component (floating-vue).
There are some behavior making the tests do not work anymore.

I will send a PR to nextcloud-vue to address some of that behaviors.

juliushaertl:
@vinicius73 I pushed an adjustment to the cypress selectors and it seems to be happy about that. Also ran another smoke test an couldn't find any obvious issues. Maybe you can later list your findings, so we have that tracked.

Voting for getting this in fast, we can still follow up on smaller issues.

Raudius (approved):
Tested against server master branch, working well! :)

juliushaertl:
- [ ] Check issues mentioned by Vincent in https://github.com/nextcloud/server/pull/33516 regarding the focus-trap in actions

PR #2795, opened 2022-08-11
Disable background fade on empty rih workspace

juliushaertl:
Ensures that the color of the placeholder is a proper max contrast text color.


juliushaertl:
🏓 for review :)

vinicius73 (approved):

PR #2804, opened 2022-08-15
Avoid duplicate queries that were caused by running execute and the mapper find method

juliushaertl:
Avoid double queries on methods that would invoke execute anyways in the mapper class methods.


PVince81 (approved):
:+1: nice catch!

vinicius73 (approved):

Issue #82, opened 2019-06-21
Unified header


juliusknorr:
- [ ] Combine viewer header and the text formatting menubar (similar to how it is done in Nextcloud Office now)
- [ ] Combine the Share icon and the avatars into one overlapped thing. When we have avatars, we don’t need a share icon at all, and they don’t need individual clickable area – all can open the sidebar. (Mockup in the works)
- [ ] For Shared links, even put the controls in the blue header bar – see Dropbox Paper. (Mockup in the works)

Here are the mockups:

## Unified bar Viewer + Text when logged in
We should also combine the Share icon and the avatars into one overlapped thing. When we have avatars, we don’t need a share icon at all, and they don’t need individual clickable area – all can open the sidebar.
![Text app desktop logged in header](https://user-images.githubusercontent.com/925062/60168249-c55cdf80-9804-11e9-8a09-087b3cf8d5a3.png)

## For Shared links, even put the controls in the blue header bar
Desktop
![Text app share link header bar desktop](https://user-images.githubusercontent.com/925062/60167675-d9eca800-9803-11e9-86db-e41627bc32fe.png)
Mobile
![Text app header mobile new](https://user-images.githubusercontent.com/925062/60167676-d9eca800-9803-11e9-993f-2e9a137d34f2.png)

Smaller polishing before the rename (already done):

- [x] Flash of half-transparent black at the beginning of viewer open should be white instead. https://github.com/nextcloud/text/pull/1907
- [x] Title and icons on top right shouldn’t fade out

juliushaertl:
I've moved the mockups to the first post and clarified the scope a bit since the original ticket was containing multiple things. Now this is purely about the unified header as the rest is already addressed.

szaimen:
I wonder how the unified header would work on mobile?

jancborchardt:
For the share link view, it would look even a bit nicer if we do it similar like in Nextcloud Office where there is only a thin primary line on the top, and the header "bar" is white. Then it’s not that huge of a contrast with a blue menu bar which is a bit uncommon. :)

juliushaertl:
I'm wondering if we should actually rebuild the already existing UI or the option I would prefer is to find a way to inject the menubar and the avatars into the existing header.

Not yet sure if that is technically feasible but maybe we could somehow trick the component to render into divs outside of the main text container.

@vinicius73 Maybe you have a creative idea or can tell if that is a stupid idea :D 

One potential library for that I've found would be https://portal-vue.linusb.org

[...]

PR #2808, opened 2022-08-18
feat: unify header

luka-nextcloud:

* Resolves: #82
* Target version: master 

### Summary

https://user-images.githubusercontent.com/89908051/211894234-33bed67f-acbd-4774-aeb6-8de127e99fc9.mov

### DIFF
--- a/src/components/Editor/RightSideActions.vue
+++ b/src/components/Editor/RightSideActions.vue
@@ -1,0 +1,126 @@
+<template>
+	<div class="right-side-actions">
+		<div class="user-avatar__icon user-avatar__icon--with-avatar"
+			:style="avatarUrl ? { backgroundImage: `url(${avatarUrl})` } : null">
+		</div>
+		<a :download="basename"
+			:href="fullPathToFile">
+			<Button class="download__button"
+				type="primary">
+				Download
+			</Button>
+		</a>
+		<Actions>
+			<ActionButton icon="icon-edit"
+				:close-after-click="true"
+				@click="showSidebar">
+				Open sidebar
+			</ActionButton>
+			<ActionButton icon="icon-delete"
+				:close-after-click="true"
+				@click="onDelete">
+				Delete
+			</ActionButton>
+		</Actions>
+	</div>
+</template>
+
+<script>
+
+import axios from '@nextcloud/axios'
+import { getCurrentUser } from '@nextcloud/auth'
+import Actions from '@nextcloud/vue/dist/Components/Actions'
+import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import Button from '@nextcloud/vue/dist/Components/Button'
+import { getRootPath } from '../../helpers/files.js'
+import { getAvatarUrl } from '../../helpers'
+
+export default {
+	name: 'RightSideActions',
+
+	components: {
+		Actions,
+		ActionButton,
+		Button,
+	},
+
+	props: {
+		basename: {
+			type: String,
+			default: null,
+		},
+		relativePath: {
+			type: String,
+			default: '',
+		},
+	},
+
+	computed: {
+		avatarUrl() {
+			let currentUser = getCurrentUser() 
+			return currentUser ? getAvatarUrl(currentUser?.uid, 44) : null
+		},
+		fullPathToFile() {
+			return getRootPath() + this.relativePath
+		}
+	},
+
+	methods: {
+		async onDelete() {
+			try {
+				const url = this.fullPathToFile
+				await axios.delete(url)
+				this.closeEditor()
+			} catch (error) {
+				showError(error)
+			}
+		},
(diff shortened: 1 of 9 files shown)


src/components/Editor/RightSideActions.vue
```diff
@@ -0,0 +1,126 @@
...
+	.download__button {
+		border: 1px solid white !important;
+	}
+</style>
+
+<style lang="scss">
+	// It makes the app right sidebar appear under the Menubar instead
+	// of Hiding a part of it
+	// The value used is the max-length of the Menubar
+	#app-sidebar-vue {
```
vinicius73, line 123: Is it possible to avoid this global selector?

src/components/Menu/MenuBar.vue
```diff
@@ -33,6 +33,12 @@
 			'text-menubar--is-workspace': $isRichWorkspace
 		}">
 		<HelpModal v-if="displayHelp" @close="hideHelp" />
+		<div class="text-menubar--logo">
+			<a href="/">
+				<img src="http://nextcloud.local/core/img/logo/logo.svg">
```
vinicius73, line 38: It's an absolute and local URL

src/components/Menu/MenuBar.vue
```diff
@@ -142,6 +152,9 @@ export default {
 				children: this.hiddenEntries,
 			}
 		},
+		fileName() {
+			return this.relativePath.substring(this.relativePath.lastIndexOf('/') + 1)
+		},
```
vinicius73, line 157: You do not need the prop `repativePath`

https://github.com/nextcloud/text/blob/4c7baddf9631a80f27a34ca189fcc1c36306902e/src/components/Editor.provider.js#L68-L79

src/components/Menu/MenuBar.vue
```diff
@@ -215,22 +228,37 @@ export default {
 }
 </script>
 
-<style scoped lang="scss">
+<style lang="scss">
+	$color-white: white;
```
vinicius73, line 232: Can we use CSS variables instead of sass variables?
It will prevent theme issues.

src/helpers/files.js
```diff
@@ -43,6 +44,14 @@ const optimalPath = function(from, to) {
...
 }
 
+const getRootPath = function() {
+	if (getCurrentUser()) {
+		return generateRemoteUrl(`dav/files/${getCurrentUser().uid}`)
+	} else {
+		return generateRemoteUrl('webdav').replace('/remote.php', '/public.php')
+	}
+}
+
```
vinicius73, line 54: You can avoid call `getCurrentUser` twice.

```suggestion
const getRootPath = () => {
    const user = getCurrentUser()
	if (user) {
		return generateRemoteUrl(`dav/files/${user.uid}`)
	}
	
	return generateRemoteUrl('webdav').replace('/remote.php', '/public.php')
}

```

vinicius73: `getRootUrl` do not solve your case?

https://github.com/nextcloud/nextcloud-router/blob/8913e9fb9450d4a105a93b306ad297785e1e59fe/lib/index.ts#L190

src/components/Menu/ActionEntry.scss
```diff
@@ -14,7 +16,7 @@
 		border: 0;
 		// opacity: 0.5;
 		position: relative;
-		color: var(--color-main-text);
+		color: $text-menubar-action-white-color;
```
vinicius73, line 19: Can we avoid SASS variables in favor of CSS Variables?

src/components/Editor/RightSideActions.vue
luka-nextcloud: @vinicius73 I don't think so. We need to interact with the sidebar from the component. Any suggestion?

src/components/Menu/MenuBar.vue
luka-nextcloud: It's default value. Can you please elaborate?

luka-nextcloud:
@juliushaertl I've uploaded a video to show the status.

juliushaertl:
@luka-nextcloud I think we agreed on trying to use portal-vue as a way to not re-implement the header bar, but instead try to only inject the menubar parts into the existing header. 

https://github.com/nextcloud/text/issues/82#issuecomment-1152315458

If there have been any obstacles, maybe you can provide a summary of what you tried and why this didn't work.

This would be a flexible way to also take the [mockups](https://github.com/nextcloud/text/issues/82) into account which are different for the logged in users where the viewer is used with the white header bar and the public share links that use the blue existing bar. We should try to avoid reimplementing the parts that are already in there by default like the close button/actions/download button.

juliushaertl:
@luka-nextcloud Let me close this for now as it is on hold and needs a different approach anyways as we talked about the possible usage of vue-portal.

PR #2832, opened 2022-08-24
Allow opening files outside of the file system in viewer

juliushaertl:
Some early PoC for displaying text file attachments from mail in viewer

- [ ] requires https://github.com/nextcloud/server/pull/39171
- [ ] requires https://github.com/nextcloud/viewer/pull/1835
- [x] BUG variables css missing
- [x] Properly check if the file outside of the filesystem was passed
- [x] Wrap logic into dedicated component to handle styling
  - [x] Needs a scroll area
  - [x] Header style is missing with blue bar
- [x] BUG: When opening a file in mail the file name is not displayed properly (probably something for mail/viewer) -> needs retesting

## Follow up / separate
- [x] Fix menubar size calculation when showing side by side comparison https://github.com/nextcloud/text/pull/4637
- [ ] Cypress tests
- [ ] Read only mode does not have syntax highlighting (probably needs moving hljs styles from Editor.vue to prosemirror.scss)

cc  @ChristophWurst


juliushaertl:
I've updated the checklist above, when giving it another test run together with the versions PR i think there are a few more topics to address, added to the description.

juliushaertl:
@ChristophWurst FYI we will merge and backport this for 27.1. Iirc you had a mimetype filter in mail so might be a quick win to enable preview for markdown attachments then

src/components/ViewerComponent.vue
```diff
@@ -67,19 +80,86 @@ export default {
...
+		},
+	},
+
+	watch: {
+		source() {
+			this.loadFileContent()
+		},
+		active() {
+			this.loadFileContent()
+		},
```
max-nextcloud, line 115: I don't quite understand why this is needed. The `content` should already be fetched during the `mounted` handler. Why do we need to fetch it again if `active` is toggled?
Do we also need to fetch it if `active` turns false?

I assume this won't matter anymore if the `files` list in the viewer is empty as i suspect active is toggled when a `previous` file becomes `current`.

max-nextcloud (approved):
Just a minor remark. Other than that... good to go from my side.

src/components/ViewerComponent.vue
juliushaertl: Yes, this seemed a leftover from the initial attempt. Removed now

mejo-:
/backport 246f1584c21b7a64eb6a8afb635206852c77b8bb,67f01011cd16b62d153a8593bc49fc55b3c10b20,b35a36968cc33270502133fb44d0a187864b7d4e to stable27

PR #2849, opened 2022-08-31
Do not try to get a displayname for guests

juliushaertl:

* Resolves: # 
* Target version: master


max-nextcloud:
cypress tests are still failing but now they are exposing a different issue. Somehow there's no menubar and the editor is in a funny state when loaded from the share token. I have not been able to reproduce this outside of cypress yet though.

max-nextcloud:
Looks like the editable share doesn't do what i thought it does. May have to turn this into two requests or so.

max-nextcloud (approved):
Fixes the issue i ran into when trying to fix the cypress tests.

juliushaertl:
Awesome, thanks a lot.

PR #2852, opened 2022-09-02
Fix missing sharing token

vinicius73:


cypress/e2e/share.spec.js
```diff
@@ -95,16 +95,17 @@ describe('Open test.md in viewer', function() {
 			})
 	})
 
-	it('Shares a folder as a public read only link', function() {
+	it.only('Shares a folder as a public read only link', function() {
```
max-nextcloud, line 98: Pretty sure the only should not be here anymore.

cypress/e2e/share.spec.js
```diff
@@ -95,16 +95,17 @@ describe('Open test.md in viewer', function() {
...
 		cy.shareFile('/folder')
 			.then((token) => {
-				cy.logout()
-				cy.visit(`/s/${token}`)
+				return cy.logout()
+					.then(() => token)
 			})
+			.then((token) => cy.visit(`/s/${token}`))
```
max-nextcloud, line 104: Is this change really needed?
At some point i need to learn how these thenables work in cypress.
I was under the impression that the visit worked fine before.

cypress/e2e/share.spec.js
```diff
@@ -95,16 +95,17 @@ describe('Open test.md in viewer', function() {
...
 			.then(() => {
-				cy.get('#rich-workspace').should('contain', 'Hello world')
+				cy.get('[data-text-el="editor-content-wrapper"]').should('contain', 'Hello world')
 				cy.openFile('test.md')
-				cy.get('#editor-container').should('be.visible')
+				cy.get('[data-text-el="editor-content-wrapper"]').should('be.visible')
 				cy.get('#editor .ProseMirror').should('contain', 'Hello world')
 				cy.get('#editor .ProseMirror h2').should('contain', 'Hello world')
```
max-nextcloud, line 110: I think we should make sure that we are actually looking at the modal and not the workspace here.

max-nextcloud (commented):
Thanks a lot for working on this.
The approach looks good. I have some comments on the test.

vinicius73:
> I have some comments on the test.

Thanks!
I've changed the test.

max-nextcloud (approved):
Thanks for addressing the comments. Looks great now!

Issue #2859, opened 2022-09-05
Anchor links: headlines hidden behind toolbar


mejo-:
**Describe the bug**
When opening an anchor link, the linked headline is hidden behind the toolbar.

**To Reproduce**
1. Create a Text document with page anchors
2. Click on a page anchor
3. See the linked headline being hidden behind the toolbar

**Expected behavior**
The linked headline should be visible.

**Screenshots**
![2022-09-05T11:20:05,892427523+02:00](https://user-images.githubusercontent.com/3582805/188415436-8c17bdcf-7175-487d-8313-800dbbfd9fb3.png)

@susnux maybe you could have a look since you implemented the anchor link feature (thanks again for that, btw) :blush:

PR #2862, opened 2022-09-05
Fix headlines being hidden behind toolbar (anchor links)

susnux:
* Resolves: #2859
* Target version: master 

### Summary
This adds a scroll margin to the target elements when using the anchor links to scroll to an element.
Fixing an issue where the heading is hidden behind the toolbar.


max-nextcloud (approved):
Looks good to me. I wonder if this also fixed the failing cypress test.

Issue #2785, opened 2022-08-10
Missleading keyboard behaviour in emoji picker


mejo-:
**Describe the bug**
There's room for improvement in the keyboard navigation in emoji picker:

* [ ] When emoji picker is open, <kbd>Esc</kbd> should only close emoji picker, not the full editor modal
* [ ] <kbd>Tab</kbd> should insert the selected emoji
* [ ] When emoji picker is open, <kbd>Esc</kbd> followed by <kbd>Return</kbd> inserts the last selected emoji. This should not happen. It's only reproducible in public shares or Collectives right now, due to the fact that <kbd>Esc</kbd> closes the whole editor modal in other cases.

PVince81:
unclear whether this needs a fix in the emoji picker or on the text app side (ex: prevent event bubbling) ?

for emoji picker issues specifically, please raise in the [nextcloud-vue repo](https://github.com/nextcloud/nextcloud-vue)

mejo-:
Upstream issue regarding the unexpected behaviourt with <kbd>Esc</kbd> followed by <kbd>Return</kbd>: https://github.com/ueberdosis/tiptap/issues/3063

Issue #2199, opened 2022-02-23
Escape key should not close the editor


PVince81:
**Describe the bug**
Don't close the editor unexpectedly when hitting escape.

**To Reproduce**
1. Type some text
2. Select a portion of the text
3. Hit escape to unselect the text
4. Be surprised, where did the editor go ?

**Expected behavior**
Editor stays open at all times when hitting Escape

**Client details:**
 - OS: openSUSE Tumbleweed
 - Browser: Firefox
 - Version: 97.0
 - Device: desktop

**Server details:**
c.nc.com

juliushaertl:
@jancborchardt @nimishavijay What do you think about this one? I agree that initial escape should probably be catched during editing so that one could get the keyboard focus back to other ui elements, but not sure if we should then on the second escape not still close the viewer with text open.

PVince81:
does the escape key also close Collabora ?

for a viewer I think escape key closing is fine but not something you can work actively in

mejo-:
I think that Text in Viewer is a bit different than a picture, video or PDF. If I open a file for editing, I wouldn't expect the editor to disappear when hitting <kbd>Esc</kbd>. Even not when hit a second time. That's why I would suggest to catch <kbd>Esc</kbd> in general and don't let it close the viewer.

nimishavijay:
I think it's important for keyboard accessibility to have an easy way to close the editor using the keyboard. If nothing is selected/highlighted/no popovers or menus are open then I'd say it would be okay to close the editor. Since it feels like something "opened" like a modal, using `Esc` key to close it makes sense. 
First `Esc` key could close any action menu or popover/unselect text and clicking the `Esc` key twice could close the editor.

jancborchardt:
> First `Esc` key could close any action menu or popover/unselect text and clicking the `Esc` key twice could close the editor.

Fully agree on that.

Also tested with some other things:
- Gmail doesn’t do any 2-step thing, Esc directly closes the composer and saves the draft
- Notion "new page" does it similar to Gmail when you are in the title, but has a weird 2-step when in the content, selecting the "container"/"module" of the text
- Any textarea or input field in the browser unselects on Esc, so yes it is behavior we should respect (LibreOffice unselects on Esc as well, for the record)

juliushaertl:
Might also we an accessibility feature if we let esc bring the focus back to the page outside the editor field.

susnux:
[...]

PR #2863, opened 2022-09-06
Fix emoji autocompletion keyboard behaviour

mejo-:
* [ ] ~~When emoji picker is open, <kbd>Esc</kbd> should only close emoji picker, not the full editor modal~~ (This needs to be tackled seperately, we have #2199 to track it.)
* [x] <kbd>Tab</kbd> should insert the selected emoji
* [x] When emoji picker is open, <kbd>Esc</kbd> followed by <kbd>Return</kbd> inserts the last selected emoji. This should not happen. It's only reproducible in public shares or Collectives right now, due to the fact that <kbd>Esc</kbd> closes the whole editor modal in other cases.

Resolves: #2785


mejo-:
> * [ ]  When emoji picker is open, Esc should only close emoji picker, not the full editor modal

This needs to be tackled seperately, we have #2199 to track it.

max-nextcloud (approved):
Code looks good. I'll try it and then merge.

mejo-:
/backport 1bcf04181832c7b514ba9d4af07165bb5f18964c,f090c52317aebfdc36a6d965d2307fdf100139ee,6eb324f144b3761fc153dacc75b14e3e9b35d0cf to stable24

Issue #112, opened 2019-06-27
Support printing notes (first print CSS, then export to PDF)


rkaraba:
Please add PDF export because people sometimes need print their notes.

jancborchardt:
@juliushaertl what do you think? Does this require some PDF generation library?

Another thing we can do is look at a proper print stylesheet which hides the top bar and adjusts some other things. @rkaraba would that solve the problem?

juliushaertl:
Maybe we can use https://github.com/MrRio/jsPDF to build the pdf file, but that needs a deeper look.

I agree that a proper print sylesheet work actually be a good first step.

rkaraba:
@jancborchardt 
Your solution is better than nothing so yes it will help

nicostuhlfauth:
I already worked with jsPDF before. It's very easy to use, well documented and performant.

jancborchardt:
@nicostuhlfauth cool, do you want to help implement this? :)

@juliushaertl maybe also helpful: for the Mail app @sourenaraya implemented a print stylesheet, and PDF export can then be done through the browsers/OS own "Print as PDF" functionality: https://github.com/nextcloud/mail/pull/1804

juliushaertl:
Did not make it for 18

Jab2870:
Is this still on your radar? Printing would be excellent, printing to pdf is (IMO) secondary.

jancborchardt:
@Jab2870 yes, are you interested in contributing a fix?

See https://github.com/nextcloud/mail/pull/1804 where @sourenaraya implemented a print stylesheet for the Mail app. This could be done just the same here. :)

joshdmwood:
Hi, just wanted to add that I encountered this issue today when attempting to print my markdown document. I could not a find working md-to-pdf or print from md feature within nextCloud that actually worked and I was not able to get the pdf flow working either. just wanted to thank you in advance for making this possible and keeping the beautiful css of nextCloud while printing. Thanks!

ImaCrea:
Hello ! It'd be very helpful to be able to print or export as pdf notes. I see the discussion here stopped a while ago. Is there any chance we see that feature shipped at some point please? 🙏

WilliamDEdwards:
In NextCloud, notes are stored as - and therefore also accessible as - regular files. To print a note before this issue is solved, without having to copy + paste it to some application manually, you could download the file from the file browser, or open it from your NextCloud client, and print it using your text editor.

ImaCrea:
[...]

PR #2867, opened 2022-09-06
Add printing style sheet, enables printing of text files

susnux:
* Resolves: #112
* Target version: master 

### Summary
Currently printing will include the menu bar and longer text files will be cropped after the first page.
This enables printing of text documents without any major styling changes,
slightly adjusted margins and added table borders (they are quite hard to read without on paper).

This implements the CSS part of #112


css/print.scss
```diff
@@ -0,0 +1,83 @@
...
+	#app-sidebar-vue {
+		display: none;
+	}
+
+	// Hide Content behind modal
+	#content {
+		display: none;
+	}
+
+	#viewer {
```
vinicius73, line 17: We should avoid global selector like that.

```suggestion
	#viewer[data-handler="text"] {
```

susnux: Fixed that

susnux:
I had to add the rule for hiding the content behind the modal as an event listener to prevent hiding the content if the editor was loaded for *rich workspace*.
So this should now work for all cases.

vinicius73:
Looks good @susnux 

But only the "first page" is printable.

https://user-images.githubusercontent.com/1561347/189185433-fb556c0f-cee4-4fa7-a6e6-ae3a85ba5662.mp4

This is expected?

File used into this test:

[demo.md](https://github.com/nextcloud/text/files/9529021/demo.md)

susnux:
@vinicius73 
> But only the "first page" is printable.
> Is this expected?

No, this happened because the `body` was set to `position: fixed` and `overflow: hidden`. Added a rule to fix this.

vinicius73 (approved):
Looks awesome.

https://user-images.githubusercontent.com/1561347/189776390-9ffb7e1c-b276-4665-8460-da83ecfb60ea.mp4

Issue #2869, opened 2022-09-07
MenuBubble: Detection whether link is active or not is broken


mejo-:
**Describe the bug**
The detection whether current selection is a link or not is broken in MenuBubble.vue.

Seems like the result of [`isActive` in `MenuBubble.vue`](https://github.com/nextcloud/text/blob/e542d159f98a9b4761ea95ff2760035b94899531/src/components/MenuBubble.vue#L174) somehow gets cached. When watching it with a debugger, `isActive` is not called again during selecting new text.

I was able to reproduce the bug on both Nextcloud 24 and current master (25).

**To Reproduce**
1. Select text in a document and make it a link. The options change from "Add Link" to "Update Link" (as expected)
2. Select some other text in the document. The linking options stay "Update Link" despite no link being selected.

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/3582805/188850653-8314b7ae-5a4b-4d3e-9558-e399a1c84a03.png)

mejo-:
The tiptap MenuBubble example doesn't seem to have this problem: https://tiptap.dev/api/extensions/bubble-menu/

PR #2872, opened 2022-09-08
fix: update link menububble on editor updates

max-nextcloud:
* Resolves: #2869 
* Target version: master, stable24 ?

## Summary

Ensure the menububble always shows the correct entries.

The menububble shows different buttons if the selected text is a link. However its content often became out of sync with the editor:
* when new text was selected the bubble was still cached
* when the selection stayed active but was expanded / shrinked the menu bubble also stayed the same
* when someone else editing the same page changed the current selection into a link

Recompute the menububbles `active` data on every editor change.

The code used for this is mostly taken from the menu bars mechanism for the same purpose.
See updateState in src/components/Menu/BaseActionEntry.js.


src/components/MenuBubble.vue
```diff
@@ -113,11 +114,21 @@ export default {
...
 		return {
+			isActive: false,
 			linkUrl: null,
 			linkMenuIsActive: false,
 			isUsingDirectEditing: loadState('text', 'directEditingToken', null) !== null,
 		}
 	},
+	mounted() {
+		this.$_updateIsActive = debounce(this.updateIsActive.bind(this), 50)
+		this.$editor.on('update', this.$_updateIsActive)
```
max-nextcloud, line 125: I'm wondering if we actually need this.
Any change to the content in question (i.e. someone else turning the selection into a link) should also result in changes to the absolute position of the selection in the text because new tags are inserted. Not sure if that triggers a `selectionUpdate` though. Will need to try it out.

max-nextcloud: Okay... so i just checked and we need to act on `update` as well. Even when we change the link ourselves the menububble would not update otherwise.

mejo- (approved):
Thanks a lot, works nicely. I really like the approach with listening to changes from the editor.

max-nextcloud:
/backport 7bd2e54e2b11803e7c2e46a1b82e1de8420fcaab to stable24

PR #2875, opened 2022-09-08
[stable24] fix: update link menububble on editor updates

mejo-:
Backport of #2872

Ensure the menububble always shows the correct entries.

The menububble shows different buttons if the selected text is a link. However its content often became out of sync with the editor:
* when new text was selected the bubble was still cached
* when the selection stayed active but was expanded / shrinked the menu bubble also stayed the same
* when someone else editing the same page changed the current selection into a link

Recompute the menububbles `active` data on every editor change.

The code used for this is mostly taken from the menu bars mechanism for the same purpose.
See updateState in src/components/Menu/BaseActionEntry.js.

* Resolves: # 
* Target version: master

### DIFF
--- a/src/components/MenuBubble.vue
+++ b/src/components/MenuBubble.vue
@@ -39,23 +39,23 @@
 
 		<template v-else>
 			<button class="menububble__button"
-				:class="{ 'is-active': isActive('link') }"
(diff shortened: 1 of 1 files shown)


mejo-:
Closing, as #2874 exists.

Issue #2368, opened 2022-05-05
Multiple elements with the same html id


vinicius73:
**Describe the bug**

the same editor component runs almost 2 times:
- Rich Workspaces
- Modal/Full editor

https://github.com/nextcloud/text/blob/ab7dcc2b607c00b56a044e8a9a36a999eeb36e80/src/components/EditorWrapper.vue#L24

It's natural to have the same ID more than one time in this scenario.
Although the `#editor` appears 3 times.

![image](https://user-images.githubusercontent.com/1561347/166925169-cd06144f-115e-407a-8c12-b7ac39567b84.png)

**Expected behavior**
We must do not repeat the same ID on all pages, it can introduce unexpected behaviors.
In my first investigation, looks like it is impacting the possibility to print the documents. (#112)

vinicius73:
One of these IDs is from server.
https://github.com/nextcloud/server/blob/master/apps/files/templates/list.php#L78

![image](https://user-images.githubusercontent.com/1561347/168392448-519dffe1-36bb-48bf-8a9c-b81ccb2c6ca5.png)

PR #2876, opened 2022-09-08
Fix: Multiple elements with the same html id

susnux:
* Resolves: #2368
* Target version: master 

### Summary
Remove IDs on elements which might be present multiple times in the DOM. This might happen because they are used in the rich workspace but also in the editor viewer.
Replaced the ID usage with the class name.


vinicius73 (approved):
Thanks for taking care of this.

PR #2906, opened 2022-09-12
fix: solve escape key closing the editor issue

luka-nextcloud:

* Resolves: #2199
* Target version: master


vinicius73 (approved):
Works well while you are with editor in focus (typing/editing)

But when your focus move to a dropdown menu (title, callouts...) `esc` key stil closing the modal.

I guess we can solve it in another issue.

juliushaertl:
> But when your focus move to a dropdown menu (title, callouts...) esc key stil closing the modal.

Popovers are actually part of the body and not within the modal, but we might to prevent the escape key event bubbling in the actions in general in `@nextcloud/vue`. There currently the keydown handler is registered globally.

juliushaertl (approved):

Issue #2868, opened 2022-09-06
Change heading error


vinicius73:
Related to: #2841 

https://user-images.githubusercontent.com/1561347/188709155-07918164-6129-4714-b3d9-6564aad7d731.mp4

vinicius73:
@susnux I will revert #2841 until I find a solution for this.
It seems related to the `addNodeView`  and `toggleNode` in https://github.com/ueberdosis/tiptap/blob/5bcfb322ceb30575430bd8797f2787d65de80f2c/packages/extension-heading/src/heading.ts#L81

susnux:
The issue is caused by the VueNodeViewRenderer, I have implemented the node view using plain js and that version is not affected by the issue.

susnux:
@vinicius73 I think I found a solution, have a look at https://github.com/nextcloud/text/tree/feat/page-anchors

It seems the problem is prosemirror rendering the children into the `contentDOM` before the tiptap vue renderer updates the nodeview, which will replace the `contentDOM` with a new one as the wrapper element changes (e.g. from `h2` to `h1`).

I solved it by enforcing tiptap to remount the NodeView if the level changes. I tested it and now it works, also added some more tests for this issue.

vinicius73:
Works perfectly @susnux 
Please go ahead and open a PR pls.

PR #2914, opened 2022-09-13
Feat/page anchors

susnux:
* Resolves: #2868
* Target version: master 

### Summary
Bring back the link anchors for headings. Also added tests for the issue where changing the heading level removed the heading content (#2868)


vinicius73 (approved):
Thanks for take care of this.

Issue #2904, opened 2022-09-12
Table of contents: add button to close outline from itself


mejo-:
**Describe the bug**
The table of contents can only be opened and closed by navigating to the headings submenu in the menubar. While this is fine for opening the outline, it feels quite cumbersome for closing it. We should add a button to the outline view to close it from itself.

mejo-:
@vinicius73

mejo-:
Sorry, I only tested the outline on Collectives and there the heading and close button were hidden by the menubar. I fixed this now by adding the following in Collectives:

```scss
[data-text-el='editor-container'] .editor--outline {
	z-index: 10021;
}
```

mejo-:
Well, the better solution would be to remove `top: 104px` from https://github.com/nextcloud/text/blob/9a45c935dd4451a83c8cc0b9f6a6575262ad7c54/src/components/Editor/EditorOutline.vue#L66. What was the reason that you added it @vinicius73?

vinicius73:
I'm working in this approach.

https://user-images.githubusercontent.com/1561347/190250989-696a4ea5-d932-48c2-8c51-13076a383667.mp4

vinicius73:
> Well, the better solution would be to remove `top: 104px` from
> 
> https://github.com/nextcloud/text/blob/9a45c935dd4451a83c8cc0b9f6a6575262ad7c54/src/components/Editor/EditorOutline.vue#L66

> . What was the reason that you added it @vinicius73?

I've removed it 😉

PR #2920, opened 2022-09-14
Able to show outiline in readonly mode

vinicius73:
* Resolves: #2904 
* Target version: master 

### Readonly

https://user-images.githubusercontent.com/1561347/190261995-20c63bc2-5c7b-49b3-a6e6-5cc34a480f42.mp4

### Public

https://user-images.githubusercontent.com/1561347/190262053-cd9f71de-af29-4735-98a9-4db1ded0d385.mp4


max-nextcloud (commented):
Code looks good.

However if i click an entry in the outline it does not move to that heading. That's not what i'd expect because the entry is highlighted and looks like a link to me.

I also managed to trigger some funny behavior when clicking at the end of headings. I was not able to produce the same behavior on the editable view of the owner of the file.
[Bildschirmaufzeichnung vom 15.09.2022, 08:19:51.webm](https://user-images.githubusercontent.com/97337118/190329639-3aeb07cc-7ab9-45c8-9bf1-4359edc32544.webm)

src/components/Menu/ReadonlyBar.vue
```diff
@@ -0,0 +1,35 @@
...
+	components: { ActionEntry },
+	setup() {
+		return {
+			visibleEntries: entries,
+		}
+	},
+})
+</script>
+
+<style scoped>
```
susnux, line 30: To be consistent with the edit mode the menu bar should be `sticky`.

jancborchardt (commented):
Nice! Another point in addition to @susnux:
The "x" close button is a bit floating in the air. Would be better if it’s directly after the "Outline" text instead of floating on the right.

mejo-:
Thanks a lot for looking into this @vinicius73 :heart:

Unfortunately using the menubar space in read-only mode conflicts with Collectives, where we already use this space for Collectives-specific content (last changed user and timestamp, in future also things like count of backlinks and attachments). Sorry, I don't have a quick solution for this. Maybe we should discuss it in our next team call?

vinicius73:
> Thanks a lot for looking into this @vinicius73 :heart:
> 
> Unfortunately using the menubar space in read-only mode conflicts with Collectives, where we already use this space for Collectives-specific content (last changed user and timestamp, in future also things like count of backlinks and attachments). Sorry, I don't have a quick solution for this. Maybe we should discuss it in our next team call?

What do you think about make able to inject a slot that space and expose some editor methods?

So that way you will be able to not only inject more options also access internal editor methods in a safe way

vinicius73:
@jancborchardt what do you think about this?

image

mejo-:
> What do you think about make able to inject a slot that space and expose some editor methods?
> 
> So that way you will be able to not only inject more options also access internal editor methods in a safe way

That would be the way to go indeed :blush:

jancborchardt:
@vinicius73 it's better if the x is on the right if the text, so that the text is left-aligned with the headings below. :) Just not all the way over away from the text.

blizzz:
master is 26, please trigger backport if necessary.

vinicius73:
@mejo- I've rewritten the all PR, check if you can please

mejo-:
Nice, thanks a lot @vinicius73. Works like expected. I removed the `top: 104px` as discussed [here](https://github.com/nextcloud/text/issues/2904#issuecomment-1247244100). It doesn't solve the problem for Collectives yet, but at least it doesn't break Collectives any longer - and it adds outline support for read-only Text shares, which is great :heart_eyes:

mejo- (approved):

mejo-:
/backport 103d8ef26df9ec6c86098eecbf16304f08248f7d,7d9a18b15c56608fa38492ec9d7f877130922488 to stable25

Issue #2008, opened 2021-12-15
Open embedded image in original size


philsch:
**Is your feature request related to a problem? Please describe.**
Embedding images is very useful. However, I often have photos of whiteboards in my notes and need to open the photo in full resolution to be able to read the content. It would be nice if this would be possible out of the box.

**Describe the solution you'd like**
There are different possible solutions:

* being able to control the size of the embedded image (to embed a higher resolution in the first place)
* opening a popup when clicking the image
* opening the original image (in a new browser tab?)

**Describe alternatives you've considered**
My current workaround is to add a link on top of the image, but this way I have to select the image source twice (first to add the image itself, second to add the link to the file).

![Screenshot 2021-12-15 at 17 07 53](https://user-images.githubusercontent.com/5527838/146224114-3230ccdb-1e06-4feb-a244-bc4ab214bdbc.png)

ya-d:
Thanks for raising this issue, I stumbled open the same problem. I think both of these options should be possible:

* Control how images are **displayed within the document**. Maybe limited options like "full width", "half width", "quarter width" are sufficient and require less effort wrapping text around or implementing free form controls.
* Clicking an image should **open the regular full page image viewer**. (Which in turn might get zooming, paging (last/next image) or even simple editing capabilities (e.g. rotate, crop) in the future.)
* Opening the original image is always possible via context menu (right click / two finger click / long press / …). This is sufficient in my view.

Its not only going bigger, sometimes you want images to appear smaller. If you write a documentation in Text or Collectives and insert screenshots of certain parts of the UI they can get very big quickly and consume much of your layout which hinders readability. This ist especially true for high DPI screens (2x) where your screenshot appears twice as big.

I therefore propose to change the title to "Handle size of embedded images" or something similar.

juliushaertl:
Also relevant for collectives https://gitlab.com/collectivecloud/collectives/-/issues/364

juliushaertl:
@jancborchardt Maybe you can write some quick spec about how this should look from the UX perspective, so @luka-nextcloud can look into this :)

jancborchardt:
Yep, so for the first version, it should simply be possible to click the image to open it in the Viewer.

As an added bonus, it would be nice to be able to use the viewer next-previous control arrows on the sides to go through the images inside the current document.

Any further display controls as @ya-d suggested are something for a separate issue. (Great proposal though! :)

PR #2924, opened 2022-09-16
feat: show images in a document with the viewer

luka-nextcloud:

* Resolves: #2008
* Target version: master


max-nextcloud:
@luka-nextcloud Please fix the eslint issues. Thanks!
You can also check for them locally by running `npm run lint`.

eneiluj (commented):
* The directory path for the first PROPFIND is wrong. It seems it looks for `STORAGE_ROOT/ATTACHMENT_DIR_NAME` so it only works for documents in user's storage root.
* This will not work when accessing the document via a shared access. We could use the endpoints provided by the attachment controller. These are public endpoints using the Text edition session to authenticate. I can help in this department if needed.

You could use `#getImageAttachmentUrl()` in `src/services/AttachmentResolver.js` which ends up getting the image from `AttachmentController::getImageFile()`.
Problem is that this method only returns a preview for now. We could add a parameter to choose between raw content or preview.
There is still the special case of images which can't be displayed but for which there is a preview provider. So in this case, the preview would be a fallback even if asking the raw image.

eneiluj:
https://github.com/nextcloud/text/pull/2974 might help here. You could get the image URL with `#getImageAttachmentUrl(FILE_NAME, true)`.

eneiluj:
@luka-nextcloud https://github.com/nextcloud/text/pull/2974 is merged now :grin:

src/components/ImageView/ShowImageModal.vue
```diff
@@ -0,0 +1,76 @@
...
+			default() {
+				return []
+			},
+			validator(imagesList) {
+				return (imagesList.length === 0)
+					? true
+					: imagesList.findIndex(({ basename, source }) => !(basename && source)) !== -1
+			},
+		},
+		startIndex: {
```
eneiluj, line 39:
```suggestion
		imageIndex: {
```
Because `startIndex` suggests that it is definitive and that the prop is not reactive but it is! :+1:

src/components/ImageView/ShowImageModal.vue
```diff
@@ -0,0 +1,76 @@
...
+		},
+		showPreviousImage() {
+			this.currentImageIndex = this.currentImageIndex <= 0
+				? this.images.length - 1
+				: this.currentImageIndex - 1
+			this.currentImage = this.images[this.currentImageIndex]
+		},
+	},
+}
+</script>
```
eneiluj, line 76:
```suggestion
</script>

<style scoped lang="scss">
.modal__content {
	height: 100%;
	padding: 0 50px;
	display: flex;
	justify-content: center;
	img {
		height: 100%;
		width: 100%;
		object-fit: contain;
	}
}
</style>
```
So that we always see the entire image and the aspect ratio is preserved.

https://user-images.githubusercontent.com/11291457/194531704-c38b75e6-2f0f-4d3d-bb57-813993e8e719.mp4

eneiluj (changes requested):
Nicely done!

* [ ] A few remaining eslint warnings and errors (`npm run lint`)
* [ ] The image could use all the available space and avoid overflow. See inline basic style suggestion.
* [ ] The dav helper is not used anymore and could be removed IMO
* [ ] Pressing escape closes the modal and the document (the viewer). @max-nextcloud @vinicius73 Do you know if there is any way to avoid that?

+ some other inline comments

luka-nextcloud:
@julien-nc Can you please check again?

julien-nc (approved):

juliushaertl:
@luka-nextcloud Would be great to have a cypress test for this :)

PR #2970, opened 2022-09-20
Add support for @user mentions to RichTextReader.vue

### LINKED ISSUE #2919: Add new features to RichTextReader
**Describe the bug**
We need to add the new Text features to RichTextReader in order to use them in read-only mode (e.g. in Collectives):

* [x] rendering of user mentions (#2970)
* [ ] rendering of file attachments
* [x] attributes for page anchor links (#3345)
* [x] Rich link previews
* [x] Allow toggeling outline

mejo-:
* Partly resolves #2919
* Target version: master


max-nextcloud (approved):
Haven't tested it myself. Codewise looks good.

Good catch that the config needs to be different in edit and readonly mode.

Issue #2847, opened 2022-08-29
Y.js backend drop in replacement


juliushaertl:

max-nextcloud:
Next steps afterwards...

- [ ] put compressed update on server when saving file
- [ ] understand awareness protocol
- [ ] use notify push
- [ ] make use of awareness
- [ ] user cursor
- [ ] user color

juliushaertl:
Reordered them a bit on what I would consider the order of priority, hope that makes sense ;)

Issue #1042, opened 2020-09-15
Cache autosave document to reduce writes to disk


juliusknorr:
- Temporarily dump the document to save into a cache
- Write it back less often than every 10s to reduce possible disk writes (also db load when triggering activity)

Downside: File changes might take longer to be represented in the actual files, which i'd say is acceptable since in most cases the editing will happen mainly in the text app then. We can still force save to disk when users actively close the session or press ctrl+s

Benefit: Less disk access

cc @rullzer

jospoortvliet:
Less often than every 10 seconds would be good. Indefinitely is not - maybe a once-every-five-minutes or so, IF there are edits of course. You do want the changes to get synced to the clients, just not too often...

juliushaertl:
Yeah we can basically persist it with the cronjob then, if no user is connected anymore, if users are active we then just save less regularly.

mejo-:
Nowadays we only write the document every 30 seconds. I'll close this issue given that the topic didn't come up in the last years.

PR #2971, opened 2022-09-20
Minimal yjs based collaboration

max-nextcloud:
* Target version: master 
* Fixes https://github.com/nextcloud/text/issues/2847


src/components/Editor.vue
```diff
@@ -69,10 +69,10 @@
 import Vue, { set } from 'vue'
 import escapeHtml from 'escape-html'
 import moment from '@nextcloud/moment'
-import { getVersion, receiveTransaction } from 'prosemirror-collab'
-import { Step } from 'prosemirror-transform'
 import { getCurrentUser } from '@nextcloud/auth'
 import { loadState } from '@nextcloud/initial-state'
+import { Collaboration } from '@tiptap/extension-collaboration'
+import * as Y from 'yjs'
```
vinicius73, line 75: Try to do not import `*` from any package.

If this package supports [tree shaking](https://webpack.js.org/guides/tree-shaking/), it will not be applied.

lib/Service/DocumentService.php
```diff
@@ -125,11 +124,11 @@ public function createDocument(File $file): Document {
 
 			// Do not hard reset if changed from outside since this will throw away possible steps
 			// This way the user can still resolve conflicts in the editor view
-			if ($document->getLastSavedVersion() !== $document->getCurrentVersion()) {
+			$stepsVersion = $this->stepMapper->getLatestVersion($document->getId());
+			if ($stepsVersion && ($document->getLastSavedVersion() !== $stepsVersion)) {
```
juliushaertl, line 128: If we drop the step version that check against the last saved version will no longer work, so maybe we need to find a different approach to determine the collision case.

lib/Db/StepMapper.php
```diff
@@ -51,6 +52,24 @@ public function find($documentId, $fromVersion, $lastAckedVersion = null) {
 		return $this->findEntities($qb);
 	}
 
+	public function getLatestVersion($documentId): ?string {
```
juliushaertl, line 55: Maybe we can return an int here?

```suggestion
	public function getLatestVersion($documentId): ?int {
```

lib/Service/ApiService.php
```diff
@@ -93,9 +91,13 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
...
 			} elseif ($fileId) {
-				$file = $this->documentService->getFileById($fileId);
+				try {
+					$file = $this->documentService->getFileById($fileId);
+				} catch (NotFoundException $e) {
+					return new DataResponse([], Http::STATUS_NOT_FOUND);
+				}
 			} else {
-				return new DataResponse('No valid file argument provided', 500);
+				return new DataResponse('No valid file argument provided', 412);
```
juliushaertl, line 100:
```suggestion
				return new DataResponse('No valid file argument provided', Http::STATUS_PRECONDITION_FAILED);
```

lib/Service/ApiService.php
```diff
@@ -127,19 +129,26 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
 		}
 
 		$session = $this->sessionService->initSession($document->getId(), $guestName);
+
 		try {
 			$baseFile = $this->documentService->getBaseFile($document->getId());
 			$content = $baseFile->getContent();
-
-			$content = $this->encodingService->encodeToUtf8($content);
```
juliushaertl, line 134: Removed by accident?

lib/Service/DocumentService.php
```diff
@@ -182,59 +192,68 @@ public function get($documentId) {
...
+				array_push($stepsToInsert, $step);
+			} else {
+				array_push($querySteps, $step);
+			}
 		}
+		if (sizeof($stepsToInsert) > 0) {
+			$newVersion = $this->insertSteps($documentId, $sessionId, $stepsToInsert, $version);
+		}
+		// If there were any queries in the steps send the entire history
+		$getStepsSinceVersion = sizeof($querySteps) > 0 ? 0 : $version;
```
juliushaertl, line 213:
```suggestion
		$getStepsSinceVersion = count($querySteps) > 0 ? 0 : $version;
```

I'd prefer using count as sizeof has different meanings in other languages so this may lead to confusion.

lib/Service/DocumentService.php
```diff
@@ -182,59 +192,68 @@ public function get($documentId) {
...
-			$validStepTypes = ['addMark', 'attr', 'removeMark', 'replace', 'replaceAround'];
-			foreach ($steps as $step) {
-				if (array_key_exists('stepType', $step) && !in_array($step['stepType'], $validStepTypes, true)) {
-					throw new InvalidArgumentException('Invalid step data');
-				}
-			}
-			$oldVersion = $document->getCurrentVersion();
-			$newVersion = $oldVersion + count($steps);
+			$stepsVersion = $this->stepMapper->getLatestVersion($document->getId());
+			$newVersion = $stepsVersion + count($steps);
```
juliushaertl, line 239: If we get rid of the lock we may want to wrap this in a transaction so that reading the current version, increasing and writing steps is an atomic operation.

Should be easy with https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/database.html#transactions

juliushaertl (commented):
First half of the review for backend bits

lib/Service/ApiService.php
max-nextcloud: * removed on purpose when dropping the BaseFile
* not brought back when bringing it back
* remove BaseFile again as i am now confident we don't need it

So should be fine now. It does not apply to the state file because that contains base64 encoded content and that should not have utf8 content.

max-nextcloud:
Locally only `sync`, `attachments` and the flaky `Links` tests are failing (with `npx cypress run`).

juliushaertl:
Pushed a bit of polishing for the backend code to make CI happy and a fix for direct editing to pass the initial session data to the web socket polyfill. Let's see what CI says.

juliushaertl:
One thing I noticed locally is that getContent was failing due to a timeout in the cypress tests sometimes. It seems loading the editor might take a bit longer then before.

juliushaertl:
Few more things to fix according to the cypress tests where some js errors are thrown

> this.connection.insertAttachment is not a function
> Cannot read properties of undefined (reading 'lastSavedVersion')

juliushaertl:
@max-nextcloud So far for my week I pushed couple of fixes and polishing. I ended up also refactoring the PollingBackend a bit to clean this up. Cypress tests pass now though there are two draft commits that we should discuss on Monday about on how to best handle that.

- [f232328](https://github.com/nextcloud/text/pull/2971/commits/f2323289297f4b2045bb69f5fd77c4666db35df2) It seemed that there was still an issue where longer running sessions ended up playing ping pong with messages which lead to heavy load. I figured out that we may not need to send back steps during the push operations as this was originally more a shortcut to avoid another polling round. I feel this is more stable and also makes more sense to just have one channel where steps get sent from the server to the clients. This was they can also continue to fetch them in order of the database
- [00879a6](https://github.com/nextcloud/text/pull/2971/commits/00879a6c0989f382a43b3e433a25b53b400882f9) As we no longer store the version of the document the left TODO statement `// TODO: figure out when to autosave` caused that the files were not saved. In order to have a working branch I pushed this draft commit to always send the documentState and autosaveContent with each sync request, but we should probably be smarter about that.

max-nextcloud:
@juliushaertl Wow... thanks for looking into this and pushing the pr forward like that!
:heart:

max-nextcloud:
@juliushaertl Given your last commit i wonder what happens if...
* two users edit the doc
* it's update on the server through a different mechanis -> conflict
* one user resolves the conflict

Ideally the conflict would get resolved for the other user as well, right? In particular so they do not pick different strategies.
Is that the case already, or what happens?
Did we account for that in the past?

juliushaertl:
> Ideally the conflict would get resolved for the other user as well, right? In particular so they do not pick different strategies.
Is that the case already, or what happens?
> Did we account for that in the past?

We didn't have any handling for different strategies in the past. My additional commit would already solve this by just moving any resulution to happen trough y.js, so individual users can also undo. One thing that might need further work is to timely propagate the resolved conflict, but I'd not consider that a blocker for the merge as it has been worse before ;)

lib/Service/DocumentService.php
```diff
@@ -193,8 +193,12 @@ public function addStep($documentId, $sessionId, $steps, $version): array {
 		}
 		// If there were any queries in the steps send the entire history
 		$getStepsSinceVersion = count($querySteps) > 0 ? 0 : $version;
+		// TODO: To discuss
+		// above causes increasing requests, maybe we need to filter out query steps in the above response
+		// Likely steps returned here will also get returned in the next fetch request, so maybe we just skip this
+		// as this was only a shortcut for the old handling to avoid another conflict/sync/push
```
juliushaertl, line 199: - Could be the old awareness messages that cause issues here where old awareness messages come in and update the state again
  - This PR: Maybe we can also filter awareness steps before returning them here
  - Follow up: Maybe store the last awareness message in oc_text_sessions/distributed memory cache (per session or generally per document)
  -

src/services/PollingBackend.js
```diff
@@ -124,8 +124,10 @@ class PollingBackend {
 
 		this.#pollActive = true
 
-		const autosaveContent = shouldSave ? this.#syncService._getContent() : null
-		const documentState = shouldSave ? this.#syncService.getDocumentState() : null
+		const shouldAutosave = true // FIXME: Figure out when we should autosave
```
juliushaertl, line 127: - [ ] For this PR: introduce a time check here to only send this every 30 seconds
- Follow up: 
  - Move save request out of the sync request to a dedicated endpoint
- Maybe later: 
  - Maybe not directly write the content to the file but put it in cache/db for persisting later if all connections are closed #1042 
  - See if we can notify other clients about autosave via awareness messages/simple document state

juliushaertl:
> https://cloud.cypress.io/projects/hx9gqy/runs/8373/overview/f4c7b201-4088-409a-8d43-af032463e2e9?reviewViewBy=FAILED&utm_source=github&utm_medium=failed&utm_campaign=view%20test

Fails because the save is not happening fast enough, maybe we workaround that with a 30s delay for now and then move to the separate request as discussed earlier @max-nextcloud ?

julien-nc:
I didn't check the implementation yet but from user testing:
1. Text enters some kind of request loop when opening a file that was created with the old/current sync mechanism
2. We talked about that and I don't remember so here it is again :grin: : Do you have a plan to manage/prevent this use case if/when we make it possible to have a websocket at some point:
    * user A share a file with B (with write permissions)
    * A and B open the file and edit it
    * A removes the write permission to the share
    * B will be able to push the changes through the websocket

    When everything goes through the server, fine, I was glad to see the 403s.
    With the websocket, it seems costly to validate each step on each client when receiving it.

    I guess one way to put the question is: Where do we put the information required to do proper access control? On the websocket server entity? On the clients?

julien-nc:
Oh and I forgot to mention it again: This is awesome :tada:

max-nextcloud:
> 1. Text enters some kind of request loop when opening a file that was created with the old/current sync mechanism

In general i think we prevent this by cleaning up all sessions during the update. But both for development as well as robustness I will investigate this to make sure we reach a workable state.

max-nextcloud:
> We talked about that and I don't remember so here it is again grin : Do you have a plan to manage/prevent this use case if/when we make it possible to have a websocket at some point:
> 
>     * user A share a file with B (with write permissions)
> 
>     * A and B open the file and edit it
> 
>     * A removes the write permission to the share
> 
>     * B will be able to push the changes through the websocket
> 
> 
> When everything goes through the server, fine, I was glad to see the 403s. With the websocket, it seems costly to validate each step on each client when receiving it.
> 
> I guess one way to put the question is: Where do we put the information required to do proper access control? On the websocket server entity? On the clients?

I think there's two answers to this:
* as a first step i would like to use `notify_push` to send out the changes. This will give us the main benefit of a websocket - no need to poll anymore and pushing changes would still only happen through the old api - so no added attack surface.
* As a pure websocket yjs server i'd look into hokuspokus. It has an [authentication hook](https://tiptap.dev/hocuspocus/guide/authentication-and-authorization) that we'd need to implement.

Another attack vector comes to my mind speaking of this. The websocket yjs provider that we are building upon also syncs between browser tabs. So by taking control of one browser tab i could get access to a document in a different tab. Maybe we can use a secret as the room name to prevent that.

max-nextcloud:
> Fails because the save is not happening fast enough, maybe we workaround that with a 30s delay for now and then move to the separate request as discussed earlier @max-nextcloud ?

I don't fully understand what's happening here. We explicitely save with `ctrl+s`, right? So the `AUTOSAVE_INTERVAL` should not interfere with this.

I'll see if i can reproduce this locally.

juliushaertl (approved):

lib/Service/DocumentService.php
max-nextcloud: I created a follow up issue in #3726

szaimen:
🎉🎉🎉🎉🎉

PR #2974, opened 2022-09-20
Add "preferRawImage" parameter to image attachment endpoint

eneiluj:
This can help https://github.com/nextcloud/text/pull/2924

The logic was to return a preview if possible and the raw content as a fallback.
* If the new parameter is true, we return the raw image (only if it is supported by browsers)
* If it's false, we do as before: try preview, raw as fallback

`#getImageAttachmentUrl()` has a new related parameter.


lib/Service/AttachmentService.php
```diff
@@ -80,43 +83,50 @@ public function __construct(IRootFolder $rootFolder,
 	 * @throws \OCP\Files\InvalidPathException
 	 * @throws \OCP\Files\NotPermittedException
 	 */
```
mejo-, line 85: Nitpicking, but seems like you forgot to update the annotations here.

mejo- (approved):
I didn't test it, but looks sane codewise :+1:

lib/Service/AttachmentService.php
eneiluj: :+1:

lib/Controller/AttachmentController.php
```diff
@@ -190,19 +201,21 @@ private function getUploadedFile(string $key): array {
 	 * @param string $sessionToken
 	 * @param string $imageFileName
 	 * @param string|null $shareToken
+	 * @param int $preferRawImage
```
luka-nextcloud, line 204: I suggest that we use bool here for `prefereRawImage` as you did for the other methods.

eneiluj: I don't trust passing boolean values in GET params. But maybe it's fine, do you actually do it somewhere else? Do you pass the "true" or "false" string and the middleware converts it to a boolean?

I would still prefer keeping it like that and passing 1 for true and 0 for false :grin:.

luka-nextcloud (approved):

PR #2975, opened 2022-09-20
[stable24] Make it possible to insert heif/heic images

eneiluj:
Manual backport of https://github.com/nextcloud/text/pull/2972

I don't know why but explicit heic/heif mime types have to be added in the `<input>`.
They are not shown with `image/*`.

If such files are inserted and the preview provider is not enabled (or does not work), Text behaves just like if we insert a corrupted image.


src/components/MenuBar.vue
```diff
@@ -24,7 +24,7 @@
 	<div class="menubar" :class="{ 'show': isVisible, 'autohide': autohide }">
 		<input ref="imageFileInput"
 			type="file"
-			accept="image/*"
+			accept="image/*, image/heic, image/heif"
```
vinicius73, line 27: Why it is necessary?

eneiluj: > I don't know why but explicit heic/heif mime types have to be added in the `<input>`.
They are not shown with image/*.

juliushaertl (approved):

PR #2977, opened 2022-09-21
Link preview widgets

juliushaertl:
Implement link preview widgets for paragraphs that only contain a link.

Screenshot 2022-09-21 at 15 55 12


eneiluj (changes requested):
* I get 404 responses for open-graph link previews (/core/references/preview/3097fca9b1ec8942c4305e550ef1b50a)
* The RenderReferenceEvent is not emitted by Text so the reference frontend stuff is not loaded with Text so we get `Widget for rich object type integration_github not registered` for every provider.

mejo-:
Whoohoo, super nice :heart_eyes: Works like a charm.

I wonder though, whether there should be an option to remove the preview widget. Imagine e.g. a list of links. Legibility would decrease with the huge preview widgets in between each item in my eyes.

package.json
```diff
@@ -80,6 +81,7 @@
     "@tiptap/extension-underline": "^2.0.0-beta.195",
     "@tiptap/suggestion": "^2.0.0-beta.195",
     "@tiptap/vue-2": "^2.0.0-beta.195",
+    "lodash.debounce": "^4.0.8",
```
mejo-, line 84: What's the reason to use lodash.debounce instead of debounce? As far as I know, we use debounce in many other apps.

src/nodes/ParagraphView.vue
```diff
@@ -0,0 +1,94 @@
...
+			this.text = this.getTextReference(newNode?.textContent)
+		}, 500)
+		this.text = this.getTextReference(this.node?.textContent)
+	},
+	beforeUnmount() {
+		this.debouncedWatch?.cancel()
+	},
+	methods: {
+		getTextReference(text) {
+			const PATTERN = /(^)(https?:\/\/)?((?:[-A-Z0-9+_]+\.)+[-A-Z]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)($)/ig
```
vinicius73, line 66: we can move it to outside of component methods

mejo-:
Another frontend comment: I think it would look better if we displayed *either* the link *or* the widget. Now the link is always duplicated. Once the typed text and a second time at the bottom of the widget.

Even though, thinking about it, no idea how the link could be updated if only the widget is displayed.

src/nodes/ParagraphView.vue
```diff
@@ -0,0 +1,94 @@
...
+:deep(.widget-default--details) {
+	overflow:hidden;
+	p {
+		margin-bottom: 4px !important;
+	}
+}
+</style>
+<style lang="scss">
+@import '@nextcloud/vue-richtext/dist/style.css'
+</style>
```
vinicius73, line 94: We can import it direct in script statement

```html
<script>
import "@nextcloud/vue-richtext/dist/style.css"
</script>
``` 
Or using src attribute. 
https://v2.vuejs.org/v2/guide/single-file-components.html#What-About-Separation-of-Concerns

```html
<style src="@nextcloud/vue-richtext/dist/style.css"></style>
``` 

Also, we do not need `lang="scss"`

juliushaertl:
> I get 404 responses for open-graph link previews (/core/references/preview/3097fca9b1ec8942c4305e550ef1b50a)

Do you have memcache enabled?

> The RenderReferenceEvent is not emitted by Text so the reference frontend stuff is not loaded with Text so we get Widget for rich object type integration_github not registered for every provider.

Ah good point, will add that.

juliushaertl:
> Another frontend comment: I think it would look better if we displayed either the link or the widget. Now the link is always duplicated. Once the typed text and a second time at the bottom of the widget.

> Even though, thinking about it, no idea how the link could be updated if only the widget is displayed. 

Yes, updating was the main reason to always show it. About the option to hide it, I'd agree that it would be nice to have, but we cannot really persist it in markdown, so it would need a per user setting that we store or something like this.

jancborchardt:
Good stuff! :) Design feedback:
- [ ] There seems to be some slight left margin of the container, making it not fully left-aligned with the text
- [ ] The container has the same mrgin to top and bottom, making it unclear which link it belongs to. It should have a bit less margin topl the top and a bit more to the bottom. (E.g. --default-grid-baseline on top, and *3 for bottom. – or *2 top and *4 bottom, whatever looks better.)
- [ ] The container has more padding below the last line of text (link) than above the title. Maybe we need to reduce the image size a bit to fix this?

Possibly these have to be fixed in the component instead?

eneiluj:
> > I get 404 responses for open-graph link previews (/core/references/preview/3097fca9b1ec8942c4305e550ef1b50a)
> 
> Do you have memcache enabled?
> 

I realized right after writing my comment that i had commented the caching stuff in the ReferenceManager. All good now. Sorry for the noise.

> > The RenderReferenceEvent is not emitted by Text so the reference frontend stuff is not loaded with Text so we get Widget for rich object type integration_github not registered for every provider.
> 
> Ah good point, will add that.

I tried dispatching the RenderReferenceEvent event in `lib/Listeners/LoadViewerListener.php` right after loading the 'text-viewer' script and it works fine.

jancborchardt:
@mejo- agree, we discussed before there should ideally be a "close x" on the top right to hide the preview, but decided it's something for a future iteration.

And we also talked about showing both vs only showing the preview, and came to the same conclusion as you with the editing issue.

In a future version, we could make it so that only the preview is shown, and it has a 3-dot menu for "Edit link" and "Hide preview"

mejo-:
> About the option to hide it, I'd agree that it would be nice to have, but we cannot really persist it in markdown, so it would need a per user setting that we store or something like this.

Yeah, I see the point. I wondered whether it would be an option to add the preview as a separate node to markdown. There's e.g. [junkawa/markdown-it-link-preview](https://github.com/junkawa/markdown-it-link-preview). The preview would be auto-generated when inserting an URL, but once it got removed manually it would stay away persistently in the document. What do you think about it?

> And we also talked about showing both vs only showing the preview, and came to the same conclusion as you with the editing issue.
> In a future version, we could make it so that only the preview is shown, and it has a 3-dot menu for "Edit link" and "Hide preview"

That's a great idea!

jancborchardt (changes requested):
Good stuff! :) Design feedback:
- [ ] There seems to be some slight left margin of the container, making it not fully left-aligned with the text
- [ ] The container has the same mrgin to top and bottom, making it unclear which link it belongs to. It should have a bit less margin topl the top and a bit more to the bottom. (E.g. --default-grid-baseline on top, and *3 for bottom. – or *2 top and *4 bottom, whatever looks better.)
- [ ] The container has more padding below the last line of text (link) than above the title. Maybe we need to reduce the image size a bit to fix this?

Possibly these have to be fixed in the component instead?

src/nodes/ParagraphView.vue
juliushaertl: Let me export that from the `@nextcloud/vue-richtext` one separately

eneiluj (approved):
woop woop

juliushaertl:
I could not reproduce the failures reliably locally and they seem to be unrelated, so I'd consider failing cypress tests (outside of the link related ones) as not blocking for the merge.

Issue #2905, opened 2022-09-12
User mentions: improve search results


mejo-:
**Describe the bug**
There's two issues with the results provides by user mentions:
1. The current user is always listed additionally to the search results. When searching for 'admi' as user 'jane', I get 'jane' and 'admin' as results.
2. Exact matches seem to be missing. Searching for 'admi' still lists 'admin', but searching for 'admin' doesn't list it anymore.

@luka-nextcloud

luka-nextcloud:
@mejo- I cannot reproduce the first issue.

PR #3068, opened 2022-09-28
fix: mention search issues

luka-nextcloud:

* Resolves: #2905
* Target version: master

### DIFF
--- a/lib/Controller/UserApiController.php
+++ b/lib/Controller/UserApiController.php
@@ -8,21 +8,18 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\Collaboration\Collaborators\ISearch;
 use \OCP\IRequest;
 use OCP\IUserManager;
-use OCP\IUserSession;
 use OCP\Share\IShare;
 
 class UserApiController extends ApiController {
 	private ISearch $collaboratorSearch;
-	private IUserSession $userSession;
 	private IUserManager $userManager;
 	private SessionService $sessionService;
 
-	public function __construct($appName, IRequest $request, SessionService $sessionService, ISearch $ISearch, IUserManager $userManager, IUserSession $userSession) {
+	public function __construct($appName, IRequest $request, SessionService $sessionService, ISearch $ISearch, IUserManager $userManager) {
 		parent::__construct($appName, $request);
 
 		$this->sessionService = $sessionService;
 		$this->collaboratorSearch = $ISearch;
-		$this->userSession = $userSession;
 		$this->userManager = $userManager;
 	}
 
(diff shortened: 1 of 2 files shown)


vinicius73:
LGTM;

Is it possible to have tests for it?

lib/Controller/UserApiController.php
```diff
@@ -10,6 +10,7 @@
 use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Share\IShare;
+use Psr\Log\LoggerInterface;
```
mejo-, line 13: Looks like a leftover to me :)

mejo- (commented):
Thanks for looking into it @luka-nextcloud! I agree with @vinicius73 that we should have tests for that.

lib/Controller/UserApiController.php
mejo-: Did you see this comment @luka-nextcloud?

mejo-: Sorry, just realized that you fixed it in a follow-up commit.

mejo- (approved):
Thanks a lot @luka-nextcloud, works as expected now :heart:

tests/unit/Controller/UserApiControllerTest.php
```diff
@@ -0,0 +1,115 @@
...
+				]
+			]);
+
+			$users = $this->userApiController->index(
+				$documentId,
+				$sessionId,
+				$sessionToken,
+				$filter
+			);
+			$this->assertInstanceOf(DataResponse::class, $users);
```
mejo-, line 105: Well, this test only tests if `$users` is a `DataResponse` object, right? It seems to miss the actual tests on the content of the response? :thinking:

mejo- (changes requested):

mejo-:
@luka-nextcloud also, please fix the code style issues reported by php-cs-fixer (see PHP linting CI job).

luka-nextcloud:
@mejo- Please check again :)

lib/Controller/UserApiController.php
```diff
@@ -44,24 +44,25 @@ public function index(int $documentId, int $sessionId, string $sessionToken, str
...
 
-			foreach ($result['users'] as ['label' => $label, 'value' => $value]) {
+			foreach ($userSearch as ['label' => $label, 'value' => $value]) {
 				if (isset($value['shareWith'])) {
 					$id = $value['shareWith'];
 					$users[$id] = $label;
 				}
 			}
-
-			$user = $this->userSession->getUser();
```
mejo-, line 63: Since you no longer use `userSession`, could you also remove it from the constructor @luka-nextcloud?

tests/unit/Controller/UserApiControllerTest.php
mejo-: Fixed myself now :)

mejo- (approved):

PR #3112, opened 2022-10-01
Fix escaping of `|` inside table cells

susnux:
### Summary

`|` is a special character inside tables and need to be escaped if used inside table cells.
Without this tables get broken if a cell contains something like `a | b`.


vinicius73:
Works!
Thanks to take care of it.

Can you prepare a test for it? To prevent eventual regression.

juliushaertl (approved):
Thanks a tons, tested and works. 👍 

A small unit test would be great :)

susnux:
@vinicius73 @juliushaertl Added a test to ensure table cells are escaped

vinicius73:
/backport 9b3ddaeeb731721a6da4760fc623fcabeee7bfbd,61df5f4fbdbb2962d15f34bb0b74af046646c324 to stable25

PR #3153, opened 2022-10-11
Improve test

vinicius73:
Remove table test from workspace.
Prevent errors from hide menu.


mejo- (approved):
Thanks a lot @vinicius73, that will make our all lifes less painful again :heart:

Issue #3121, opened 2022-10-04
Direct editing issue on 25


juliushaertl:
Seems something is broken with direct editing on iOS and Android with 25:

<img src="https://user-images.githubusercontent.com/3404133/193809195-944806f3-ce17-4050-b829-2e3b7417d70e.png" width="300" />

juliushaertl:
Seems to work through the browser but the new UI is also creating an issue there:

<img width="1680" alt="Screenshot 2022-10-04 at 13 44 11" src="https://user-images.githubusercontent.com/3404133/193811327-46273e5a-46b1-4cbd-a012-d8dd61b23591.png">

juliushaertl:
```
curl --request POST \
  --url 'http://admin:admin@nextcloud.dev.local/ocs/v2.php/apps/files/api/v1/directEditing/open?editorId=text&creatorId=textdocument&format=json' \
  --header 'Content-Type: multipart/form-data' \
  --header 'OCS-APIRequest: true' \
  --form path=/Readme1.md
```

mejo-:
Not sure whether this is fixed by #3148. Someone with a mobile device connected to their test environment would need to check.

PR #3175, opened 2022-10-12
Do not attempt to load a non-existing script

juliushaertl:
This fixes flaky loading failures when using direcrt editing, as sometimes the redirect of the not found resource is triggered too early which then leads to an updated CSRF token due to the redirect to the login page when requesting the not found style.

I think the redirect should be addresses separately and we should return a proper 404 in server for those cases, but this should fix the issues with text.

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -1,6 +1,5 @@
 <?php
 script('text', 'text-text');
-style('text', 'icons');
 ?>
 <div id="app-content">
 	<div id="maineditor"></div>


mejo-:
Nice finding :heart_eyes:

mejo-:
So this will fix #3121?

juliushaertl:
> So this will fix https://github.com/nextcloud/text/issues/3121?

Yes :)

Issue #3150, opened 2022-10-11
Enter after headline removes the text


juliushaertl:
- Open a document with a headline
- Move the cursor to the end of the headline
- Press Enter
- 💥 The line vanishes

This seems to happen only in Firefox, not Chromium

@susnux Any idea about this? I feel I've seen this happening only after https://github.com/nextcloud/text/pull/2841 so maybe related to the node view, but just a guess.

juliushaertl:
Maybe related https://github.com/ueberdosis/tiptap/issues/3200

mejo-:
It seems like @susnux already opened an issue upstream at Tiptap: https://github.com/ueberdosis/tiptap/issues/3186

susnux:
> It seems like susnux already opened an issue upstream at Tiptap: [ueberdosis/tiptap#3186](https://github.com/ueberdosis/tiptap/issues/3186)

No this is a different issue.

This issue is caused by using a `div` tag inside the `h1` / `h2`... tags which is not allowed and triggers this issue.
It can be fixed by using `span` as the content container.

But there is a different issue (at least on Firefox): Navigating using the arrow keys and pressing enter works, but clicking into the heading and pressing enter will split the heading but also copying the heading anchor link resulting in weird headings.

PR #3182, opened 2022-10-12
Fix pressing the enter key removes heading content

susnux:
* Resolves: #3150
* Target version: master 

### Summary
`div` is not an allowed child element of a `hX` tag, this leads to invalid behavior on firefox so we need to use the `span` tag as the NodeView content container.

Also using `tabindex=-1` on the NodeView will trigger weird behavior when clicking inside the heading and pressing the enter key. It will split the node but also copy the heading anchor link of the NodeView.


juliushaertl:
I figured out that the `-webkit-user-modify` seems to cause quite some side-effects, so I pushed a fix to revert this change.

juliushaertl:
Regarding cypress it seems that the way we clear the content editable is not very reliable in combination with the heading node view, so I moved the failing tests over to start with an empty file by default to avoid that. Something we should investigate separately in a follow up to maybe unify that logic or find a more stable way to clear the content.

juliushaertl (approved):

juliushaertl:
File as follow up issues:
- [ ] See if we need a different approach for https://github.com/nextcloud/text/pull/2712 or if we can live with that
- [ ] Unify cypress test setup to start with empty files
- [ ] Check if there is a better way to clear the content editable within cypress

max-nextcloud (approved):

juliushaertl:
Thanks a lot @susnux for the quick fix. Your contributions are highly appreciated. ❤️ 

On a separate note, we have a few public chat rooms on our Nextcloud instance, one also for text development, to stay in closer contact with community contributors and I thought it would also be cool to have you there. In case you are interested, just drop me a mail with your email address, then I can send you an invite.

PR #3189, opened 2022-10-13
fix: keep editing while editor scrolls down

max-nextcloud:
Cypress considers the editor hidden when it starts scrolling. While typing this is fine.
But when a new type command starts cypress checks again and thinks the editor is hidden.

Join the two edit commands as a quick fix.

* Resolves: [ci failures](https://github.com/nextcloud/text/actions/runs/3239797387/jobs/5309623756) like in #3188 
* Target version: master

### DIFF
--- a/cypress/e2e/sections.spec.js
+++ b/cypress/e2e/sections.spec.js
@@ -93,8 +93,7 @@ describe('Content Sections', () => {
 			// Insert content above link
 			cy.getContent()
 				.type('{moveToStart}\n{moveToStart}# top \n')
-				.type('lorem ipsum \n'.repeat(25))
-				.type('{moveToEnd}\n')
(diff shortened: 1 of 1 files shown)


juliushaertl:
Still seems flaky:

https://dashboard.cypress.io/projects/hx9gqy/runs/6820/test-results/36130dcb-1cf0-4289-9984-0781181d1f39

juliushaertl:
Ah seems a different thing, I guess then we should scroll the top link into view before clicking:

https://github.com/nextcloud/text/pull/3189/files#diff-8e8d87b74c5ec8c1a4001543ef5d641a25c99122adbe8ad64be0c534a82237f2L102

max-nextcloud:
I think i also saw this happening locally. Will investigate - but this is not a fix.
My impression was that either no link was created or it was added before all the 'lorem ipsum's

PR #3193, opened 2022-10-13
Cancel other jobs for the same ref

juliushaertl:
Saves us some CI time as with this only one cypress job will be running for a pull request, still pending ones from previous pushes will be canceled.


max-nextcloud (approved):
Nice... did not know this was possible.

PR #3236, opened 2022-10-14
Pin vue version for stable branches

juliushaertl:
stable24, stable23 should stay on vue 2.6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,10 +80,10 @@
         "slug": "^8.2.2",
         "tippy.js": "^6.3.7",
         "uuid": "^9.0.0",
-        "vue": "^2.7.13",
+        "vue": "^2.7.13 <3",
         "vue-click-outside": "^1.1.0",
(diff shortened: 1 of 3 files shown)


mejo- (commented):
I wonder whether it would be a better approach to pin the version in `package.json` on the respective branches.

e.g. on `stable23`:

```json
"dependencies": {
    "vue": "^2.6"
},
"devDependencies": {
    "@nextcloud/webpack-vue-config": "^4.1"
}
```

or if we want to have particular versions in `package.json` (as renovate does it at the moment):

```json
"dependencies": {
    "vue": "^2.6.14 <2.7"
},
"devDependencies": {
    "@nextcloud/webpack-vue-config": "^4.1.0 <4.2"
}
```

juliushaertl:
Then I guess for stable branches we should stick to the "in-range-only" range strategy https://docs.renovatebot.com/configuration-options/#rangestrategy

vinicius73 (approved):

Issue #3110, opened 2022-10-01
Mentions cannot be copied and pasted


max-nextcloud:
**Describe the bug**
If you copy a text containing a mention and insert it again it will just be plain text.

**To Reproduce**
1. Open a .md text file
2. Insert a line with mentions
3. Copy and past the line to a different part of the document
4. See error - pasted mention is only plain text.

**Expected behavior**
Mentions should be inserted the way they look before the copy & paste

### Server

Nextcloud 25 RC 2 on cloud.nextcloud.com

PR #3265, opened 2022-10-17
Fix copy and paste of user mentions

mejo-:
Fixes: #3110

* Resolves: # 
* Target version: master


juliushaertl (approved):

vinicius73 (approved):

vinicius73:
@mejo- check if it solves this CI https://github.com/nextcloud/text/pull/3266

max-nextcloud:
The tests are failing because markdownit will still produce html output with `data-type="user"`. It takes the type argument from the first path element inside the `mention://user/...` url.

I named the tiptap extension `user` so it's consistent with the that.

mejo-:
@max-nextcloud unfortunately your approach leads to other undesired sideeffects. I went with a different solution now. Also pushed some further minor improvements to user mentions along the way.

@vinicius73 @max-nextcloud @juliushaertl @luka-nextcloud would be nice if you could do a second review :grin:

max-nextcloud (approved):
Looks good to me.
Didn't test though.

mejo-:
/backport ee9c9417bfff65b1b5b82dc2ddf384f664dddbba,492e14ec3e78bd4565c2b9d9f23cef82120767bc,5aad842b0c1677d8ff28291a3b57135ed4d4ed1c,5bca96084e39fbd536b9ddfdee63087d0de589d7 to stable25

max-nextcloud:
@mejo- I tried it and it worked. Only the tab completion did not work - it focussed the sidebar / the close modal button. I think it might be missing a focustrap if it should work to autocomplete the mention.

mejo-:
> Only the tab completion did not work - it focussed the sidebar / the close modal button. I think it might be missing a focustrap if it should work to autocomplete the mention.

Interesting. For me, tab completion works now - both for emojis and for user mentions. I tested on latest master (server and text) :thinking:

max-nextcloud:
> Interesting. For me, tab completion works now - both for emojis and for user mentions. I tested on latest master (server and text) thinking

Interesting - in both cases tab is moving to the modal controls for me - both in firefox and chromium. I updated server, viewer and circles to the current state of master.

max-nextcloud:
> Interesting. For me, tab completion works now - both for emojis and for user mentions. I tested on latest master (server and text) thinking

Interesting - in both cases tab is moving to the modal controls for me - both in firefox and chromium. I updated server, viewer and circles to the current state of master.

Issue #2873, opened 2022-09-08
Attachments are formatted together with the next paragraph


max-nextcloud:
**Describe the bug**
When changing the format below an attachment it also affects the attachment itself.

**To Reproduce**
1. Open a text file
2. Upload an attachment
3. type somethign below
4. change format below - heading, list, todo list

**Expected behavior**
The attachment stays the way it is. Formatting only affects the section below

**Screencast**
[Bildschirmaufzeichnung vom 08.09.2022, 11:23:23.webm](https://user-images.githubusercontent.com/97337118/189086210-6bb44713-b605-495f-a5db-420784778032.webm)

max-nextcloud:
I suspect attachments are inline elements and therefor form part of the next paragraph. I think they should be block elements in their own right.

max-nextcloud:
@eneiluj Do you have an idea about this?

vinicius73:
Seems it is related to #2868

mejo-:
The reason is that we intentionally [set `inline: true` for image nodes](https://github.com/nextcloud/text/blob/b901530e68e32756faabde9e2b3d0dc30823d9f4/src/extensions/RichText.js#L86). Not sure whether that's a good thing though, given that images (and attachments) are complex structures with nested `div` elements anyway in our [`ImageView.vue`](https://github.com/nextcloud/text/blob/b901530e68e32756faabde9e2b3d0dc30823d9f4/src/nodes/ImageView.vue).

`inline: true` in the Image node configuration got introduced by the upgrade to TipTap2, specificially in this commit: edfbb70f1578dad76ff869b9f2ca0f3c027232d2. The commit message says:

> Tiptap v1 had inline images.
> v2 keeps them outside of paragraphs by default.
> Configure Image node to use inline images as markdownit creates inline images right now.

I'm not sure about the implications of switching to `inline: false` for image nodes (default in TipTap v2). A quick test at least fixed this bug :wink:

@max-nextcloud, @juliushaertl and @vinicius73 curious about your thoughts :blush:

vinicius73:
As we are using _complex_ render strategies to render attachments, it can't be handled inline mode.

mejo-:
> As we are using complex render strategies to render attachments, it can't be handled inline mode.

I agree. I wonder though, what implications switching from inline to block mode has. Given that markdown-it creates inline images, would that result in modified markdown source code after opening and closing documents with images?

PR #3282, opened 2022-10-19
Make images block nodes

mejo-:
Use markdown-it-image-figures to render standalone images inside
`<figure>` elements instead of `<p>` when markdown source gets parsed
into HTML. This allows to treat images as block elements by TipTap.

Also keep support for inline images to not break markdown files that
have inline images.

* Resolves: #2873
* Target version: master


mejo-:
Mh, seems like inserting an image as first element in a new document is broken now :cry:

src/nodes/ImageView.vue
```diff
@@ -391,6 +424,12 @@ export default {
 
 .image__main {
 	max-height: calc(100vh - 50px - 50px);
+
+	&.image__inline {
+		// Make inline images fit into a text line
+		max-height: 40px;
```
juliushaertl, line 430:
```suggestion
		max-height: 1em;
```

How about that to make that a bit adjusted to the line height?

mejo-: Mh, I agree that it should a relative value would be better, but `1em` is way too small for most pictures. Maybe `2em`?

That would be an inline image with `max-height: 2em`:

2022-10-24T15:01:07,662887695+02:00

mejo-:
So @julien-nc and me spent some three more hours debugging the main remaining issue with this PR: inserting a (block node) image as first element into an empty editor breaks due to us doing `this.$editor.chain().setImage().insertContent('<br />').focus().run()` [here](https://github.com/nextcloud/text/blob/bc216160cf55063cd586bc960aecef928d2b76a0/src/components/Editor/MediaHandler.vue#L197).

It turns out to be a Tiptap (or maybe even ProseMirror) bug. I reported https://github.com/ueberdosis/tiptap/issues/3355 upstream.

juliushaertl:
Can you quickly summarize the remaining issue, so we can file this as a follow up?

mejo-:
> Can you quickly summarize the remaining issue, so we can file this as a follow up?

Sure, sorry for not doing so in the first place :see_no_evil:

So the fix involves moving the cursor two steps further after inserting the image and before inserting the hard break (if the image is not inserted as first element, which needs other special treatment). This is necessary because per default the cursor is *in front of* the image after inserting it, but we want to add the hard break afterwards.

If you insert an image in between text, this results in the linebreak being inserted two characters further for the moment. I think that's an acceptable tradeoff given that images are block nodes now and most likely it will not happen that often that people insert them in the middle of text anyway.

Here's the relevant code:

https://github.com/nextcloud/text/blob/02dc2ba6837b705b61d214d31140aad8bbc0e546/src/components/Editor/MediaHandler.vue#L199-L209

If you have a quick idea how to detect that the cursor (i.e. the focus, `this.$editor.view.state.selection`) is in the middle of some text and *not* at the end of a paragraph, we could add this as another special handling. But instead of waiting longer and spending another bunch of hours on this minor issue, my suggestion would be to merge this PR as is now :wink:

max-nextcloud (approved):

max-nextcloud:
@mejo- There's still a [test failing](https://github.com/nextcloud/text/actions/runs/3461054327/jobs/5778205072).

I'm not sure what to make of this. I think we are adding two new lines on purpose in the `toMarkdown` function of the block image node.

Is this really required. I think it's unfortunate if we change the markdown syntax without any changes being made to the document. Why are those newlines needed? Can't we say every line that only contains an image is considered a block image even if no empty lines follow?

Ahh... i guess markdown parsers will then think the line is part of  a larger paragraph... Hmm... difficult.

max-nextcloud (changes requested):
We need to at least adjust the test - but i think it would be better if we could aviod adding the newlines.

max-nextcloud:
I also played with using the gapcursor instead of the addition `<br/>` some. So far worked pretty well for me.

mejo-:
> Ahh... i guess markdown parsers will then think the line is part of a larger paragraph... Hmm... difficult.

Exactly. The easiest fix would be to adjust our tests and just add the two newlines to the argument given to `toBe()` in the " Markdown though editor › images" test.

I see though that it's all kinda hacky. Still, I don't see another option because we

* Want to treat images added by us as block nodes, both in Tiptap and in serialized markdown.
* Want to keep inline images that were added by some thirdparty editor as inline images both in Tiptap and in serialized markdown.

max-nextcloud:
> 



> I see though that it's all kinda hacky. Still, I don't see another option because we
> 
>     * Want to treat images added by us as block nodes, both in Tiptap and in serialized markdown.
> 
>     * Want to keep inline images that were added by some thirdparty editor as inline images both in Tiptap and in serialized markdown.

Well... the markdown in question already parsed as a block image before. So theoretically it's not necessary to add an empty line after it. However I think we have the same issue with tables at the end of the doc and other block elements. So we can move this to a separate issue and carry on.

src/nodes/ImageInline.js
```diff
@@ -0,0 +1,74 @@
...
+	},
+
+	// Empty input rules, we want only those from (block) Image extension
+	addInputRules() {
+		return []
+	},
+
+	addNodeView() {
+		return VueNodeViewRenderer(ImageView)
+	},
```
max-nextcloud, line 66: Somehow the image does not get rendered for me.
I'll add a failing test.

max-nextcloud (changes requested):

src/nodes/ImageInline.js
max-nextcloud: Okay... turns out i did not have the file in place. But there also was no placeholder icon. We might want to add that.

max-nextcloud (commented):
The main thing i am testing here and i am somewhat worried about is migration of existing .md files.
If you add an image to a .md file now... or  a bunch of attachments for that matter they will most likely not live on their own inside a paragraph:
* write some text, add one enter
* upload an image
* write some more text

results in markdown in this form right now:

```md
Write some text.

![Bildschirmfoto vom 2022-11-14 19-14-11.png](.attachments.6740564/Bildschirmfoto%20vom%202022-11-14%2019-14-11.png)  
Write some more text.
```
Since there is no line break between the image and the following text this will be interpreted as an inline image.

Same goes for uploading multiple attachments at the same time.
Here's the result:

```md
![image between lines(1).md](.attachments.6740564/image%20between%20lines%281%29.md)
![image between lines(2).md](.attachments.6740564/image%20between%20lines%282%29.md)  
![image between lines(3).md](.attachments.6740564/image%20between%20lines%283%29.md)  
![image between lines.md](.attachments.6740564/image%20between%20lines.md)
```

I see some problems here:
* inline images render very differently from what they used to
* it's hard to convert an inline image to a separate image because it will just live in a paragraph of it's own happily without turning into a block image unless the editor session ends
* inline images do not perform the file resolution magic that block images use to resolve attachments etc.

I'm really not sure how to move on with this. We can create follow up issues or try and address this now. Turns out this is quite a complex problem.

max-nextcloud:
Here's a bit of an illustration of what i am referring to:

| before | after|
|----------|-------|
|Bildschirmfoto vom 2022-11-15 07-22-16 | Bildschirmfoto vom 2022-11-15 07-23-51 |
|Bildschirmfoto vom 2022-11-15 07-21-53 ||

max-nextcloud:
I think i will exclude [2f7ff70](https://github.com/nextcloud/text/pull/3282/commits/2f7ff70e525ba8f11b8a7988fa083782c2632a90) for now so images still render the way they used to even if inline.

max-nextcloud (approved):
I think this can be merged now. I created issues for the remaining things and even if they do not get fixed this is an improvement over the status quo.

max-nextcloud:
@juliushaertl @mejo- I'm not sure if we should backport this.

Right now in 25 images and attachments are often in the same paragraph as the following content. Turning that content into say a heading then also toggles the image as it's part of that paragraph. I think that's pretty bad. At the same time this change is pretty profound and we have not really tested it so much. It will also require a bunch of follow up changes. I don't think they would need to be backported - but still.

I wonder if there would be a less intrusive fix we could apply to stable25.

mejo-:
/backport 0c496faff4e5081a13a8e6e1e7d09cae5a45032e,4b63cb684897f5512d86ed00cf290b3fffa1df43,5ee1d6b7c6e9edc03222ae470c456209e90b1a7b,59b2b68337bda059a71fbf81e2748f6d7c4d647b,47a203748157f922c17eb297d495d0b247998df0,f4b660d893ca3471310ac65a0a0ea9a475913ea9,8f9a6c4d42814d20916c8bec0a4f916b282a982b to stable25

Issue #1525, opened 2021-03-15
Toggeling checkboxes is still possible during idle timeout


scaredmushroom:
**Describe the bug**
When the session timed out due to inactivity, it is still possible to toggle the checkboxes of checklists.
Because it will not sync these changes users get confused and their changes get lost on resync.

**To Reproduce**
1. Open a text file
2. Create a checklist
3. Wait until idle timeout
4. Toggle checkboxes

**Expected behavior**
On timeout the checkboxes should be read-only.

marcelklehr:
Confirmed with Nextcloud 25

PR #3288, opened 2022-10-21
TaskItem: Don't allow toggling while editor is readonly

marcelklehr:
* Resolves: #1525
* Target version: master 

### Summary
TaskItem: Don't allow toggling while editor is readonly

My first text PR :v:


juliushaertl:
Maybe the input could also be set to disabled in the renderHTML function?

marcelklehr:
I wanted to do that, but I think we don't get the editor object passed in the renderHTML function to know whether we're readonly, do we?

juliushaertl:
I think it should be available in the current object scope as `this.editor`, but not 100% sure.

marcelklehr:
image

mejo-:
> I wanted to do that, but I think we don't get the editor object passed in the renderHTML function to know whether we're readonly, do we?

If I didn't oversee anything, `this.editor` should be available in the scope of the Tiptap Node/Extension and so it should be in `renderHTML()`. So you should be able to use `this.editor.options.editable` there, no?

marcelklehr:
Debugger says no. :D

mejo- (approved):
So in the end your approach is probably the best @marcelklehr.

I quickly tried the approach to implement a `isReadonly` flag in storage of the extension. While this works well, setting the checkbox to disabled in `renderHTML()` brings other problems and doesn't work as expected since `handleClick()` alters the `checked` attribute without respecting the `disabled` attribute anyway. So we would have to change the logic of `handleClick()`, which is probably not worth the effort.

juliushaertl (approved):

Issue #579, opened 2020-01-16
Link open behavior is buggy in chrome


juliushaertl:
When opening a link on a public share, it will be opened twice in read-only mode.

jospoortvliet:
yay bumped into this today 😭

PR #3290, opened 2022-10-21
Fix link opening in readonly mode

marcelklehr:
* Resolves: #579
* Target version: master 

### Summary

In readonly mode it seems prosemirror doesn't register a click handler override, which is why link clicks go right through to the browser handler. This commit registers a clickPreventer plugin that preventDefault's clicks when the editor is readonly


src/marks/Link.js
```diff
@@ -60,6 +60,8 @@ const Link = TipTapLink.extend({
 		...mark.attrs,
 		href: domHref(mark),
 		rel: 'noopener noreferrer nofollow',
+		// Disable default browser opening handler (only relevant in readonly mode)
+		onclick: "return false;"
```
juliushaertl, line 64: Is this still needed with the clickPreventer plugin? I'm wondering if this might be an issue with our CSP if that is considered inline javascript

marcelklehr: ah, thanks. fixed it

mejo- (approved):
Works as expected and looks sensible to me codewise.

Issue #3209, opened 2022-09-08
Rich workspaces: If there is no Readme.md, don’t show editor placeholder but move into "+" menu


jancborchardt:
Currently, every folder has the empty "Add notes, lists or links …" on the top. This leads to various issues such as:
- Always uses a lot of vertical space, for every single folder
- People can accidentally click into it
- People end up finding the feature annoying and disabling it completely even though they might find it useful on a folder-by-folder basis

Instead, we should only show it directly on top of the file list if a Readme.md is present. If not, we show an entry "Add description" at the bottom of the "+ New" menu. (This entry of course does not need to show if a Readme.md is present.)

Before | After (if a Readme.md is present, of course still show it directly)
-|-
![image](https://user-images.githubusercontent.com/925062/189092679-ee99ff79-0d60-4c7e-aa44-fc6f3b6c7912.png)|![image](https://user-images.githubusercontent.com/925062/189093430-0a54fcda-63ed-4f8a-b89d-86d5cfe98508.png)

cc @eneiluj @mejo- @max-nextcloud @juliushaertl 

As discussed @karlitschek

juliushaertl:
@jancborchardt If you can provide a example files Readme text, we can add that :)

Besides that should all be doable, but moving to 26 as we have final freeze for 25 today.

juliushaertl:
- [ ] Register new file menu entry to add the description
  - [ ] Check if we can show/hide it depending on the current folder status. If not this may need some extension of the server api for the new files menu
  - [ ] Only show the rich workspace area if a file is present
- [ ] Rename from ui parts from rich workspace to description

FYI @tobiasKaminsky to align this on the Android and iOS apps as well.

jancborchardt:
> @jancborchardt If you can provide a example files Readme text, we can add that :)

@juliushaertl how about this:

> # Welcome to Nextcloud! 📱☁💻
>
> Here you can add a description or any other info relevant for the folder. It will show as a "Readme.md" and in the web interface also embedded nicely up at the top.

dartcafe:
Isn't it simply just a folder description? Common users may have a problem in understanding, what a "rich workplace" may be. It needs to be explained. Simply naming it to "Folder Description" seems to be more understandable and needs no further explanation.

Tummerhore:
[...]

PR #3291, opened 2022-10-21
Update rich workspace visibility

luka-nextcloud:

* Depends on https://github.com/nextcloud/server/pull/35474
* Resolves: #3209
* Target version: master


src/views/RichWorkspace.vue
```diff
@@ -22,17 +22,6 @@
 
 <template>
 	<div v-if="enabled" id="rich-workspace" :class="{'icon-loading': !loaded || !ready, 'focus': focus, 'dark': darkTheme, 'creatable': canCreate, 'empty': showEmptyWorkspace}">
```
juliushaertl, line 24: We probably don't need the empty class anymore and can drop showEmptyWorkspace

src/views/RichWorkspace.vue
```diff
@@ -221,6 +190,72 @@ export default {
...
+											.createFile('Readme.md', { scrollTo: false, animate: false })
+											.then(() => {
+												menu.removeMenuEntry('rich-workspace-init')
+												return self.getFileInfo()
+											})
+									},
+								})
+							},
+						}
+						OC.Plugins.register('OCA.Files.NewFileMenu', newRichWorkspaceFileMenuPlugin)
```
juliushaertl, line 249: I think this does not work when changing directories as the plugin would be registered each time or the folder info would only be fetched for the first one. 

I think the best would be to move this to https://github.com/nextcloud/text/blob/81d7af739e8c365793888aaa40105ac03f79843e/src/files.js#L60

juliushaertl: Show dynamically showing or hiding we can probably use the same events as in https://github.com/nextcloud/text/blob/e286194c690c15c5d66b3a7198ee76b9366f1201/src/helpers/files.js#L177-L182

khoipv-1315: I have a question, how can I hide the menu?
- If using Document Select, it won't find the menu because it doesn't show the menu yet if it doesn't click the menu button
- Is there a way to add a Menu item using FileList? I only know how to use OCA.Files.NewFileMenu

src/helpers/files.js
```diff
@@ -139,14 +139,48 @@ const registerFileActionFallback = () => {
...
+			fileType: 'file',
+			actionHandler() {
+				return window.FileList
+					.createFile('Readme.md', { scrollTo: false, animate: false })
+					.then(() => {
+						menu.removeMenuEntry('rich-workspace-init')
+					})
+			},
+			shouldShow() {
+				const descriptionFile = 'Readme.md'
```
juliushaertl, line 165:
```suggestion
				const descriptionFile = t('text', 'Readme') + '.' + loadState('text', 'default_file_extension'),
```

src/helpers/files.js
```diff
@@ -139,14 +139,48 @@ const registerFileActionFallback = () => {
...
+		if (fileList.id !== 'files' && fileList.id !== 'files.public') {
+			return
+		}
+
+		// register the new menu entry
+		menu.addMenuEntry({
+			id: 'rich-workspace-init',
+			displayName: t('text', 'Add description'),
+			templateName: t('text', 'Readme') + '.' + loadState('text', 'default_file_extension'),
+			iconClass: 'icon-filetype-text',
```
juliushaertl, line 155:
```suggestion
			iconClass: 'icon-rename',
```

To match the design mockup and have distinguishable icons:

Screenshot 2022-12-08 at 14 27 06

src/helpers/files.js
```diff
@@ -139,14 +139,48 @@ const registerFileActionFallback = () => {
 
 }
 
+const newRichWorkspaceFileMenuPlugin = {
+	attach(menu) {
+		const fileList = menu.fileList
+		// only attach to main file list, public view is not supported yet
+		if (fileList.id !== 'files' && fileList.id !== 'files.public') {
```
juliushaertl, line 146: Public file lists could still be supported. I think we also cover that with out existing implementation, but needs double checking.

src/helpers/files.js
```diff
@@ -171,22 +206,24 @@ const FilesWorkspacePlugin = {
...
 				vm.active = OCA.Files.App.getCurrentFileList() === fileList
 			})
-
 			fileList.$el.on('urlChanged', data => {
 				vm.path = data.dir.toString()
 			})
 			fileList.$el.on('changeDirectory', data => {
 				vm.path = data.dir.toString()
 			})
+			fileList.$table.on('DOMSubtreeModified', () => {
```
juliushaertl, line 224:
```suggestion
			fileList.$table.on('afterChangeDirectory', () => {
```

I think this should also work and is only triggered when the list of files is actually updated

src/helpers/files.js
```diff
@@ -160,9 +194,10 @@ const FilesWorkspacePlugin = {
 		if (fileList.id !== 'files' && fileList.id !== 'files.public') {
 			return
 		}
-
+		addMenuRichWorkspace()
```
juliushaertl, line 197: I'd say we can just move the `menu.addMenuEntry` call in here and avoid the two plugins being nested to just have one for the rich workspace.

src/helpers/files.js
```diff
@@ -171,22 +206,24 @@ const FilesWorkspacePlugin = {
...
 			})
-
 			fileList.$el.on('urlChanged', data => {
 				vm.path = data.dir.toString()
 			})
 			fileList.$el.on('changeDirectory', data => {
 				vm.path = data.dir.toString()
 			})
+			fileList.$table.on('DOMSubtreeModified', () => {
+				vm.hasDescriptionFile = !!fileList.findFile(descriptionFile)
```
juliushaertl, line 225: I'd prefer to have an event that is triggered through https://github.com/nextcloud/nextcloud-event-bus on the newFileAction after creating the file. The RIchWorkspace.vue component could listen to that and reuse the existing way to check for a readme file through the backend https://github.com/nextcloud/text/blob/master/src/views/RichWorkspace.vue#LL145

src/views/RichWorkspace.vue
```diff
@@ -256,42 +225,42 @@ export default {
 		color: var(--color-text-maxcontrast);
 	}
 
-	#rich-workspace:deep(div[contenteditable=false]){
+	#rich-workspace div[contenteditable=false] {
```
juliushaertl, line 228: What is the reason for that change? Should not be needed

src/helpers/files.js
luka-nextcloud: How do you suggest that we proceed here?

src/helpers/files.js
luka-nextcloud: This event is not triggered when we add the README.md file to the home directory. Do you have any alternative way?

src/helpers/files.js
juliushaertl: Noticed that we also exclude this in the existing plugin so let's skip that for now.

src/helpers/files.js
juliushaertl: That should be fine if we rely on the existing mechanism to check for a readme existence in the component as noted below https://github.com/nextcloud/text/pull/3291#discussion_r1043361054

src/helpers/files.js
luka-nextcloud: I definitely agree with that. There is a problem though. When I tried, I noticed that the UI takes a long time before updating the rich workspace block. I guess because of the event-bus.

juliushaertl: It should take the same amount of time as it is currently. There is an extra request but that should not change from how it is right now. Event bus in general should be fast ;)

src/views/RichWorkspace.vue
luka-nextcloud: I noticed that they were not applied to the editor and I was not able to work with the editor normally.

src/helpers/files.js
luka-nextcloud: The problem is `DOMSubtreeModified` is triggered almost all the time so it's always triggering a new event (either `Text::showRichWorkspace` or `Text::hideRichWorkspace`)

juliushaertl: Yes, I think that is triggered way too often, therefore I'd go with the above mentioned event bus approach.

src/views/RichWorkspace.vue
juliushaertl: Can you maybe check if that works after running `npm ci` I think this might only be the case if you have some outdated dependencies.

juliushaertl: If that is still an issue we should still address it in a separate PR I think.

luka-nextcloud: Ok, let me check after npm ci.

src/helpers/files.js
luka-nextcloud: The event bus approach is already implemented. The actual problem is the right event to use. We're using `DOMSubtreeModified` right now.

src/helpers/files.js
```diff
@@ -139,14 +139,46 @@ const registerFileActionFallback = () => {
...
+		}
+
+		// register the new menu entry
+		menu.addMenuEntry({
+			id: 'rich-workspace-init',
+			displayName: t('text', 'Add description'),
+			templateName: descriptionFile,
+			iconClass: 'icon-rename',
+			fileType: 'file',
+			useInput: descriptionFile,
```
juliushaertl, line 158:
```suggestion
			useInput: false,
```

To make sure that the input is not shown

src/helpers/files.js
luka-nextcloud: @juliushaertl Is there any event we can listen to when the README.md file is deleted?

src/views/RichWorkspace.vue
luka-nextcloud: @juliushaertl It actually doesn't work.

src/helpers/files.js
luka-nextcloud: @juliushaertl Could you please check again?

src/views/RichWorkspace.vue
luka-nextcloud: > If that is still an issue we should still address it in a separate PR I think.

I will remove the changes from the current PR and open a new one for the fix.

luka-nextcloud: Removed the changes for now.

src/helpers/files.js
juliushaertl: Nothing yet. We could skip this for now as I know there is some ongoing work in the files2vue migration.

Otherwise a server PR with triggering an event around here would be the way to go I think:

https://github.com/nextcloud/server/blob/5b9a8f0407a9e3fe7e00f0fc9284ea986905f1b5/apps/files/js/filelist.js#L2494

juliushaertl: Something like: `this.$el.trigger(jQuery.Event('removeFile', path));`

juliushaertl:
@luka-nextcloud Server pr was merged, so rebasing should run cypress against the latest server master branch.

cypress/e2e/workspace.spec.js
```diff
@@ -62,27 +62,33 @@ describe('Workspace', function() {
...
 			.should('be.visible')
-			.should('contain', 'Hello')
 	})
 
 	it('formats text', function() {
 		cy.visit(`apps/files?dir=/${encodeURIComponent(currentFolder)}`)
 		cy.openWorkspace()
-			.type('Format me')
-			.type('{selectall}')
+			.type('Format me', { force: true })
```
juliushaertl, line 90: Would be good to have a way without force. Maybe we need to focus the workspace and wait for the menubar to appear first.

cypress/e2e/workspace.spec.js
```diff
@@ -104,8 +110,8 @@ describe('Workspace', function() {
 	it('creates headings via submenu', function() {
 		cy.visit(`apps/files?dir=/${encodeURIComponent(currentFolder)}`)
 		cy.openWorkspace()
-			.type('Heading')
-			.type('{selectall}')
+			.type('Heading', { force: true })
```
juliushaertl, line 113: Same as above.

cypress/support/commands.js
```diff
@@ -301,7 +301,7 @@ Cypress.Commands.add('getMenuEntry', (name) => {
 Cypress.Commands.add('getSubmenuEntry', { prevSubject: 'optional' }, (subject, parent, name) => {
 	return (subject ? cy.wrap(subject) : cy.getMenu())
 		.getActionEntry(parent)
-		.click()
+		.click({force: true})
```
juliushaertl, line 304: Seems unrelated

cypress/support/commands.js
```diff
@@ -337,8 +337,9 @@ Cypress.Commands.add('clearContent', () => {
 })
 
 Cypress.Commands.add('openWorkspace', () => {
-	cy.get('#rich-workspace .empty-workspace').click()
-	cy.getEditor().find('[data-text-el="editor-content-wrapper"]').click()
+	cy.createDescription()
+	cy.get('#rich-workspace .editor__content').click({force: true})
```
juliushaertl, line 341: This looks like the right place to wait until the editor content is visible. instead of forcing actions all over the place ;)

src/helpers/files.js
```diff
@@ -139,14 +139,47 @@ const registerFileActionFallback = () => {
...
+			fileType: 'file',
+			useInput: false,
+			actionHandler() {
+				return window.FileList
+					.createFile(descriptionFile, { scrollTo: false, animate: false })
+					.then(() => emit('Text::showRichWorkspace'))
+			},
+			shouldShow() {
+				if (fileList.findFile(descriptionFile)) return false
+				return true
```
juliushaertl, line 166:
```suggestion
				return !fileList.findFile(descriptionFile)
```

src/helpers/files.js
```diff
@@ -139,14 +139,47 @@ const registerFileActionFallback = () => {
 
 }
 
+const newRichWorkspaceFileMenuPlugin = {
+	attach(menu) {
```
juliushaertl, line 143: Can we merge this into the existing FilesWorkspacePlugin below? There should be no need for two separate plugins

juliushaertl (changes requested):

juliushaertl:
@luka-nextcloud Cypress tests seem good now, can you remove the remaining force parts in the spec files and fix the eslint errors?

src/helpers/files.js
luka-nextcloud: The plugin below gets the FileList as argument for the attach method but we need the menu instead.

juliushaertl: Right ... sorry I mixed that up then 👍

juliushaertl (approved):

Issue #1091, opened 2020-10-17
Argument 1 passed to OCP\\AppFramework\\Db\\Entity::fromRow() must be of the type array, null given


solracsf:
- NC 19.0.4
- PHP 7.4.11
- Text 3.0.1

```
{
  "reqId":"s1ZLDjEcvER3Vy9EkZv1",
  "level":3,
  "time":"2020-10-17T14:38:00+02:00",
  "app":"index",
  "method":"POST",
  "url":"/apps/text/session/sync",
  "message":{
    "Exception":"Exception",
    "Message":"Argument 1 passed to OCP\\AppFramework\\Db\\Entity::fromRow() must be of the type array, null given, called in /apps/text/lib/Service/SessionService.php on line 155",
    "Code":0,
    "Trace":[
      {
        "file":"/lib/private/AppFramework/App.php",
        "line":137,
        "function":"dispatch",
        "class":"OC\\AppFramework\\Http\\Dispatcher",
        "type":"->"
      },
      {
        "file":"/lib/private/AppFramework/Routing/RouteActionHandler.php",
        "line":47,
        "function":"main",
        "class":"OC\\AppFramework\\App",
        "type":"::"
      },
      {
        "function":"__invoke",
        "class":"OC\\AppFramework\\Routing\\RouteActionHandler",
        "type":"->"
      },
      {
        "file":"/lib/private/Route/Router.php",
        "line":297,
        "function":"call_user_func"
      },
      {
        "file":"/lib/base.php",
        "line":1011,
        "function":"match",
        "class":"OC\\Route\\Router",
        "type":"->"
      },
      {
        "file":"/index.php",
        "line":37,
        "function":"handleRequest",
        "class":"OC",
        "type":"::"
      }
    ],
    "File":"/lib/private/AppFramework/Http/Dispatcher.php",
    "Line":110,
    "Previous":{
      "Exception":"TypeError",
      "Message":"Argument 1 passed to OCP\\AppFramework\\Db\\Entity::fromRow() must be of the type array, null given, called in /apps/text/lib/Service/SessionService.php on line 155",
      "Code":0,
      "Trace":[
        {
          "file":"/apps/text/lib/Service/SessionService.php",
          "line":155,
          "function":"fromRow",
          "class":"OCP\\AppFramework\\Db\\Entity",
          "type":"::"
        },
        {
          "file":"/apps/text/lib/Service/SessionService.php",
          "line":176,
          "function":"getSession",
[...]

PR #3312, opened 2022-10-26
Prevent caching `false` in SessionService

Raudius:
* Resolves: https://github.com/nextcloud/text/issues/1091
* Target version: master 

### Summary
If we receive non-serializable data from the session mapper (e.g. `NULL`) we should not try to cache it. Trying to cache it results in value `false` getting cached which later fails to be decoded.


mejo- (approved):
Looks good codewise :)

I didn't test it.

juliushaertl (approved):

Issue #2344, opened 2022-04-27
empty lines added to the end of code <> blocks on exit


noseshimself:
### Steps to reproduce
1. open .md file containing code blocks
2. type something somewhere
3. close it
4. open it again

### Expected behaviour
Code blocks look like before

### Actual behaviour
An empty line may be added to the end of the code block.

## Server configuration detail

**Operating system:** Linux 5.10.0-12-amd64 #1 SMP Debian 5.10.103-1 (2022-03-07) x86_64

**Webserver:** Apache/2.4.52 (Debian) (apache2handler)

**Database:** mysql 10.7.1

**PHP version:** 

8.0.15
Modules loaded: Core, date, libxml, openssl, pcre, sqlite3, zlib, ctype, curl, dom, fileinfo, filter, ftp, hash, iconv, json, mbstring, SPL, PDO, pdo_sqlite, session, posix, Reflection, standard, SimpleXML, Phar, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, apache2handler, apcu, bcmath, exif, gd, gmp, imagick, intl, ldap, memcached, pcntl, pdo_mysql, pdo_pgsql, redis, sodium, zip, Zend OPcache

**Nextcloud version:** 23.0.0 - 23.0.0.10

**Updated from an older Nextcloud/ownCloud or fresh install:** both

**Where did you install Nextcloud from:** unknown

<details><summary>Signing status</summary>

Array
(
    [secsignid] => Array
        (
            [EXTRA_FILE] => Array
                (
                    [screenshots/screenshot_onboarding.png] => Array
                        (
                            [expected] => 
                            [current] => 6b238ed855580bb32218f11bd275ea1a33a07aaf5b5477bb3cff79c7dfea6cb2d6f7661f7b16bbe26339ed3e057eadc935e83c134b300df4cc2dea15035edfbd
                        )

                    [screenshots/screenshot_access_pass.png] => Array
                        (
                            [expected] => 
                            [current] => cdce3d7aa8105cf8453ecd9d197785b48798eeda286cd86d2fe2cdfa8b69e2ac7e2d4dcecbc273e4558b556e93ff93d44fd662b8eb8f34e34dd0b811f156d559
                        )

                    [screenshots/screenshot_personal_settings.png] => Array
                        (
                            [expected] => 
[...]

PR #3321, opened 2022-10-27
Remove trailing newline from code blocks

mejo-:
Markdown-it serialized code blocks to `<pre><code>...\n</pre></code>`, which leads to trailing newlines in the code blocks as rendered by Tiptap.

Let the Tiptap CodeBlock node remove the trailing newline in parseHTML in order to avoid this.

There's another issue with trailing newlines that got added to code blocks on purpose being removed by prosemirror-markdown, but that's not affected by this change.

Fixes: https://github.com/nextcloud/text/issues/2344

* Resolves: # 
* Target version: master


juliushaertl (approved):
Nice, thanks for the fix 🥳

susnux:
This breaks the common mark tests as
````
```
code
```
````

is expected to render to
```
<pre><code>code
</code></pre>
```

But generally the last `\n` should not affect the markdown serialization as `prosemirror-markdown` checks if the last character of the content of the code block is a newline and only inserts one if it is not a new line.
So for the example above no new line would be inserted, but for `<pre><code>code</code></pre>` a newline would be inserted.

---

I noticed that sometimes a new line is added visually (but not in the markdown). I do not know why, but prosemirror adds sometimes a `<br>` tag with class `ProseMirror-trailingBreak`.

**Comment by mejo-** [comment]
> This breaks the common mark tests

You're right. I broke commonmark on purpose in this case because the linebreak before closing HTML tags seems to cause ProseMirror to add the HTML break element. But I forgot to disable/edit the respective commonmark tests.

With my changes, the newline doesn't get added anymore.

Another solution would be to find out why ProseMirror adds the break element, but to be honest for me it was an acceptable tradeoff to break commonmark because I don't see a reason to have a linebreak before the closing HTML tags, but maybe I'm missing something?

Would you prefer another solution @susnux?

**Comment by susnux** [comment]
But it does not always add the newline, and the newline is only added visually, so are you sure it does not appear anymore?
Because currently the added break is only added randomly, at least I could not figure out any pattern so it could even be something else somewhere deeper.

**Comment by mejo-** [comment]
So after some further debugging, my understanding is that [`createDocument()`](https://github.com/ueberdosis/tiptap/blob/2b69f344c713befecd4ec606df5e9ba680aa2ce8/packages/core/src/Editor.ts#L267) from Tiptap gets `<pre><code>something\n</code></pre>` for a code block (as serialized by markdown-it). The `\n` gets transformed into a `<br>` tag by prosemirror-view and I'd argue that's even expected behaviour.

I'm not sure yet why the break tag is only added sometimes by prosemirror-view, but at least this PR reproducibly fixes the issue for me.

Whenever a newline is added to the end of a code block, this is reproducible for the same document. Closing the document, resetting the Text session and opening the document again reliably adds the new line again to the same code blocks where it's been added before. And everytime, applying the PR in between removes the newline.

@susnux

**Comment by mejo-** [comment]
After digging into this again with @juliushaertl I went with another approach.

So first, it's reproducible that the newline from markdown-it results in the rendered extra newline in Tiptap. Everytime you close the editor, reset the session and open the editor again, the trailing newline is back.

So that means that we can assume the HTML as given from markdown-it always contains one trailing newline too much.

That's why I went with removing one trailing newline from the code block content in `parseHTML()` of `CodeBlock` node now.

File: src/nodes/CodeBlock.js [hunk 1]
```diff
@@ -2,6 +2,19 @@ import TiptapCodeBlock from '@tiptap/extension-code-block'
...
 
 const CodeBlock = TiptapCodeBlock.extend({
+	parseHTML() {
+		return [
+			{
+				tag: 'pre',
+				preserveWhitespace: 'full',
+				// Remove trailing newline from code blocks (Github issue #2344)
+				getContent: (node, schema) => {
+					console.debug('node', node)
```
juliushaertl, line 12:
```suggestion
```

juliushaertl (approved):
Small debugging leftover, otherwise 👍 as discussed :)

src/nodes/CodeBlock.js
mejo-: Thanks, fixed :ok_hand:

PR #3322, opened 2022-10-27
Add method to log editor content to debugger console

mejo-:
Logs editor content in three formats
* In markdown as serialized from editor state by prosemirror-markdown et al
* In HTML as serialized from markdown by markdown-it
* In HTML, as rendered in the browser by Tiptap

This should be useful for debugging issues with the different serializations happening in Nextcloud Text.

* Resolves: # 
* Target version: master


src/components/Editor.vue
```diff
@@ -490,6 +490,7 @@ export default {
 							this.$syncService.startSync()
 						},
 						onUpdate: ({ editor }) => {
+							// this.debugContent(editor)
```
juliushaertl, line 493: Very nice idea, we could always do this if the instance is in debug mode through the config.php option even (This is exposed as `window.oc_debug`)

mejo-: I fear that it might become confusing if you're debugging other stuff (not related to serialization) :wink:

juliushaertl: Also fine with me :)

mejo-: Still you have a point in making the debug function easier to discover. Maybe add a note to `README.md`?

susnux: > Very nice idea, we could always do this if the instance is in debug mode through the config.php option even (This is exposed as `window.oc_debug`)

Or simply use `logger.debug` as that variable is already checked by the logger and if not set, the message is discarded.

src/components/Editor.vue
```diff
@@ -686,6 +687,23 @@ export default {
...
+		 * Helper method to debug serialization of content between markdown and HTML
+		 *
+		 * @param {object} editor The Tiptap editor
+		 */
+		debugContent(editor) {
+			const proseMirrorMarkdown = this.$syncService.options.serialize(editor.state.doc)
+			const markdownItHtml = markdownit.render(proseMirrorMarkdown)
+
+			logger.debug('markdown, serialized from editor state by prosemirror-markdown')
+			console.debug(proseMirrorMarkdown)
```
susnux, line 701: Those console statements are not necessary, you could simply do it like
```js
logger.debug('markdown, serialized from editor state by prosemirror-markdown', { markdown: proseMirrorMarkdown })
logger.debug('HTML, serialized from markdown by markdown-it', { html: markdownItHtml })
logger.debug('HTML, as rendered in the browser by Tiptap', { html: editor.getHTML() })
```

Or you could even put it all together in one message:
```js
logger.debug('Current editor content', {
	html: editor.getHTML(),
	markdown: proseMirrorMarkdown,
	'markdown-it': markdownItHtml,
})
```

mejo-: When using `logger.debug` for logging the variables themselves, the variable value is folded, which makes it harder to read and compare while debugging. That's why I went with `console.debug()` for the variables themselves. Do you see a problem with my approach? Would you still prefer `logger.debug()`? :blush:

juliushaertl (approved):

src/components/Editor.vue
susnux: No real problem, just personal preference (it looks cleaner to me and easier to copy only one message).
But if you like it this way more, that is fine :)

susnux (approved):

Issue #3328, opened 2022-10-28
Langage removed from markdown code blocks


juliushaertl:
When opening a markdown file text will remove the language in code blocks:

Example markdown file:

<pre><code>
## Test file

```python
num = 29
print(num, "is a number")
```

This is some more content
</code></pre>

PR #3330, opened 2022-10-28
Fix code blocks loosing the language hint on serialization

susnux:
* Resolves: #3328
* Target version: master 

### Summary
`prosemirror-markdown` uses `params` attribute for the language hint whereas `tiptap` uses the `language` attribute.


max-nextcloud (approved):
Thanks for looking into this.
Fix looks right on point.
Haven't tried it myself - but since you added a test I'm confident it works. :D

Issue #2702, opened 2022-07-10
CommonMark compatibility (Tracker bug)


susnux:
### Background
Currently there are "commonmark" tests, but they only test `markdown-it` to output correct HTML for markdown (which is already tested upstream), but they are not testing our pipeline to keep the markdown valid.
Meaning validating that the input and output are at least equivalent (both result in same HTML, as the mapping is surjective but not injective).

`md ==> markdown-it ==> HTML ==> TipTap ==> prosemirror-markdown ==> md`

### Problems
I created a branch with adjusted tests, currently there are a lot of failures, which need to be triaged (real bugs, whitespace problems, test issues...), so this is a list of all bugs blocking full CommonMark support:

- [x] :lady_beetle: https://github.com/nextcloud/text/issues/2699
- [ ] :up: https://github.com/ProseMirror/prosemirror-markdown/issues/78
- [ ] :up: https://github.com/ProseMirror/prosemirror-markdown/issues/77
- [ ] :up: https://github.com/nextcloud/text/issues/969 (Waiting for release of https://github.com/ProseMirror/prosemirror-markdown/issues/76 )
- [ ] :lady_beetle: Multiple (allowed) (white)spaces are removed
- [ ] :lady_beetle: Text within links is not properly escaped

### Opinions
There are already issues about breaking the formatting of markdown files, like https://github.com/nextcloud/text/issues/593 and https://github.com/nextcloud/text/issues/2577 but from my point of view it is hard to provide that kind of compatibility, at least the way `text` is build.
But I do not think it is needed, if I need a markdown editor with that behavior, I would recommend to use https://github.com/icewind1991/files_markdown
Of cause you as the developers of this app should decide the scope of application, but for me `text` is a collaborative text editor.

susnux:
Those test results show where CommonMark markdown is still changed, so it results in semantically unequal code: https://github.com/nextcloud/text/actions/runs/3345044724/jobs/5540096347

mejo-:
We're on ProseMirror 0.10.1 now, so "Lines that start with a colon are escaped upon save" is resolved. Thanks again for debugging, reporting and fixing @susnux :heart:

PR #3331, opened 2022-10-28
WIP: Test also CommonMark serialization

susnux:
### Summary
As two different markdown strings can be semantically the same, we can test if the resulting rendered HTML are syntactically equal.

Currently about hundred test cases fail because of a known limitation of `prosemirror-markdown` not able to serialize nested marks correctly, see https://github.com/ProseMirror/prosemirror-markdown/issues/82

So the CI test results can be used as a reference for identifiying problem for #2702 and also what needs to be fixed so that markdown files stay unchanged (#593)


susnux:
I do not think we need this anymore :question:

juliushaertl:
I think we can revive when there is some more time to dive into the further markdown differences 👍

Issue #2871, opened 2022-09-07
Author color is cleared when accent is added


drebs:
**Describe the bug**

If an accent is inserted in the end of a line colored with an author color, that line's color is cleared out.

I could only reproduce in Firefox, and in Chromium the problem doesn't happen.

It seems related to dead keys, as the color is cleared after an accent key is pressed but before the next key (the one that will receive the accent) is pressed.

The color is only cleared "locally" (that is, someone else editing the same document will not see the issue manifest).

When the accent is inserted in the middle of the line, then the color is not cleared.

**To Reproduce**

1. Create a .mdwn document and open it for editing.
2. Enable "Show author color" in the top-right "Active people" menu.
3. Type some text without accents.
4. Try to add an accented letter in the end of the line (eg. á).
5. Witness the author color be cleared for that line.

**Expected behavior**

The author color for that line should not be cleared and remain the same.

**Client details:**
 - OS: Linux Debian 11 (Bullseye)
 - Browser: Firefox
 - Version: 91.12.0esr
 - Device: Desktop computer

<details>
<summary>Server details</summary>

**Text app version:** (see Nextcloud apps page)

3.5.1

**Operating system:**

Linux Debian 11 (Bullseye)

**Web server:**

Nginx version 1.18.0-6.1+deb11u2

**Database:**

MariaDB version  1:10.5.15-0+deb11u1

**PHP version:**

PHP version 7.4.30-1+deb11u1

**Nextcloud version:** (see Nextcloud admin page)

24.0.4

</details>

#### Browser log

Nothing shows up in the console.

mejo-:
Hey @drebs, thanks for your report. I can confirm that the bug still exists on latest development branch.

mejo-:
So this happens with all dead keys like acute (`'`), backtick/grave, circumflex (`^`), cedilla (`¸`), etc.

When typing a dead key as first character in a new paragraph, the author color doesn't get applied until the second key that accompanies the dead key is typed (even though the diacrit is displayed already).

mejo-:
I guess that we have to check for [editor.view.composing](https://prosemirror.net/docs/ref/#view.EditorView.composing) and only apply the transaction that sets the author colors if it's not true. See https://discuss.prosemirror.net/t/plugins-and-characters-with-a-diacritic/2674/3

PR #3350, opened 2022-11-02
Don't apply author annotations when in composition

mejo-:
When typing dead keys (like diacritics), the browser is in composition mode[1] until the accompanioning character is typed. This breaks author annotations.

We have to hold back the transaction when in composition and only apply it afterwards, as suggested at [2]. So check for `view.composing` before applying the transaction.

Fixes: #2871

[1] https://w3c.github.io/uievents/#events-compositionevents  
[2] https://discuss.prosemirror.net/t/plugins-and-characters-with-a-diacritic/2674/3

* Resolves: # 
* Target version: master


mejo-:
Maybe it would be good to have tests for this to prevent regressions. Probably needs an e2e cypress test. Would be curious if you think it's testable in jest.

juliushaertl:
Might not even be testable in cypress if using a chromium based browser. Iirc firefox introduced a slightly different handling of compose keys.

max-nextcloud:
I checked and the issue only seems to occure in firefox. So yes... will be hard to test in chrome.

max-nextcloud:
Also had a brief look at tiptap tests and prosemirror tests.

**Tiptap** runs all tests in cypress - but sometimes they are just plain js tests.
You could create an editor with the user color extension and then try to test that directly. But i don't know how to get the editor into composition mode.
You could also try and get a hold of the apply function of the Prosemirror plugin and somehow inject a stub for the editorview. But that will only unit test the function itself. If the composition behavior of the browsers change we will not notice.

**Prosemirror** has an explicit test for composition: https://github.com/ProseMirror/prosemirror-view/blob/master/test/webtest-composition.ts
However that seems pretty hard to mimic as it uses it's own testing toolset etc.

So i don't see a way of testing this right now that would be somewhat reasonable to achieve.

Issue #2456, opened 2022-05-27
Detect session timeout and redirect to login


juliushaertl:
In cases when the text session is timing out and the user session expires, we should detect and redirect to the login page instead of doing nothing.

PR #3376, opened 2022-11-07
Redirect to login if attempting to create sync session while unauthenticated

Raudius:
* Resolves: https://github.com/nextcloud/text/issues/2456
* Target version: master 

### Summary

When a user attempts to reconnect, if the response is "Unauthorized (401)", we redirect the user to the login page. After logging in the user will be redirected back to the document.


src/components/Editor.vue
```diff
@@ -341,6 +342,10 @@ export default {
 				initialSession: this.initialSession,
 			}).catch((e) => {
 				this.hasConnectionIssue = true
+				if (e.response.status === 401) {
+					const redirectUrl = '/index.php/apps/files/?dir=' + encodeURIComponent(this.currentDirectory || '/') + '&openfile=' + this.fileId
```
juliushaertl, line 346: Maybe we can check here if the files app is actually active, as the text app might be rendered in other apps as well through viewer or directly in collectives.

juliushaertl (approved):

Issue #3370, opened 2022-11-06
Hard line break broken


XueSheng-GIT:
**Describe the bug**
Hard line breaks (shift+enter) are shown as regular line breaks if a file is re-opened.

**To Reproduce**
1. Create a markdown (text) file.
2. Enter some text with a hard line break (shift+enter), see "demo file" below.
3. Close file
4. Open file again

Demo file:
```
Hard line break...  
There should be no space between those lines!

New paragraph (regular line break).
```

**Expected behavior**
Hard line breaks should be shown correctly (without additional space between the lines).

**Screenshots**
While entering the text:
![Hard line break (while entering text)](https://user-images.githubusercontent.com/4077683/200182353-6f9ca35c-41ed-45ec-ab9f-444e3f9892a6.png)

On re-open:
![Hard line break (on re-open)](https://user-images.githubusercontent.com/4077683/200182371-0c1ca263-73c9-41d3-a7ed-4a5edc7346a2.png)

**Client details:**
 - OS: Ubuntu 22.04
 - Browser: Firefox
 - Version: 106.0.2
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.6.0

**Operating system:** Ubuntu 20.04

**Web server:** apache2 (2.4.54)

**Database:** mariadb (Ver 15.1 Distrib 10.3.34-MariaDB)

**PHP version:** 7.4.32

**Nextcloud version:** 25.0.1

</details>

juliushaertl:
I can reproduce this. FYI @nextcloud/office Might be something we should look into before the the next maintenance release.

PR #3388, opened 2022-11-08
Fix hard line breaks displayed with additional new line between

susnux:
* Resolves: #3370
* Target version: master 

### Summary
To preserve soft breaks within markdown files we preserve all whitespaces, so we must not introduce additional new lines after `<br>` elements.


max-nextcloud:
Looks like the common mark specs expect a newline after a `<br />` tag. The approach here seems fine - but i wonder how / if we can keep the specs.

I wonder if we could still keep this and use some css or so to ignore / remove the extra blank line.

juliushaertl:
Thanks again for the fix @susnux 

It would be great to have a test case for this case

juliushaertl (approved):

PR #3401, opened 2022-11-09
Add tests for direct editing and cleanup dead code

juliushaertl:
Remove the DirectSessionController which was never used and add some basic cypress tests for direct editing.

I wanted to check if the file has properly been saved, but that fails currently due to the fact that the save is not happening fast enough. Something to be looked in sepeartely though.


cypress/e2e/directediting.spec.js
```diff
@@ -0,0 +1,105 @@
+import { initUserAndFiles, randHash } from '../utils/index.js'
+
+const randUser = randHash()
+const fileName = 'empty.md'
```
marcelklehr, line 4: Can you clean up these lint errors?

marcelklehr (approved):

cypress/e2e/directediting.spec.js
juliushaertl: Done

juliushaertl:
Follow up issue for the fixme comment and saving problem: https://github.com/nextcloud/text/issues/3404

mejo- (approved):
Not sure about the failing `mobile.spec.js` Cypress tests, but they seem unrelated to this PR.

PR #3423, opened 2022-11-14
Refactor vuex store

mejo-:
Move text store into a namespaced module and load it if we already have a store - e.g. when using the `RichTextReader` component via NPM module in a vue app that has a store on its own (like Collectives).

This is a prerequisite for using features that require the store (e.g. outline view) in Collectives.


mejo-:
Cypress tests seem unrelated to me (same tests failing on other branches and no failures when running them locally), but I'll look into them.

PR #3424, opened 2022-11-14
Update pull_request_template.md

juliushaertl:

* Resolves: # 
* Target version: master


juliushaertl:
Feedback from @max-nextcloud 

We might not want to use checkboxes for anything that is covered by CI as the state would need to be updated manually.

PR #3443, opened 2022-11-17
ImageView: Show caption in full length when in readonly mode

susnux:
### 📝 Summary

Instead of truncating the caption after 200px the full caption is shown when in edit mode.
Currently even hovering the caption does not show the full caption in read-only mode so the user is not able to read the full caption without digging into the HTML sources.
This improves usability e.g. for collectives.

About the changes on the scss section: I also reordered the scss part to group rules by classes.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
caption is truncated | caption is fully shown

### 🏁 Checklist


susnux:
I am not sure about the backporting policy, would this be a thing to backport to 25?
As it is at least quite annoying that you can not read the full caption when files are shared read-only.

Issue #3428, opened 2022-11-15
Preserve number of newlines at the end of md file


max-nextcloud:
This is a follow up to #3282.

With the new block images a file that only contains an image link will have an empty line added to the end.

The markdown in question already parsed as a block image before. So theoretically it's not necessary to add an empty line after it.

I think we have the same issue with tables at the end of the doc and other block elements. In order to ensure they are not added to the following paragraph we add an empty line below. However this is not needed at the end of the file. There we want to keep an existing empty line but avoid adding new ones.

_Originally posted by @max-nextcloud in https://github.com/nextcloud/text/issues/3282#issuecomment-1314189934_

max-nextcloud:
part of #3327

PR #3445, opened 2022-11-17
Fix block images always add additional newlines on the end of the file

susnux:
### 📝 Summary

* Resolves: #3428

This fixes block images adding additional newlines, without this images at the end of a file will add two additional newlines.
This is fixed by using `closeBlock` within the `toMarkdown` function instead of always enforcing two newlines.

The tests are adjusted accordingly.

### 🏁 Checklist


src/tests/markdown.spec.js
```diff
@@ -120,9 +105,11 @@ describe('Markdown though editor', () => {
 		expect(markdownThroughEditor('[bar\\\\]: /uri\n\n[bar\\\\]')).toBe('[bar\\\\](/uri)')
 	})
 	test('images', () => {
+		// Inline images
 		expect(markdownThroughEditor('text ![test](foo) moretext')).toBe('text ![test](foo) moretext')
-		// regression introduced in #3282. To be fixed in #3428.
-		expect(markdownThroughEditor('![test](foo)')).toBe('![test](foo)\n\n')
+		// regression introduced in #3282. See issue #3428.
```
max-nextcloud, line 110: This comment is outdated now :smiley: - we don't need it anymore.
```suggestion
```

max-nextcloud (approved):
Thanks a lot for addressing this. It fixes an annoying issue and the approach is much cleaner.

src/tests/markdown.spec.js
susnux: I let the comment stay there to have some information why a test was added, but I can remove it as well :)

max-nextcloud: Okay... cool. I meant the comment as documenting a regression - now it's documenting a fixed regression - i guess that's fine. Either way it points to a relevant issue.

PR #3462, opened 2022-11-21
Add APIs for outline state to RichTextReader and WrapperComponent

mejo-:
### 📝 Summary

* Related: #2919

Adds APIs to interfere with outline state to `RichTextReader` and `WrapperComponent` components.

Required for https://gitlab.com/collectivecloud/collectives/-/merge_requests/649

### 🚧 TODO

- [x] Change height of outline container height (not 100%, only content + a bit of padding on top and bottom)

### 🏁 Checklist


max-nextcloud (commented):
So far looks good to me. I guess the handing down props and propagating up events could be avoided by using provide and inject - but from what i see it's only two levels. So I'm not sure if that's worth it.

max-nextcloud (approved):
Had a quick look at the code again.

Issue #2201, opened 2022-02-24
Wrap cypress helpers into js library for reuse across apps


max-nextcloud:
Wrap all those nice helpers into a small js library, so that they could be reused across apps when writing cypress tests.

_Originally posted by @juliushaertl in https://github.com/nextcloud/text/pull/2193#r812626438_

mejo-:
@max-nextcloud is this rather about moving some of our custom cypress commands to `@nextcloud/cypress`?

max-nextcloud:
Yes. But given that more and more apps are using playwright it might not be worth the effort. I'd say let's close this and move helpers if we notice we keep copying them to other apps.

mejo-:
I agree!

PR #3465, opened 2022-11-22
test: start using @nextcloud/cypress

max-nextcloud:
### 📝 Summary

Use `@nextcloud/cypress` to enable code sharing with other apps.

For now just using the new fast login method.
This reduces the time it takes to login by 6 sec. on my machine.

* See also: #2201

### 🏁 Checklist

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -25,4 +25,5 @@ module.exports = defineConfig({
 		// do not retry in `cypress open`
 		openMode: 0,
 	},
+	"numTestsKeptInMemory": 5,
 })
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -21,6 +21,7 @@
  */
 
 import axios from '@nextcloud/axios'
+import { login } from '@nextcloud/cypress/commands'
 
 // eslint-disable-next-line no-unused-vars,n/no-extraneous-import
 import regeneratorRuntime from 'regenerator-runtime'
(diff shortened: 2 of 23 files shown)


skjnldsv:
I would migrate the cypress config files to typescript.
Or add cypress to eslint ignore.

https://github.com/import-js/eslint-plugin-import/issues/2495

cypress/support/commands.js
```diff
@@ -21,24 +21,19 @@
...
-	cy.session(user, function() {
-		cy.visit(route)
-		cy.get('input[name=user]').type(user)
-		cy.get('input[name=password]').type(password)
-		cy.get('form[name=login] button[type=submit]').click()
-		cy.url().should('include', route)
-	})
-	// in case the session already existed but we are on a different route...
+Cypress.Commands.add('login', (userId, password, { route, onBeforeLoad } = {}) => {
+	route = route || '/apps/files'
```
skjnldsv, line 35:
```suggestion
Cypress.Commands.add('login', (userId, password, { route =  '/apps/files', onBeforeLoad } = {}) => {
```

cypress/support/commands.js
```diff
@@ -21,24 +21,19 @@
...
+	cy.loginWithoutVisit({ userId, password })
 	cy.visit(route, { onBeforeLoad })
```
skjnldsv, line 37: I would personally do it the other way around, `loginAndVisit`, and `login` :)

max-nextcloud: I agree... actually i prefer these to be two commands. But I wanted to check if i could just use `@nextcloud/cypress`s login command without touching all the tests.

juliushaertl:
CI seems to take the usual time but not sure why there is no difference

cypress.config.js
```diff
@@ -25,4 +25,5 @@ module.exports = defineConfig({
 		// do not retry in `cypress open`
 		openMode: 0,
 	},
+	"numTestsKeptInMemory": 1,
```
skjnldsv, line 28: Oooof, we should all reduce that number, yes! :see_no_evil:

max-nextcloud: I had issues with heap overflows when testing this. Problem also is heap does not get cleaned up when rerunning a test or moving to the next.

This also means loosing the ability to inspect previously run tests - but the benefit still outweights the cost for me here.

I'll still increase this to 5. One test typically takes up 40 MB of heap space. So 5 are 200 MB - so heap is full after 10 test runs.

max-nextcloud:
> CI seems to take the usual time but not sure why there is no difference

Yes. We only log in once per test spec. Afterwards the cookies are reused by the session managment. So even if you have a 3 minute test spec, this will only cut off 6 seconds - which is lost in the noise.

It's more relevant when working on a single test maybe even using `it.only` - because if your test takes only 15 seconds those 6 seconds are a big difference.

skjnldsv:
> It's more relevant when working on a single test maybe even using `it.only` - because if your test takes only 15 seconds those 6 seconds are a big difference.

We have a lot of tests in viewer too, there was a lot of login in and out :)
That should explain why there was such a drastic change there

skjnldsv (approved):

PR #3472, opened 2022-11-22
Use GH ref as cypress run title

skjnldsv:
image


juliushaertl:
Failure unrelated

juliushaertl:
Oh I see, it would be rather this one then or?

https://github.com/nextcloud/viewer/blob/master/.github/workflows/cypress.yml#L82

skjnldsv:
Ah, I added the type of PR, push or pull, but not the title! :rocket: 
Thanks Julius :)

skjnldsv:
https://github.com/nextcloud/text/pull/3473

PR #3476, opened 2022-11-23
Fix direct editing loading of txt files

juliushaertl:
### 📝 Summary

Fix loading txt files through direct editing

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [x] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [x] Documentation ([README](https://github.com/nextcloud/text/blob/master/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required

PR #3479, opened 2022-11-24
[stable25] Fix direct editing loading of txt files

juliushaertl:
Backport of #3476


src/services/SyncService.js
```diff
@@ -95,12 +95,10 @@ class SyncService {
 			document: this.document,
 			session: this.session,
 		})
-		const content = connectionData.content
-			|| await this._fetchDocument()
 		this.emit('loaded', {
 			document: this.document,
 			session: this.session,
-			documentSource: '' + content,
+			documentSource: '' + connectionData.content,
```
marcelklehr, line 101: Isn't this the same thing we removed above? I'm confused. :sweat_smile:

Raudius, line 101: Minor nitpick, but could we just cast it?

```js
documentSource: String(content)
```

juliushaertl: Yes, which is fine still just without the fetch fallback

juliushaertl: The main problem was that an empty string (from a file without content) would always trigger the fetch request which is actually not necessary anymore as we always pass the content earlier.

Raudius (approved):

src/services/SyncService.js
Raudius: Nevermind, this is already a backport :see_no_evil: 

Approved

PR #3502, opened 2022-11-30
Keep markdown syntax of hard line breaks

susnux:
### 📝 Summary

* Part of #593 

Do not change the markdown syntax used for hard line breaks in documents, e.g. if an existing document contains backslashes as hard line breaks those were replaced with double spaces.

This will keep the used syntax while also default to use double spaces.
Also adjusted the tests for this new behavior.

### 🏁 Checklist


max-nextcloud (approved):
Thanks a lot for taking care of this @susnux!

Code looks good. Tests included so i did not test myself.

juliushaertl:
🎉

Issue #2725, opened 2022-07-16
Allow line breaks in table rows


theroch:
**Is your feature request related to a problem? Please describe.**
We want to have line breaks in table rows to format the text.
Othe markdown viewers/editors allow this by using `<br>` or `<br />` tags.

**Describe the solution you'd like**
Support `<br>` tags or other formating tags to allow line breaks in table rows like this:
```
| **Org** | **Name** | **Contact** |
|-----|------|--------------|
| Org | Name | E-Mail: my@mail.com <br> Tel: 01/ 123456789 |
```

**Describe alternatives you've considered**
We already tested with double spaces, `\`, `\n`, <br> nothing works. 

**Additional context**
See also https://stackedit.io/app#

juliushaertl:
Sounds reasonable to support that, even if that is not something that makes the output markdown table very readable.

Any thoughts @max-nextcloud ?

susnux:
We should support the `<br>` syntax as this is the most common one, `\` is only supported by pandoc (as far as I know).

mejo-:
Not sure about the `<br>` syntax as that feels like a first step to introduce HTML into our markdown set. What do others think? @vinicius73, @max-nextcloud, @juliushaertl, @juliushaertl, @jancborchardt?

juliushaertl:
My main concern about HTML would be that we properly need to ensure to purify/sanitize it, though I think tiptap/prosemirror might already take care of that quite well.

max-nextcloud:
The [security recommendations of markdown-it](https://github.com/markdown-it/markdown-it/blob/master/docs/security.md) say:

> `markdown-it` provides 2 possible strategies to produce safe output:
> 1. Don't enable HTML. Extend markup features with [plugins](https://www.npmjs.org/browse/keyword/markdown-it-plugin). We think it's the best choice and use it by default.
     - That's ok for 99% of user needs.
     - Output will be safe without sanitizer.
> 2. Enable HTML and use external sanitizer package.

I'd like to avoid enabling HTML. But I think we could support `<br>` without enabling it. It would require us to parse `<br>` and `<br />` inside a table as a token of it's own and turn it into a `<br>` tag when parsing the markdown. I think tiptap would handle that `<br>` tag just fine.

We'd also need to serialize `<br>` tags in tables to that token in the `toMarkdown` function of table cells or so.

Issue #2170, opened 2022-02-08
Parse HTML tags


xplosionmind:
**Is your feature request related to a problem? Please describe.**
HTML tags are displayed as plain text and not rendered, as it should happen since it is a superset of Markdown. Example below

<img width="1440" alt="Screenshot 2022-02-08 at 6 47 21 PM" src="https://user-images.githubusercontent.com/42862428/153045857-c4dada09-f09b-4c18-8c2f-f6b9638d3f3d.png">

**Describe the solution you'd like**
HTML tags should be rendered normally, or at least tags should be stripped and only the text inside them should be displayed

**Describe alternatives you've considered**
There is none

**Additional context**
I do not know what I could add but I am available for clarifications

luka-nextcloud:
It's still reproduceable.

max-nextcloud:
The underlying issue here is that there are a lot of ways html tags can be used to perform actions on behalf of anyone who opens the page. These attacks are called [XSS - cross site scripting](https://owasp.org/www-community/attacks/xss/) attacks.
The most simple example would be injecting a `<script>` tag that can perform arbitrary javascript such as sharing a file, deleting data, leaking the users session cookie.

So we could exclude `<script>` tags, but there are a LOT of ways to make the browser execute javascript. There are sanitizers for this but so far we are sticking with the [recommendation of markdown-it](https://github.com/markdown-it/markdown-it/blob/master/docs/security.md):
> Don't enable HTML. Extend markup features with [plugins](https://www.npmjs.org/browse/keyword/markdown-it-plugin). We think it's the best choice and use it by default.
> * That's ok for 99% of user needs.
> * Output will be safe without sanitizer.

PR #3504, opened 2022-12-01
Preserve HTML newlines and add support for newlines within table cells

susnux:
### 📝 Summary

* Resolves: #2725 and partly #2170

First of all this allows adding line breaks inside of table cells, for this we can not use the normal markdown syntax so as markdown is a superset of HTML it is common to use `<br>` tags for this.
But as be disabled HTML for security reasons, this PR introduces a special markdown-it parser only enabling `<br>` HTML tags.
Meaning at least support for `<br>` is added (tiny part of #2170).

About UX:
* Pressing enter within a table cell will create a newline
* Pressing enter twice will move into the next row instead (like before, so this behavior is mostly the same)

#### 🖼️ Screenshots

https://user-images.githubusercontent.com/1855448/204934084-fa23de0f-8ae9-4526-88bc-4a6e310ea891.mp4

### 🚧 TODO

- [x] add cypress e2e test

### 🏁 Checklist

- [x] [Signoff message](https://probot.github.io/apps/dco/) is added to all commits


max-nextcloud:
# Yay! :tada: 

Screencast looks really good. Will check out code and then rebase etc.

src/markdownit/hardbreak.js
```diff
@@ -26,7 +26,7 @@ import markdownitEscape from 'markdown-it/lib/rules_inline/escape.js'
 /**
  * Add information about used markdown syntax to HTML hard breaks
  *
- * @param {object} md Markdown object
+ * @param {import('markdown-it')} md Markdown object
```
max-nextcloud, line 29: uhh... nice. I learned something today. :smile:

max-nextcloud (approved):
Code looks good. Also tested copy and paste between table cells.

PR #3519, opened 2022-12-06
Cypress: refactor user management

max-nextcloud:
Unify and simplify user management in cypress tests.


cypress/e2e/FrontMatter.spec.js
```diff
@@ -18,18 +18,17 @@
  *
  */
 
-import { User } from '@nextcloud/cypress'
-import { initUserAndFiles, randHash } from '../utils/index.js'
+import { initUserAndFiles, randUser } from '../utils/index.js'
 
-const randUser = new User(randHash(), 'password')
+const me = randUser()
```
juliushaertl, line 23: I find the `me` variable naming a bit unexpected, would still prefer something like user or testUser, but might just be me 😉

juliushaertl (approved):

cypress/e2e/FrontMatter.spec.js
max-nextcloud: I'm fine with `user`. I don't like `user1` and `user2` as they are easy to confuse. But i can use meaningful names there like `sender` and `recipient` or so.

PR #3520, opened 2022-12-06
Prevent dangling debounced extractHeadings function

mejo-:
### 📝 Summary

Our Tiptap heading extension registers a debounced extractHeadings function for onUpdate. If the editor gets removed but a debounced extractHeadings call is still pending, this leads to errors (which sometimes break tests).

So let's track whether the editor is still alive and only call debounced extractHeadings in this case.

### 🏁 Checklist


juliushaertl (approved):
Makes sense 👍

max-nextcloud:
I wonder if this also applies to the other debounced functions we have. But then that's a different step / fix i guess.

mejo-:
/backport 72a1c48ce04345a362810f7a97e022f81581093e to stable25

Issue #3304, opened 2022-10-25
Tables don't support right justified columns


phippodoplis:
**Is your feature request related to a problem? Please describe.**
Any tables with columns of numbers look a lot better with those columns right justified. The markdown syntax is to use |-:| for the column, |:-:| is for centering, but that's less important to me.

**Describe the solution you'd like**
|---:| results in a right justified column, I don't mind if the header cell is right justified or not.

**Describe alternatives you've considered**
Adding spaces doesn't work.

**Additional context**

PR #3521, opened 2022-12-06
Feature: Support table column alignment

susnux:
### 📝 Summary

* Resolves: #3304

This adds support for markdown tables using column alignment.
 Besides supporting alignment from markdown files this also
 adds a UI for changing the alignment.
    
 The TipTap commands for adding table rows are extended so they preserve the text align attribute on the new created row.

*Please note the "ugly" classes for the align buttons in `TableHeaderView.vue` is only a workaround for https://github.com/nextcloud/nextcloud-vue/issues/3547 so when that is fixed we can remove the classes and stick with the data attribute*

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_20221206_134021 | Screenshot_20221206_133957

### 🏁 Checklist


max-nextcloud (approved):
This looks really good!

Thanks also for improving the table spec and adding e2e tests for the new behavior.

I'll rebase and compile again and also give it a try while doing that. Codewise this looks excellent. :100: :wink:

max-nextcloud:
The right and middle alignment conflict with the buttons in the heading. But that's something for another issue.

PR #3530, opened 2022-12-07
Refactor code to not have NcActions within a NcActions

susnux:
### 📝 Summary
Instead of nesting `NcActions`, which is not officially supported by `@nextcloud/vue`,
the code is rewritten to use a simple wrapper component for inline actions within the action popover.

### 🏁 Checklist


max-nextcloud (approved):
Code looks good.

skjnldsv:
Hey!
That does not change anything.
The actions menu have lots of scripts for proper accessibility and keyboard navigation.

Considering this is not at all part of the accepted components, you're breaking the usage.
Where is this being used exactly? Is this released yet?

skjnldsv:
Ah, found it, it's not released.
Well, using the trick made in this PR works, arrows and tab navigation are compliant again as we're basically still displaying only ActionsXxxx :)

Nonetheless, we never did this until now, so before we introduce this, i'd like to make sure the design team approved a clear design so we can add this officially (or not) and document it on the vue components :)
image

juliushaertl:
cc-ing @nimishavijay and @jancborchardt then for a after-merge review :)

nimishavijay:
This looks great! :) I would suggest one change: Reduce the margin so that each item width is 1/3rd of the width of the action menu and the icon can be centred in each item. Right now it looks like each item is rather small, with lots of margin between them.

susnux:
Yes currently each button has the default size of `44px * 44px`, but I work on implementing your review comments.
Implemented review comments, buttons take whole width

Issue #394, opened 2019-11-02
Support more file extensions


The1AndMany:
**Is your feature request related to a problem? Please describe.**
When I was using files_texteditor, it allowed me to open files with pretty much any file extension, so long as the data was stored in plaintext format. This was especially useful for editing my website's .html and .css files on the fly. Since upgrading to Nextcloud 17, this functionality seems to have been removed when the default text editor changed. This means I must have that file synced to my computer and open it locally, but there are some instances where I might be using a public computer at a library, so that wouldn't be an option.

**Describe the solution you'd like**
Ideally, if a file format stores its data in a plaintext-like format (.html, .css, .json, .xml, etc.), Text should be able to open it

**Describe alternatives you've considered**
Switch back to files_texteditor, but I'm not sure that's a possibility anymore

**Additional context**
The only other context I can think of is specifically stating that I keep my website saved to another folder on my server, which I connect to using local external storage.

gt2416:
+1 for this, I used to be able to code my .cpp files on nextcloud for my school projects, now it just downloads the file.

juliushaertl:
As a quick fix we can extend the list in https://github.com/nextcloud/text/blob/master/src/helpers/mime.js#L38 

Ideally the viewer should maybe also support wildcards just as the files apps file actions which also allows only matching for the mime part (`text`)

FastBlit:
I would like to open asciidoc files without hiding the syntax.

GottHeit:
> 
> 
> As a quick fix we can extend the list in https://github.com/nextcloud/text/blob/master/src/helpers/mime.js#L38
> 

I tried to add that fix as I would like do edit some files which are editable yet. But I don´t have that folder/file mentioned? Cannot find it, even with a search. Maybe you can give me a hint? (Not a nextcloud pro, so maybe too stupid ^^)

juliushaertl:
> I tried to add that fix as I would like do edit some files which are not editable yet. But I don´t have that folder/file mentioned? Cannot find it, even with a search. Maybe you can give me a hint? (Not a nextcloud pro, so maybe too stupid ^^)

Those file are not available in the release bundle since we use webpack to bundle all files in the `src/` directory into javascript bundles. Checkout the development setup in our readme: https://github.com/nextcloud/text#-development-setup

Torenga:
Please make this an opt-in/out feature. I want to click my HTML-files, start a download and open them directly in a new tab. No hosting, no editing.

[...]

PR #3535, opened 2022-12-07
Add text/asciidoc mimetype

bjornfor:
### 📝 Summary

This allows editing AsciiDoc files, as plain text.

This change depends on
https://github.com/nextcloud/server/commit/582f07c8fc47d4a775f05fbaad32269753d944f0
("Add text/asciidoc mimetype").

Signed-off-by: Bjørn Forsman <bjorn.forsman@gmail.com>

~Disclaimer: this is a trivial change, but I haven't tested (I don't know how).
I based this on
https://github.com/nextcloud/text/issues/394#issuecomment-1188860327~ It has a test now.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


max-nextcloud:
Hi @bjornfor,

Thanks a lot! That seems like a useful addition. Code also looks good.

I added a test for two mimetypes we already support here: https://github.com/nextcloud/text/pull/3540.
Would that work as a basis for adding tests for you?

I'd be curious to see if you can add a test with this. If you get stuck or miss some documentation please let me know so we can make this work more smoothly.

bjornfor:
Hi @max-nextcloud, thanks, I can probably make a copy of the plain text test, without fully understanding it :smile:

bjornfor:
@max-nextcloud: I rebased and added a test, but I think `cy.getContent().find('pre').should('exist')` part isn't quite right.

So `find()` looks for HTML tags (right?), and in the "empty.md" test you added I think the test just checks that the overall HTML page contains a `pre` element. In the "markdown" test, the code checks for a `h2` tag that comes from converting markdown to HTML. But for this AsciiDoc code, we don't get any special markup (it's treated as plain text), but we would perhaps like to check that the text *content* got rendered?

I guess I'd like to say `cy.getContent().SOMETHING_HERE.should('contain', 'Hello world')`, but don't know what `SOMETHING_HERE` should be.

bjornfor:
> I guess I'd like to say `cy.getContent().SOMETHING_HERE.should('contain', 'Hello world')`, but don't know what `SOMETHING_HERE` should be.

Does `cy.getContent().find('pre').should('contain', 'Hello world')` work?

susnux:
The `getContent()` part restricts the following `find` to the editor content (whereas `get()` always searches the whole document).
So this is working as expected.

max-nextcloud (approved):
Hey @bjornfor 

Thanks for adding the test. Congratulations for figuring it out without understanding it.
:key: achievement unlocked :wink: 

> Does `cy.getContent().find('pre').should('contain', 'Hello world')` work?

Yes - that should do the trick. As @susnux explained `getContent` only looks at the content of the editor.
Feel free to add the hello world check - test also seems okay for me as is because it tests that text is loading the file.

Your last commit seems to lack a signoff. Other than that i think this is ready to go.

The failing `node` test is because the js still needs to be build here by commenting with `/compile`. It's a mess and we're working on fixing it. I can take care of that once you signed off the second commit.

bjornfor:
@susnux @max-nextcloud: Thanks. I added `Signed-off-by:` and update the test with `cy.getContent().find('pre').should('contain', 'Hello world')`.

max-nextcloud:
CI failure is not related.

bjornfor:
@max-nextcloud: Thanks!

PR #3544, opened 2022-12-09
Fix design of `InlineActionsContainer` - resolv design review comments

susnux:
### 📝 Summary

* Resolves design review comments of @nimishavijay https://github.com/nextcloud/text/pull/3530#issuecomment-1343955035

Items within the container are now sized equally and use the whole width.
Icons are centered within the actions as suggested in the comment.

I decided to do it using a more generic approach, being reactive on the default slot, so the component can be reused more easily.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
buttons use 44px fixed width | buttons with 33% width

### 🏁 Checklist


src/components/InlineActionsContainer.vue
```diff
@@ -19,7 +19,7 @@
   -->
 <template>
 	<li class="inline-container-base">
-		<ul>
+		<ul class="inline-container-content">
 			<slot />
 		</ul>
 	</li>
```
skjnldsv, line 25: Could we have that component exposed and documented? :)

src/components/InlineActionsContainer.vue
```diff
@@ -34,17 +34,45 @@ import { defineComponent } from 'vue'
...
+		 *
+		 * @param {Node} target The mutation target
+		 */
+		onMutation(target) {
+			const children = [...target.childNodes].filter(node => node instanceof HTMLElement)
+			children.forEach(node => {
+				node.style.width = `${Math.round(100 / children.length)}%`
+			})
+		},
+	},
```
skjnldsv, line 62: This seems overly complicated?
Let me suggest a css only fix

src/components/InlineActionsContainer.vue
```diff
@@ -34,17 +34,45 @@ import { defineComponent } from 'vue'
...
-	}
+ul.inline-container-content {
+	display: flex;
+	justify-content: space-between;
+
 	.action-button {
+		// Fix action buttons beeing shifted to the left (right padding)
 		padding: 0 !important;
+		// We force a with other than 44px so we need to center the items again
+		justify-content: center;
```
skjnldsv, line 75:
```suggestion
ul.inline-container-content {
	display: flex;
	justify-content: space-between;
	li {
		flex: 1 1;
	}

	.action-button {
		// Fix action buttons beeing shifted to the left (right padding)
		padding: 0 !important;
		width: 100%;
		display: flex;
		justify-content: center;
```

skjnldsv (changes requested):

src/components/InlineActionsContainer.vue
skjnldsv: https://github.com/nextcloud/text/pull/3544#discussion_r1044635494

nimishavijay (commented):
Looks great design-wise! :) 🚀

nimishavijay (approved):

src/components/InlineActionsContainer.vue
susnux: Thank you!

src/components/InlineActionsContainer.vue
susnux: Submitting to nextcloud vue components?

max-nextcloud (approved):

src/components/InlineActionsContainer.vue
skjnldsv: Yes or something like that

skjnldsv (approved):

Issue #3527, opened 2022-12-07
It is possible to insert hard breaks into headlines


susnux:
**Describe the bug**
Hard breaks are not allowed within headlines (following the commonmark specifications).
But on the editor it is possible to insert hard breaks into headlines by pressing `CTRL+ENTER` inside the headline.

This results in a markdown file containing the hard break after the headline, but as this is not valid common mark, when opening the file again and trigger a save the second line would be converted to a new paragraph.

**To Reproduce**
1. Open new file
2. Insert a headline
3. Press `CTRL` + `ENTER`
4. See linebreak within headline 

**Expected behavior**
I would expect the cursor to jump out of the headline, the same as when pressing enter without `CTRL`.

PR #3568, opened 2022-12-13
Feat: backspace converts paragraph break to hard line break

susnux:
### 📝 Summary
A follow up on #2775 (see comments)
* Supersedes: #2775
* Resolves: #3527 
* Resolves: #2809 
* Resolves: #2761 

This allows to use hard line breaks on mobile devices by using the same workflow as for lists: Pressing enter results, as before, in a paragraph break. But if you now press backspace on the begin of the paragraph, the paragraph break is converted into a hard break.
Besides converting paragraph breaks into hard line breaks this PR also fixes an issue where users could insert hard line breaks into headings, which is not supported by markdown.

### 🏁 Checklist


max-nextcloud (approved):
Code looks good. I will play with this some.
I'm also unsure if the failing cypress test is related. I have not seen that in the past.

max-nextcloud (changes requested):
I think the cypress failures are related to the changes - at least the same failure happened twice on different runs of the tests and i never saw it before. So would be great if that could be fixed.

susnux:
> I think the cypress failures are related to the changes - at least the same failure happened twice on different runs of the tests and i never saw it before. So would be great if that could be fixed.

I do not know why it was working, there was:
type: `# heading 1{enter}{backspace}{backspace}2`
but this should have always resulted int
`# heading2` instead of `heading-2`.

I removed the second backspace and now it works.

max-nextcloud:
> type: `# heading 1{enter}{backspace}{backspace}2`

Well the first backspace undoes the enter - so you are back at the end of the heading. The second one will remove the digit.
I think that's also how it works with paragraphs right now.

susnux:
@max-nextcloud You are absolutely right, sorry :facepalm: 
I found the issue, the reason was some incorrect sibling query.
This is fixed so only paragraphs are joined with hard breaks (and not headlines + paragraph).

For every other nodes than paragraphs the default action of `backspace` should be used.

max-nextcloud (approved):

max-nextcloud:
I played with it some and it's really nice.
There's one downside to it though...
When in a list and i press enter but then i decide i want to go back to the previous list item I have to press backspace three times:
* remove the new list item, end up on a second paragraph in the first
* turn the second paragraph into a new line on the first paragraph
* get into the first line of the list item.

I think it would be nice if backspacing into a list item would also land you after a hard break rather than in a second paragraph in that list item.

PR #3569, opened 2022-12-13
Use cypress helpers from `@nextcloud/cypress`

susnux:
### 📝 Summary
* Resolves only partly #2201 

Use user management commands from `@nextcloud/cypress` instead maintaining them by our self.

### 🏁 Checklist

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -10,8 +10,6 @@ cy.getContent()
 ### Available custom commands
 | Command              | Function               | Parameters                          |
 | -------------------- | ---------------------- | ----------------------------------- |
-| `nextcloudUpdateUser`| Update user setting    | `user`, `password`, `key`, `value`  |
-| `nextcloudDeleteUser`| Delete user            | `user`                              |
 | `uploadFile`         | Upload file            | `fileName`, `mimeType`, `target`    |
 | `createFile`         | Create file            | `target`, `content`, `mimeType`     |
 | `moveFile`           | Move a file            | `path`, `destinationPath`           |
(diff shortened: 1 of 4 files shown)


cypress/README.md
```diff
@@ -35,10 +33,9 @@ cy.getContent()
 
 ### Commands from `@nextcloud/cypress`
 
-We also use some commands from `@nextcloud/cypress`:
+We also use some commands from [`@nextcloud/cypress`](https://github.com/nextcloud/nextcloud-cypress):
 * `login(user)`
 * `logout()`
 * `createRandomUser().then(user => login(user))`
 * `createUser(user)`
```
max-nextcloud, line 40: Could you add the two commands here?
I think we should either keep this list complete or remove it entirely.
```suggestion
We also use some commands from [`@nextcloud/cypress`](https://github.com/nextcloud/nextcloud-cypress):
* `login(user)`
* `logout()`
* `createRandomUser().then(user => login(user))`
* `createUser(user)`
* `modifyUser(user, key, value)`
* `deleteUser(user)`
```

max-nextcloud (approved):
Code looks good. I suggest adding the commands to the list in the docs.

cypress/README.md
susnux: I would vote for removing and only keep a reference of our own commands + a note about the nextcloud package.
(so keep the table but remove the list and only keep `We also use some commands from [`@nextcloud/cypress`](https://github.com/nextcloud/nextcloud-cypress).`

max-nextcloud: Fine with me.

Issue #1715, opened 2021-07-12
Adjustable Timer for Idle Timeout of #1027


tdoerschel:
Hi,
we use a simple shared text file as a status board (who is in which room right now, who is on vaccation, what is going on etc.),
so everyone read+writes it during the day.

The 30 minute idle timeout is really annoying (see #1027). Also, because it hits often while writing in the text file.

Is there a way to adjust the timeout, or generally increase it to 10 hours?

juliushaertl:
Yes, we should definitely improve this. I' wondering if we could just use the timer for expiring if the window is not in the foreground which should be possible with https://developer.mozilla.org/de/docs/Web/API/Page_Visibility_API

kkplein:
This would be very convenient if it were implemented yes..!

max-nextcloud:
Copying what @mejo- said in #3375:

> Some ideas how to improve idle session handling:
> 
> * Only start idle timer when [page becomes invisible](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)
> * Set session to idle after 30 minutes of page being invisible
> * Reset timer once [page becomes visible again](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)
> * Automatically reconnect once [page becomes visible](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) if session got idle in the meantime

PR #3596, opened 2022-12-21
Increase idle timeout to one day (1440 minutes)

mejo-:

### 📝 Summary

As discussed, let's change the session idle timeout to 1 day.

Related to #1715.

### 🏁 Checklist


juliushaertl:
We could also consider bumping https://github.com/nextcloud/text/blob/d8e4490db81318eaca8da50e304e5a23f8db4adb/lib/Service/SessionService.php#L39 so that sessions are also kept if someone has a network disconnect for longer than 5 minutes.

mejo-:
/backport 8c9eaf47b45d3c65f417d0353a5cadf67e7886fd to stable25

mejo-:
> We could also consider bumping
> 
> https://github.com/nextcloud/text/blob/d8e4490db81318eaca8da50e304e5a23f8db4adb/lib/Service/SessionService.php#L39
> so that sessions are also kept if someone has a network disconnect for longer than 5 minutes.

Good idea. What do you think would be a sensible value? Also a day? Or maybe shorter because catching up with a session after one day would not be a good idea? So maybe three hours?

fwsmit:
Sorry to jump in at a merged PR, but I noticed you haven't finished discussing the session timeout. I think it's best to leave the session timeout this way and make sure the session can reconnect automatically if the network is restored.
I'm not that knowledgeable about this app, but a session timeout of 3 hours doesn't seem helpful and can only cause bugs.

PR #3615, opened 2022-12-29
Editor frontend API

juliushaertl:
- [x] Based on #3700 

### 📝 Summary

Very early draft for implementing a frontend API so that the text editor can be used by other apps.

There is a [small demo page](https://github.com/nextcloud/text/pull/3615/files#diff-ba1781b9dbf5178b4ca0741e91667a5fe53a90cc5b5438d084b9704de3145ea8R50-R66) temporarily available at `/apps/text/` that shows both usages as a file editor (e.g. for collectives) and as a editor based on providing markdown content (e.g. for deck).

For now I've introduced a new MarkdownContentEditor component that is being used for editing text that is not sourced from files. This makes use of most components that are already used in the Editor component (which is currently the one for files).

## Can be tested with

- https://github.com/nextcloud/deck/pull/4399
- https://github.com/nextcloud/notes/pull/958

### 🚧 TODO

- [x] Proof of concept for exposing the editor through a plain JS API
- [x] Implement MarkdownFileEditor as wrapper around Editor and RichTextReader
- [x] Disable non-functional parts
  - [x] Attachment upload
  - [x] File picker (still to test - should work in theory)
  - [x] Mentions
- [x] Settle on API
- [x] Check on all requirements of external apps
- [x] Discuss on potential input/output formats (e.g. tables currently uses html) -> We should move all to markdown to keep things simple
- [x] Cleanup before merge
  - [ ] Remove demo page commit ([62ec547](https://github.com/nextcloud/text/pull/3615/commits/62ec5478ee6d6be725f5f0e7283c7be4be9d58c0))
- [ ] Document limitations of the content only mode

## Follow up
- [ ] Probably wrap API to typed package
- [ ] Things that collectives triggers
  - [ ] Save through syncService.save()
  - [x] focusEditor

### 🏁 Checklist

- [ ] Documentation ([README](https://github.com/nextcloud/text/blob/master/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required


juliushaertl:
@mejo- @max-nextcloud Pinging for some early feedback on the general concept. I refactored quite a bit in terms of how the tiptap plugins are being gathered, but this should not have any breaking change for now in terms of the existing file editor and the components exposed to the npm packages.

max-nextcloud (commented):
Looks like a great start to me. :+1: 

At some point we might want a switch in collectives to toggle editing / read only. But from my point of view that does not have to land in a first iteration.

I really like the cleanup of the EditorFactory, moving the tiptap extensions into RichText.

susnux:
So if I understand it correctly this provides editor related code under `OCA....` global variable?
What is the benefit over the already existing `@nextcloud/text` package?

Would it not make more sense to use `@nextcloud/text` also for the text app and provide it through *federated modules* (supported by webpack and by vite) so other apps can use without needing to provide it?

juliushaertl:
Before the merge:
- [ ] Remove demo page
- [ ] Disable mentions if no callback is provided

mejo-:
> So if I understand it correctly this provides editor related code under `OCA....` global variable? What is the benefit over the already existing `@nextcloud/text` package?
> 
> Would it not make more sense to use `@nextcloud/text` also for the text app and provide it through _federated modules_ (supported by webpack and by vite) so other apps can use without needing to provide it?

Sorry for the late reply @susnux. We discussed the pros and cons of different ways to provide Nextcloud Text as a global editor for different apps back and forth several times.

The main advantage of the approach with providing the editor via `OCA.Text` global is that there will be only one Text version on a particular instance. Using a NPM module sooner or later would result in different Text versions with an incompatible featureset being used.

juliushaertl:
Seems I failed on the rebase, will have another look

mejo-:
I rebased and force-pushed. For some reason, Github doesn't recognize my force-push. It lists it, but doesn't start CI nor gets the list of commits in UI updated :thinking:

In manual testing, it mostly worked. A few issues I spotted:

* Once I got yjs synchronisation errors in a session. Not sure whether this was a one-off.
* In demo app (commit 02b0b30ad4d442ee906b7b4e79d8084ed83bf8df, needs to be removed before merge), I get errors that undo/redo commands cannot be found in `MarkdownContentEditor`. Didn't investigate futher yet.
* In demo app, I cannot type a space in `Editor`. Everything else seems to work (including Emojis, Mentions and Link Picker), but when I type <kbd>Space</kbd>, nothing happens.

README.md
```diff
@@ -57,3 +57,56 @@ Or you might set the `CYPRESS_baseUrl` environment variable for a custom nextclo
...
+
+```
+window.OCA.Text.createEditor({
+	el: document.getElementById('my-editor-div'),
+	fileId: 12345,
+	filePath: '/Readme.md',
+}).then((editor) => {
+	// Once ready you can access the editor instance and call methods like:
+
+	editor.setContent('new content')
```
mejo-, line 87: Maybe put a remark here that this will overwrite the content from the source file?

README.md
```diff
@@ -57,3 +57,56 @@ Or you might set the `CYPRESS_baseUrl` environment variable for a custom nextclo
...
+
+```php
+if (class_exists(LoadEditor::class)) {
+	$this->eventDispatcher->dispatchTyped(new LoadEditor());
+}
+```
+
+### Integrate a file editor
+
+Make sure to check if OCA.Text is available as the text app needs to be enabled. You will need to provide an editor fallback on your own in this case.
```
mejo-, line 76:
```suggestion
Make sure to check if OCA.Text is available as the Text app needs to be enabled. If you want your app to work without Text being installed, you will need to provide an editor fallback on your own.
```

mejo- (approved):
Huge thanks to @juliushaertl for all the work here :heart:

juliushaertl (commented):
Demo still needs to be dropped

Issue #3598, opened 2022-12-21
Treat newlines as hard linebreaks (not paragraph breaks) when pasting plaintext


mejo-:
**Describe the bug**

Pasting plaintext currently transforms each newline into a paragraph break.

**To Reproduce**
1. Open a Text document
2. Copy some multi-line plaintext to the clipboard
3. Paste the plaintext into Text (<kbd>Ctrl</kbd>-<kbd>V</kbd> in Nextcloud 25 and below, <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>V</kbd> in Nextcloud 26)
4. See each newline being converted into a paragraph break.

**Expected behavior**
Newlines should be regarded as hard linebreaks.

susnux:
Should every newline be converted to hard line breaks or should empty lines be converted to paragraph breaks?
E.g.:
```
Hello
world
```
being `<p>Hello<br>world</p>`
vs 
```
Hello

world
```
being `<p>Hello</p><p>World</p>` ?

BTW I think you meant <kbd>CTRL</kbd>+<kbd>V</kbd>?

mejo-:
@susnux that's a good question. Empty newlines should probably be treated as paragraph breaks like in your example.

mejo-:
@susnux @juliushaertl @max-nextcloud do you have an idea where to change this? I might take a look into it the next days but don't have a good idea where to start. So any hint would be appreciated :heart:

juliushaertl:
I think it could basically use the same hook through clipboardTextParser as in https://github.com/nextcloud/text/pull/2716/files

Prosemirror splits them be newlines and wraps in paragraphs, so as soon as we return a slice there we can overrule the default behaviour. This we currently only do for the markdown paste but could easily be done for the plaintext paste with shift as well.

https://github.com/ProseMirror/prosemirror-view/blob/49903c758b510aa937e2f19d16e9ff4bc5bef4f8/src/clipboard.ts#L51-L61

The backportable fix would just skip the markdown conversion.

susnux:
I think changing [Markdown.js#L85](https://github.com/nextcloud/text/blob/master/src/extensions/Markdown.js#L85) to something like this could fix it:
```js
clipboardTextParser(str, $context, _, view) {
	const parser = DOMParser.fromSchema(view.state.schema)
	const doc = document.cloneNode(false)
	const dom = doc.createElement('div')
	if (shiftKey) {
[...]

PR #3623, opened 2022-12-30
Treat newlines as linebreaks when pasting as plaintext

mejo-:
### 📝 Summary

Until now, every newline was transformed into a paragraph break when pasting plaintext.

Instead, newlines will be treated as linebreaks. Two newlines in a row will be treated as paragraph break.

Fixes: #3598

The second commit in this PR adds a change to catch shiftKey when pasting with middle click.

### 🏁 Checklist


src/extensions/Markdown.js
```diff
@@ -78,19 +78,27 @@ const Markdown = Extension.create({
...
-						if (shiftKey) {
-							return
-						}
+						const parser = DOMParser.fromSchema(view.state.schema)
 						const doc = document.cloneNode(false)
 						const dom = doc.createElement('div')
-						dom.innerHTML = markdownit.render(str)
+						if (shiftKey) {
+							// Treat single newlines as linebreaks and double newlines as paragraph breaks when pasting as plaintext
+							dom.innerHTML = '<p>' + str.replaceAll('\n', '<br />').replaceAll('<br /><br />', '</p><p>') + '</p>'
```
max-nextcloud, line 97: I'm wondering if we should ignore trailing new lines.
If i paste an entire line including the trailing new line I would not expect to have a hard break at the end of the inserted paragraph.

max-nextcloud (approved):
Approach looks good. Please check my comment and play with that a bit. I don't know of the top of my head how the code will behave. Feel free to merge as is if the behavior makes sense.

**update:** I did so myself.

src/extensions/Markdown.js
max-nextcloud: I guess the same goes for leading newline chars.

max-nextcloud: I tried it myself... works good enough. Plus this seems like a super rare scenario given that this paste only happens on shift+ctrl+v.
So let's stick to the most generic approach.

PR #3625, opened 2022-12-31
[stable25] Treat newlines as linebreaks when pasting as plaintext

mejo-:
Until now, every newline was transformed into a paragraph break when pasting plaintext.

Instead, newlines will be treated as linebreaks. Two newlines in a row will be treated as paragraph break.

Fixes: #3598

Manual backport of #3623

### 🏁 Checklist


max-nextcloud (approved):
code looks good.

PR #3626, opened 2023-01-01
Save some time on cypress

juliushaertl:
- As we don't really care about the hash mechanism during tests we can use PASSWORD_DEFAULT which is faster than argon2. Saves a bit of time during password authenticated requests.
- Clone server, viewer and build text once and cache the result for the cypress runners


max-nextcloud (approved):
Looks good to me.

Issue #3543, opened 2022-12-09
Files get saved even if they have not been altered


max-nextcloud:
_@meonkeys reported in https://github.com/nextcloud/text/issues/2337#issuecomment-1343249174 that saving files without changes is back_

Files will be saved even if they have not been touched if they contain a bare link link such as: 

```md
https://nitter.net/meonkeys
```

Links such as this are fine:

```md
[Adam Monsen on Twitter, via Nitter](https://nitter.net/meonkeys)
```

Note that this URLs target HTML data includes OpenGraph metadata.

**update:** Headings also lead to the same effect.

Files with zero bare links **and no headings** are correctly left unmodified if I open them and make zero changes. Files with bare links **or headings** *are* marked as modified.

meonkeys:
I was able to repro with a fresh 25.0.2.3 server created with `docker run -d -p 8080:80 --name nextcloud nextcloud`. I opened `Documents/Readme.md`, made no changes, then closed it, and it showed up in the activity feed as "changed".

![Screenshot from 2022-12-14 08-42-52](https://user-images.githubusercontent.com/50639/207655870-a7af5be6-7551-484f-b943-4835b3605f6a.png)

![Screenshot from 2022-12-14 08-43-18](https://user-images.githubusercontent.com/50639/207655894-7401f7d7-dd11-4dc0-84e7-c43076e6d067.png)

Here are the entire contents of `Readme.md`:

```
# Documents

Nextcloud works well with all the common document formats. You can even collaborate with others on ODT and Markdown files!
```

135 bytes. No EOL at the end of the file. SHA-1 checksum is `eb37821f747d23af2524a69892aa5e2b7399a204`, before and after opening the file.

max-nextcloud:
@meonkeys Thanks, I was able to reproduce that too. I tried with another file i created that did not contain a heading and it did not get saved. I was able to reproduce it on another server and again it would only be saved automatically when it contained a heading.

meonkeys:
Will you remove `Link preview` from the title of this bug? Since we have a repro without that.

See also:

#2686 

#2337

Also, not sure if this is related, but I'm seeing an issue when I use [Temporary files lock](https://apps.nextcloud.com/apps/files_lock). I try to edit a file locally, so it is locked. I make a change locally while I am definitely the only one making changes, but I still get a local conflict. Makes me think the root cause is something in the server inappropriately changing file last modified dates (as opposed to in the Text app). Just another data point, I guess. I hope this is helpful.

[...]

PR #3644, opened 2023-01-05
fix: only save if needed when closing

max-nextcloud:
Including tests that open and close do not save unedited files. Test is currently still failing for open.

See #3543.
- [x] Documentation ([README](https://github.com/nextcloud/text/blob/master/README.md) is not required


max-nextcloud:
Okay... turns out a stable sync between client and server can take even longer than 4 roundtrips because the server only saves the doc every so and so often if it's an autosave and the first save was already triggered by the transaction caused by the headings.

max-nextcloud:
This PR misses the point. The server will only persist changes if the version has been bumped. So that should in itself already prevent saving without changes.
However the version gets bumped by process headings - so that's what we need to fix.

Issue #3658, opened 2023-01-12
Add authoritative autoloaders


juliushaertl:
Ref https://github.com/nextcloud/server/pull/36112

Provides some general performance improvement to the instance if more apps make use of this.

PR #3663, opened 2023-01-13
perf: Add authorative autoloader

juliushaertl:
Use a authoritative classmap for the PSR-4 namespace of the text app as this has a small performance gain (the more apps that make use of this, the better for the overall nextcloud performance).

I used a separate composer.json file as the server autoloader will use the composer/ directory to check this and to have separation between dev dependencies as that may get annoying during development if any composer install command would change the apps class map.

Fixes https://github.com/nextcloud/text/issues/3658


marcelklehr (approved):
tests still pass :tada:

PR #3683, opened 2023-01-19
Always provide original GIF file

julien-nc:
Previews of GIF files are static images so we might prefer to always provide the raw file to the editor.


max-nextcloud (approved):
Thanks! :heart_eyes_cat:

PR #3700, opened 2023-01-26
Link picker

juliushaertl:

### 📝 Summary

- [x] Refactor existing suggestion extensions to make code reusable
- [x] Implement a link picker plugiun
  - [x] Prepare extension
  - [x] Fetch link provider list from the backend
  - [x] Trigger link picker on provider
  - [x] Make sure that result gets inserted
  - [x] Adapt mention cypress tests to new component layout

### Depends on

- [x] https://github.com/nextcloud/server/pull/35557
- [x] https://github.com/nextcloud/vue-richtext/pull/863

### 🏁 Checklist


julien-nc:
Awesome!

I pushed a commit adding the glue to use the link picker in your suggestion plugin.

Text needs to be linked with `@nextcloud/vue-richtext` 's "enh/noid/reference-picker" branch.

Remaining issues:
* [ ] Suggestions (provider list) still visible when the link picker modal is displayed
* [ ] If we insert the link with a trailing space, rendering is not triggered. I guess we could insert a link node instead of the raw text (problem: we might get some text meant to be inserted but not as a link).
* [ ] When opening a "search" provider, there is a focus loop issue slowing down the browser for a few seconds. Seems to be a focus trap issue.

juliushaertl:
@julien-nc Pushed a few very minor fixes for the issues you found. The focus trap issue I could not reproduce so I assume it is a side effect of another app that you have enabled on your setup.

julien-nc:
@juliushaertl Nice!
I've pushed a few adjustments:
* make the provider icons smaller to make it consistent with those in RichContentEditable (which size was set to be consistent with the emojis)
* increase suggestion list max-width to 400px because provider names can be long

Are you ok with those changes?

julien-nc:
> @julien-nc Pushed a few very minor fixes for the issues you found. The focus trap issue I could not reproduce so I assume it is a side effect of another app that you have enabled on your setup.

After frantically updating `@nextcloud/vue` in all my apps I realized I hadn't pull viewer for a long time...all good for the focus trap now.

julien-nc:
"Open link picker" action is there. The icon might not be the best one.

juliushaertl:
Ready for review and good to get in from my side 👍 

Let's file follow up tickets for:
- [ ] Avoid to close text on esc key press
- [ ] Cypress tests

mejo-:
Very nice and works well :heart: 

Unfortunately it brings back a bug that I had fixed in 6eb324f144b3761fc153dacc75b14e3e9b35d0cf before: when opening the emoji picker, closing it with <kbd>Esc</kbd> and hitting enter, the emoji that was selected last still gets inserted. That's particularly annoying in languages like french, where single colons are often written.

The same bug exists with mentions (`@`), but there it's less annoying because single @ is not used as often.

julien-nc (commented):
Awesome. The Text operating system is on its way.

Can confirm the bug mentioned by @mejo- for all users, emojis and link providers.

juliushaertl:
Pushed a fix that applies for all autocomplete components now.

mejo- (approved):
Awesome. Works great and the latest fix by @juliushaertl also works :heart:

Issue #3369, opened 2022-11-06
Black table on light theme


XueSheng-GIT:
**Describe the bug**
If light theme is chosen, tables are shown in black.

**To Reproduce**
1. Choose light theme
2. Open a markdown (text) file
3. Create a table

**Expected behavior**
Table should be "normally" visible (black lines/letters on white ground).

**Screenshots**
![Black table](https://user-images.githubusercontent.com/4077683/200181839-5697214f-8c6a-4825-b008-89e3dff07bcf.png)

**Client details:**
 - OS: Ubuntu 22.04
 - Browser: Firefox
 - Version: 106.0.2
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.6.0

**Operating system:** Ubuntu 20.04

**Web server:** apache2 (2.4.54)

**Database:** mariadb (Ver 15.1 Distrib 10.3.34-MariaDB)

**PHP version:** 7.4.32

**Nextcloud version:** 25.0.1

</details>

juliushaertl:
I cannot reproduce this, can you maybe check if this is some browser caching that still sticky to a dark version? Does it also happen when you try loading in a private browser window?

XueSheng-GIT:
Just logged in with a different user on another client (PC) with private browsing (firefox). This time the theme defaulted to light.
Opening the markdown file, the table was shown correctly (white for the light theme). Closing the markdown file, changing the theme to dark and opening the markdown file again result now in the opposite isssue (white table on dark theme).
![Markdown Table](https://user-images.githubusercontent.com/4077683/200250134-63f62826-97a7-4450-842e-7c02e7295c51.png)

This is reproduceable for me and doesn't seem to be related to an old cache on the first view.
@juliushaertl Did you try to change the theme?

juliushaertl:
How and where did you change the theme? I tried with the OS native dark/light mode on macOS and the table properly adjusts there.

XueSheng-GIT:
I used the personal theme settings. It defaults to "system standard" and was manually changed to either light or dark.

![Theme](https://user-images.githubusercontent.com/4077683/200253544-779e30f7-afd0-42b4-906f-ee7ca0c0c74c.png)

XueSheng-GIT:
[...]

PR #3733, opened 2023-02-01
Fix background color of tables

susnux:
### 📝 Summary

* Resolves: #3369

The server css variables are set on the body tag currently, so if we use those variables, e.g. for aliasing them as done here, they need to be evaluated on the same or a child node. This means we can not define aliases on the `:root` element.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
black background on bright theme, unreadable | white background, good readable

### 🏁 Checklist

### DIFF
--- a/css/prosemirror.scss
+++ b/css/prosemirror.scss
@@ -226,14 +226,12 @@ div.ProseMirror {
 	}
 
 	// table variables
-	@at-root :root {
-		--table-color-border: var(--color-border);
-		--table-color-heading: var(--color-text-maxcontrast);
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

mejo- (approved):

mejo-:
Awesome, thanks a lot for looking into this @susnux.

susnux:
/backport 3700a0fed8d579709c3f381f3fc61db1ec9fe057 to stable25

PR #3735, opened 2023-02-01
chore(import): Avoid importing NcModal through other export

juliushaertl:
Otherwise we ship the Modal component twice in the bundles.


juliushaertl:
Screenshot 2023-02-02 at 08 20 46

luka-nextcloud (approved):

Issue #3741, opened 2023-02-02
Odd patterns on sync request


juliushaertl:
Having just one browser window open for a longer time I've noticed that sync seems to run still way too often if nothing is happening:

![Screenshot 2023-02-02 at 21 30 50](https://user-images.githubusercontent.com/3404133/216444040-34f6d659-fa76-4b33-a74c-3260cdf4f32b.png)

Also there seems some way of double fetching still ongoing:

![Screenshot 2023-02-02 at 21 32 03](https://user-images.githubusercontent.com/3404133/216444053-29c7e759-e9a2-426c-b932-4680fd9ad9fb.png)

max-nextcloud:
I think one part of the problem is that awareness messages are also creating steps now. So Polling backend acts as if someone was typing because steps came in and resets the refetch timer:
![grafik](https://user-images.githubusercontent.com/97337118/217035006-a2196bc6-b486-40f0-8a37-17cbebced2a5.png)

With a single user awareness messages are send once per 15 seconds. So the timer goes down to 300, then up to 600, 1200, 2400, 4800 and just when it reached 5k the next awareness step arrives and it starts at 300 again.

One way around this would be to check the message type. Another option might be to reset the refetch timer whenever the document changes - not when steps come in.

juliushaertl:
Right, maybe we can already solve this when looking into #3732 if the awareness state is not a step anymore?

max-nextcloud:
#3776 fixed this with respect to the awareness messages. They are now transported via the session data rather than the steps so the refetchTimer won't be reset anymore because of awareness messages.

If new odd patterns in sync show up please reopen this issue or file a new one.

PR #3769, opened 2023-02-06
fix: collaborator disconnect timeout

max-nextcloud:
Perform all computations / comparisons in milliseconds.

The collaborator disconnect timeout is in ms as are all intervals.
So we can directly subtract it from `Date.now()` - which also is in ms.
The only time that is not expressed in ms is the last contact timestamp.

See also #3741

### 🏁 Checklist

- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) - this is a first take at a minimal fix. It's somewhat hard to test in the current state. Will look into refactoring later.


max-nextcloud:
Even though this improves the behavior of the `PollingBackend` by increasing the timeout to 5s right away it still alternates between 5sec. and 300ms (after every message received from the user itself.)

juliushaertl (approved):

PR #3770, opened 2023-02-06
Fix focus becoming impossible to get after escape pressed in the link picker modal

julien-nc:
requires https://github.com/nextcloud/vue-richtext/pull/897 (will most likely be included in the next vue-richtext release)

We can now tell the link picker when it is created while the Viewer is displayed so an event is emitted to deal with the focus trap issue.


max-nextcloud (approved):
I was not able to reproduce the behavior described as without this fix escape would close the entire viewer for me. Maybe I'm running an outdated server or viewer version.

With this fix the behavior seems correct.

Code changes seem reasonable as well. Thanks :tada:

juliushaertl (approved):

Issue #3513, opened 2022-12-04
Prefix heading ids to prevent naming conflicts


arnowelzel:
## This is a https://github.com/nextcloud/text/labels/good%20first%20issue if you...

* have a basic understanding of javascript.
* have a working nextcloud development environment to test your changes.

If you would like some guidance feel free to ping @max-nextcloud or join the [public text channel](https://cloud.nextcloud.com/call/9tq2th3p).

## Problem
If you enter the text "Nextcloud" and format as heading, it will be indented.
This is because css that targets `#nextcloud` is applied to it.

## Solution
* Add a `heading-` prefix to the id in html to avoid id naming conflicts in https://github.com/nextcloud/text/blob/main/src/nodes/Heading/HeadingView.vue#L22
* Make sure the links in the outline **do not** have the prefix and still work.

## To Reproduce
1. Create a new markdown file
2. Open the file in the text editor
3. Enter "# Nextcloud"
4. The word "Nextcloud" will be outpus as heading but also be indendet.

## Expected behavior
The word "Nextcloud" will just be a heading without indentation.

## Screenshots
![image](https://user-images.githubusercontent.com/6613614/205507777-005be970-4c9b-4c80-8381-27d0cef2fe54.png)

## Context
<details>
<summary>Client details</summary>
 - OS: Windows 10
 - Browser: Vivaldi 5.5.2805.50, Firefox 107.0.1
 - Device: desktop
</details>

 - Nextcloud version: 25.0.1

## Side note
https://github.com/nextcloud/text/issues/3513#issuecomment-1336485033 outlines a workaround which should be ignored for resolving the matter but can be used by anyone finding this issue and looking for a temporary solution until a fix is shipped with the test app itself.

arnowelzel:
A workaround is to add the following CSS using the ["Custom CSS"](https://apps.nextcloud.com/apps/theming_customcss) app. It seems there is a global style for `#nextcloud` which adds a padding to the left for every element with the ID "nextcloud":

```
.text-editor__wrapper div.ProseMirror #nextcloud {
  padding-left: 0 !important;
}
```

juliushaertl:
Thanks for reporting.

Seems we need some kind of prefix to avoid id naming conflicts in https://github.com/nextcloud/text/blob/master/src/nodes/Heading/extractor.js#L13

max-nextcloud:
:clap: such a good catch! Congratulations.

max-nextcloud:
[...]

PR #3775, opened 2023-02-07
fix: prefix heading ids

snehar97:
### 📝 Summary

* Resolves: #3513
prefix heading ids to prevent naming conflicts.

#### 🖼️ Screenshots
Capture1

### 🏁 Checklist


max-nextcloud:
@snehar97 thanks for your pull request.
The node test was failing - but that's to be expected because the javascript needs to be compiled on the CI.
I triggered the compilation and also aborted the cypress tests as they will be run after the compilation again anyway.

max-nextcloud:
Okay... looks like i cannot trigger the compile command on someone elses branch. I started the cypress run anyway now so we can see how that goes.

julien-nc:
> Okay... looks like i cannot trigger the compile command on someone elses branch. I started the cypress run anyway now so we can see how that goes.

@max-nextcloud It's because the branch is in a fork of which you're not a member.

src/nodes/Heading/HeadingView.vue
```diff
@@ -19,7 +19,7 @@
   -->
 
 <template>
-	<NodeViewWrapper :id="node.attrs.id"
+	<NodeViewWrapper :id="'heading-' + node.attrs.id"
```
susnux, line 22: I think we should not change the ID only on the node view, but directly on the node props.
So probably here:
https://github.com/nextcloud/text/blob/33ae28d3444195ff2a18610eeaccaecde26d66d5/src/nodes/Heading/extractor.js#L27

So that the ID property of the node is the real node and one would not have to check multiple files to get the Dom id/ have multiple IDs.

susnux:
Thanks for looking into this 😊
I left a comment, as I think this should be fixed in the node rather than the node view.
But you will have to adjust the cypress tests as well, as they expect the old heading ids:
https://github.com/nextcloud/text/blob/main/cypress/e2e/sections.spec.js

src/nodes/Heading/HeadingView.vue
max-nextcloud: I was thinking the same initially. But then i figgured that this would also change the `href` attribute of the resulting tag.
I'd prefer to keep the `href` as is because `Our-Garden#tomatoes` looks better to me then `Our-Garden#heading-tomatoes`.

susnux: That's the way how anchors work: The href **must** match the ID of the heading so that the browser scrolls to the element.

max-nextcloud: Ohhh... you are right. I was confused there. :facepalm:

susnux: > ... because `Our-Garden#tomatoes` looks better to me then `Our-Garden#heading-tomatoes`

If this is a problem, maybe then use a shorter prefix, like `h-` or similar.
But after releasing this we should try to **not** change the prefix as this would annoy many users, for example if you set up links to sections all links are now invalid any you have to change every link.

mejo-: I like the idea to change the prefix to `h-` to make it shorter. What do you think @snehar97? Would you be up to this change?

Issue #3732, opened 2023-02-01
Only store last awareness messages on the backend


max-nextcloud:
There's no need to keep all the awareness messages around. Each of them contains the entire awareness state.
Put awareness messages into a separate table and only keep one entry for each session.

juliushaertl:
@max-nextcloud Could you maybe take some time to look into this one? Seems like this could be quite an improvement if that works easily.

PR #3776, opened 2023-02-07
enh: store awareness messages separately

max-nextcloud:
### 📝 Summary

* Resolves: #3732 

### 🚧 TODO

- [x] fix migration

### 🏁 Checklist


src/services/WebSocketPolyfill.js
```diff
@@ -88,19 +88,33 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 		}
 
 		#initiateSending() {
-			let steps
+			let outbox
 			syncService.sendSteps(() => {
-				steps = this.#queue.map(s => encodeArrayBuffer(s))
+				outbox = this.#queue
```
juliushaertl, line 93: When testing it seems that debouncing needs to be improved now for the push request, it seems on my test with two browsers the push is triggering every 200ms through the following stack trace even if the cursor positions do not change:

Screenshot 2023-02-08 at 12 10 59

lib/Service/ApiService.php
```diff
@@ -184,11 +184,12 @@ public function close($documentId, $sessionId, $sessionToken): DataResponse {
 	 * @throws NotFoundException
 	 * @throws DoesNotExistException
 	 */
-	public function push($documentId, $sessionId, $sessionToken, $version, $steps, $token = null): DataResponse {
+	public function push($documentId, $sessionId, $sessionToken, $version, $steps, $awareness, $token = null): DataResponse {
 		if (!$this->sessionService->isValidSession($documentId, $sessionId, $sessionToken)) {
 			return new DataResponse([], 403);
 		}
 		$session = $this->sessionService->getSession($documentId, $sessionId, $sessionToken);
+		$this->sessionService->updateSessionAwareness($documentId, $sessionId, $sessionToken, $awareness);
```
juliushaertl, line 192: Would be good to return early here if no steps are passed along.

src/services/WebSocketPolyfill.js
max-nextcloud: I suspect this is due to us replaying old awareness messages and creating an inconstent state.
I created #3785 to track it.

juliushaertl (approved):
👍 

Also tested to collaborate a bit with @miaulalala and no collaboration issues encountered 🚀

PR #3789, opened 2023-02-11
fix(SessionApi): only send content if no session is open

max-nextcloud:
If a session is open - even if no state has been saved yet the content of the document will be recovered from the steps.

In this case do not send the content in the response to the create request as it would be duplicated:
* once created when the editor is initialized with the content
* then again when the steps are replayed.

Include tests for this corner case and another one.
Fixes the [failing cypress tests in collectives](https://cloud.cypress.io/projects/b6myzg/runs/266/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github).

### 🏁 Checklist

- [ ] Documentation is not required


max-nextcloud:
Interesting - this actually breaks the frontmatter and conflict specs. Will need to investigate. Code reviews would still be appreciated. :)

juliushaertl (approved):

PR #3842, opened 2023-02-27
Fix document state issues

max-nextcloud:
### 📝 Summary

* Use a different naming scheme than the old initial document to not trap over old files
* Only make use of the version if document state is provided.

- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing.


lib/Service/DocumentService.php
```diff
@@ -152,17 +152,45 @@ public function createDocument(File $file): Document {
...
-		return $this->appData->getFolder('documents')->getFile((string) $document);
+		return $this->appData->getFolder('documents')->getFile($filename);
+	}
+
+	/**
+	 * @param $documentId
+	 * @return ISimpleFile
+	 * @throws NotPermittedException
+	 */
+	public function createStateFile($documentId): ISimpleFile {
```
mejo-, line 172: Would be nice to have the parameters typed.

```suggestion
	public function createStateFile(int $documentId): ISimpleFile {
```

Also `@param int $documentId` in the function annotations.

lib/Service/DocumentService.php
```diff
@@ -152,17 +152,45 @@ public function createDocument(File $file): Document {
...
+		$filename = (string)$documentId . '.yjs';
+		return $this->appData->getFolder('documents')->newFile($filename);
+	}
+
+	/**
+	 * @param $documentId
+	 * @param $content
+	 */
+	public function writeDocumentState($documentId, $content): void {
```
mejo-, line 181: Same here:

```suggestion
	public function writeDocumentState(int $documentId, string $content): void {
```

lib/Service/DocumentService.php
```diff
@@ -152,17 +152,45 @@ public function createDocument(File $file): Document {
...
+		try {
+			$documentStateFile = $this->getStateFile($documentId);
+		} catch (NotFoundException $e) {
+			$documentStateFile = $this->createStateFile($documentId);
+		} catch (NotPermittedException $e) {
+			$this->logger->error('Failed to create document state file', ['exception' => $e]);
+			return;
```
mejo-, line 188: Is it intended that `writeDocumentState()` doesn't fail in case of `NotPermittedException`? I'm just unsure here.

mejo- (approved):
Code looks good to me. Just minor comments.

lib/Service/DocumentService.php
max-nextcloud: Yes... it's an optimization - if we cannot write the state file it's most likely because we cannot create the documents folder or so.

That's a problem with the setup that will prevent using the document state to load an editing session - but text can recover from that. So no use failing hard here I think.

juliushaertl (approved):

PR #3843, opened 2023-02-27
Add repair step to reset all sessions before Yjs migration

mejo-:
Text sessions with data from our old collaboration backend cause issues after the upgrade to Text with the new Yjs backend. Therefore let's force-reset all sessions when upgrading from a Text version smaller than 3.7.1.

### 🏁 Checklist

### DIFF
--- a/lib/Migration/ResetSessionsBeforeYjs.php
+++ b/lib/Migration/ResetSessionsBeforeYjs.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace OCA\Text\Migration;
+
+use OCA\Text\Db\SessionMapper;
+use OCA\Text\Service\DocumentService;
+use OCP\IConfig;
+use OCP\Migration\IOutput;
+use OCP\Migration\IRepairStep;
+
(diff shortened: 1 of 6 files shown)


lib/Migration/ResetSessionsBeforeYjs.php
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * @param IOutput $output
+	 *
+	 * @return void
+	 */
+	public function run(IOutput $output): void {
+		$appVersion = $this->config->getAppValue('text', 'installed_version');
+
+		if (!$appVersion || version_compare($appVersion, '3.7.1') !== -1) {
```
juliushaertl, line 39: Should we still do this once also for instances upgraded to 26 during beta phase? Then I'd to a increase to 3.7.2 in appinfo and use that instead

mejo-: Good point! Changed now.

juliushaertl (approved):

PR #3850, opened 2023-02-28
Update `@nextcloud/vue-richtext` to avoid circular dependency issue

julien-nc:
This issue basically makes it impossible to use `@nextcloud/vue-richtext` in a script which also imports `@nextcloud/vue` >= 7.0.0 (when we introduced the circular dependency).


mejo- (approved):

juliushaertl (approved):

julien-nc:
No clue why it fails.

max-nextcloud:
https://github.com/nextcloud/text/actions/runs/4295726255/jobs/7486535732 says:

```
ERROR in ./node_modules/@nextcloud/vue-richtext/dist/index.js 16072:15-67
Module not found: Error: Can't resolve '@nextcloud/calendar-js' in '/home/runner/work/text/text/node_modules/@nextcloud/vue-richtext/dist'
```

and

```
ERROR in ./node_modules/@nextcloud/vue-richtext/dist/index.js 16775:54-79
Module not found: Error: Can't resolve 'linkify-string' in '/home/runner/work/text/text/node_modules/@nextcloud/vue-richtext/dist'
resolve 'linkify-string' in '/home/runner/work/text/text/node_modules/@nextcloud/vue-richtext/dist'

```

I suspect this marks them as external:
https://github.com/nextcloud/vue-richtext/blob/master/vite.config.js#L18

So they need to be provided as peer dependencies by whatever uses vue-richtext.

max-nextcloud:
> I suspect this marks `@nextcloud/calendar-js` and `linkify-string` as external.

It does not. They are not even in the `dependencies` - quite the contrary. They are only dependencies of the `@nextcloud/vue` dependency which is explicitely not treated as external.

:shrug:

juliushaertl:
Even weirder that the build passes locally without any issues :/

julien-nc:
I can reproduce locally. This happens when text is not inside a server clone. Apparently `@nextcloud/calendar-js` and `linkify-string` are provided by the server's node_modules.

juliushaertl:
I think this is due to the bad tree shaking abilities of web pack. I currently see two workarounds:
- We add both dependencies temporarily to text
- Switching all imports in @nextcloud/vue-richtext to use the dist/ folder of @nextcloud/vue - This seems to work locally but I have no good explanation yet on what the difference is

julien-nc:
We can also wait for #3859 which is a better fix anyway.

mejo-:
Should be good to close now that #3859 landed, right?

PR #3856, opened 2023-03-01
Widget fixes

juliushaertl:
- fix: Limit paragraph spacing to the content
- fix: Only attempt to show reference widgets for logged in users

### 📝 Summary

Avoid rendering issues with reference previews on paragraphs and public share links.


mejo- (commented):
Code looks good to me.

mejo- (approved):

mejo-:
I wonder though why the Node CI job didn't fail - this PR contains changes to the JS sources that should require a compile, no? :thinking:

juliushaertl:
Indeed, I first had suspected https://github.com/nextcloud/.github/issues/124 but we only have js/scss files here.

juliushaertl:
Ah I think it is because css/* changes are also not part of the path checks for node. :/

PR #3859, opened 2023-03-01
Use NcRichText component from `@nextcloud/vue`

julien-nc:
:tada: Circular dependency is definitely going down with the next `@nextcloud/vue` release which will includes `NcRichText` and everything needed for the link picker and the reference widgets.

refs https://github.com/nextcloud/nextcloud-vue/pull/3841 https://github.com/nextcloud/nextcloud-vue/issues/3812

I'll update `@nextcloud/vue` and set this to "ready for review" once the new release is out.


mejo- (changes requested):
Given that we import files from `@nextcloud/vue/dist/.*` now, we need to add a rule to `vite.config.js`:

```diff
diff --git a/vite.config.js b/vite.config.js
index 0b7885ce8..a1ee24e76 100644
--- a/vite.config.js
+++ b/vite.config.js
@@ -21,6 +21,7 @@ export default defineConfig({
                                /markdown-it\/.*/,
                                /@tiptap\/pm\/.*/,
                                /vue-material-design-icons\/.*/,
+                               /@nextcloud\/vue\/.*/,
                        ],
                        output: {
                                globals: { vue: 'Vue' },
```

If you want me to, I can push a commit to this PR.

julien-nc:
@mejo- I don't exactly know the implications of doing that but fully trust your judgement :grin:.
Yes you can add a commit to this PR.

mejo-:
> @mejo- I don't exactly know the implications of doing that but fully trust your judgement grin. Yes you can add a commit to this PR.

The reason for that is that we only treat imports that exactly match names of our dependencies from `package.json` as external in vite per default - all other imports lead to the imported files are vendored into the resulting JS asset by `vite build`. Therefore we need to list regexes for direct imports of files from our dependencies explicitely. See PR #3821for further details.

mejo-:
I rebased on current main and pushed my commit.

Issue #3785, opened 2023-02-09
Prevent awareness message avalanche


max-nextcloud:
When testing it seems that debouncing needs to be improved now for the push request, it seems on my test with two browsers the push is triggering every 200ms through the following stack trace even if the cursor positions do not change:

<img width="1213" alt="Screenshot 2023-02-08 at 12 10 59" src="https://user-images.githubusercontent.com/3404133/217513863-136b06b8-7e5e-4b3a-ac8b-9508003543d1.png">

_Originally posted by @juliushaertl in https://github.com/nextcloud/text/pull/3776#discussion_r1099987022_

PR #3868, opened 2023-03-02
fix: only load editor once even if connection is recovered

max-nextcloud:
### 📝 Summary

Only load the editor if it has not been loaded yet.
~~Move `createEditor` out of the `onLoaded` function.~~

`onLoaded` may be called multiple times.
When the connection to the server is lost and restored the collaboration extension reloads the websocket polyfill which in turn opens the SyncService
which will trigger the `onCreated` and `onLoaded` callbacks.

* Resolves: #3785 

### 🚧 TODO

When loosing the connection the yjs websocket provider will try to reconnect automatically
However we still show the reconnect button. The automatic reconnection succeeded for me - but the button fails.

- [ ] Hide the reconnect button in that state
- [ ] Properly disconnect after a timeout and show reconnect dialogue then
- [ ] Make sure the reconnect actually works

### 🏁 Checklist


max-nextcloud:
This PR includes a test that cancels some requests to simulate a connection loss.
I was able to trigger the same behavior locally by stopping the nextcloud-docker-dev environment and restarting it a few seconds later. That way it's probably easier to explore this in the browser oneself.

max-nextcloud:
We will need to work on the TODO items listed above. However I think it would still be good to land this today before the RC to mitigate some of the problems we are seeing on the server right now.

src/services/SyncService.js
```diff
@@ -98,8 +98,8 @@ class SyncService {
 
 		// TODO: Only continue if a connection was made
 		this.connection = await connect
-
-		this.version = this.connection.docStateVersion
+		this.backend = new PollingBackend(this, this.connection)
+		this.version = this.connection.lastSavedVersion
```
juliushaertl, line 102: Was this reverted on purpose from https://github.com/nextcloud/text/pull/3842/files#diff-103df33bb085aaec0d6134982c312e96ecf58cf09335b9cf15964e17a45b4071 ?

max-nextcloud: good catch... no that's left from the rebase!

max-nextcloud: good catch... no that's left from the rebase!

max-nextcloud:
Okay... found a **much** simpler way of basically achieving the same.

One issue remains though:
After the automatic session recovery the old session will still be used for the mention extension.
Therefore mentions will not work anymore.

juliushaertl (approved):

mejo- (approved):
Code looks good to me :ok_hand:

I don't fully understand the impact of moving up the initialization of `this.backend` above the emits.

Issue #3871, opened 2023-03-02
Use href of links instead of `node.textContent` for link previews


mejo-:
**Describe the bug**
Currently we use `node.textContent` to determine whether a paragraph is a link that warrants a link preview. Instead, we should check whether the paragraph has a link mark and use its href.

Reasons for this change would be:
* We would make sure to always show the preview of the *link target*, not of the description text. Both may differ, which has security implications.
* Links with a custom description would get a link preview as well.
* Apparently url-encoded spaces in `textContent` of links get decoded when transformed to markdown. Spaces are not regarded in [the regex that checks for URLs](https://github.com/nextcloud/text/blob/905310124d45da86e5f37c596d67eef568af9d4a/src/nodes/ParagraphView.vue#L78). Therefore URLs with spaces loose their link preview once the Text session is closed.

I'll take a look into implementing this change.

@julien-nc would be interested in your opinion on that.

PR #3873, opened 2023-03-02
Use the href of links for link previews, not `node.textContent`

mejo-:
Until now we used node.textContent to determine whether a paragraph is a link that warrants a link preview. Instead, we now check whether the paragraph has a single text node wich is a link and use its href.

Text nodes with empty textContent are ignored in order to allow whitespaces before and after the link.

This way we ensure to always show the preview of the link target, not of the description text. Both may differ, which has security implications.

Also, links with a custom description get a link preview as well.

And last but not least, it fixes link previes for URLs with spaces. (Background: for some reason, url-encoded spaces in textContent of links get decoded when they're transformed to markdown and written to a file. Therefore URLs with spaces lost their link preview once the Text session was closed prior to this commit)

Fixes: #3871

### 🏁 Checklist


juliushaertl (approved):
Nice, thanks for the quick fixing :)

mejo-:
/backport 35db3d4d96f8bd68be07ef3da8d59d00f181c61f to stable25

PR #3881, opened 2023-03-03
chore: Add Nextcloud 27 support on main

nickvergessen:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@nextcloud/text",
   "description": "Collaborative document editing",
-  "version": "26.0.0-alpha.7",
+  "version": "26.0.0-alpha.8",
   "authors": [
     {
       "name": "Julius Härtl",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@nextcloud/text",
   "description": "Collaborative document editing",
-  "version": "26.0.0-alpha.7",
+  "version": "3.8.0",
```
juliushaertl, line 4: Needs double checking with @mejo- about the npm package that is still being built for collectives

mejo-: Yeah, better to keep it as is for now or bump it to `27.0.0-alpha.1`. `3.8.0` would be lower than the current version. And we need to do a new Text NPM release soon anyway in order to fix a few bugs in Collectives.

mejo-: Changed to `26.0.0-alpha.8` for now.

mejo- (approved):

PR #3893, opened 2023-03-05
fix: only initialize editor once

max-nextcloud:
### 📝 Summary

When we recover a session the editor already exists.
Make sure editing can be resumed:
* Start the syncing again as the polling backend is created from scratch.
* Do not attach further events to the editor
* Do not reset the content.

* Resolves: [Test failures in sync spec](https://cloud.cypress.io/projects/hx9gqy/runs/8877/test-results/0a44c2b1-2e31-465f-9229-9d6a21c1638a?statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D)


mejo- (approved):
Code looks sensible to me.

mejo-:
Is the failing Cypress test the one you would expect to have fixed with this PR?

>   1) Sync
>        recovers from a lost connection:
>      AssertionError: Timed out retrying after 10000ms: expected '## Hello world\n\n- Saving the doc saves the doc state\n- ' to include 'after the lost connection'
>       at Context.eval (webpack://@nextcloud/text/./cypress/e2e/sync.spec.js:92:4)

max-nextcloud:
> Is the failing Cypress test the one you would expect to have fixed with this PR?

Yes. I saw it fail occasionally locally - but three failures in a row is way too much. Will need to investigate some more.

juliushaertl (approved):

max-nextcloud:
I made the tests more robust i think. Let's see if they turn flaky eventually. The actual fix to the underlying problem would be #3899 .

juliushaertl:
back ported manually with #3900

PR #3883, opened 2023-03-03
chore(CI): Adjust testing matrix for Nextcloud 26 on stable26

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #3894, opened 2023-03-05
chore(ci): Set server branch to `stable26` in Cypress workflow

mejo-:


mejo-:
Pushed to https://github.com/nextcloud/text/pull/3883 instead.

Issue #3875, opened 2023-03-02
 Session is invalid exception should only be a warning/info level


juliushaertl:
```
[index] Error: OCP\AppFramework\Db\DoesNotExistException: Session is invalid at <<closure>>

0. /var/www/cloud.nextcloud.com/nextcloud/apps/text/lib/Service/SessionService.php line 242
   OCA\Text\Db\SessionMapper->find("*** sensitive parameters replaced ***", "*** sensitive parameters replaced ***", "*** sensitive parameters replaced ***")
1. /var/www/cloud.nextcloud.com/nextcloud/apps/text/lib/Service/ApiService.php line 197
   OCA\Text\Service\SessionService->updateSessionAwareness("*** sensitive parameters replaced ***")
2. /var/www/cloud.nextcloud.com/nextcloud/apps/text/lib/Controller/PublicSessionController.php line 86
   OCA\Text\Service\ApiService->push("*** sensitive parameters replaced ***")
3. /var/www/cloud.nextcloud.com/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 230
   OCA\Text\Controller\PublicSessionController->push("*** sensitive parameters replaced ***")
4. /var/www/cloud.nextcloud.com/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 137
   OC\AppFramework\Http\Dispatcher->executeController(["OCA\\Text\\Con ... "], "push")
5. /var/www/cloud.nextcloud.com/nextcloud/lib/private/AppFramework/App.php line 183
   OC\AppFramework\Http\Dispatcher->dispatch(["OCA\\Text\\Con ... "], "push")
6. /var/www/cloud.nextcloud.com/nextcloud/lib/private/Route/Router.php line 315
   OC\AppFramework\App::main("OCA\\Text\\Cont ... r", "push", ["OC\\AppFramewo ... "], ["text.PublicSession.push"])
7. /var/www/cloud.nextcloud.com/nextcloud/lib/base.php line 1050
   OC\Route\Router->match("/apps/text/public/session/push")
8. /var/www/cloud.nextcloud.com/nextcloud/index.php line 36
   OC::handleRequest()

POST /apps/text/public/session/push
```

PR #3901, opened 2023-03-06
fix: 403 when session is closed during push

max-nextcloud:
### 📝 Summary

* Resolves: #3875 

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -194,20 +194,31 @@ class ApiService {
 			return new DataResponse([], 403);
 		}
 		$session = $this->sessionService->getSession($documentId, $sessionId, $sessionToken);
-		$this->sessionService->updateSessionAwareness($documentId, $sessionId, $sessionToken, $awareness);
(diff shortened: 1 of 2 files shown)


lib/Service/ApiService.php
```diff
@@ -194,20 +194,31 @@ public function push($documentId, $sessionId, $sessionToken, $version, $steps, $
 			return new DataResponse([], 403);
 		}
 		$session = $this->sessionService->getSession($documentId, $sessionId, $sessionToken);
-		$this->sessionService->updateSessionAwareness($documentId, $sessionId, $sessionToken, $awareness);
+		if (!$session) {
+			return new DataResponse([], 403);
+		}
+		try {
+			$this->sessionService->updateSessionAwareness($documentId, $sessionId, $sessionToken, $awareness);
```
juliushaertl, line 201: In the long term maybe we should only fetch the session once and pass it along on further calls, but for now this change seems good 👍

juliushaertl (approved):

PR #3908, opened 2023-03-07
Avoid cleanup on close to not run into race conditions with other sessions on the same document

juliushaertl:
- Add more debug logging around document handling
- Add logging for saving empty files (replaces #3872)
- fix: Rework save locking
- fix: Avoid document cleanup when closing sessions
- fix: Make sure to still cleanup attachments on document reset

## Follow up

- The database showed some leftovers which I cleaned up manually but we should tackle that with https://github.com/nextcloud/text/issues/3915 at some point
- https://github.com/nextcloud/text/issues/3916

### DIFF
--- a/lib/Cron/Cleanup.php
+++ b/lib/Cron/Cleanup.php
@@ -79,7 +79,7 @@ class Cleanup extends TimedJob {
 				$this->logger->error('Document ' . $session->getDocumentId() . ' has not been reset, as an error occured', ['exception' => $e]);
 			}
 		}
-		$removedSessions = $this->sessionService->removeInactiveSessions(null);
+		$removedSessions = $this->sessionService->removeInactiveSessionsWithoutSteps(null);
 		$this->logger->debug('Removed ' . $removedSessions . ' inactive sessions');
 	}
 }
--- a/lib/Cron/Cleanup.php
+++ b/lib/Cron/Cleanup.php
@@ -28,8 +28,6 @@ declare(strict_types=1);
 
 namespace OCA\Text\Cron;
 
-use OCA\Text\Db\Session;
-use OCA\Text\Exception\DocumentHasUnsavedChangesException;
 use OCA\Text\Service\DocumentService;
 use OCA\Text\Service\AttachmentService;
 use OCA\Text\Service\SessionService;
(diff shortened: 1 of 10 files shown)


max-nextcloud (approved):
All looks good to me. Thanks a lot for taking care of this!

juliushaertl:
@max-nextcloud I pushed two more commits to still unlock the file on close and some cleanup of resetDocument

juliushaertl:
I'd still like to introduce locking around document creation and resetting but would do this in a follow up to also take handling this in the frontend properly into account. My idea would be to:

- Wrap each operation create/sync/push/close with a shared lock
- Cover document creation and reset with an exclusive lock
- Any API call that cannot obtain the shared lock (meaning that there is a create/reset currently ongoing) should fail with a 423 Locked status
- The document can then only be reset if no other request is currently working with it (in addition to the existing check for active sessions)
- Helps to avoid issues if two users open a file where the document creation is triggered

lib/Service/ApiService.php
```diff
@@ -113,26 +112,15 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
...
-					$this->attachmentService->cleanupAttachments($file->getId());
-				} catch (DocumentHasUnsavedChangesException $e) {
-					$this->logger->debug('Failed to reset document during session create for ' . $file->getId(), [
-						'remainingSessions' => $remainingSessions,
-					]);
-				}
+			$document = $this->documentService->getDocument($file);
+			$freshSession = $document === null;
+
+			if ($document === null) {
```
mejo-, line 118: Maybe use `if ($freshSession) {` here for better readability?

lib/Service/ApiService.php
```diff
@@ -113,26 +112,15 @@ public function create($fileId = null, $filePath = null, $token = null, $guestNa
 			$readOnly = $this->documentService->isReadOnly($file, $token);
 
 			$this->sessionService->removeInactiveSessionsWithoutSteps($file->getId());
-			$remainingSessions = $this->sessionService->getAllSessions($file->getId());
-			$freshSession = false;
-			if ($forceRecreate || count($remainingSessions) === 0) {
```
mejo-, line 118: Seems like `$forceRecreate` is no longer used anywhere. Is there no need for it anymore? Then I'd also remove it from the function parameters.

lib/Cron/Cleanup.php
```diff
@@ -57,28 +55,24 @@ public function __construct(ITimeFactory $time,
...
-				$this->logger->error('Document ' . $session->getDocumentId() . ' has not been reset, as an error occured', ['exception' => $e]);
+			if ($this->documentService->hasUnsavedChanges($document)) {
+				continue;
 			}
+
+			$this->documentService->resetDocument($document->getId());
 		}
+
+		$this->logger->debug('Run cleanup job for text sessions');
 		$removedSessions = $this->sessionService->removeInactiveSessionsWithoutSteps(null);
```
mejo-, line 76: Could it be that the document doesn't have sessions left after `removeInactiveSessionsWithoutSteps`? Then we should move this in front of the code that resets documents without sessions, no?

lib/Cron/Cleanup.php
```diff
@@ -57,28 +55,24 @@ public function __construct(ITimeFactory $time,
...
-		$this->logger->debug('Run cleanup job for text sessions');
-		$inactive = $this->sessionService->findAllInactive();
-		/** @var Session $session */
-		foreach ($inactive as $session) {
-			$activeSessions = $this->sessionService->getActiveSessions($session->getDocumentId());
-			if (count($activeSessions) > 0) {
+		$this->logger->debug('Run cleanup job for text documents');
+		$documents = $this->documentService->getAll();
+		foreach ($documents as $document) {
+			$allSessions = $this->sessionService->getAllSessions($document->getId());
```
max-nextcloud, line 61: I'd think we could even do this for all documents with no *active* Sessions. But maybe we can also do that later.

lib/Cron/Cleanup.php
juliushaertl: I haven't touched those intentionally for now to have an extra buffer of 1 cronjob run to clean up documents where inactive sessions have been removed. Still worth to check in a follow up.

lib/Cron/Cleanup.php
juliushaertl: Thinking about this again, wouldn't this bring back the problem where the state is reset while a session is inactive but opens again later? However I think #3921 would be the better way of handling this case and then we could also cleanup more aggressively

PR #3909, opened 2023-03-07
Get mimeIconUrl for media attachments without a session

mejo-:
### 📝 Summary

In order to have non-image attachments rendered in editors without a session, `AttachmentResolver.resolve()` should return a candidate with the mimeType icon as url as last resort - because all endpoints from Text `AttachmentController` require a session. This fixes rendering of non-image attachments in RichTextReader.
    
Fixes: #2919

### 🏁 Checklist


src/components/ImageView/ShowImageModal.vue
```diff
@@ -33,7 +33,7 @@ export default {
 			validator(imagesList) {
 				return (imagesList.length === 0)
 					? true
-					: imagesList.findIndex(({ basename, source }) => !(basename && source)) !== -1
+					: imagesList.findIndex(({ basename, source }) => !(basename && source)) === -1
```
mejo-, line 36: @luka-nextcloud could you take a look here? I'm not sure whether I got the logic of the code right, but it seemed to me that you wanted to check if all images in `imageList` have both `basename` and `source` set, right?

src/services/AttachmentResolver.js
```diff
@@ -249,6 +260,12 @@ export default class AttachmentResolver {
 		return pathNormalize(f)
 	}
 
+	async #getMimeUrl(src) {
+		const headResponse = await axios.head(src)
```
mejo-, line 264: When getting the HEAD response, we could actually store more metadata (in particular the filesize) that can be reused in https://github.com/nextcloud/text/blob/0f6805c90210699390f639fa285764d83056b127/src/nodes/ImageView.vue#L46.

My current take would be to register a `metadata` key to the `results` entry from `resolve` above and check at https://github.com/nextcloud/text/blob/0f6805c90210699390f639fa285764d83056b127/src/nodes/ImageView.vue#L315 whether that's set before doing the extra request. What do others think?

src/components/ImageView/ShowImageModal.vue
luka-nextcloud: The condition (with !==) returns false if we find any entry which doesn't have basename and/or source set. Please look carefully, the condition is evaluated on the value returned by findIndex. So it should be !== I think.

mejo-: Mh, not sure whether I can follow. You want this statement to return `true` if all children of `imagesList` have `basename` and `source` set, right? And if one of the children misses one of the two keys, you want it to return `false`, right?

So this was my test:

```js
[{ basename: 'a', source: 'b' }].findIndex(({ basename, source }) => !(basename && source)) === -1 // returns true
[{ basename: 'a' }].findIndex(({ basename, source }) => !(basename && source)) === -1 // returns false
```

vs.

```js
[{ basename: 'a', source: 'b' }].findIndex(({ basename, source }) => !(basename && source)) !== -1 // returns false
[{ basename: 'a' }].findIndex(({ basename, source }) => !(basename && source)) !== -1 // returns true
```

mejo-: To put your statement in words: `image.findIndex(...)` returns `-1` if no child is found where `basename` or `source` are unset. So the check for `image.findIndex(...) !== -1` is true if all children are valid, no?

src/services/AttachmentResolver.js
mejo-: Done this now.

src/components/ImageView/ShowImageModal.vue
max-nextcloud: Let me propose yet another approach. You could either use the `.some` or `.every` functions on Array:
```suggestion
					: imagesList.every(image => image.basename && image.source)
```

I think you can even leave out the special handling for the empty list because `[].every(...)` is always true.

max-nextcloud (approved):

luka-nextcloud (approved):

src/components/ImageView/ShowImageModal.vue
mejo-: Thanks, much easier to parse @max-nextcloud. I changed the code to your suggestion.

Issue #3849, opened 2023-02-28
Opening non-image attachments is broken


mejo-:
**Describe the bug**
Opening non-image attachments is broken in latest Text. I'd expect them to be opened in the Viewer, but instead a modal is displayed that tries to load them as image and displays them as broken image.

Probably related: https://github.com/nextcloud/text/pull/2924 @luka-nextcloud

**To Reproduce**
1. Open a Text file, upload a non-image attachment
2. Click on the attachment
3. See the modal showing a broken image

**Expected behavior**
The attachment should be opened in the viewer.

mejo-:
After discussing this a bit more I'm not sure anymore whether we want the attachment to be opened in Viewer - because Text itself might already be open in Viewer. :thinking:

mejo-:
https://github.com/nextcloud/text/blob/bf6c6bb8df03037dc9d95dc2f62550d2a61d84ed/src/nodes/ImageView.vue#L38 is the relevant part in code.

luka-nextcloud:
@mejo- @juliushaertl 
I would suggest to show an icon based on the file type of attachments in the preview modal.
Or just a file icon for all non-image file, it may be a simple fix for now.

mejo-:
>  I would suggest to show an icon based on the file type of attachments in the preview modal.

I'm not sure whether that's useful from a user perspective. How about opening non-image attachments in the files app in a new tab (`target="_blank"`) instead. That way the files app could take care of opening the attachment in viewer.

Maybe we would just make the whole non-image NodeView a link to the file view (`/f/<fileId>`) and let the default `linkHandler` do it's job. That way we could take care of opening it in Viewer from Collectives with its own `linkHandler` later on :thinking:

luka-nextcloud:
@mejo- Sounds good to me, totally agree with you.

hcaihao:
Does this issue sloved? How to update & test?

PR #3924, opened 2023-03-10
fix: open non-image attachment

luka-nextcloud:
### 📝 Summary

* Resolves: #3849

#### 🖼️ Screenshots

🏚️ Before
image

🏡 After
image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


mejo-:
Thanks for looking into this @luka-nextcloud. From the screenshot I wonder why you changed the layout of non-image attachments? I think we should preserve the way they were rendered before, no?

mejo-:
@luka-nextcloud I would have envisioned a simple `<a href="/f/<fileId>">[...]</a>` around the non-image attachment node here:
 https://github.com/nextcloud/text/blob/25f2f5fad1c43efc2c1a444416a792d94f731b61/src/nodes/ImageView.vue#L36-L48

Additionally, the registered `@click` handler in this block would need to be removed. That would allow Text/Collectives to treat the non-media attachment node as simple link with their registered linkHandlers. What do you think about that approach?

luka-nextcloud:
> Thanks for looking into this @luka-nextcloud. From the screenshot I wonder why you changed the layout of non-image attachments? I think we should preserve the way they were rendered before, no?

Actually, I didn't change the layout, this layout for non-image was already there at the beginning. It wasn't shown before because of some missing data (mime type of file was missing before). I just fixed it.

luka-nextcloud:
> @luka-nextcloud I would have envisioned a simple `<a href="/f/<fileId>">[...]</a>` around the non-image attachment node here:
> 
> https://github.com/nextcloud/text/blob/25f2f5fad1c43efc2c1a444416a792d94f731b61/src/nodes/ImageView.vue#L36-L48
> 
> Additionally, the registered `@click` handler in this block would need to be removed. That would allow Text/Collectives to treat the non-media attachment node as simple link with their registered linkHandlers. What do you think about that approach?

I think it's already done here. Please check: https://github.com/nextcloud/text/blob/main/src/nodes/ImageView.vue#L108

mejo-:
@luka-nextcloud I think there's a missunderstanding here.

To my understanding, the following block is used for attachments that have a set, but unsupported mimetype:
https://github.com/nextcloud/text/blob/dd0f8261e28c44ce75048743a96f706ed91c7969/src/nodes/ImageView.vue#L105-L123

I would like to continue using the following block for media attachments (non-image attachments) instead (where it's rendered properly as attachment node), just making this node a link to the files URL of the attachment:

https://github.com/nextcloud/text/blob/dd0f8261e28c44ce75048743a96f706ed91c7969/src/nodes/ImageView.vue#L36-L59

luka-nextcloud:
@mejo- Could you please check again?

src/nodes/ImageView.vue
```diff
@@ -34,13 +34,15 @@
...
 						<div v-if="isMediaAttachment"
-							class="media"
-							@click="handleImageClick(src)">
+							class="media">
 							<div class="media__wrapper">
-								<img v-show="loaded"
-									:src="imageUrl"
-									class="image__main"
-									@load="onLoaded">
+								<a :href="internalLinkOrImage"
```
mejo-, line 39: I would put the `a` tag around the div, not inside it. So that the metadata div becomes part of the link.

src/nodes/ImageView.vue
```diff
@@ -224,16 +226,13 @@ export default {
...
 			return this.loaded && this.imageLoaded
 		},
-		imageFileId() {
-			return getQueryVariable(this.src, 'fileId')
-		},
 		isSupportedImage() {
-			return typeof this.mime === 'undefined'
-				|| IMAGE_MIMES.indexOf(this.mime) !== -1
+			return typeof this.mimeType === 'undefined'
+				|| [...IMAGE_MIMES, ...MEDIA_MIMES].indexOf(this.mimeType) !== -1
```
mejo-, line 231: I don't think this is necessary. Just leave it as is.

src/nodes/ImageView.vue
```diff
@@ -224,16 +226,13 @@ export default {
...
 		},
 		internalLinkOrImage() {
-			if (this.imageFileId) {
-				return generateUrl('/f/' + this.imageFileId)
+			if (this.fileId) {
+				return generateUrl('/apps/files/f/' + this.fileId)
```
mejo-, line 235: Same here, no need for the change. The URL `/f/<fileId>` works just fine, no need for '/apps/files/f/<fileId>`, no?

src/nodes/ImageView.vue
```diff
@@ -257,6 +256,12 @@ export default {
 				})
 			},
 		},
+		fileId() {
+			return getQueryVariable(this.src, 'fileId')
+		},
+		mimeType() {
+			return getQueryVariable(this.src, 'mimeType')
+		},
```
mejo-, line 264: This change is not required either if I'm not mistaken. Really just add the `a` tag above and that's it, no?

mejo- (changes requested):
Thanks, looks better already but still I think most of your changes are not necessary :wink:

Also, could you squash all commits in the PR into one?

src/components/Editor/MediaHandler.vue
```diff
@@ -185,7 +185,7 @@ export default {
 			const src = dirname + '/'
 				+ encodeURIComponent(name).replace(/[!'()*]/g, (c) => {
 					return '%' + c.charCodeAt(0).toString(16).toUpperCase()
-				})
+				}) + '?fileId=' + fileId + '&mimeType=' + encodeURIComponent(mimeType)
```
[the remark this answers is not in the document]
mejo-: Sorry, I see the `fileId` parameter is still necessary in order to have it available later on. The `mimeType` parameter causes issues though and should be removed.

src/nodes/ImageView.vue
luka-nextcloud: @mejo- `/f/<fileId>` will not work, I have checked this.

src/nodes/ImageView.vue
luka-nextcloud: @mejo- It is required to detect if this file type is supported or not.

src/components/Editor/MediaHandler.vue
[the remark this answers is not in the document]
luka-nextcloud: @mejo- The mimeType param looks fine to me, what issues do you see?

src/nodes/ImageView.vue
luka-nextcloud: Will fix it.

src/nodes/ImageView.vue
luka-nextcloud: Will fix it.

mejo- (changes requested):
Thanks @luka-nextcloud. I think that the PR has to wait until #3974 got resolved.

As said, the option to inject `fileId` as query param to the markdown in `insertAttachment()` of `MediaHandler.vue` brings several problems and should be avoided in my eyes:

* It doesn't work for attachments that were uploaded prior to this PR.
* It will break whenever the text document gets copied - fileIds of the attachments will change in that case.
* It will not work for attachments that got inserted manually (arguably neglectable though)

luka-nextcloud:
@mejo- 
> It doesn't work for attachments that were uploaded prior to this 
For attachments already saved before, there is no way to detect it's fileId. The saved content before will look like this.
image

> It will break whenever the text document gets copied - fileIds of the attachments will change in that case.
Not sure what it means...

> It will not work for attachments that got inserted manually (arguably neglectable though)
Do you mean using "Upload from computer"?
image

mejo-:
Hey @luka-nextcloud,

> For attachments already saved before, there is no way to detect it's fileId. The saved content before will look like this.

Exactly, and that's why I suggested to update the logic of `AttachmentResolver` as described in #3974 :wink: 

> > It will break whenever the text document gets copied - fileIds of the attachments will change in that case.
> 
> Not sure what it means...

I mean the following: create a Text document and add an attachment. Then you will have the attachment `fileId` hardcoded in its markdown. Afterwards, copy the document in the Nextcloud Files app ("Move or copy" in three-dot-menu, then "Copy"). The resulting copy will have the attachment copied over as well, but the `fileId` in markdown will be wrong.

ralfrupf1976:
I just uploaded a PDF in Kollektive - still PDFs are not shown correctly. Any Updates on this?

Screenshot

mejo-:
> I just uploaded a PDF in Kollektive - still PDFs are not shown correctly. Any Updates on this?

@ralfrupf1976 we're working on a proper fix for this. It can take a bit though because it requires bigger internal code changes.

juliushaertl:
@mejo- I guess this will be covered by #5042 then so let's close this one for now I'd say.

PR #3946, opened 2023-03-15
fix: Ensure to not persist user put into the session for direct editing

### LINKED ISSUE #3848: User session might be persisted
https://github.com/nextcloud/text/blob/main/lib/Controller/SessionController.php#L110 should get reset after controller method calls that make use of it as otherwise this may load to the current session getting updated with a different user.

I've seen this happening locally with:
- tab 1: Having a tab open with a text editor
- tab 2: Logout in a second tab
- tab 2: Login as a different user
- tab 2: Navigate to another page
- tab 2: See that you're logged in as the previous user

juliushaertl:
Fixes #3848

### DIFF
--- a/lib/Controller/SessionController.php
+++ b/lib/Controller/SessionController.php
@@ -31,6 +31,7 @@ use OCA\Text\Service\SessionService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\IRequest;
+use OCP\IUser;
 use OCP\IUserManager;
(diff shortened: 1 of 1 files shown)


marcelklehr (approved):
Makes sense!

sascha-wi:
How do I know if this issue will also be fixed in v24?

HumBelele:
> How do I know if this issue will also be fixed in v24? Or is v24 not affected at all by this?

NC 24 is out of active Support. So no further Updates for Version 24.

Issue #3961, opened 2023-03-17
Conflict resolution is broken


juliushaertl:
CLicking one of the options results in an endless loading, but reopening the document works to see the changes

PR #3972, opened 2023-03-21
Conflict fixes

juliushaertl:
Couple of fixes around conflict handling and file save, split into individual commits for easier review:

- test: Add cypress tests for resolving conflicts
- fix: store actually used mtime
- fix: Properly restart sync after conflict resolution
- fix: Avoid writing file content if it has not changed
- fix: Fix styling of collision dialog and move it out of the editor
- fix: Do not disconnect during conflict to be able to catch up on a resolution

There is still some work left that probably involves design work to come up with better handling of the conflict display on mobile devices, but that is left out for now to have a more easily back portable change.

### 📝 Summary

* Resolves: #3961 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2023-03-22 at 18 44 30 | Screenshot 2023-03-22 at 18 44 09

### 🏁 Checklist


juliushaertl:
Will check the cypress test tomorrow, seems we should await the sync request after pushing the state as it passed before my squashing push.

mejo- (approved):
Nice, thanks for tackling this! Code looks good to me :ok_hand:

PR #3975, opened 2023-03-22
Emit events from editor when image nodes get added or deleted

mejo-:
To be used in Collectives in the attachment list.

### 🏁 Checklist


src/components/Editor.vue
```diff
@@ -658,15 +661,23 @@ export default {
...
 			this.$emit('blur')
 		},
 
-		async close() {
+		onAddImageNode() {
+			this.$parent.$emit('add-image-node')
+		},
 
+		onDeleteImageNode(imageUrl) {
+			this.$parent.$emit('delete-image-node', imageUrl)
```
juliushaertl, line 677: Not blocking but maybe we can annotate those through [emits](https://v3-migration.vuejs.org/breaking-changes/emits-option.html) though I'm not sure if that necessary the right place since we emit on the parent

PR #4026, opened 2023-04-05
tmp: Run cypress against nextcloud/server#37588

juliushaertl:
https://github.com/nextcloud/server/pull/37588


juliushaertl:
Adjustment also part of https://github.com/nextcloud/text/pull/3999

PR #4028, opened 2023-04-06
fix: Properly emit ready event on conflicts with the editor API

juliushaertl:
### 📝 Summary

* Resolves: https://github.com/nextcloud/notes/issues/1008

Fixes loading the editor integrated through the editor API when a file conflict occurs during the initial loading. In this case the ready event was never emitted on the root component that would pass the state to the notes app.

I'd love to have a more generic way there but haven't found a proper way to avoid emitting on both this and parent, so the second commit cleans up the emit handling in Editor.vue to make sure we always emit on both.

### 🏁 Checklist


juliushaertl:
Also fixes this issue on Collectives as there the parent component is the `ViewerComponent`

mejo- (approved):
Thanks for tackeling this @juliushaertl. Code looks good to me :+1:

Issue #4010, opened 2023-03-30
NC26: author tags are shown although no other user is in the file


jakobroehrl:
**Describe the bug**
When I enter a md file, some author tags are shown of users who earlier edit this file.
After I wait for 30 seks the tags are gone

![grafik](https://user-images.githubusercontent.com/24530680/228915475-31f99fee-a6a9-475f-b965-034c6cddc827.png)

fin444:
In my experience, this issue occurs because the editor displayed closed the tab directly via the browser instead of using the X in the top right. The tag will be left at the exact location that their cursor was at when they closed the tab, and will persist until the server restarts.

jakobroehrl:
> 

Could be true, I think. 
But a server reboot does not delete this tags.

PR #4053, opened 2023-04-14
fix: Only handle recent awareness messagess from sessions

juliushaertl:
fixes https://github.com/nextcloud/text/issues/4010


max-nextcloud (approved):
Makes sense to me. I'm wondering if we should only use the last awareness message but i need to dig into the awareness protocol more for that.
This is definitely a step in the right direction. So let's get it in.

PR #4063, opened 2023-04-17
Adjust smart picker menu action

julien-nc:
* Adjust the menu item label
* Adjust the icon to use the same as in Mail


julien-nc:
IMO this can be backported to stable26 but this is not a bug fix...

max-nextcloud (approved):

max-nextcloud:
> IMO this can be backported to stable26 but this is not a bug fix...

I'll create a backport request. Stable26 is blocked for most of the week anyway and we can discuss if we want to backport in the backport PR.

PR #4068, opened 2023-04-20
chore(maintainers): Update CODEOWNERS file

nickvergessen:
* Ref https://github.com/nextcloud/documentation/pull/10049


juliushaertl:
@nickvergessen Shall we backport this as well or is only the main/master branch taken into account ?

nickvergessen:
I fixed it last week after a first test run to always only consider master/main

PR #4069, opened 2023-04-20
enh(log): use level warning for saving empty docs

max-nextcloud:
### 📝 Summary

Saving an empty document can happen from time to time but if it happens to often it may indicate some problem. Keep an eye on these without needing to log all debug things in text.

### 🏁 Checklist

- [ ] Tests - we don't test our logging.


datenangebot (approved):
✓

Issue #4045, opened 2023-04-13
Editing image description fails if enter isn't pressed


provokateurin:
**Describe the bug**
You need to press enter in order to save an image description

**To Reproduce**
1. Paste an image
2. Edit the description
3. Click somewhere else

**Expected behavior**
Description should be saved, but it is not. Pressing enter before clicking somewhere else makes it work.

**Client details:**
 - OS: Linux
 - Browser: Firefox
 - Version: 112
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.7.2

**Nextcloud version:** 26

</details>

PR #4071, opened 2023-04-24
fix: update ImageView

luka-nextcloud:
### 📝 Summary

* Resolves: #4045

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


max-nextcloud (approved):
I wonder why `@blur` did not do the trick. But this also seems fine.

juliushaertl:
The `@blur` seems to work just fine, however I noticed that the actual issue seems to be that the component might be re-rendered if the user moves the mouse out which then restores the old value even before the blur was triggered

https://github.com/nextcloud/text/blob/5730fc775d0aa3927ce5b4fd07f23f3c3b7be454/src/nodes/ImageView.vue#L32

Anyways I think always updating seems more safe to me 👍

juliushaertl (approved):

max-nextcloud:
shall we backport this?

juliushaertl:
@luka-nextcloud Please backport manually.

Issue #3348, opened 2022-11-02
The Text app creates another Readme.md on top of the existing one


james10424:
**Describe the bug**
The Text app creates `Readme.md` if you already have a readme.

**To Reproduce**
1. Create a `README.md`, doesn't matter if it has content or not. Easiest way is `touch README.md`
2. Wait for it to sync
3. Go to the browser UI and find `README.md` and `Readme.md`. But you didn't create any `Readme.md`
4. The behaviour goes away after disabling the Text app

This causes an error in case-insensitive OS like macOS to throw out errors like "Readme.md could not be downloaded because of a local file name clash!"

**Expected behavior**
I do not expect `Readme.md` to be created along with my own readme file.

**Client details:**
 - OS: macOS Mojave
 - Browser: Safari 14.1.2
 - Version: 3.4.4
 - Device: iMac

<details>
<summary>Server details</summary>

**Text app version:** 3.5.1

**Operating system:** Debian 10.2.1-6

**Web server:** Apache2

**Database:** mysql

**PHP version:** 7.4

**Nextcloud version:** 24.0.6

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

juliushaertl:
https://github.com/nextcloud/text/pull/834 was there to address this, but maybe this is not working properly anymore

mejo-:
I'm unable to reproduce this on latest development branch. @james10424 could you please test the following:

1. Create a new subfolder 'readmetest' and move into it in the Files app
2. Create a new markdown file 'README.md' from the + button on top, add some content
3. Navigate back to the folder, force-reload the page and see whether the content is displayed in the rich workspace area above the file list
4. See whether an extra file 'Readme.md` is created

Also, please send us a list of apps that you have enabled on your instance (`php occ app:list`).

Is it possible that the `Readme.md` file had been created *before* you manually created the `README.md` file?

james10424:
I did the above steps and yes, the `Readme.md` is created after creating `README.md` using the + button in the web UI.

The apps that are enabled:
```
Enabled:
  - accessibility: 1.10.0
  - activity: 2.16.0
  - bruteforcesettings: 2.4.0
  - circles: 24.0.1
  - cloud_federation_api: 1.7.0
  - comments: 1.14.0
[...]

PR #834, opened 2020-05-19
Fix rich workspaces on case insensitive filesystems

juliushaertl:
Fixes #807 

This fixes multiple files showing up on case-insensitive filenames as otherwise
the IFolder::get call creates multiple entries in the filecache for the same file. In addition to the fix now the behaviour is actually the correct one where the first matching filename of supported names is picked up for creating the file and obtaining in case there are multiple names matching.

PR #4104, opened 2023-05-02
Fix double readme files getting created on case-insensitive storages

szaimen:
Fix https://github.com/nextcloud/text/issues/3348

Apparently this got missing from https://github.com/nextcloud/text/pull/834 when the rebase was done.

See https://github.com/nextcloud/text/compare/1e600fb26111f0c850f134cf17aad96b1c8c59d3..7d6569cdf4c01e3467fe4b066a5f80bb56c48d9e

But actually @max-nextcloud and me tested a bit and this only seems to fix when creating new readme files from the workspace. The rootcase would probably need to be fixed somewhere in getFileInfo in View.php in server but we don't know much about that so adding this change here seems like a quick win.

### DIFF
--- a/lib/Service/WorkspaceService.php
+++ b/lib/Service/WorkspaceService.php
@@ -12,8 +12,8 @@ class WorkspaceService {
 	private IL10N $l10n;
 
 	private const SUPPORTED_STATIC_FILENAMES = [
-		'README.md',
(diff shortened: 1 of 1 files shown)


juliushaertl:
Thanks a lot for that :)

PR #4107, opened 2023-05-02
Fix/cypress share failures

max-nextcloud:
### 📝 Summary
* Handle failing connections without exceptions in the console.
* Avoid unnecessary call to `getSubmenuEntry` which opens the menu in question.

* Resolves: Cypress failures such as https://cloud.cypress.io/projects/hx9gqy/runs/9592/overview/c18c26ab-d84d-48eb-9569-ee0c1b193bbd?roarHideRunsWithDiffGroupsAndTags=1
* Resolves: Cypress failures such as https://cloud.cypress.io/projects/hx9gqy/runs/9611/test-results/195f49e9-2077-4322-973f-d678a7f5e640

### 🏁 Checklist


src/services/SyncService.js
```diff
@@ -91,17 +91,22 @@ class SyncService {
 	}
 
 	async open({ fileId, initialSession }) {
-		this.on('change', ({ sessions }) => {
+		const onChange = ({ sessions }) => {
 			this.sessions = sessions
 		})
```
juliushaertl, line 96:
```suggestion
		const onChange = ({ sessions }) => {
			this.sessions = sessions
		}
```

juliushaertl (changes requested):

src/services/SyncService.js
max-nextcloud: Thanks! :heart: 
And that is why one should always test again after making changes - even if they seem simple :wink:

juliushaertl (approved):

Issue #4112, opened 2023-05-03
Text 3.7.2: Files mixed or emptied


jakobroehrl:
Since NC16 I have some problems with the textfiles. 
I have mainly 2 differnt textfiles that a my users view and some of them edit it. 
I noticed several times that both files are mixed, so the complete contents of both files are stored in both files. 
Or both files are completly empty after a user edited it. 
Till now I cant reproduce it, but I think it's a bug, we work since a long time with this textfiles without this problems.

max-nextcloud:
Hi @jakobroehrl 

We changed the collaboration backend in Nextcloud 26 in order to allow for more people to work on the same file at the same time.

We also noticed some empty files after that and isolated and fixed a problem with the sync mechanism. One fix has been released with 26.0.1. However it seems like the fix did not cover for all cases. We're logging the saving of empty files with a warning from the next release on to be able to better investigate this.

However i have not heard about an issue of mixing the content of two files. That really seems odd. Is there anything special about these files:
* are the Readme.md files?
* are they stored on a external storage?
* are they inside a groupfolder?

Anything that connects these two files?

juliushaertl:
Also can you clarify since when this issue shows? "Since NC16 I have some problems with the textfiles." this sounds like it is an issue for along time already but "we work since a long time with this textfiles without this problems" sounds different.

juliushaertl:
Hi @jakobroehrl With #4100 we found that mobile apps might have been one of the reasons for this happening a potential fix is pushed to #4124. Can you tell if your users used the Android/iOS apps for editing and if that might be related?

jakobroehrl:
Hi Max und Julius,

> are the Readme.md files?

no

> are they stored on a external storage?

no

> are they inside a groupfolder?

no

> Anything that connects these two files?

I think no :)

> Since NC16 I have some problems with the textfiles.

I meant NC26, sorry 

> Can you tell if your users used the Android/iOS apps for editing and if that might be related?

yes, we use a lot the mobile Clients, both Android and iOS.

rurseekatze:
Hello,

since I upgraded my Nextcloud to 26 a few days ago, I also observed the problem of mixed text files described by @jakobroehrl. In my case there were 4 different files mixed together.

[...]

Issue #4100, opened 2023-05-01
user position indicators triggers different bugs (what's that feature correctly called?)


kolAflash:
The new _"user position indicator"_ feature triggers different bugs.

<details>
<summary>screenshot</summary>

![Screenshot_20230501-222621_Nextcloud](https://user-images.githubusercontent.com/3355089/235526420-7c7984cb-9470-42a7-aaa0-8d97140c8523.png)
</details>

&nbsp;

### Questions, Problems and Bugs

Questions:
- What's the correct name for that feature!? (I just call it _"User position indicator"_ because I don't know any other name for it)
- Can the feature be disabled?
  - It's a nice feature. But for me it's not worth the problems and bugs it causes.
- The indicators seem to go away 30 seconds after opening a document. Can that time span be changed?

&nbsp;

Bugs:
- Sometimes all text following an indicator is deleted. This seems to happen when editing text at the position of the indicator. (couldn't determine the exact circumstances until now) 
- In the browser (Firefox) empty lines containing an indicator can't be clicked.
- Same user is listed multiple times, although he's just logged in from one client. (counted up to five)

&nbsp;

Problems:
- When an user position indicator overlaps text, the text can't be read.
- Sometimes the same indicator seems to be multiple times at the same position. Clicking the first indicator in the screenshot (see above) on Android makes my keyboards auto completion show `kolAflashkolAflashkolAflashkolAflashkolAflash`. (5 times)

&nbsp;

&nbsp;

### Tested platforms

- PC (Firefox-112.0.2)
- Android-10 (Nextcloud App 3.24.2)
- Android-13 (Nextcloud App 3.24.2) 

Server:
Nextcloud: 26.0.1
Text App: 3.7.2
OS: Debian-12 (Apache-2.4.57, php-fpm-8.2.4, mariadb-10.11.2)

juliushaertl:
Thanks for reporting, for the first issue it would be quite helpful to have more detailed steps, maybe also a short screen recording to see when this happens.

> Same user is listed multiple times, although he's just logged in from one client. (counted up to five)

This has been addressed with https://github.com/nextcloud/text/pull/4053 already and will be fixed in the next maintenance release.

kolAflash:
@juliushaertl  
Sure. Here are some, showing a wide range of problems.  
Including the bug which deletes all text following the cursor.

[...]

PR #4124, opened 2023-05-05
fix: pass file id for direct editing and fail y.js provider setup if none was passed

juliushaertl:
### 📝 Summary

When setting up the WebsocketProvider we still used to create them even if no file id was passed. This is actually the case for direct editing trough android/iOS and would mean that if some browser states in the webview is still present it could mix up documents.

This is basically a follow up to https://github.com/nextcloud/text/commit/b2186da8610a2762f4b155d1452402eac3423cdd

* Possibly fixes https://github.com/nextcloud/text/issues/4112
* Possibly fixes parts of https://github.com/nextcloud/text/issues/4100

### 🏁 Checklist


juliushaertl:
Note that the error behaviour is currently quite bad as when this happens the Editor.vue is just empty, will see if we can improve that as well.

src/services/SyncServiceProvider.js
```diff
@@ -33,6 +33,9 @@ import { logger } from '../helpers/logger.js'
  * @param {object} options.initialSession - initialSession to start from
  */
 export default function createSyncServiceProvider({ ydoc, syncService, fileId, initialSession }) {
+	if (!fileId) {
```
mejo-, line 36: Maybe a short comment would be helpful here for future reference?

mejo- (commented):
Just a minor comment.

And to be honest I don't fully understand the situation. So directediting sessions didn't pass a `fileId` before the change to `DirectEditing.vue` from this PR, which lead to `WebsocketProvider` instances with similar identifiers, which in turn could result in content from leftover sessions getting mixed into the new session. Is this roughly the scenario that this PR aims to fix?

mejo-:
/backport 23a46cd582cca8224d0fa6beca60b9e9c76f80c5 to stable26

max-nextcloud:
Great catch @juliushaertl and thanks for the quick review and backport @mejo- ! :heart:

PR #4178, opened 2023-05-12
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.


mejo- (approved):

juliushaertl:
Let's keep this for 27 for now as I'm not fully sure we backported all recent adaptions of the css variables to those branches.

juliushaertl (approved):

Issue #3907, opened 2023-03-07
Edit icons not updated properly


datenangebot:
Steps to reproduce :

- Open web app
- Edit a text file
- Write a text
- Click on "Bold" or "Italic" or "Underline" button
- Obtained Results : The buttons are not highlighted until we write something after clicking on them

Browser used

- Chrome Windows
- Firefox Windows

PR #4211, opened 2023-05-19
fix(menubar): show selection status correctly

Hephi2:
Buttons in the menubar like bold, italic, underline and strikethrough are show now if they are selected or not right after the selection.

### 📝 Summary

* Resolves: #3907

Before the fix the is-active status of the icon buttons for bold, italic, underline and strikethrough did not change after selecting. Now the status is updated after the click to show if the modifiers are active even before starting to write with those.


juliushaertl:
Nice, change looks good. I think this is also a good case to cover with a simple cypress test that asserts that the button has the active state class. 

@max-nextcloud Maybe you can look into that together with @Hephi2 next week and guide him a bit on the first steps with cypress :)

cypress/e2e/workspace.spec.js
```diff
@@ -84,17 +84,19 @@ describe('Workspace', function() {
...
 
-	it('formats text', function() {
+	it('formats text', function() {		
 		cy.visit(`apps/files?dir=/${encodeURIComponent(currentFolder)}`)
-		cy.openWorkspace().type('Format me')
-		cy.getContent().type('{selectall}')
-		;[
+		cy.openWorkspace()
+
+		;const buttons = [
```
max-nextcloud, line 91: The extra `;` at the beginning of the line was here because the line started with an array:
```js
cy.getContent().type('{selectall}')
;[
```
Without a `;` Javascript  would interpret the `[` as an index into an array (as in  `cy.getContent().type('{selectall}')[1]`)

Now that you are using a variable (much better) there's no more need for the semicolon.

cypress/e2e/workspace.spec.js
```diff
@@ -323,3 +325,21 @@ function testButton(button, tag, content) {
...
+ * @param {string} button Name of the button to click.
+ * @param {string} tag Html tag expected to be toggled.
+ * @param {string} content Content expected in the element.
+ */
+function testButtonUnselected(button, tag) {
+	cy.getMenuEntry(button).click().should('have.class', 'is-active')
+	cy.getContent().type('Format me{selectall}')
+		.find(`${tag}`)
+		.should('contain', 'Format me').type('{del}')
+	cy.getMenuEntry(button).click().should('have.class', 'is-active').click().should('not.have.class', 'is-active')
```
max-nextcloud, line 340: Cypress eslint extension is complaining about these lines because you are chaining off commands.
What does it mean by that:
```javascript
cy.getContent().type('Format me{selectall}').find(...)
```
`.type()` is a command. (So is `.click()`).

The problem is rooted in how cypress works. For queries like `find()` it will look for the element. If it does not exist yet it will retry until the element exists or it times out.
If the `find()` command is chained from the `type()` command there's no safe way for cypress to retry without calling the `type()` command multiple times.

So whenever you have a command that actually changes something you need to end the chain:
```suggestion
	cy.getContent().type('Format me{selectall}')
	cy.getContent().find(`${tag}`)
		.should('contain', 'Format me').type('{del}')
	cy.getMenuEntry(button).click()
	cy.getMenuEntry(button).should('have.class', 'is-active').click()
	cy.getMenuEntry(button).should('not.have.class', 'is-active')
```

This is only checked since a short time - so you will still find old documentation and even code that does not follow this rule.

max-nextcloud (commented):
All in all looks great. Thanks a lot for looking into this. I added a few comments to the eslint complaints. Hope that makes it easier to understand and fix them.

When in doubt please ask in the talk channel.

Hephi2:
Thanks for the detailed feedback!

max-nextcloud (approved):
Looks really good. Thanks for diving into this!

:test_tube: Bonus points for tests included  :100: :wink:

max-nextcloud:
Failing cypress test is known to fail from time to time.

blizzz:
if that's a bugfix that has to go to 27, it needs backporting for we are behind branch-off-

PR #4237, opened 2023-06-02
[wip] feat: welcome typescript and cypress in docker

skjnldsv:
Ongoing effort, some adjustments might be needed :wink: 

Please take over Text team! :rocket:


max-nextcloud:
Thanks a lot for looking into this @skjnldsv !

cypress.config.ts
```diff
@@ -0,0 +1,82 @@
...
+			})
+
+			// Remove container after run
+			on('after:run', () => {
+				stopNextcloud()
+			})
+
+			// Before the browser launches
+			// starting Nextcloud testing container
+			return startNextcloud(process.env.BRANCH)
```
juliushaertl, line 68: Would be nice to have an option to still run this towards a local instance for easier debugging. And the currently used docker image doesn't come with arm architecture support.

cypress/dockerNode.ts
```diff
@@ -0,0 +1,242 @@
...
+	console.log('\nConfiguring nextcloud...')
+	const container = docker.getContainer(CONTAINER_NAME)
+	await runExec(container, ['php', 'occ', '--version'], true)
+
+	// Be consistent for screenshots
+	await runExec(container, ['php', 'occ', 'config:system:set', 'default_language', '--value', 'en'], true)
+	await runExec(container, ['php', 'occ', 'config:system:set', 'force_language', '--value', 'en'], true)
+	await runExec(container, ['php', 'occ', 'config:system:set', 'default_locale', '--value', 'en_US'], true)
+	await runExec(container, ['php', 'occ', 'config:system:set', 'force_locale', '--value', 'en_US'], true)
+	await runExec(container, ['php', 'occ', 'config:system:set', 'enforce_theme', '--value', 'light'], true)
```
juliushaertl, line 141: Ideally also have `"memcache.local"=>"\OC\Memcache\APCu","hashing_default_password"=>true` for faster runs

cypress.config.ts
skjnldsv: Do we support arm officially?

juliushaertl: Depends on what you mean with supporting, at least community has high demand with single board computers.

Also this would block developers on apple arm chips from running the tests in a reasonable time.

juliushaertl: At least providing a way to just pass CYPRESS_baseUrl as an environment variable and then not setup docker would be nice, but I can look into that part :)

skjnldsv: Well, I'm guessing the current behaviour also breaks on windows. 
I think it's an interesting topic, but I don't think it's worth spending more time on it. :see_no_evil:

juliushaertl: Yes, Hephi2 had problems running the server tests on windows as well with WSL+Docker. Let's go for the manual URL approach then as a fallback/workaround which should be easy to add

juliushaertl:
Closing this for now as redoing is likely easier then resolving conflicts.

PR #4241, opened 2023-06-06
fix(UserApiController): Fix warning during test run with PHP 8.2

juliushaertl:
Fix test failure that only happens on PHP8.2

> stripos(): Passing null to parameter #1 ($haystack) of type string is deprecated

### DIFF
--- a/lib/Controller/UserApiController.php
+++ b/lib/Controller/UserApiController.php
@@ -41,7 +41,7 @@ class UserApiController extends ApiController {
 			$sessionUserId = $session['userId'];
 			if ($sessionUserId !== null && !isset($users[$sessionUserId])) {
 				$displayName = $this->userManager->getDisplayName($sessionUserId);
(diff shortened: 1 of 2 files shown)


max-nextcloud (approved):

max-nextcloud:
Cypress failures cannot be related - but I wonder where the block merges during freezes is coming from...

And i wonder why i could still merge this.

juliushaertl:
Merge freeze seems like I overseen a change when updating the actions to adapt the server branch when text is using main:

Screenshot 2023-06-06 at 09 04 03

juliushaertl:
Cypress failure looks a bit like it could be a upstream issue with loading the files app:

https://cloud.cypress.io/projects/hx9gqy/runs/e149dd86-8742-4946-b5fb-7e5277760429/test-results/45605c10-1ad8-4149-84fa-1fc611744854/screenshots

juliushaertl:
Pushed a fix for the merge block https://github.com/nextcloud/text/pull/4243

Issue #3816, opened 2023-02-19
Console error on public share auth page


juliushaertl:
When opening the share link asking for a password:

```
Cannot read properties of null (reading 'value')
```

https://github.com/nextcloud/text/blob/main/src/public.js#L51

https://juliushaertl.sentry.io/share/issue/c9a7f6638e6e460d8bb9868465713456/

PR #4253, opened 2023-06-07
Don't expect HTML element with ID `mimetype` in public share

mejo-:
If the share is password-protected, the authentication page doesn't have the property set yet.

Fixes: #3816


juliushaertl (approved):

mejo-:
/backport 4f17eeeb99d2fd8fe4e12b9829f590c34433bf7c to stable27

PR #4254, opened 2023-06-07
Prevent cumulative layout shift when rich workspace is loading

mejo-:
### 📝 Summary

Instead of only increasing editor height with growing document length, let's stick to a fixed height (25vh) and use that one already while loading the document.
    
The editor height still increases when it gets focused.

* Fixes: #2834
* Fixes: #2803

#### 🖼️ Screencasts

🏚️ Before | 🏡 After
---|---
[recording1.webm](https://github.com/nextcloud/text/assets/3582805/0cdd5427-3bd0-42e3-af6b-e71c6939fc71) | [recording2.webm](https://github.com/nextcloud/text/assets/3582805/7f060f9a-a7f6-42d7-893f-24ca2459afba)

### 🚧 TODO

- [ ] Wait for feedback by @nextcloud/designers

### 🏁 Checklist

### DIFF
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -259,11 +259,13 @@ export default {
 	}
 
 	#rich-workspace.focus {
+		min-height: 25vh;
 		max-height: 50vh;
 	}
 
-	#rich-workspace:not(.focus) {
-		max-height: 30vh;
+	#rich-workspace:not(.focus), #rich-workspace.icon-loading {
+		min-height: 25vh;
+		max-height: 25vh;
 		position: relative;
 		overflow: hidden;
 	}
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -203,7 +203,7 @@ export default {
 		z-index: 61;
 		position: relative;
 		&.creatable {
-			min-height: 100px;
+			min-height: 150px;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):
Nice improvement!

juliushaertl:
The initial jumping is trickier as we'd need to know the availability of a rich workspace before the file list is loaded. We'd need a hook in the files app code to have a synchronous pass over of the propfind data to text before rendering.

mejo-:
> The initial jumping is trickier as we'd need to know the availability of a rich workspace before the file list is loaded. We'd need a hook in the files app code to have a synchronous pass over of the propfind data to text before rendering.

Indeed, I just discovered the same. So inside Text the only thing we can do, is reserve a fixed size for the rich workspace regardless of whether the `Readme.md` exists or not.

How about always limiting the height of non-focused rich workspace to `150px` and make the toolbar `display: none` if not focused? Only on focus, the toolbar takes space and the editor is expanded? I pushed these changes to the PR now. See the following screencasts:

| On large screen | On mobile screen|
|---|---|
| [recording2.webm](https://github.com/nextcloud/text/assets/3582805/711463d9-2bf3-4477-8355-ac6d60a325a1) | [recording1.webm](https://github.com/nextcloud/text/assets/3582805/dfe2b84c-73fd-4c49-ab22-982dc8884421) |

juliushaertl:
I think we should try to find a way to prefetch the information, then we need no layout shift at all:

Some code paths that could be worth to investigate:

Point when the file list is loaded and being setup before rendering, possibly a good time to emit an event with the file list data? https://github.com/nextcloud/server/blob/8de29843497efed3eff05db1f90a15937d0a504c/apps/files/js/filelist.js#L2284

Maybe there is a way to extend the list of properties dynamically which is taken from https://github.com/nextcloud/server/blob/8de29843497efed3eff05db1f90a15937d0a504c/apps/files/js/filelist.js#L2199 when the file list does the PROPFIND.

We do something similar in group folders but I wouldn't reuse the code there due to the complexity and in depth dependency on files internals: https://github.com/nextcloud/groupfolders/blob/master/src/client.js#L141

juliushaertl:
Pushed a first draft of my suggested approach to https://github.com/nextcloud/text/pull/4776

Seems to work fine except for some quirks with the file list headers and dav properties when navigating around.

mejo-:
Awesome, let's close this PR in favour of #4776 then.

PR #4240, opened 2023-06-05
[stable26] ci: Backport relevant action changes from main

juliushaertl:
Manually backporting recent relevant changes of the github action workflow files to stable26

PR #4255, opened 2023-06-07
Limit cypress to github runners

juliushaertl:
Until we know how to make it work this should bring back more reliable cypress runs


mejo- (approved):

juliushaertl:
Manually picked in #4239 #4240

PR #4260, opened 2023-06-07
chore: Simplify issue template

### LINKED ISSUE #4177: Issue template version
I'm always confused which version is requested in https://github.com/nextcloud/text/blob/ad181c5a361ef0a054ab2afb3a805a7652c959c2/.github/ISSUE_TEMPLATE/Bug_report.md?plain=1#L26
I can open a PR fix this but for that I would need to know what the intention was. Judging from the example value it would be the server version, but I'd say the browser version is probably the interesting value?

juliushaertl:
Fixes #4177 

Simplify the relevant server details a bit as the version of text is currently bound to the server releases anyways. Having it outside of a details block also makes it more visible.


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -20,32 +20,17 @@ A clear and concise description of what you expected to happen.
...
 
+**Server details:**
+- Nextcloud version: [e.g. 26.0.1]
+- PHP Version: [e.g. 8.2]
+- Database: [e.g. MariaDB 10.6]
+
 **Client details:**
  - OS: [e.g. iOS]
  - Browser: [e.g. chrome, safari]
  - Version: [e.g. 22]
```
provokateurin, line 31: The issue I opened was specifically about this line. Can you updated to either say "browser version" or "os version"? I think it's still confusing.

provokateurin (changes requested):

.github/ISSUE_TEMPLATE/Bug_report.md
juliushaertl: Ah thanks, will clarify :)

PR #4261, opened 2023-06-07
chore: Simplify issue template

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/4177

Simplify the relevant server details a bit as the version of text is currently bound to the server releases anyways. Having it outside of a details block also makes it more visible.

Another attempt since somehow amending to the commit in #4260 failed through github code spaces and the branch got into a weird state also locally.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -20,31 +20,16 @@ A clear and concise description of what you expected to happen.
 **Screenshots**
 If applicable, add screenshots to help explain your problem.
 
+**Server details:**
+- Nextcloud version: [e.g. 26.0.1]
+- PHP Version: [e.g. 8.2]
+- Database: [e.g. MariaDB 10.6]
+
 **Client details:**
- - OS: [e.g. iOS]
- - Browser: [e.g. chrome, safari]
- - Version: [e.g. 22]
+ - OS: [e.g. Windows/macOS/Ubuntu]
+ - Browser: [e.g. Firefox, Chrome]
+ - Browser version: [e.g. 22]
  - Device: [e.g. iPhone6, desktop]
-
-<details>
-<summary>Server details</summary>
-<!--
-You can use the Issue Template application to prefill most of the required information: https://apps.nextcloud.com/apps/issuetemplate
--->
-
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -20,32 +20,17 @@ A clear and concise description of what you expected to happen.
...
+**Server details:**
+- Nextcloud version: [e.g. 26.0.1]
+- PHP Version: [e.g. 8.2]
+- Database: [e.g. MariaDB 10.6]
+
 **Client details:**
- - OS: [e.g. iOS]
- - Browser: [e.g. chrome, safari]
- - Version: [e.g. 22]
+ - OS: [e.g. Ubuntu]
```
provokateurin, line 29: Maybe also list Windows and MacOS. Non-technical users might not understand this.

provokateurin (changes requested):

.github/ISSUE_TEMPLATE/Bug_report.md
juliushaertl: Added 👍

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -20,32 +20,17 @@ A clear and concise description of what you expected to happen.
...
+- Database: [e.g. MariaDB 10.6]
+
 **Client details:**
- - OS: [e.g. iOS]
- - Browser: [e.g. chrome, safari]
- - Version: [e.g. 22]
+ - OS: [e.g. Windows/macOS/Ubuntu]
+ - Browser: [e.g. Firefox, Chrome]
+ - Browser version: [e.g. 22]
  - Device: [e.g. iPhone6, desktop]
```
provokateurin, line 32: wouldn't something like `desktop, mobile, tablet` represent the info better? it doesn't really matter what exact device it is, just it's class

juliushaertl: The exact device might be relevant for mobile issues where for example on Android the on screen keyboard might introduce different behavior based on the vendors to the contenteditable HTML attribute

provokateurin: I see

PR #4266, opened 2023-06-08
Introduce middleware for session handling

juliushaertl:
### 📝 Summary

This PR introduces a middleware for checking if a valid text session is present. Controllers can annotate this requirement using the `#[RequireDocumentSession]` attribute and furtheron rely on the `getSession` method of the `ASessionAwareController` base class.

For now only two controllers are migrated as the others still pass around the request parameters so this will require some more refactoring in a follow up.

Integration test in cypress is added for the user api controller to cover the check for a valid session.

Also did some code cleanup with new PHP features along the way.

### 🏁 Checklist


cypress/e2e/api/UsersApi.spec.js
```diff
@@ -0,0 +1,110 @@
+/*
+ * @copyright Copyright (c) 2022 Max <max@nextcloud.com>
+ *
+ * @author Max <max@nextcloud.com>
```
max-nextcloud, line 4: I assume this is your copyright - or has this test only been moved?

max-nextcloud (approved):
Great work. So much nicer than before.

cypress/e2e/api/UsersApi.spec.js
juliushaertl: I started with a copy of the other API one ;) Will update to keep us both

mejo-:
/backport f6f2b804342bc82f18d5c63e1b0c05e70113ec3f to stable27

Issue #2708, opened 2022-07-12
Issue pasting certain tables


juliushaertl:
- Save the following table as table.html
- Open it in the browser
- Copy it and paste it into text

```
<table>
<tr style="height: 63px;">
<th style="width: 15.8024%; height: 63px;">
Country
</th>
<th style="width: 31.7902%; height: 63px;">
Partner / Service Providers
</th>
<th style="width: 26.8211%; height: 63px;">
Website
</th>
<th style="width: 10.0771%;">
Type
</th>
<th style="width: 16.8673%;">
Customer type
</th>
</tr>
<tr>
<td style="width: 15.8024%;">Australia</td>
<td style="width: 31.7902%;">Foo</td>
<td style="width: 26.8211%;"><a href="http://www.nextcloud.com.com/">http://www.nextcloud.comm/</a></td>
<td style="width: 10.0771%;">VAR</td>
<td style="width: 16.8673%;">&nbsp;</td>
</tr>
</table>
```

![Screenshot 2022-07-12 at 08 15 46](https://user-images.githubusercontent.com/3404133/178421566-4f15664e-9a8c-4cc0-b17d-e0f7dd415fa7.png)

juliushaertl:
More simplified repro:

```
<table>
<tr>
<th>Country</th>
<th>Partner</th>
<th>Website</th>
<th>Type</th>
<th>Customer</th>
</tr>
<tr>
<td>Australia</td>
<td>Foo</td>
<td>dfg</td>
<td>VAR</td>
<td>sdf</td>
</tr>
</table>
```

![Screenshot 2022-07-12 at 08 26 35](https://user-images.githubusercontent.com/3404133/178423160-610092c7-e3c5-4e34-bf76-73486c9b9265.png)

juliushaertl:
@max-nextcloud Any idea on why that could happen?

max-nextcloud:
I don't get it. This looks pretty much just like our table definition.

juliushaertl:
Seems that browsers add a tbody to the Dom automatically:
![Screenshot 2022-07-13 at 06 54 50](https://user-images.githubusercontent.com/3404133/178653667-4fdd6ef6-d1d1-43ff-bba3-a27168066b25.png)

max-nextcloud:
We should also have a rule for that:
```
{ 
  tag: 'table tbody :first-child th',
  priority: 60
}
```

[...]

PR #4285, opened 2023-06-12
Fix/2708 pasting tables

max-nextcloud:
### 📝 Summary

Fixes https://github.com/nextcloud/text/issues/2708.

Try to read the first table row as headings.

Pasting works via `insertContent` while opening uses `setContent`.
They use the schema in different ways.
So we also need to make sure to test both for some corner cases.

`setContent` is fairly flexible in turning the input
into a valid document structure.
`insertContent` however fails to resolve structures
that would require picking lower priority parent elements.

Note: Some tests in src/tests/nodes/Table.spec.js
fail when using `insertContent` instead of `setContent`.
Pasting the correponding html table is fixed never the less.

### 🏁 Checklist


src/tests/markdown.spec.js
```diff
@@ -1,7 +1,11 @@
 import spec from "./fixtures/spec"
 import markdownit from './../markdownit'
 import { typesAvailable } from './../markdownit/callouts'
-import { markdownThroughEditor, markdownThroughEditorHtml } from "./helpers";
+import {
+	markdownThroughEditor,
+	markdownThroughEditorHtml,
+	markdownFromPaste
+} from "./helpers";
```
mejo-, line 8: Nitpicking, but this probably needs adjustments to please the linter (single quotes, no semicolon).

mejo-:
Probably we should backport this all the way back to stable25.

src/tests/markdown.spec.js
max-nextcloud: Good catch. Linter is not complaining though... funny.

src/components/PlainTextReader.vue
```diff
@@ -37,7 +39,7 @@ export default {
 		renderHtml(content) {
 			return '<pre>' + escapeHtml(content) + '</pre>'
 		},
-		extensions: () => [PlainText],
+		extensions: () => [PlainText, CodeBlock],
```
juliushaertl, line 42: Is this change related or shouldn't this be a separate PR?

juliushaertl (commented):
Seems good just a comment about if the code block change is related or rather something for a separate PR.

src/components/PlainTextReader.vue
max-nextcloud: Good point... I'll split!

juliushaertl (approved):

mejo-:
/backport 69d58894a94909068e1915179042a02c398a2e98,92a3dcf9eddf616b92c203c1843500624faa9b93 to stable27

mejo-:
/backport 69d58894a94909068e1915179042a02c398a2e98,92a3dcf9eddf616b92c203c1843500624faa9b93 to stable26

mejo-:
/backport 92a3dcf9eddf616b92c203c1843500624faa9b93 to stable25

mejo-:
Manual backport to stable25: https://github.com/nextcloud/text/pull/4326

PR #4286, opened 2023-06-12
fix(sync): Save even if versions match

### LINKED ISSUE #3404: File might not be saved if close is happening too fast
Cypress test showing the issue:

https://github.com/nextcloud/text/blob/f7596d29f3e17e52f5c9825b8128fbb3117e6130/cypress/e2e/directediting.spec.js#L56-L78

There is a sync request issued before close but that may not trigger the save due to the short timespan.

We might want to always send the markdown content with the close request to save if nothing changed.

max-nextcloud:
### 📝 Summary

Fixes #3404.

Apply saves even if the client and server version match.

The client version only reflects the steps which the client has received back from the server.
It may leave out the steps the client just send itself.

So if the versions match - save the file to be sure to include changes from the client.

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/directediting.spec.js
+++ b/cypress/e2e/directediting.spec.js
@@ -61,15 +61,10 @@ describe('direct editing', function() {
 			})
 		const closeRequestAlias = 'closeRequest'
 		cy.intercept({ method: 'POST', url: '**/session/close' }).as(closeRequestAlias)
-		cy.intercept({ method: 'POST', url: '**/apps/text/session/sync' }).as('sync')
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):

juliushaertl:
> Alternatively we could also count the version number up on the client side when we push changes that directly apply as we know we're synced to that point.

Let's consider this as a possible follow up. 

As discussed previously this might cover at least the cases where the push happened but the sync would be still pending. In the long term we may just split out the save.

Issue #3902, opened 2023-03-06
Refactor ApiService to fetch session once


max-nextcloud:
In the long term maybe we should only fetch the session once and pass it along on further calls

_Originally posted by @juliushaertl in https://github.com/nextcloud/text/pull/3901#discussion_r1126241615_

PR #4290, opened 2023-06-13
Make use of session middleware in remaining controllers

juliushaertl:
Another larger backend changeset as a second part to #4266 to actually make use of it in the editing related controllers among a few cleanup tasks I did along the way.

Overall this should have some positive impact on performance as we reduce queries to fetch the session and document data to just one call in the middleware. Further way down in the the controller methods we just pass the objects around.

Let's see what cypress things about this.

- chore: Annotate controller methods in routes.php
- chore: Cleanup workspace controller
- chore: Refactor remaining controllers to make use of the session middleware
- chore: Cleanup in direct editing backend
- chore: Cleanup psalm baseline

### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/3902

### 🏁 Checklist


juliushaertl:
Need to check why the api test is flaky, could be either due to caching of the session or maybe the close request is not properly waited for

blizzz (approved):

PR #4297, opened 2023-06-14
Emit sync event after successful push to pick up syncing again

juliushaertl:
- fix: Emit sync event after successful push to pick up syncing again
- tests(cypress): Refactor reconnect test to wait more reliably

### 🚧 TODO

- [x] Needs #4286 

### 🏁 Checklist


juliushaertl:
Found two other reconnect bug when testing but those would be something for a different round:


- Open text
- Set to offline in network tab of browser console
- Click reconnect
- 🐛 It seems to show an error but request still happen and there is a continuous loading spinner

--- 

- Open text
- Delete the oc_text_session entry
- 🐛 See an empty editor

juliushaertl:
>  Still need to figure out how to stabilize the save after reconnect

@max-nextcloud Do you have any suggestion or code pointer how we do that in other cypress tests? Might be worth to look into separating the save request soonish ;)

max-nextcloud:
> @max-nextcloud Do you have any suggestion or code pointer how we do that in other cypress tests? Might be worth to look into separating the save request soonish ;)

Looks like the `save if versions match` commit addressed this, right?

We have this in a bunch of places and my usual workaround was to wait for at least two syncs. Used to be so that first sync ensures steps by others are synced and second sync could then make sure we are up to date with our own steps.

I think if we fix the actual issue and maybe separate the save and sync requests we could get rid of a few of those waits.

max-nextcloud:
> I think if we fix the actual issue and maybe separate the save and sync requests we could get rid of a few of those waits.

The only example of that pattern i could find outside the directediting spec is unrelated:
https://github.com/nextcloud/text/blob/d32b251883057af456679de90322eeb1e716618a/cypress/e2e/nodes/FrontMatter.spec.js#L76C1-L77

So maybe we got rid of them already.

cypress/e2e/sync.spec.js
```diff
@@ -64,28 +64,28 @@ describe('Sync', () => {
 			.should('include', 'saves the doc state')
 	})
 
-	it('recovers from a lost connection', () => {
+	it.only('recovers from a lost connection', () => {
```
juliushaertl, line 67: To remove

mejo- (approved):
Thanks a lot for looking into this @juliushaertl! Code changes look sensible to me (apart from `.only` in the Cypress test).

Issue #3925, opened 2023-03-10
Translate design feedback


juliushaertl:
From reviewing with @jancborchardt 

- Put full text into the input when nothing is selected
- Change layout to be more like google translate/deepl (side-by-side) – of both the dropdowns as well as the big textareas
- Add placeholder text to language selectors
- Small info text that selected text could be used (To translate individual parts of the text, select it before using the translate function.)
- Texareas should not be resizable
- Keep initial focus on the first input field

thgoebel:
> Texareas should not be resizable

The text areas should automatically resize to fit its contents. The default 2.5 lines is not realistic imho, most texts are longer. It's super crammed:

![image](https://github.com/nextcloud/text/assets/33295590/93732845-1af3-41a0-bbeb-d28efa793cec)

juliushaertl:
@luka-nextcloud Please check if the description makes sense so far or if you need further input from the design team. Also see if you manage to setup translations either with deepl https://github.com/juliushaertl/integration_deepl/ or https://github.com/nextcloud/translate. The first one might be easier in docker based envirnoments

luka-nextcloud:
@juliushaertl 
- I was able to setup https://github.com/nextcloud/translate
- `Add placeholder text to language selectors` -> please provide placeholder values
- `Small info text that selected text could be used (To translate individual parts of the text, select it before using the translate function.)` -> I don't really get this point. I have checked translate functionality. It is already working as described (select a text then use translate function).
- `Textareas should not be resizable` -> what is the final decision?
- `Keep initial focus on the first input field` -> It's already done, right? (currently, first input field is always focused when opening the translate popup form)

jancborchardt:
- Placeholder text should be "Select language" for both
- The fields should as said not be manually resizable, and as @thgoebel said use autosize (think we discussed that as well @juliushaertl). Note with the side-by-side layout, the height of the text boxes can also increase a bunch, to fill out the rest of the modal height.
- The suggested info text for below the heading is written in the original issue:
  > To translate individual parts of the text, select it before using the translate function.

PR #4299, opened 2023-06-15
Feature/translate modal

luka-nextcloud:
### 📝 Summary

* Resolves: #3925 

#### 🖼️ Screenshots

| Desktop |
|---|
| grafik |

Mobile1 | Mobile2
---|---
grafik | grafik

### 🏁 Checklist


src/components/Modal/Translate.vue
```diff
@@ -172,20 +190,42 @@ export default {
...
 </script>
 
 <style lang="scss" scoped>
 .translate-dialog {
 	margin: 24px;
+
+	.wrapper {
+		display: grid;
+		grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
+		grid-column-gap: 10px;
```
marcoambrosini, line 214:
```suggestion
		grid-column-gap: 16px;
		margin-top: calc(var(--default-grid-baseline) * 6) 
```

src/components/Modal/Translate.vue
```diff
@@ -172,20 +190,43 @@ export default {
...
+
```
marcoambrosini, line 215: language-selector {
    font-weight: bold;
    flex-wrap: wrap;
    gap: var(--default-grid-baseline);
}

marcoambrosini (commented):
Nice work, I just dropped a few style suggestions. 

Maybe we could also change the hint into something like: 
You can translate parts of the document by selecting the text before using the translate function.

juliushaertl:
A few more small issues that I think would be worth to address:

- [ ] Limit the maximum height of the text areas so that the buttons at the bottom are still visible without scrolling
- [ ] The result text area should probably still be scrollable as currently the translated text is being cut off if it is longer then the source
Screenshot 2023-06-20 at 15 17 50

nimishavijay:
Looks good! :) I would be careful about small screens here, it should change to a single column layout. 

Also a note on the labels: "source language" and "target language" seem a bit technical, and on Google Translate as well as Deepl, there are no labels at all. 
- What do you think about using "Translate from" and "Translate to"? 
- Alternatively, we could get rid of labels entirely, make the multiselects to full width and use a [right arrow icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aarrow_forward%3A) in between them to indicate the source and target language. 

What do you think @nextcloud/designers ?

juliushaertl:
@mejo- Could you give this a review? I'd say we could merge this already and do a follow up for the remaining open question to be discussed by the design team.

src/components/Modal/Translate.vue
```diff
@@ -20,26 +20,44 @@
...
+					</div>
+					<textarea ref="input"
+						v-model="input"
+						autofocus="true"
+						@input="autosize"
+						@focus="onInputFocus" />
+				</div>
+				<div class="col">
+					<div class="language-selector">
+						<label for="toLanguage">{{ t('text', 'Translate to') }}</label>
```
mejo-, line 46: How about changing this label to "to" only. Not sure whether this is acceptable from accessibility perspective, but it would read way nicer "Translate from <language> to <language>". @nimishavijay @marcoambrosini

src/components/Modal/Translate.vue
```diff
@@ -20,26 +20,44 @@
...
 		<div class="translate-dialog">
 			<h2>{{ t('text', 'Translate') }}</h2>
-			<textarea v-model="input" />
-			<div class="language-selector">
-				<label for="fromLanguage">{{ t('text', 'Source language') }}</label>
-				<NcSelect v-model="fromLanguage"
-					input-id="fromLanguage"
-					:options="fromLanguages"
-					:append-to-body="false" />
+			<em>{{ t('text', 'You can translate parts of the document by selecting the text before using the translate function.') }}</em>
```
mejo-, line 26: The text should be "To translate individual parts of the text, select it before using the translate function." as mentioned by @jancborchardt [here](https://github.com/nextcloud/text/issues/3925#issuecomment-1566903096)

src/components/Modal/Translate.vue
```diff
@@ -172,20 +191,67 @@ export default {
...
 }
 
 textarea {
 	display: block;
 	width: 100%;
 	margin-bottom: 12px;
+	height: auto;
+	resize: none;
+	box-sizing: border-box;
+	overflow: hidden;
```
mejo-, line 237:
```suggestion
	overflow-y: auto;
```

src/components/Modal/Translate.vue
```diff
@@ -172,20 +191,67 @@ export default {
...
+	max-height: 60vh;
+
+	&.result {
+		overflow: scroll;
```
mejo-, line 241: Why just for the result? Let's remove the `.result` class and set `overflow-y: auto` for both, no?

src/components/Modal/Translate.vue
```diff
@@ -94,6 +112,7 @@ export default {
 			canDetect: loadState('text', 'translation_can_detect'),
 			loading: false,
 			error: null,
+			disableFromLanguageSelect: true,
```
mejo-, line 115: What's the purpose of `disableFromLanguageSelect`? It seems to be set to `false` immediately by `onInputFocus` when the modal is opened. So maybe a leftover that can go away?

src/components/Modal/Translate.vue
```diff
@@ -20,26 +20,44 @@
...
+						<NcSelect v-model="fromLanguage"
+							input-id="fromLanguage"
+							:placeholder="t('text', 'Select language')"
+							:options="fromLanguages"
+							:disabled="disableFromLanguageSelect"
+							:append-to-body="false" />
+					</div>
+					<textarea ref="input"
+						v-model="input"
+						autofocus="true"
```
mejo-, line 40: `autofocus` doesn't support a value, see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

```suggestion
						autofocus
```

mejo- (changes requested):
Nice work @luka-nextcloud!

Could you address the two issues mentioned by Julius [here](https://github.com/nextcloud/text/pull/4299#issuecomment-1598768059)? Both textareas should have `overflow-y: auto` and the buttons should always be visible, even with small window height.

Some extra comments:

* [ ] I don't find it very intuitive that I have to press 'Translate' after both source and target language have been selected. Why not remove the button and instead start translating immediately after both source and target have been selected? What do you think @nimishavijay and @marcoambrosini?
* [ ] On mobile view, label and selection of languages are not aligned properly due to different label length. Maybe always put them in separate lines on mobile? See screenshot: 
grafik

juliushaertl:
>  I don't find it very intuitive that I have to press 'Translate' after both source and target language have been selected. Why not remove the button and instead start translating immediately after both source and target have been selected?

Especially with the local translation model this can take quite some time and server resources so unless we can always ensure a fast response I'd try to avoid auto triggering that request

nimishavijay:
> On mobile view, label and selection of languages are not aligned properly due to different label length. Maybe always put them in separate lines on mobile? 

True! There definitely needs to be some sort of rearrangement on mobile. We could either:
1. Show the labels and dropdowns on separate lines
2. Shorten the labels
3. Not show external labels at all, instead show internal labels "Translate from German" "Translate to English" inside the dropdowns in the closed state and make the dropdowns full width. 

I would vote to show solution 3 (only on mobile), what do you think? @nextcloud/designers

src/components/Modal/Translate.vue
nimishavijay: It would definitely read nicer on desktop. However, more than a11y I am concerned about translation to other languages, and it may be confusing on mobile as well, what do you think @marcoambrosini ?

marcoambrosini: I think that it's fine to have just from and to since the heading of the dialog already says translate

src/components/Modal/Translate.vue
luka-nextcloud: It is for fixing the autofocus of first select box since we expect to autofocus on the first input textarea.

luka-nextcloud: @mejo- Is it all good now?

luka-nextcloud:
@mejo- Please check again.

src/components/Modal/Translate.vue
```diff
@@ -105,7 +127,7 @@ export default {
 					set.add(item.from)
 					result.push({
 						id: item.from,
-						label: item.fromLabel,
+						label: !this.$isMobile ? item.fromLabel : t('text', 'Translate from ' + item.fromLabel),
```
mejo-, line 130:
```suggestion
						label: !this.$isMobile ? item.fromLabel : t('text', 'Translate from {language}', { language: item.fromLabel }),
```

This works better for translations into languages which don't have the language prepended to the end.

src/components/Modal/Translate.vue
```diff
@@ -129,7 +151,7 @@ export default {
 					set.add(item.to)
 					result.push({
 						id: item.to,
-						label: item.toLabel,
+						label: !this.$isMobile ? item.toLabel : t('text', 'Translate to ' + item.toLabel),
```
mejo-, line 154:
```suggestion
						label: !this.$isMobile ? item.toLabel : t('text', 'Translate to {language}, { language: item.toLabel }),
```

same here.

mejo- (changes requested):

jancborchardt:
Are the screenshots current or could you post updated ones @luka-nextcloud?
Ideally both for desktop (side by side layout) and mobile (single column layout). :)

mejo-:
@jancborchardt @luka-nextcloud I took the liberty to now update the screenshots in the original PR description.

mejo-:
@thgoebel this PR didn't get backported to Nextcloud 27. It will be released with Nextcloud 28.

PR #4309, opened 2023-06-19
fix: Pass origin along the initial state update

juliushaertl:
### 📝 Summary

* Resolves: # 

This ensures that the initial update that is applied to set the document state is not added as another step.

https://github.com/yjs/y-websocket/blob/master/src/y-websocket.js#L331

### 🏁 Checklist


mejo- (approved):
Code changes look sensible to me. Good catch. Is this merely a performance improvement or does it fix a user-facing bug?

juliushaertl:
Not fully sure yet also if we should backport.

I could imagine this may solve problems where the file is getting version bumps and then random saves without actual file content change. Other than that I can so far not think about a bug scenario.

mejo-:
If there is a chance that this fixes unsolicited file saves (even if just in rare corner cases), then I'd say we should backport it given that people are so unhappy about them :wink:

mejo-:
/backport ac5bc7b0f0637331b2da32793178819a81a33f5f to stable27

mejo-:
/backport ac5bc7b0f0637331b2da32793178819a81a33f5f to stable26

PR #4326, opened 2023-06-20
[stable25] fix: paste tables into document

### LINKED ISSUE #2708: Issue pasting certain tables

mejo-:
Manual backport of #4285.

Fixes #2708.

Try to read the first table row as headings.

Pasting works via `insertContent` while opening uses `setContent`. They use the schema in different ways.
So we also need to make sure to test both for some corner cases.

`setContent` is fairly flexible in turning the input into a valid document structure.
`insertContent` however fails to resolve structures that would require picking lower priority parent elements.

Note: Some tests in src/tests/nodes/Table.spec.js
fail when using `insertContent` instead of `setContent`. Pasting the correponding html table is fixed never the less.

### 🏁 Checklist


datenangebot (approved):
Code looks good, but I don't have deeper knowledge here.

PR #4328, opened 2023-06-20
chore(dependencies): Skip tiptap/prosemirror updates for 25

juliushaertl:
As discussed with @max-nextcloud recently and to avoid reopening of https://github.com/nextcloud/text/pull/3752

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -62,6 +62,15 @@
       "enabled": false
     },
     {
+      "groupName": "tiptap",
+      "matchBaseBranches": ["stable25"],
+      "matchPackagePrefixes": [
+        "@tiptap/",
+        "prosemirror-"
+      ],
+      "enabled": false
+    },
+    {
       "matchPackageNames": ["vue"],
       "allowedVersions": "<3"
     },


renovate.json
```diff
@@ -61,6 +61,20 @@
...
+      "groupName": "tiptap",
+      "matchBaseBranches": ["stable25"],
+      "matchPackagePrefixes": [
+        "@tiptap/",
+        "prosemirror-"
+      ],
+      "enabled": false
+    },
+    {
+      "matchUpdateTypes": ["major"],
```
mejo-, line 74: Do we need this block twice?

juliushaertl: No, copy paste seems harder than expected 🙈 Fixed.

mejo-: Hehe, how about swapping the two blocks, so that the two tiptap blocks are one after each other? Sorry for the nitpicking, feel free to ignore :zany_face:

mejo- (approved):

renovate.json
juliushaertl: I thought about that but wanted to rather group by stable branches from new to old ;)

Issue #4126, opened 2023-05-05
Hide user cursors


juliushaertl:
User cursors might be in the way for reading text. Ideally we only show the names on hover or when people are editing. Other than that a checkbox to disable them in the collaborator menu might be worth as well.

What do you think @max-nextcloud @nimishavijay ?

mejo-:
I like the idea to always show the cursors but only display the names on hover and while they're typing (with a timeout, maybe something like 3-5 seconds).

md-work:
Please add an option to completely disable the user cursors.

I've just been in a meeting where a lot of people were annoyed by all the user cursors blinking around (optical noise).  
And even if text was just overlapped for seconds many people were disturbed when reading text.

max-nextcloud:
I agree it should be possible to hide them. I wonder if it would be a middle ground to only show them when people are actually typing - but that also seems harder to accomplish.

nimishavijay:
@jancborchardt Good point about removing the setting :) 

for 1) it is enabled by default and people can switch it off from there, and for 3) it wouldn't be a per document setting, ideally? it would be a per user setting so once you unset it it would remain like that. What do you think? @max-nextcloud @mejo- would this work?

kolAflash:
> I agree it should be possible to hide them. I wonder if it would be a middle ground to only show them when people are actually typing - but that also seems harder to accomplish.

**I think showing the user cursor labels when typing is NOT a good middle ground.**  

Maybe showing a pure colored cursor without label is OK.  
But **if there's a label it will always overlap other text. Doesn't matter if someone is typing or not.**

If you really want to have some feature like this, do it like Etherpad Lite. Coloring the text background and provide some user list view at the side.  
https://etherpad.org/img/etherpad_demo.gif

&nbsp;

Please consider:  
There's finite space. And it has to be decided how to use it.  
**You can't have everything without making compromises!**  
**I'd say first priority has to display the text, because that's why users are here for.**

And additionally:  
**Please carefully test before releasing.**  
The user cursors [caused a mess of troubles](https://github.com/nextcloud/text/issues/4100) in Nextcloud-26.0!  
**If the basics are not working stable, everything additional like users cursors becomes worthless!**

PR #4336, opened 2023-06-21
Fade out user cursor labels after five seconds of inactivity

mejo-:
### 📝 Summary

Fade out user cursor labels after inactivity.

Implementation details:
* Add a custom CollaborationCursor Tiptap extension
* Add CSS to fade out the cursor label after some time.
* Listen for Yjs updates.
  - If it's a doc change by ourself, update the timestamp of own user in
    awareness state.
 - If it's a remote awareness update, add back the CSS class to the
   corresponding cursor.
 - Wait 50ms before showing the cursor in the DOM to account for cases
   where the cursor gets re-rendered by y-prosemirror.

Fixes: #4126

### 🏁 Checklist


src/helpers/yjs.js
```diff
@@ -42,3 +55,29 @@ export function applyDocumentState(ydoc, documentState, origin) {
...
+	const decoder = decoding.createDecoder(update)
+	const type = decoding.readVarUint(decoder)
+	if (type !== yjsMessageTypes.awareness) {
+		return []
+	}
+
+	const len = decoding.readVarUint(decoder)
+	const clientIds = []
+	for (let i = 0; i < len; i++) {
+		clientIds.push(decoding.readVarUint(decoder))
```
juliushaertl, line 80: I think we still need to read the other values here as well to continue at the correct byte position for the next client if.

src/components/Editor.vue
```diff
@@ -642,6 +644,35 @@ export default {
...
+
+		updateCursorLabel: debounce(function(user) {
+			// Remove CSS class that fades out and hides the cursor label and bring it back after five seconds.
+			// Limit this to once every second to prevent it from happening with each awareness update.
+			const cursorLabelEls = document.getElementsByClassName(`collaboration-cursor__label__${user.name}`)
+			for (const el of cursorLabelEls) {
+				el.classList.remove('collaboration-cursor__label_fadeout')
+				clearTimeout(this.$cursorLabelTimeouts[user.clientId])
+				this.$cursorLabelTimeouts[user.clientId] = setTimeout(() => {
+					el.classList.add('collaboration-cursor__label_fadeout')
```
juliushaertl, line 671: Could there be cases where the element no longer exists here after the timeout?

mejo-: Good point. It's possible (because the session got closed). I'll check for the element and only apply the changes in case it exists.

mejo-: Well, given that I already get the list of elements with `getElementsByClassName` and iterate over them, the code should be fine as is, no?

src/helpers/yjs.js
mejo-: Good point. Tried to fix the code now. But given that I for now didn't see an update message with more than one awareness update, I was not able to test this in practice. Do you have a good idea how to create a test scenario?

src/components/Editor.vue
```diff
@@ -642,6 +644,35 @@ export default {
 			this.emit('delete-image-node', imageUrl)
 		},
 
+		onYjsUpdate(update, _origin, _doc, _tr) {
+			// Update cursor labels for clients from awareness message
+			if (getMessageType(update) === yjsMessageTypes.awareness) {
```
juliushaertl, line 649: Quickly looking at it again we can maybe just exclude our own updates with checking against the orign like in https://github.com/nextcloud/text/pull/4309

juliushaertl: Also regarding the 60 and 62 message types it would be interesting to see if they get triggered somehow from the initial document state that is applied from the state in the lines from the same PR.

juliushaertl (changes requested):
Generally the change looks good to me, animation behaviour seems very stable and looks nice. However updating the timestamp on the awareness state seems to cause two push requests for typing an individual character (one for the awareness, one for the step). Maybe we can combine them in some way?

src/services/SyncService.js
```diff
@@ -165,7 +162,7 @@ class SyncService {
 					this.#sendIntervalId = null
 					this._sendSteps(getSendable).then(resolve).catch(reject)
 				}
-			}, 200)
+			}, 50)
```
juliushaertl, line 165: This could have an impact on the resource demand due to increased push requests on the server. I think the PR itself should also be fine with the existing 200ms. Any reason to decrease that specific to the cursor hiding?

juliushaertl (approved):
Tested and works, one question regarding reducing the push interval inline

src/services/SyncService.js
max-nextcloud: This is a bit tricky as we change the initial waiting. Before if there was no request pending we would just send it immediately - so the 200 ms had no effect at all. Now we also delay this initial message.

I remember that 200ms are roughly what the human brain starts to notice as a delay. But that is only if one is waiting for a direct response from the UI. This will only delay the sync between different parties. So 200 ms are probably okay there as well.

I'd say we go with the 200 ms and if we notice usability issues with it we can change it to 50 ms on the initial request and 200 when another request is pending.

max-nextcloud: Okay... so i gave this a try locally with 200 and noticed that the polling delay still is the main delay one notices. If I type fairly fast 200ms will result in two keystrokes getting combined in one push request. That seems totally reasonable. If this actually reduces the number of requests send - even better.

max-nextcloud:
Funny that lint passes even though we introduced a warning.
I think the workflow probably was updated from a template.

max-nextcloud:
/backport b7baceb6c648b20f56270f2bdaed5c55f2caab6c,928fe3527a7467c78612e7db43fe35dead65819a to stable27

max-nextcloud:
backport b7baceb6c648b20f56270f2bdaed5c55f2caab6c,928fe3527a7467c78612e7db43fe35dead65819a to stable26

max-nextcloud:
I'm not sure about the backports. 27 is pretty clear to me also as this does not break any api and 28 will still be quite some time into the future. But 26 I don't know.

mejo-:
I think we discussed this already and decided to backport this to stable26 as well, because it fixes an annoying usability bug.

max-nextcloud:
backport b7baceb6c648b20f56270f2bdaed5c55f2caab6c,928fe3527a7467c78612e7db43fe35dead65819a to stable26

max-nextcloud:
/backport b7baceb6c648b20f56270f2bdaed5c55f2caab6c,928fe3527a7467c78612e7db43fe35dead65819a to stable26

Issue #3844, opened 2023-02-27
Strange behaviour of background when direct editing on narrow screens


AlvaroBrey:
**Describe the bug**
When direct editing on a narrow screen (i.e. mobile), and typing a line that's longer than the screen width, the white background for the text editor is extended only around the longer lines, but keeps the screen width elsewhere. See screenshot below as it's easier to understand.

**To Reproduce**
1. Open a file through direct editing in a narrow screen (mobile, or just a resized desktop browser window)
2. Ensure there's some text lines that are wider than the viewport
3. Those lines will overflow to the right: scroll horizontally to the right to see the background mismatch

**Expected behavior**
I expected either:
a) the text to wrap around to the width of the viewport, or
b) The entire background expanding to the width of the text, to avoid this effect

Additional info: this does not happen in the company instance, the entire background is white there.

**Screenshots**

<details>

![Screen Shot 2023-02-27 at 16 29 29](https://user-images.githubusercontent.com/8677743/221607066-186e74ea-5682-41bb-a74e-d88627748f0c.png)

</details>

**Client details:**
 - OS: Linux, Android
 - Browser: Desktop Firefox, Android Chrome, Android Webview
 - Version: -
 - Device: -

<details>
<summary>Server details</summary>

**Text app version:** main

**Operating system:** debian bullseye

**Web server:** apache2

**Database:** SQLite

**PHP version:** 8.x

**Nextcloud version:** master (26.0.0 beta5)

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log

</details>

juliushaertl:
@AlvaroBrey Did you test that with #3839 in place already? Haven't tried yet but could fix this already

luka-nextcloud:
It seems like it was fixed already.

![image](https://github.com/nextcloud/text/assets/89908051/4a10c090-247f-4532-b858-ed0438a20112)

mejo-:
@luka-nextcloud did you intend to close the issue?

juliushaertl:
Just looked into this with @luka-nextcloud and it is still reproduicable.

- Create a welcome.txt file in the web interface
- Have a long line without spaces e.g. `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`
[...]

PR #4337, opened 2023-06-21
fix: max width of editor container on mobile

luka-nextcloud:
### 📝 Summary

* Resolves: #3844

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/components/Editor/ContentContainer.vue
```diff
@@ -67,6 +67,12 @@ export default {
 		width: 100%;
 	}
 
+	@media (max-width: 670px) {
```
juliushaertl, line 70: It wasn't immediately obvious to me that this is the value of `--text-editor-max-width`, maybe add a code comment to this.

Maybe we can just set a max-width in the above general `.editor_content` rule like this to keep it simple?

```css
.editor__content {
	max-width: min(var(--text-editor-max-width, calc(100vw - 16px));
```

juliushaertl (changes requested):

src/components/Editor/ContentContainer.vue
luka-nextcloud: Ok, please check again.

juliushaertl (approved):

juliushaertl:
/backport 8765b4f8ef1dd6a7ded01b357d31f8e9a8c9328d to stable27

juliushaertl:
/backport 8765b4f8ef1dd6a7ded01b357d31f8e9a8c9328d to stable26

PR #4350, opened 2023-06-23
debug: cypress failure on upload / CSRF / Session failure with 401

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [ ] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [ ] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required

PR #4357, opened 2023-06-24
ci/flaky runners

juliushaertl:
ready to merge changes from tests in #4350 

- use postgres for cypress which seems to stabilize the random failures a bit
- Change step order in mysql setup to give it some more time to be ready
- Use only github runners for oci for now

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -75,10 +75,20 @@ jobs:
         node-version: [16]
         containers: [1, 2, 3, 4, 5, 6, 7, 8]
         php-versions: [ '8.1' ]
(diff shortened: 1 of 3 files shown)


juliushaertl:
Will merge already to see how well this works during dependable updates. Reviews still welcome in case there is feedback I'll follow up.

mejo- (commented):
Thanks, looks good to me :ok_hand:

mejo-:
@juliushaertl shall we do manual backports of this to stable27/26/25?

juliushaertl:
Yes, maybe combined with the cypress rework once merged

PR #4364, opened 2023-06-24
Rework cypress commands

juliushaertl:
This is a set of changes to cypress commands that should increase stability over CI runs. There still seem to be rare failures where some requests return a 401 on the backend which I'm continuing to follow debugging in #4350 however I think this would already be good to get in fast:

- Move from axios usage to cy.request as I suspect that the axios promise was sometimes causing timing issues with requests
- Use cy.selectFile instead of the 3rdparty helper
- Rework recovery after lost connection test to await the session recreation
- Avoid some extra navigation steps

### DIFF
--- a/cypress/e2e/attachments.spec.js
+++ b/cypress/e2e/attachments.spec.js
@@ -21,7 +21,6 @@
  */
 
 import { initUserAndFiles, randHash, randUser } from '../utils/index.js'
-import 'cypress-file-upload'
 
 const user = randUser()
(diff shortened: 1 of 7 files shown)


mejo- (approved):
Thanks a ton for looking into this @juliushaertl! All changes look good, just the first commit message seems to be wrong and a bit cluttered.

juliushaertl:
Cleaned up the commit messages :)

Issue #3899, opened 2023-03-06
Ensure files are saved on close


max-nextcloud:
**Describe the bug**
Sometimes closing the editor won't save the file.
The sync request goes out but since the version has not increased it does not save anything.

**To Reproduce**
1. Open file in text
2. Type something and click `x` to close viewer directly after
3. Download the file
4. Sometimes observe last edits missing

**Expected behavior**
Closing the viewer should always save the file

Will provide logs once i recorded them.

### Background

When typing the changes are send out via the push endpoint.
However the version counter will only increment once a sync request happened.
In the meantime save requests will still have the same version and be ignored.

#### Increment version on push

We could increase the version whenever we push - but this would require including all intermediate changes by others in the answer to the push request. The version counter is mainly used to not fetch steps over and over again - so we cannot just increase it as that would lead to ignoring relevant steps.

#### Sync before saving

We could also use two requests for saving - one to ensure the current version is up to date and then one to actually save the file.

max-nextcloud:
Sync before saving is what we did in the past and i think it's the least intrusive and easiest change - even though it will require another roundtrip.

blizzz:
still a thing.

DavidMikeSimon:
I'm also still noticing this behavior

blizzz:
@DavidMikeSimon actually i cannot reproduce this behaviour anymore. Do you have some reliable reproduction steps perhaps?

max-nextcloud:
Thanks @DavidMikeSimon.
I was able to reproduce this. I think there are two issues at play here:
* Closing the viewer should remove the editor view and save the content - need to investigate what's happening there.
* We don't save anything if there have not been any changes by the current user - but we should probably only do that when it comes to autosaving - but still save on click.

juliushaertl:
@blizzz Will unassign you as it looks more like a frontend issue ;)

bentuna:
Under 28.0.9, we can still reproduce the issue sometimes: When you close a text document without clicking save, a part of your unsaved contents is lost. And: that does *not* only happen on external storage.

mejo-:
This should be solved nowadays. Please report if you still experience this problem.

PR #4394, opened 2023-06-26
fix(sync): ensure files are saved on close

blizzz:
### 📝 Summary

* Resolves: #3899 

The close logic was already in place, albeit missing a small detail. Without `manualSave` set to true, the backend would only save the file, when `AUTOSAVE_MINIMUM_DELAY` (10s) has passed since the last save. The flag would override it and trigger the save.

### 🏁 Checklist

- [X] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required


blizzz:
@juliushaertl we were talking about splitting the save from sync. It still makes sense, but for the plain nature of this is fix, I'd say it's worth to keep it separated and have this as a "quick" fix though (quotation marks because i took time to get familiar with the code, also about the splitting).

src/services/SyncService.js
```diff
@@ -270,7 +270,7 @@ class SyncService {
 	async close() {
 		this.backend?.disconnect()
 		const timeout = new Promise((resolve) => setTimeout(resolve, 2000))
-		await Promise.any([timeout, this.save()])
+		await Promise.any([timeout, this.save({ manualSave: false })])
```
juliushaertl, line 273: Shouldn’t the flag then be true here?

juliushaertl: Ah this is the default, then I don't get why turning off the manual save flag would fix this as now we would hit the `AUTOSAVE_MINIMUM_DELAY` https://github.com/nextcloud/text/blob/3b7877b14e4dc0b4decb2a5d212d15dd22b5d031/lib/Service/DocumentService.php#L343

blizzz: > Shouldn’t the flag then be true here?

Err… absolutely

blizzz: No, something's not right here. Seems I also got the repro steps not as reliable as is should, especially manualSave was even by default set to true.

blizzz: Now that's embarrassing. I cannot reproduce the original issue at all anymore. The `false` must have been a mistake from me yesterday due to cleaning up (had it at a different position first, plus debug notes). Anyway, embarrassing.

juliushaertl: Yeah, I think reproduction steps are loose and probably depend a lot on timing and if there were previous steps added to the document, so could well be that on the first look it seemed like the change worked.

The separate save request might not help directly, but would allow us to better observe what is happening in failure cases, make cypress tests more reliable as we could await the save request and maybe also helps with creating reproduction test cases later on where we artificially block individual requests upfront to simulate timing problems.

PR #4406, opened 2023-06-28
chore: update relativeci workflow and webpack usage

skjnldsv:
- Move the relativeci at the same stage as the node workflow
  image
- Fix webpack usage in package.json


juliushaertl (approved):

juliushaertl:
Cypress failure unrelated

PR #4424, opened 2023-06-29
Split sync of ApiService to sync and save

blizzz:
### 📝 Summary

As discussed with @juliushaertl when talking about #3899 it might be worthwhile to split the sync route into dedicated sync and save endpoints.

### 🚧 TODO

- [x] probably have to adjust cypress
- [x] take care of PublicSessionController

### 🏁 Checklist


src/services/SyncService.js
```diff
@@ -242,7 +242,7 @@ class SyncService {
 	async save({ force = false, manualSave = true } = {}) {
```
juliushaertl, line 242: We may still want to trigger the autosave somehow regular. I think this is no longer the case as only sync is called in an interval. Might be something to check with @max-nextcloud next week on how to best achieve that.

blizzz: It's actually still being triggered regularly. At least yesterday it was :see_no_evil:

blizzz: And that's the 30s autosave: https://github.com/nextcloud/text/blob/0d088d669fad3861387b27ca4e1b0cc51c147cab/src/services/SyncService.js#L88

blizzz:
Pushed 2ac9c6c9ccc81a8a5289f1ef0a589e2421074d7b and 8f74dec8f but they don't appear here :shrug: :octocat::zzz:

lib/Service/ApiService.php
```diff
@@ -233,14 +252,14 @@ public function sync(Session $session, Document $document, $version = 0, $autosa
 		}
 
 		try {
-			$result['document'] = $this->documentService->autosave($document, $file, $version, $autosaveContent, $documentState, $force, $manualSave, $token, $this->request->getParam('filePath'));
-		} catch (DocumentSaveConflictException $e) {
+			$result['document'] = $this->documentService->autosave($document, $file, $version, $autosaveContent, $documentState, $force, $manualSave, null, $this->request->getParam('filePath'));
```
max-nextcloud, line 255: I think you'd still want to hand over the `token` - now called `shareToken` as provided as an argument.

```suggestion
			$result['document'] = $this->documentService->autosave($document, $file, $version, $autosaveContent, $documentState, $force, $manualSave, $shareToken, $this->request->getParam('filePath'));
```

max-nextcloud (changes requested):
Thanks a lot for tackling this!
I have only one small remark.
Just went over the code did not test locally or go through untouched code to see if further changes would be needed.

lib/Service/ApiService.php
blizzz: Absolutely, good catch

max-nextcloud: I applied the suggestion as a fixup to see if that helps with cypress.

max-nextcloud (approved):

blizzz:
squahsed and rebassed

blizzz:
The remaining cypress errors about conflicts are rooted that this information was not transported on `/sync` anymore. This is fixed now. It works locally, but Cypress is still not convinced.… will continue there.

The handler we have, reacts to this status only on polling! Saves would be ignored, you could see the failures only on the console (before and after the changes). With the polling itself it's not a big deal.

blizzz:
The remaining issue with cypress conflicts tests was an accidentally inverted condition, so that the proper exception was not thrown.

Issue #4414, opened 2023-06-28
Lost content in collaborative session


mejo-:
**Describe the bug**
We have reports of lost content in collaborative editing sessions. The changes by one client are not visible to the other clients.

Current assumption is that the client stops sending its steps after something went wrong.

In one case we got the console log of the faulty client. It contained the following relevant log lines:

```
[ERROR] text: failed to apply steps due to collission, retrying.
[ERROR] text: Failed to fetch steps due to other reason.
[ERROR] text: Failed to save document.
```

**To Reproduce**
No clear reproducer yet.

**Server details:**
- Nextcloud version: 26.0.2 and 27.0.0 (company instance)

mejo-:
Description of the situation by the user whose changes didn't propagate:

We were working on a document collaboratively and some text written by one member of the session (me) did not synchronize properly to the document for all other members. The content was there, being added in real time, then all a sudden ALL of the additions added by me - even the previously visible-to-all-members text vanished (me believes he use a shortcut key for formatting, tabs, shift-enter, etc in an unsorted list when it happened).

Text visible/added by me was then copied to an offline doc (backup!) then the browser window was refreshed (i.e. logs might be incomplete).

Then, the previous backed up text (only partial section of document) was pasted back into the doc, visible to all members.

Upon continuing to live-edit in the document, only one other member seemed to not be able to see the new text being added. All other members are believed to have normal behaviour, save one member. That member refreshed, and had visibility now into the new edits, but then it was decided to not trust the document live-editing and record subsequent notes offline.

gerbrent:
> Did you by chance see a Nextcloud notification saying "Changes could not be sent yet"?


I was actively note-taking during a meeting/call, so may not have noticed, though I typically do. That said, I did not notice any such message.

gerbrent:
potential culprit worth investigating:

this scenario occurred while concurrently on an active 5 person Talk call. @gerbrent - the author of text that didn't propagate - residing on a different continent than the server, in a non-ideal internet connectivity situation. Thus latency, packet loss, bandwidth congestion are worth considering:

**Ping times to server**
* with no concurrent Talk call, i.e. low bandwidth demands:

```--- ping statistics ---
2137 packets transmitted, 2137 received, 0% packet loss, time 2136113ms
rtt min/avg/max/mdev = 195.863/207.985/353.368/11.831 ms```

Will attempt some pings in future calls in case that's helpful.

juliushaertl:
[...]

Issue #4340, opened 2023-06-21
Issues when reconnecting


juliushaertl:
Found two other reconnect bug when testing but those would be something for a different round:

- Open text
- Set to offline in network tab of browser console
- Click reconnect
- 🐛 It seems to show an error but request still happen and there is a continuous loading spinner

--- 

- Open text
- Delete the oc_text_session entry
- 🐛 See an empty editor

_Originally posted by @juliushaertl in https://github.com/nextcloud/text/issues/4297#issuecomment-1590580730_

PR #4474, opened 2023-07-10
Fix sync errors after network issues

mejo-:
### 📝 Summary

* Resolves: #4340
* Resolves: #4414

### 🚧 TODO

- [x] Look into Cypress test
- [ ] Remove `.only` from Cypress test
- [x] remove `disableBc: true` from `SyncServiceProvider.js`

### 🏁 Checklist


src/services/SyncServiceProvider.js
```diff
@@ -42,7 +42,7 @@ export default function createSyncServiceProvider({ ydoc, syncService, fileId, i
 		'ws://localhost:1234',
 		'file:' + fileId,
 		ydoc,
-		{ WebSocketPolyfill },
+		{ WebSocketPolyfill, disableBc: true },
```
juliushaertl, line 45: Nice, maybe something to still keep in debug mode?

max-nextcloud: I wonder how we could best trigger this from the dev environment.

max-nextcloud:
I could not reproduce the cypress failure locally... set a crazy long wait time :tm: for the failed cypress statement to see if it's a matter of speed.

src/services/SyncServiceProvider.js
mejo-: At another place we use `window._oc_debug` for this: https://github.com/nextcloud/text/blob/519637de1ccd75f9620060ff25b8f8f758da8a61/src/editor.js#L98

mejo-: I pushed a commit to set `disableBc` according to `window?._oc_debug`.

max-nextcloud:
Ahh... now the `SyncServiceProvider.spec` is failing because the browser broadcast kicks in.

src/services/SyncServiceProvider.js
mejo-: Interesting, this broke our `recovers from a dropped message` Cypress test.

juliushaertl:
One minor unrelated UI bug, the notice about connection issues seems to no longer be sticky, but unrelated to this change:

juliushaertl:
One minor unrelated UI bug, the notice about connection issues seems to no longer be sticky, but unrelated to this change:

Screenshot 2023-07-11 at 16 48 05

mejo-:
> One minor unrelated UI bug, the notice about connection issues seems to no longer be sticky, but unrelated to this change:

Interesting, good catch. Let's open a new issue to track it: https://github.com/nextcloud/text/issues/4483

mejo-:
/backport 6454257361d0ef158c04dab981fe949452449646,c012aac34c0a47f9b3fae67878a5244a9ac12638,2a666d1ea5958abb26f99cc39cd9372f5c802bc5,6a4489e8cce0d73bcee115fe472738cef180270a,155bbb5fa5783ef08e3ac8a6d3106c236d5c606e,0329f46f61420d5d4fbcf1c7da02d3cb71712161,58fc762c4da53ec6275513c447275b835371b67b,fa5d293a2df74b4a6189c3c4cd9f3bed41d47981,d196ae0d1fd16eb2a5ba2c08e400c92e4c7adb0f,f0396923a4ed15ec104d7e9533ac3a053cea725b,236b07ef71353083d2bafaac86bbe81973d3116e to stable27

mejo-:
Manual backports:

- stable27: https://github.com/nextcloud/text/pull/4486
- stable26: https://github.com/nextcloud/text/pull/4487

PR #4503, opened 2023-07-12
fix(sync): only save on close if changes were made

max-nextcloud:
### 📝 Summary

When opening the same file over and over again in the page-links test in collectives the file turns empty after some time.
This is due to the file getting saved before it's even been loaded. So it will be saved empty.

This fix is more broad then that. It will ensure files are only saved during close if any undoable changes have been made to them - the same logic we also apply on autosave.

* Resolves: #818 
  (This is more a sideeffect of this and #4286). If there are no changes close is fast now. If there are changes we don't wait for a sync loop anymore (since a while) and due to #4286 it will still save.

### 🏁 Checklist

- Tests -> this was discovered due to cypress tests failing in collectives.


mejo- (approved):
Very nice catch :fishing_pole_and_fish:

I'd say we should backport this to stable27 and stable26, no?

src/components/Editor.vue
```diff
@@ -326,13 +326,17 @@ export default {
...
+	async beforeDestroy() {
 		if (!this.richWorkspace) {
 			window.removeEventListener('beforeprint', this.preparePrinting)
 			window.removeEventListener('afterprint', this.preparePrinting)
 		}
 		unsubscribe('text:image-node:add', this.onAddImageNode)
 		unsubscribe('text:image-node:delete', this.onDeleteImageNode)
+		if (this.dirty) {
+			const timeout = new Promise((resolve) => setTimeout(resolve, 2000))
+			await Promise.any([timeout, this.$syncService.save()])
```
juliushaertl, line 338: Maybe I'm missing something but wouldn't this be called after the text session has been closed already?

max-nextcloud: No. The text session is closed afterwards where we destroy all the providers - one of which is the `SyncServiceProvider` which will close the session on destruction.
Only if the save takes longer than the timeout the provider will be destroyed never the less.

juliushaertl: Right, makes sense. I just got confused by the close/_close methods of the sync service and how they would be triggered as I assumed that the viewer close calls them, but that doesn't seem to be the case. 👍

juliushaertl:
Failure is unrelated

juliushaertl (approved):

juliushaertl:
/backport 0fe2f30696fdd09754d3ac5f2774d4ada9591647 to stable27

juliushaertl:
/backport 0fe2f30696fdd09754d3ac5f2774d4ada9591647 to stable26

PR #4504, opened 2023-07-12
Fix: clear pending autosync on close

### LINKED ISSUE #4080: Stop autosave when session is closed
**Describe the bug**
I noticed some 403 responses to `sync` requests during a conflict e2e test. Looks like they are triggered by calls to the `autosave` function in the `SyncService` even after the session was closed.
This probably happens because `autosave` gets triggered by an `onChange` handler that persists even after closing the session.

**To Reproduce**
1. Run the cypress tests for conflict handling - in particular noticed it during the 'keep the server version' one.
2. Open the console while the test runs and keep the window open after it finished
3. Watch the network requests and see error some seconds after the test ended.

**Expected behavior**
No 403 responses, no dangling handlers from an old session.

**Screenshots**
Will add them later.

max-nextcloud:
### 📝 Summary

This would normally happen in the save call before closing.
However if that call fails
(for example due to a conflict)
we still clear the pending autosync calls to prevent 403s.
    
Fixes #4080.

### 🏁 Checklist

- This was discovered due to 403 responses during test runs. However we do not have a test explicitly covering this. I think that's fine.


juliushaertl (approved):

juliushaertl:
Let's backport for 27.1

mejo-:
/backport 14d29b1bd33a73a8e9309310b7b3550b2e44f7d4 to stable27

PR #4507, opened 2023-07-12
tests(cypress): Refactor reconnect test to be more reliable and add second test for actual reconnect

juliushaertl:
This should make our top failure cypress test on main more reliable and also extends it with another case:

- The existing test only tests if an existing session is picket up after a short interrupt of sync/push requests
- The new test actually waits until the y.js websocket connection gets closed and checks if a full reconnect is working

### DIFF
--- a/cypress/e2e/sync.spec.js
+++ b/cypress/e2e/sync.spec.js
@@ -64,23 +64,22 @@ describe('Sync', () => {
 			.should('include', 'saves the doc state')
 	})
 
-	it('recovers from a lost connection', () => {
-		let count = 0
-		cy.intercept({ method: 'PUT', url: '**/apps/text/session/create' })
-			.as('createSession')
-		cy.intercept({ method: 'POST', url: '**/apps/text/session/*' }, (req) => {
-			if (count < 4) {
+	it('recovers from a short lost connection', () => {
+		let reconnect = false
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Awesome, thanks for tackling this :sparkles:

cypress/e2e/sync.spec.js
```diff
@@ -64,23 +64,23 @@ describe('Sync', () => {
...
-			.as('createSession')
-		cy.intercept({ method: 'POST', url: '**/apps/text/session/*' }, (req) => {
-			if (count < 4) {
+	it('recovers from a short lost connection', () => {
+		let reconnect = false
+		cy.intercept('**/apps/text/session/*', (req) => {
+			if (reconnect) {
+				req.continue()
+				req.alias = 'alive'
+				reconnect = true
```
max-nextcloud, line 73: This seems somewhat redundant as reconnect is already checked to be truethy in line 70

Or am I missing something?

max-nextcloud (approved):
Yay! Thanks for getting to the bottom of this! :diving_mask: 
Also great idea to test the close / create loop.

I have just one small remark.

max-nextcloud:
Maybe also backport to stable27...
grafik

Issue #4398, opened 2023-06-27
can not using tab in code block


happyxhw:
**Describe the bug**
Can not using tab in code block

**To Reproduce**
create a code block in .md file, then tab to indent

**Expected behavior**
can using tab key in this text app

**Screenshots**
![image](https://github.com/nextcloud/text/assets/44490504/0571e463-8eff-43be-b5fc-44fa1244bb9d)

happyxhw:
this is a feature or bug ?

happyxhw:
read the code, I find:

https://github.com/ueberdosis/tiptap/issues/1153

![image](https://github.com/nextcloud/text/assets/44490504/711b6b49-eefc-48e9-ba35-a4a4e5e0eec3)

PR #4520, opened 2023-07-13
fix: tab key in code block

luka-nextcloud:
### 📝 Summary

* Resolves: #4398

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/components/Editor.vue
```diff
@@ -717,6 +717,13 @@ export default {
 				return
 			}
 
+			if (event.key === 'Tab' && !event.shiftKey && this.$editor.isActive('codeBlock')) {
```
juliushaertl, line 720: Wouldn't this be something we could add to the CodeBlock node directly as a keyboard handler instead of here globally?

luka-nextcloud: @juliushaertl I have tried that way but it didn't work since tab key event was stoped by focus-trap js.

mejo-:
@juliushaertl had the idea to add an `aria-label` attribute to code blocks to make it visible to screen readers that you're inside a code block.

Given that it's very unlikely to end up in a code block by tabbing through the app and that you still can move out of the code block via arrow keys, please disregard my a11y doubts :relaxed:

juliushaertl:
@Pytal @susnux Since you work more on accessibility topics, do you have any hint how how this could be addressed best? 

TL;DR: Tab key in a code block should indent but then the tab key can no longer be used to move to the next focus element, arrow keys would work to move out.

Pytal:
As WCAG states "... if it requires more than unmodified arrow or tab keys or other standard exit methods, the user is advised of the method for moving focus away." (https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html) this technically meets requirements

But it would be ideal to add an `aria-describedby` on the code block pointing it to a description informing the user that the arrow keys can be used to move focus out of the code block, any additional recommendations @michaelnissenbaum?

max-nextcloud (approved):
Code changes look good to me. Did not test it though.
Maybe accessibility hint could be moved to a separate issue.

mejo- (approved):

mejo-:
/backport f1ed024d077111b3e6f99b5d2f00038d5e4701bd to stable27

mejo-:
/backport f1ed024d077111b3e6f99b5d2f00038d5e4701bd to stable26

PR #4540, opened 2023-07-17
fix: Remember cursor position when autofocus on load

### LINKED ISSUE #4140: When opening a file cursor always jumps to end of content
**Describe the bug**

In comparison to earlier versions of the text app the cursor now always jumps to the end of a file when the file is opened (either when opening the file from the file browser or following a shared link to a file. We get feedback from users that this behaviour is confusing, especially with longer files like meeting agendas or meeting notes where important content is at the beginning of a file. Users need to scroll back to the top and sometimes miss important information. Also, shared links with fragment identifiers to jump to headlines in a file currently do not seem to work.

**To Reproduce**

1. Open a text file in the current version of the app.

**Expected behavior**

When opening a file, the cursor should be placed at the start of the content and not the end of the file. Also, shared links with fragment identifiers should open a file at the position of the fragment identifier in the file.

**Client details:**
 - OS: Windows, MAC, Linux
 - Browser: Tested with current Firefox, Chrome and Safari
 - Version: 26
 - Device: Desktop

**Text app version:**

3.7.2

**Operating system:**

Debian

**Web server:**

Apache

**Database:**

MySQL

**PHP version:**

8.1

**Nextcloud version:** 

26.0.1

juliushaertl:
### 📝 Summary

* Resolves: #4140 

From being personally annoyed by it, this is a naive attempt to resolve #4140 to focus on the last cursor position. The position is remembered in the session storage per file and will be restored if present, otherwise we focus on the start (instead of the end).

Note that this may lead to cursor offsets if the document was changed in the meantime but so far I couldn't think of a proper way to get the position state relative to a node within the y.js doc and this should already improve the situation quite a bit. Even if the position is not exactly matching it seems more sane to focus close enough rather than to the end.

### 🏁 Checklist


mejo- (approved):
Very nice, works really well. Will have to adjust https://github.com/nextcloud/collectives/blob/main/src/components/Page/TextEditor.vue#L150 afterwards to make this work in Collectives.

Probably something we should backport to stable27 and stable26 because it fixes an annoying usability bug, right?

mejo-:
https://github.com/nextcloud/collectives/pull/795 is the corresponding PR for Collectives.

juliushaertl:
> Probably something we should backport to stable27 and stable26 because it fixes an annoying usability bug, right?

Yes, would definitely do that.

mejo-:
> The position is remembered in the session storage per file and will be restored if present, otherwise we focus on the start (instead of the end).

After thinking more about it, I wonder whether it would be better to store this in `localStorage` instead of `sessionStorage`. `sessionStorage` seems a bit short-living to me (as someone who usually opens documents in new tabs and doesn't stick to one tab). The main problem I see with using `localStorage` is that it will get bloated with `fileIds` over time and I cannot think of a sane garbage collector mechanism. Maybe good to go with `sessionStorage` for now and see whether we want to change/improve it later on.

juliushaertl:
Yes, the cleanup was the main reason why I picked session storage as it seemed more short living. I'd rather get this in now and then iteratively improve.

mejo-:
/backport cd552ef7de5396112e1cf3f3fc5d740bc3ecb0ae to stable27

mejo-:
/backport cd552ef7de5396112e1cf3f3fc5d740bc3ecb0ae to stable26

juliushaertl:
Think I merged too fast, https://cloud.cypress.io/projects/hx9gqy/runs/11150/test-results/0a248360-1a30-4eae-89ac-64c7a3a593d4 seems related, I'll look into that tomorrow.

Issue #3825, opened 2023-02-24
Prevent duplicate loading of yjs when workspace and viewer open text


max-nextcloud:
**Describe the bug**
When opening both the Readme.md in the workspace and a file in the viewer the following error shows up in the console:
> Yjs was already imported. This breaks constructor checks and will lead to issues! - https://github.com/yjs/yjs/issues/438

**To Reproduce**
1.Create a folder description.
2. Create a test file
3. Open console
4. See error message there

**Expected behavior**
We should only import Yjs once and avoid that error message.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/97337118/221192346-8aa26aad-b4d6-4e3d-90c4-f40d90253241.png)

luka-nextcloud:
This is a reproducible bug.

max-nextcloud:
I wonder if it is due to the dynamic loading of Editor.vue. Both the rich workspace and the viewer component load it dynamically with
Editor: () => import(/* webpackChunkName: "editor" */'./../components/Editor.vue'),
We will want to keep this i think because it reduces the size of the view component and rich workspace artifacts and they can share the editor.
But if normal imports would fix the duplicate loading it might be an idea to use normal imports for yjs and then inject the yjs doc ($ydoc) into the Editor either via props or with provide / inject.

PR #4542, opened 2023-07-17
Bugfix/duplicated load of yjs

luka-nextcloud:
### 📝 Summary

* Resolves: #3825

Reset value of `window['__ $YJS$ __']` before Editor component is mounted to avoid console error `Yjs was already imported.`

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/components/ViewerComponent.vue
```diff
@@ -68,6 +70,21 @@ export default {
...
+		return {
+			enabled: false,
+		}
+	},
+	mounted() {
+		if (!window.richWorkspaceEnabled) {
+			this.enabled = true
+		} else {
+			setTimeout(() => {
+				window['__ $YJS$ __'] = false
```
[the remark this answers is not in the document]
juliushaertl: What has been described in https://github.com/yjs/y-websocket/issues/72#issuecomment-1082003115 might be a feasible workaround for now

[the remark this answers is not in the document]
juliushaertl: Ok but there doesn't seem to be an import/require of the utils.js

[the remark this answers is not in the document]
juliushaertl: I think this is due to the fact that we use multiple entrypoints (viewer and rich workspace), so webpack keeps them separate, so mabye we need a singleton approach for the dynamic import of Editor.vue

src/components/ViewerComponent.vue
```diff
@@ -32,11 +32,12 @@
 
 <script>
 import { getSharingToken } from '../helpers/token.js'
+import getEditorInstance from './Editor.singleton.js'
 
 export default {
 	name: 'ViewerComponent',
 	components: {
-		Editor: () => import(/* webpackChunkName: "editor" */'./Editor.vue'),
+		Editor: async () => { return await getEditorInstance() },
```
juliushaertl, line 40: I think this could be simplified:

```suggestion
		Editor: getEditorInstance,
```

src/components/Editor.singleton.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Get instance of Editor component
```
juliushaertl, line 2: Might be nice to have a comment why we need this for later reference with a link to the issue or y.js code

src/components/Editor.singleton.js
```diff
@@ -0,0 +1,17 @@
...
+ * @return {Promise<*>}
+ */
+export default async function getEditorInstance() {
+	if (!window.EditorInstance) {
+		if (window.EditorInstanceImporting) {
+			await new Promise((resolve) => setTimeout(resolve, 500))
```
juliushaertl, line 8: Feels a bit risky to wait for a timeout here, with the low value it could fail on slow network loading the editor in the other entrypoint, higher would delay loading of the editor. Maybe we can check more frequently in an inverval instead?

juliushaertl:
Small comments, but the general approach looks good :) Haven't tested yet though

max-nextcloud:
@luka-nextcloud I just had a brief look at this - you marked three conversations with julius as resolved but only one of them seems to be addressed (the timeout). I don't understand why you resolved the other ones. When you resolve a conversation without changes to the code - could you add a brief comment explaining why? That would make it easier to understand what is going on. 

Thanks! :heart:

luka-nextcloud:
@max-nextcloud  All of the comments are resolved now.

src/components/Editor.singleton.js
```diff
@@ -0,0 +1,25 @@
...
+					clearInterval(intervalId)
+				}, 200)
+			})
+		} else {
+			window.EditorInstanceImporting = true
+		}
+		const Editor = await import(/* webpackChunkName: "editor" */'./Editor.vue')
+		window.EditorInstance = Editor.default
+	}
+	return window.EditorInstance
```
juliushaertl, line 24: In other app/libraries we use `_` to start the globals that should not be accessed directly. Maybe we can follow this pattern here and use those:
- `window._nc_text_editor_instance`
- `window._nc_text_editor_importing`

juliushaertl (approved):

Taikaschi:
I still have this error in my browser console after upgrade to 27.1.0 final.

max-nextcloud:
> I still have this error in my browser console after upgrade to 27.1.0 final.

As far as i can tell this has not been backported to 27. I'll trigger a backport - not sure if we want to backport but we can discuss that in the backport PR.

Issue #4269, opened 2023-06-08
Use skeleton screen instead of loading spinner 


nimishavijay:
**Is your feature request related to a problem? Please describe.**

For elements where a block of text appears, generally a skeleton screen is used to indicate loading state. Examples are Talk and Mail. It is also used in Notion as well. 

![image](https://github.com/nextcloud/text/assets/52440189/8de656f2-6134-432d-8eb6-dd2eaf4ece7a)

cc @nextcloud/designers

PR #4545, opened 2023-07-18
feat: add component SkeletonLoading

luka-nextcloud:
### 📝 Summary

* Resolves: #4269

#### 🖼️ Screenshots

🏚️ Before 
image
image

 🏡 After
image
image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,181 @@
...
+
+<style lang="scss" scoped>
+$clickable-area: 44px;
+$margin: 8px;
+$messages-list-max-width: 670px;
+
+.placeholder-main {
+	max-width: $messages-list-max-width;
+	position: relative;
+	margin-bottom: auto;
```
juliushaertl, line 80: I noticed there is some flickering during loading when the editor div appears the skeleton is hidden behind it. Adding a `z-index: 1;` here seems to fix this.

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,181 @@
...
+
+	&-text {
+		margin: 0 auto;
```
juliushaertl, line 83: Maybe we can give it a top margin to account for the menu bar being loaded. A margin of 50px seems to be sufficient to make sure that the first line of the placeholder matches the first line of the text.

juliushaertl:
I've added a quick screen recording for easier review by the designers with my proposed changes. @marcoambrosini could you have a look?

marcoambrosini (approved):
Looks much nicer than the spinner :) 
the only thing I would add is a small border radius to the rectangles, maybe 4px?

juliushaertl:
> the only thing I would add is a small border radius to the rectangles, maybe 4px?

I'd keep this separate, if this shall be done then it would be good to structure this so that every app using skeleton loaders adjusts for the same target nextcloud version :)

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,181 @@
...
+			colorPlaceholderLight,
+		}
+	},
+
+	computed: {
+		placeholderData() {
+			const data = []
+			for (let i = 0; i < this.count; i++) {
+				// generate random widths
+				data.push('width: ' + (Math.floor(Math.random() * 40) + 50) + '%')
```
mejo-, line 57: @luka-nextcloud as you might have seen, I pushed two follow-up commits to your PR in Collectives to fix a few minor issues. Maybe it would be good to use the same code here and in Collectives?

See https://github.com/nextcloud/collectives/commit/af719fcef0a62d9a4721b6b88d06889878261fb1

mejo- (changes requested):

marcoambrosini (approved):

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,187 @@
...
+}
+
+.text-placeholder {
+	width: min($messages-list-max-width, 100vw);
+	height: 6em;
+	margin: $margin auto;
+	padding: 4px 8px 0 14px;
+	display: block;
+
+	&-icon {
```
mejo-, line 125: `-icon` can go, it's not used.

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,187 @@
...
+	&-line-one,
+	&-line-two,
+	&-line-three,
+	&-line-four {
+		width: 670px;
+		height: 1em;
+	}
+
+	&-line-one {
+		y: 5px;
```
mejo-, line 142: All the `y` values with `px` can go, they're overwritten by those with `em` anyway. I also missed this in Collectives :wink:

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,187 @@
...
+				<defs>
+					<linearGradient :id="'placeholder-gradient' + suffix">
+						<stop offset="0%" :stop-color="(gradientIndex === 0) ? colorPlaceholderLight : colorPlaceholderDark" />
+						<stop offset="100%" :stop-color="(gradientIndex === 0) ? colorPlaceholderDark : colorPlaceholderLight" />
+					</linearGradient>
+				</defs>
+			</svg>
+
+			<ul :key="'list' + suffix" :class="'placeholder-list placeholder-list' + suffix">
+				<li v-for="(width, index) in placeholderData" :key="'placeholder' + suffix + index">
```
mejo-, line 15: Given that `width` from `placeholderData` is not really used anymore anyway, it should be enough to use `count` for this loop and remove `placeholderData` altogether, no?

mejo- (changes requested):

mejo- (changes requested):
@luka-nextcloud sorry for yet another round of feedback: both in `src/components/Editor/Wrapper.vue` and in `src/views/RichWorkspace.vue` there's leftovers in the CSS that refer to `.icon-loading` which is not used anymore. The ones in `src/components/Editor/Wrapper.vue` probably can be removed, the ones in `src/view/RichWorkspace.vue` look like they need to be updated to be active when `this.loading` is set, no?

mejo- (approved):

mejo-:
/backport b982efa45d5c76bccdb458290556848f06f392fb to stable27

PR #4339, opened 2023-06-21
fix: Delete inactive sessions in one query

juliushaertl:
Fixes #4218 

Delete inactive sessions without steps in one query as the oracle limit is only about 1000 expressions. Using one query also has the benefit to avoid deadlock scenarios as the query would lock the rows directly.

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [x] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [x] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required

PR #4549, opened 2023-07-19
fix: Use simple query and chunk in PHP to delete inactive sessions

juliushaertl:
reverts #4339 and #4539 except for the added tests and moves to plain PHP chunking to avoid issues with MySQL 8

Easiest to review in split mode https://github.com/nextcloud/text/pull/4549/files?diff=split

### 🏁 Checklist

### DIFF
--- a/lib/Db/SessionMapper.php
+++ b/lib/Db/SessionMapper.php
@@ -100,30 +100,36 @@ class SessionMapper extends QBMapper {
 	}
 
 	public function deleteInactiveWithoutSteps(?int $documentId = null): int {
-		$selectSubQuery = $this->db->getQueryBuilder();
-		$selectSubQuery->select('s.id')
+		$lastContact = time() - SessionService::SESSION_VALID_TIME;
+
+		$inactiveSessionBuilder = $this->db->getQueryBuilder();
+		$inactiveSessionBuilder->select('s.id')
 			->from('text_sessions', 's')
-			->leftJoin('s', 'text_steps', 'st', $selectSubQuery->expr()->eq('st.session_id', 's.id'))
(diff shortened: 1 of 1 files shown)


juliushaertl:
Cherry-picked to https://github.com/nextcloud/text/pull/4548 to validate the tests passing also on CI for MySQL 8

nickvergessen (approved):

lib/Db/SessionMapper.php
```diff
@@ -100,30 +100,37 @@ public function findAllInactive() {
...
+		$deleteBuilder = $this->db->getQueryBuilder();
+		$deleteBuilder->delete($this->getTableName())
+			->where($deleteBuilder->expr()->in('id', $deleteBuilder->createParameter('ids'), IQueryBuilder::PARAM_INT_ARRAY));
+
+		$deletedCount = 0;
+		foreach ($chunks as $ids) {
+			$deleteBuilder->setParameter('ids', $ids, IQueryBuilder::PARAM_INT_ARRAY);
+
+			$deletedCount += $deleteBuilder->executeStatement();
+			$result->closeCursor();
```
mejo-, line 131: Shouldn't this be `$deleteBuilder->closeCursor()` and rather be moved after the `foreach` loop?

mejo- (changes requested):

lib/Db/SessionMapper.php
nickvergessen: result is already closed in new line 118.
executeStatement does not return a Result object, so this can just go:
```suggestion
```

juliushaertl: Right, good catch. Pushed.

mejo- (approved):

juliushaertl:
Failure known and unreated (fixed with #4550)

PR #4550, opened 2023-07-19
ci: Fix cypress test failing due to new autofocus cursor position

juliushaertl:
Fix cypress failure after #4540


mejo- (approved):
Very nice :tada:

PR #4584, opened 2023-07-24
Stricter typing to match psalm level 2

juliushaertl:
Did some work towards higher psalm levels during todays travel, let's see what CI thinks about that

- [x] Check why psalm locally complains about stuff from vendor/nextcloud/ocp which should only be additional files 
  - No idea why this happens but added to the baseline for now as it is not text related

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -69,7 +69,7 @@ class ApiService {
 		$this->l10n = $l10n;
 	}
 
-	public function create($fileId = null, $filePath = null, ?string $token = null, $guestName = null): DataResponse {
+	public function create(?int $fileId = null, ?string $filePath = null, ?string $token = null, ?string $guestName = null): DataResponse {
 		try {
 			if ($token) {
 				$file = $this->documentService->getFileByShareToken($token, $this->request->getParam('filePath'));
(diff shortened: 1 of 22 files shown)


lib/Service/EncodingService.php
```diff
@@ -77,6 +77,7 @@ private function detectUtfBom(string $string): ?string {
 	 */
 	private function getEncodings(): array {
 		$mbOrder = mb_detect_order() ?: [];
```
mejo-, line 79: Seems like this line can go, no?

lib/Service/ApiService.php
```diff
@@ -127,6 +124,8 @@ public function create($fileId = null, $filePath = null, ?string $token = null,
 			return new DataResponse('Failed to create the document session', 500);
 		}
 
+		/** @var Document $document */
+
```
mejo-, line 128: Nitpicking, but this empty line is superfluous, no?

mejo- (approved):
Thanks a lot for looking into this :bee:

Apart from two small comments changes all look sensible to me :blush:

lib/Service/ApiService.php
juliushaertl: This is not the var of the left line but an indicator for psalm not detecting that document is never undefined in this case (which seems like a false positive).

lib/Service/EncodingService.php
juliushaertl: The return value is still needed, unfortunately the return value of mb_detect_order is strange as it is documented to return a bool|array while when calling it without an argument it will always return an array. I've updated the relevant line to require the duplicate variable assignment

Issue #1805, opened 2021-08-02
Using notify_push for syncing


Lejo1:
**Is your feature request related to a problem? Please describe.**
Currently the client frequently has to check for changes on the server creating high load also for idle files.

**Describe the solution you'd like**
The [notify_push](https://github.com/nextcloud/notify_push) app can solve this problem by pushing the changes to the client directly via a websocket. Therefore the latency will be highly reduced making the app even better for collaboration.

**Additional context**
The notify_push app is already used for file updates which are often even faster than the syncing of the text app.

juliushaertl:
Yes, that would definitely be a nice integration. @icewind1991 I guess it would make sense to have a dedicated message type implemented on the notify_push side that the text app can use to push document changes to any active editing session for a document through the websocket.

icewind1991:
This should already be possible without any changes to notify_push by using `custom` events

```php
$queue = \OC::$server->get(OCA\NotifyPush\IQueue::class);
$queue->push('notify_custom', [
	'user' => "uid",
	'message' => "my_message_type",
        'body' => ["foo" => "bar"], // optional
]);
```

and

```js
import { listen } from '@nextcloud/notify_push'

listen('my_message_type', (message_type, body) => {

})
```

max-nextcloud:
Planning to tackle this as part of the y.js based collaboration rework in #2669.
There's little use in looking into it before hand as a lot of the collaboration API will change. So marking the issue as blocked by the y.js rework.

PR #4585, opened 2023-07-26
feat: Use notify push for sync messages during editing

juliushaertl:

### 📝 Summary

* Resolves: #1805 

This is a first PoC to make use of notify_push for syncing steps back to other editors instead of regular polling. We still poll every 30 seconds as a fallback and to ensure that the autosave requests are still triggered until https://github.com/nextcloud/text/pull/4424 is there (even with that some fallback sync might still be worth).

Can be manually enabled for now with

```
occ config:app:set text notify_push --type boolean --value=1
```

### 🚧 TODO

- [x] More manual testing in different scenarios
  - [x] Check what happens with the base doc changing externally (did we use the sync to detect the conflict?)

### 🏁 Checklist

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -35,6 +35,7 @@ use Psr\Log\LoggerInterface;
 class ApiService {
 	public function __construct(
 		private IRequest $request,
+		private ConfigService $configService,
 		private SessionService $sessionService,
 		private DocumentService $documentService,
 		private EncodingService $encodingService,
@@ -193,7 +194,7 @@ class ApiService {
 	}
 
 	private function addToPushQueue(Document $document, array $steps): void {
-		if ($this->queue === null) {
+		if ($this->queue === null || !$this->configService->isNotifyPushSyncEnabled()) {
 			return;
 		}
 
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -12,6 +12,7 @@ namespace OCA\Text\Service;
 use Exception;
 use InvalidArgumentException;
 use OCA\Files_Sharing\SharedStorage;
+use OCA\NotifyPush\Queue\IQueue;
 use OCA\Text\AppInfo\Application;
 use OCA\Text\Db\Document;
 use OCA\Text\Db\Session;
@@ -32,7 +33,6 @@ use OCP\Share\IShare;
 use Psr\Log\LoggerInterface;
 
 class ApiService {
-
 	public function __construct(
 		private IRequest $request,
 		private SessionService $sessionService,
@@ -41,6 +41,7 @@ class ApiService {
 		private LoggerInterface $logger,
 		private IL10N $l10n,
 		private ?string $userId,
+		private ?IQueue $queue,
 	) {
 	}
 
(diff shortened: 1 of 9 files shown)


lib/Service/ApiService.php
```diff
@@ -210,6 +215,27 @@ public function push(Session $session, Document $document, $version, $steps, $aw
 		return new DataResponse($result);
 	}
 
+	private function addToPushQueue(Session $session, Document $document, int $version): void {
+		try {
+			$queue = Server::get(IQueue::class);
+			$syncResponse = $this->sync($session, $document, $version);
```
max-nextcloud, line 221: I would expect to just add the steps that just arrived. Say we start at version 0 and two users push their steps without syncing in between:
* user1 pushes `step1`
* user2 pushes `step2`

Now the first push will add `[step1]` and the second push will add `[step1, step2]` to the queue because user2 still has version 0.

Or in other words.. the content of the queue gets send to all users and what is added to it should not depend on the version of the user adding it.

max-nextcloud:
Thanks a lot for looking into this. I'm really curious to play with it. :heart_eyes: 

> We still poll every 30 seconds as a fallback and to ensure that the autosave requests are still triggered until https://github.com/nextcloud/text/pull/4424 is there (even with that some fallback sync might still be worth).

Autosave is independent from the polling backend these days. It's triggered by 
https://github.com/nextcloud/text/blob/14d29b1bd33a73a8e9309310b7b3550b2e44f7d4/src/services/SyncService.js#L91

I agree that having a fallback for a start might still be worthwhile.

lib/Service/ApiService.php
max-nextcloud: y-websocket has a timer that will send Sync messages of type 1 periodically:
https://github.com/yjs/y-websocket/blob/35be3431b8db14f07dfaacab5dfa1a4cdec13fac/src/y-websocket.js#L300-L311

We could enable that. But if the frequency is too high it will create quite a bit of load as all clients would query all other clients. If the frequency is too low it will take too long to resync to be helpful.

My favorite approach right now would be for the saving client to also save a Sync message type 1 alongside the current state. Then every client can get that as a baseline and apply the saved state and reply to the message type 1 with a message type 2 that will contain its local diff.

juliushaertl: This should now be resolved by just using the notify_push queue as a fast path to distribute any incoming messages to other and the own clients.

lib/Service/ApiService.php
```diff
@@ -215,6 +207,25 @@ public function push(Session $session, Document $document, int $version, array $
 		return new DataResponse($result);
 	}
 
+	private function addToPushQueue(Document $document, array $steps): void {
+		if ($this->queue === null || !$this->configService->isNotifyPushSyncEnabled()) {
+			return;
+		}
+
+		$sessions = $this->sessionService->getActiveSessions($document->getId());
+		$sessions = array_values(array_filter(array_unique(array_map(fn ($session): ?string => $session['userId'], $sessions))));
```
mejo-, line 216: Any particular reason to overwrite `$sessions` and not use a new variable `$userIds`? The latter would be easier to understand in my opinion.

juliushaertl: No reason, good call 👍

elzody (approved):
I did some more reading on the notify_push app and how it's used, and with my basic understanding of the app and websockets, it looks okay to me. I will approve in case it counts here :)

juliushaertl:
- [ ] Need to do an extra check for public share links (as I haven't yet)

PR #4587, opened 2023-07-26
feat: Expose formatting menu bar actions through slash command

juliushaertl:

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


juliushaertl:
@nextcloud/designers I'd be glad to get some early design feedback on this :)

mejo-:
Really nice, can't wait to use this myself :heart_eyes:

One thing that immediately came to my mind: should we change the order to hide the formatting options a bit? I don't think it's very intuitive to have six headings options listed as the first smartpicker options :grimacing:

juliushaertl:
Yes, maybe we could also sort the usual smart pickers first. Regarding the list of visible entries we could also increase that a bit if screen space allows. @nextcloud/designers What do you think?

src/components/Suggestion/SuggestionListWrapper.vue
```diff
@@ -128,11 +153,26 @@ export default {
 
 	min-width: 200px;
 	max-width: 400px;
+	width: 80vw;
```
luka-nextcloud, line 156: Not sure if we really need it here. It seems to look ok without this line.

marcoambrosini (commented):
Nice!
I think it would be really nice to have the 4-5 most used styles at the top of the menu so that they're immediately visible.

src/components/Suggestion/SuggestionListWrapper.vue
juliushaertl: This makes sure that on mobile devices there is still a bit of space around the picker

juliushaertl:
>  I think it would be really nice to have the 4-5 most used styles at the top of the menu so that they're immediately visible.

Not sure what the most used would be then. I've pushed a change to sort ToDo list, table, h1, h2 first to at least show that there are different options other than headings.

marcoambrosini:
Sorry, I meant most used on a per document base :)
Whichever are the used styles in the document should probably take priority over anything else

juliushaertl:
I'd rather prefer a default choice of order rather than randomly changing ones. Not only for the complexity that it adds but also users would know what to expect from the menu then. I think changing per document could be confusing and also is not very helpful on empty documents or when documents only have a simple structure like headings and paragraphs.

package.json
```diff
@@ -147,7 +147,7 @@
     ],
     "moduleNameMapper": {
       "^@/(.*)$": "<rootDir>/src/$1",
-      "\\.(css)$": "identity-obj-proxy"
+      "^.+\\.(css|less|scss)$": "identity-obj-proxy"
```
max-nextcloud, line 150: This sounds like we would be handing out `less` and `scss` files as they are.

Is that what you intended? I probably just don't understand what's going on here. :)

src/components/Suggestion/LinkPicker/suggestions.js
```diff
@@ -24,9 +24,48 @@ import LinkPickerList from './LinkPickerList.vue'
...
+		].filter(e => e?.label?.toLowerCase?.()?.includes(query.toLowerCase()))
+			.map(e => ({ ...e, suggestGroup: suggestGroupFormat })),
+	)
+}
+
 export default () => createSuggestions({
 	listComponent: LinkPickerList,
 	command: ({ editor, range, props }) => {
+		if (props.action) {
+			const commandChain = editor.chain().deleteRange(range)
```
max-nextcloud, line 64: The `deleteRange` is so this only creates new elements, right? I cannot use it to format the current selection?
(I'll try out - just reviewing the code first)

max-nextcloud (approved):
Code looks good. :sparkles: 
Will still test locally.

max-nextcloud (changes requested):
Please check if the actions are available in the current context.

juliushaertl:
@max-nextcloud Thanks for bringing this up. I pushed a fix to address this. Main change is https://github.com/nextcloud/text/pull/4587/files#diff-461b071d61c872ecc5d0c24f5227cfdbafd62c22347ef001a4d1aab55c25d58eR83-R87

package.json
juliushaertl: Without this jest will complain as we have an import of an scss file in js being now pulled into the scripts loaded in the jest tests. Since we do not care about the styling there, we can just return a proxy instead. 

less was just added for completeness as this is a pattern i found while researching what to do :D

src/components/Suggestion/LinkPicker/suggestions.js
max-nextcloud: The range contains the character(s) used to trigger the suggestions `/`.

max-nextcloud (approved):
looks good!

PR #4614, opened 2023-07-31
feat: Add support for mermaid.js rendering

### LINKED ISSUE #163: LaTeX and Mermaid chart support
Not exactly, but it will get closer to feature parody with the original text editor + markdown editor extension.

Although it should be noted that some of my current markdown documents look broken because they have latex and mermaid charts integrated into them and when viewed in this editor it makes them look bad.

I would like latex support and support for mermaid charts in markdown.  (feature parody with the markdown editor extension).  Would also be really nice if there were a way to view the mark down in source mode and/or interpreted mode, like the markdown editor app does.

Currently, I switched back to the standard text editor and markdown extension, because they support more.  

**Additional context**
LaTeX example:
Screen Shot 2019-07-11 at 2 36 25 PM

Mermaid chart example:
Screen Shot 2019-07-11 at 2 36 41 PM

juliushaertl:
### 📝 Summary

Fixes #163 

Implements mermaid js support for code blocks. By default the preview mode is used, unless the code block is empty or has faulty syntax. Users can then switch between preview/split and code mode. This also provides a way to switch the language of code blocks in the same location.

##### Mermaid code block

Screenshot 2023-08-02 at 13 33 29

Screenshot 2023-08-02 at 13 32 09

##### Regular code bock
Screenshot 2023-08-02 at 13 32 49

### 🚧 TODO

- [x] There seems to be no syntax highlighting for mermaid on highlight.js -> no highlight
- [x] Design: Figure out language selector
- [x] Design: Check mode switcher and what the default should be
- [x] Design: Maybe we shall have some border or boxing around the element
- [x] BUG: Preview should probably be not contenteditable (noticable when selecting all text with Ctrl-A)
- [x] BUG: Make sure to only render the preview in read only mode (https://github.com/nextcloud/collectives/issues/284)
- [x] Make preview the default and call it Diagram
- [x] Use NcActions Edit button -> floating on the top right, move language selector to the menu

#### Follow up tickets

- [ ] Custom mermaid theme
- [ ] Ctrl-A in code block should probably only select the code block
- [x] Clear preview if code block gets emptied
- [ ] Cypress tests

### 🏁 Checklist


juliushaertl:
@nextcloud/designers Do you have any suggestions to make the UI a bit nicer here?

src/components/Editor/MarkdownContentEditor.vue
```diff
@@ -90,7 +90,7 @@ export default {
 
 	created() {
 		this.$editor = this.createEditor()
-		this.$editor.setOptions({ editable: !this.readOnly })
+		this.$editor.setEditable(!this.readOnlyse)
```
mejo-, line 93: This looks like a typo to me :wink:

jancborchardt (changes requested):
Looks nice! Some enhancements for the diagram style:
- [ ] It is purple now, could instead use the primary color: primary-light for bg, primary for the border
- [ ] The text in the boxes doesn’t look vertically aligned but a bit too far up / too much bottom padding
- [ ] Boxes could use border-radius (3px default), might look nicer and Nextcloud-y
- [ ] The Yes/No boxes in the example could use some more padding
- [ ] If it makes sense, the Yes/No text can also be bold? Or is the other stuff more important? (Either way maybe makes sense to bold one of the types)

nimishavijay:
Super cool! :) 🎉 It looks great! I'm wondering if we can just change this to "Insert flowchart" as a separate formatting option and link to the mermaid language documentation for help with the syntax? :) so not have it as a part of the code block option, but a separate option :)

juliushaertl:
I'll see how far we can influence the diagram rendering, but this might be more complex if we need to provide a custom theme: https://mermaid.js.org/config/theming.html

Would you consider this a blocker or just nice to have?

> Super cool! :) 🎉 It looks great! I'm wondering if we can just change this to "Insert flowchart" as a separate formatting option and link to the mermaid language documentation for help with the syntax? :) so not have it as a part of the code block option, but a separate option :)

Mermaid supports all kinds of charts, but we could add a separate option "Insert diagram" with #4587 though I'd keep this as a follow up then. For the syntax documentation that seems like a nice suggestion.

mejo-:
Looks really nice and I'm looking forward to use this myself :blush: 

Regarding the menu icon, I wonder whether it would be better to have a static one to avoid confusion and mark the selected view (Source code, Both or Diagram) as highlighted in the menu.

Personally I find it confusing that clicking on an eye icon allows me to change to source code view and the other way round.

We could use the Pencil icon (that we use for editing buttons) to open the menu, what do you think?

juliushaertl:
The edit button seems more fitting indeed, but would probably not fit to indicate switching from edit to diagram view or in the regular code block to change the language.

Will think a bit about that, further input still welcome of course :)

juliushaertl:
My suggestion would be:
- In diagram view mode -> show the edit icon
- In split mode -> show the split icon 
- In code mode -> show the view icon
- In the regular code block -> show the curly braces icon as it probably fits the language setting best

src/nodes/CodeBlockView.vue
```diff
@@ -0,0 +1,214 @@
...
+		'node.textContent'() {
+			this.renderMermaid()
+		},
+	},
+	beforeMount() {
+		this.isEditable = this.editor.isEditable
+		this.editor.on('update', ({ editor }) => {
+			this.isEditable = editor.isEditable
+		})
+		this.renderMermaidDebounced = debounce(async function() {
```
mejo-, line 130: It should work to define `renderMermaid: debounce(async function() { ... })` in `methods`.

src/nodes/CodeBlockView.vue
```diff
@@ -0,0 +1,214 @@
...
+
+			const textContent = this.node?.textContent || ''
+			if (textContent.trim() === '') {
+				this.viewMode = this.isEditable ? 'side-by-side' : 'code'
+			}
+
+			try {
+				await mermaid.parse(this.node?.textContent || '')
+
+				const { svg } = await mermaid.render(this.targetId, this.node?.textContent || '')
```
mejo-, line 144: Any reason to use `this.node?.textContent` instead of `textContent` (as defined above)?

src/nodes/CodeBlockView.vue
```diff
@@ -0,0 +1,214 @@
...
+				</NcActions>
+			</div>
+		</div>
+		<div :class="{'split-view': showCode && showPreview }">
+			<pre v-show="showCode" class="split-view__code"><NodeViewContent as="code" :contenteditable="editable" /></pre>
+			<div v-show="showPreview"
+				ref="preview"
+				class="split-view__preview"
+				:contenteditable="false" />
+			<div v-show="false" :id="targetId" :contenteditable="false" />
```
mejo-, line 54: I guess the `v-show="false"` is there to initialize the element with `display: none`, right? I wonder whether vue could interfere here later on given that the condition stays `false` all the time :thinking:

mejo- (approved):
Just minor comments, otherwise looks good :blush: 

I wonder whether we should backport the `setEditable()` and `defaultLanguage: 'plaintext'` fixes to stable26 and stable27.

src/nodes/CodeBlockView.vue
juliushaertl: I didn't manage to do that as the this context was always missing then

src/nodes/CodeBlockView.vue
juliushaertl: No, will adjust, thanks :)

src/nodes/CodeBlockView.vue
mejo-: I see, then let's keep it as is for now :relaxed:

src/nodes/CodeBlockView.vue
mejo-: Ah, now I see - it's never shown. Then nevermind.

src/nodes/CodeBlockView.vue
juliushaertl: Done.

mejo-:
/backport b9ea43f5fca6a3a987ac5a9d1263eff70bba9655,62f5d3e846b604a9e7576f5b9eae85bdd82fba1f,6a8b01fa57d9ba253607257262d8c56cd7ee21ec,946ca9bcba925b9bca8809538a069d67c2303d73 to stable27

PR #4627, opened 2023-08-04
feat/text processing

juliushaertl:
- feat: Expose text processing API in text
- feat: Text processing assistant integration

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


jancborchardt (commented):
Looks really great! :) The only thing I am a bit worried about is that even with the icon on the button, "Show assistant results" in the menu then is not super obvious.

- Should we permanently show "Assistant results" as an entry on the top instead?
- We could also say "Assistant results (4 new)" or have a counter bubble there for unviewed results if possible?
- Other ideas?

nimishavijay:
Looks really nice! Super cool! Only one thing: maybe we can add a separator line before "Show assistant results" to indicate a different type of action. Other than that it looks great! :)

juliushaertl:
Fixed:

Screenshot 2023-08-09 at 14 07 04

juliushaertl:
> Looks really great! :) The only thing I am a bit worried about is that even with the icon on the button, "Show assistant results" in the menu then is not super obvious.
> Should we permanently show "Assistant results" as an entry on the top instead?
> We could also say "Assistant results (4 new)" or have a counter bubble there for unviewed results if possible?
> Other ideas?

I think the indicator about the status already gives an idea, considering the timeline of getting this merged, I'd rather get it  in and align during the demo next week if this is something we can polish next week and delay screenshot taking or keep as it is. In any case having clear mockups for any permanent display option would be appreciated then.

max-nextcloud (approved):
Code looks good. Did not try myself.

I think I'd personally compose the `Assistent.vue` out of smaller components for the Floating Menu and the Modal. But that can also happen later. I also did not check how much code reuse there is between the two. If they overlap a lot might not be worth it either.

juliushaertl:
/backport 526302d056c0c1fbf33f8a44fed526f7201889e1,130a4ec5fe5f24e0e27a63e459eeea22472770e5,5ebcd8bd7788afa300b2da098a88067b59eec9dc to stable27

PR #4637, opened 2023-08-05
fix: Use resize observer to calculate menubar icon limit

juliushaertl:

### 📝 Summary

Make the icon limit recalculation more reliable by using a resize observer on the actual element and makes the code much cleaner

`@vueuse/core` is already a dependency through `@nextcloud/vue`

Fixes https://github.com/nextcloud/deck/issues/4970

### 🏁 Checklist


src/components/Menu/MenuBar.vue
```diff
@@ -178,9 +159,7 @@ export default {
 		},
 	},
 	mounted() {
-		window.addEventListener('resize', this.getWindowWidth)
-		subscribe('files:sidebar:opened', this.redrawAfterTransition)
-		subscribe('files:sidebar:closed', this.redrawAfterTransition)
+		this.resize = useResizeObserver(this.$refs.menubar, this.onResize)
```
max-nextcloud, line 162: Is there any reason to keep a handle on the resize observer?

max-nextcloud (approved):
Code changes look good. So much cleaner indeed.
I did not give it a try. Happy to hear it also fixes things in deck.
I have one minor question in the review.
Also... backports?

src/components/Menu/MenuBar.vue
juliushaertl: Right, I missed to clean up the resize observer. Documentation is not very specific if that needs to be done, but reading https://stackoverflow.com/questions/67581868/should-i-call-resizeobserver-unobserve-for-removed-elements it felt more safe to do so.

mejo- (commented):
Thanks for looking into this! We still a listener on `resize` in `getClientWidth()` at `src/store/plugin.js` and in `src/mixins/isMobile.js`.

Also in `src/components/Editor/EditorOutline.vue` we implement our own `$resizeObserver`.

Maybe it would be good to consolidate all of them? :thinking:

juliushaertl:
Sounds like a good idea. For the isMobile mixin we can actually rely on the one from `@nextcloud/vue` as it has a shared state and doesn't add another listener per usage.

I'd do that in a follow up to have the bugfix isolated for backports

juliushaertl:
Need to double check, cypress failure looks like it could be related not showing the menubar on direct editing.

https://cloud.cypress.io/projects/hx9gqy/runs/11522/test-results/9f7c840a-7788-4409-a6d1-cfd42c58408c

juliushaertl:
Pushed a fix to set the ready state in any case as for plain text files we don't have a resize observer as the menubar isn't there.

mejo-:
/backport 180cc393a9260d3b4c650b873ace758f88d5ceed to stable27

mejo-:
/backport 180cc393a9260d3b4c650b873ace758f88d5ceed to stable26

mejo-:
* Manual backport to stable27: https://github.com/nextcloud/text/pull/4649
* Manual backport to stable26: https://github.com/nextcloud/text/pull/4650

PR #4657, opened 2023-08-09
fix: Properly handle viewer loading in collectives

juliushaertl:
- Fix emitting the loaded state in all cases
- Rework check for source based read only view to explicitly limit on versions or files with source only (without a file id)

### 🏁 Checklist


mejo- (approved):
Thanks a lot for tackling this!

PR #4663, opened 2023-08-09
Cypress tests for smart picker, code block, mermaid and version view/compare

juliushaertl:
- tests(cypress): Add smartpicker tests
- tests(cypress): Add code block and mermaid tests
- tests(cypress): Add version view and compare tests

### 🚧 TODO

- Bug found during writing tests
  - [x] https://github.com/nextcloud/server/pull/39786
  - [x] https://github.com/nextcloud/viewer/pull/1851
  - [x] https://github.com/nextcloud/server/pull/39789
  - [ ] https://github.com/nextcloud/server/issues/39791

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/SmartPicker.spec.js
+++ b/cypress/e2e/SmartPicker.spec.js
@@ -1,0 +1,83 @@
+import { initUserAndFiles, randUser } from '../utils/index.js'
+
+const currentUser = randUser()
+
+const fileName = 'empty.md'
+
+describe('Smart picker', () => {
+	before(() => {
+		initUserAndFiles(currentUser, 'empty.md')
+	})
+
+	beforeEach(() => {
+		cy.login(currentUser)
+		cy.visit('/apps/files')
+	})
+
+	it('Type and see the smart picker', () => {
+		cy.isolateTest({
+			sourceFile: fileName,
+		})
+		cy.openFile(fileName, { force: true })
+
+		cy.getContent()
+			.type('/')
+
+		cy.get('.tippy-box .suggestion-list').children().should(($children) => {
+			const entries = $children.find('.suggestion-list__item').map((i, el) => el.innerText).get()
+			expect(entries.length).to.be.greaterThan(0)
+			expect('To-Do list').to.be.oneOf(entries)
+			expect('Table').to.be.oneOf(entries)
+		})
+
+		cy.getContent()
+			.click()
+
+		cy.getContent()
+			.type('Heading{enter}Hello World{enter}')
+
+		cy.getContent()
+			.find('h1 [data-node-view-content]')
+			.should('have.text', 'Hello world')
+	})
+
+	it('Insert a link with the smart picker', () => {
+		cy.isolateTest({
+			sourceFile: fileName,
+		})
+		cy.openFile(fileName, { force: true })
+
+		cy.getContent()
+			.type('/Any')
+
+		cy.get('.tippy-box .suggestion-list').children().should(($children) => {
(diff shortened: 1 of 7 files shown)


cypress/e2e/SmartPicker.spec.js
```diff
@@ -0,0 +1,83 @@
...
+		cy.openFile(fileName, { force: true })
+
+		cy.getContent()
+			.type('/')
+
+		cy.get('.tippy-box .suggestion-list').children().should(($children) => {
+			const entries = $children.find('.suggestion-list__item').map((i, el) => el.innerText).get()
+			expect(entries.length).to.be.greaterThan(0)
+			expect('To-Do list').to.be.oneOf(entries)
+			expect('Table').to.be.oneOf(entries)
```
max-nextcloud, line 30: Is this still expected after the recent reordering that put default actions on top again?

cypress/e2e/nodes/CodeBlock.spec.js
```diff
@@ -0,0 +1,130 @@
...
+		cy.isolateTest({ sourceFile: 'codeblock.md' })
+		cy.openFile('codeblock.md').then(() => {
+			cy.clearContent()
+			cy.getContent().type('{enter}```javascript{enter}')
+			cy.getContent().type('const foo = "bar"{enter}{enter}{enter}')
+			cy.getContent().find('.hljs-keyword').first().contains('const')
+		})
+	})
+
+	it.only('See a mermaid diagram', function() {
```
max-nextcloud, line 75:
```suggestion
	it('See a mermaid diagram', function() {
```

cypress/support/e2e.js
```diff
@@ -6,6 +6,14 @@ import chaiExtension from './chai.js'
...
 before(() => {
 	chai.use(chaiExtension)
+
+	Cypress.on('uncaught:exception', (err) => {
+		if (err.message.includes('ResizeObserver')) {
+		  return false
+		}
+
+		return true
+	})
```
max-nextcloud, line 16: I ran into this one as well when working on 28 cypress tests. Any idea where it's coming from?

max-nextcloud (approved):
Looks good to me. Two `.only` need to be removed still. Approving anyway so you can move on afterwards.

cypress/support/e2e.js
juliushaertl: Could be either text itself for the menubar since #4637 or apps/files/src/components/VirtualList.vue for example, but as far as I read this is safe to just ignore.

juliushaertl: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

cypress/e2e/SmartPicker.spec.js
juliushaertl: Should be fine for the test as there is only very limited default pickers. Will check though.

cypress/e2e/SmartPicker.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+		cy.getContent()
+			.type('Heading{enter}Hello World{enter}')
+
+		cy.getContent()
+			.find('h1 [data-node-view-content]')
+			.should('have.text', 'Hello world')
+	})
+
+	it.only('Insert a link with the smart picker', () => {
```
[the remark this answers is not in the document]
juliushaertl: Great idea 🙈

juliushaertl:
Okay, needs some more fixing I guess

juliushaertl:
Ok, new failure due to the nextcloud-vue update that changed the label position

PR #4667, opened 2023-08-11
ci(cypress): Make sure to build before cypress

juliushaertl:
To have cypress tests running against the actual latest state


max-nextcloud:
ouch... yes - we should really do this. I think we also lost it in 27. I wonder how we can prevent overwriting this in the future again.

juliushaertl:
We should at some point just switch to the shared workflow template, but that needs some work first

luka-nextcloud (approved):

PR #4679, opened 2023-08-14
[stable27] Fix missing text on assistant task list

juliushaertl:
Oversight when backporting as `@nextcloud/vue` on stable27 still uses title instead of name as attributes and slot names


szaimen (approved):

juliushaertl:
Failing tests unrelated (see #4680)

Issue #2451, opened 2022-05-26
Multiple assets loading


vinicius73:
**Describe the bug**

Now is created one bundle for each "viewer" / "page".

- `main.js`
- `files.js`
- `public.js`
- `viewer.js`

Each of these files is bundled with Vue, Vuex, Nextcloud Components, Tiptap, CSS...
Although, the main issue could be the page load, it is causing other issues like "CSS collisions" and extra memory usage.

max-nextcloud:
Isn't there a vendor.js that ships most of the dependencies such as Vue, Vuex, ...?

juliushaertl:
Yep, most parts should be shared from the vendors or editor-*.js files:

```
-rw-r--r--  1 julius  staff    35K May 27 08:19 editor-collab.js
-rw-r--r--  1 julius  staff   6.3K May 27 08:19 editor-guest.js
-rw-r--r--  1 julius  staff    40K May 27 08:21 editor-rich.js
-rw-r--r--  1 julius  staff   201K May 27 08:21 editor.js
-rw-r--r--  1 julius  staff   1.0K May 27 08:21 files-modal.js
-rw-r--r--  1 julius  staff   213K May 27 08:21 text-files.js
-rw-r--r--  1 julius  staff   164K May 27 08:21 text-public.js
-rw-r--r--  1 julius  staff   121K May 27 08:21 text-text.js
-rw-r--r--  1 julius  staff    33K May 27 08:21 text-viewer.js
-rw-r--r--  1 julius  staff   2.5M May 27 08:21 vendors.js
```

But we can of course have a look if splitting like that still makes sense. The main reasoning behind the split and dynamic imports was that with the viewer we can only load the heavy parts if the user is actually opening the text app, e.g. on talk or the dashboard.


The text-public.js entrypoint we may drop once viewer is used on public pages of single shared files: https://github.com/nextcloud/viewer/issues/508

vinicius73:
Yes, there is a vendor.
But for unknown reason not all libs are shared between bundles.

![image](https://user-images.githubusercontent.com/1561347/170744870-95637004-0a78-43e5-8f73-c7ae6257eb33.png)

susnux:
A bit off topic: Maybe [ModuleFederation](https://webpack.js.org/plugins/module-federation-plugin/#specify-package-versions) of vendor packages (`vue`, `@nextcloud/...`) should be considered / discussed on a `nextcloud/core` (server) level to reduce traffic / improve loading times.

[...]

PR #4698, opened 2023-08-18
Lazy load text-viewer and text-files

artonge:
Fixes https://github.com/nextcloud/text/issues/2451
Fixes https://github.com/nextcloud/text/issues/5045

Currently, 5Mb are loaded on the main thread. The idea is to lazy load most of it, and let webpack do a proper chunking.

Also in this PR:
+ Remove vendor bundling
+ Remove chunkIds declaration as it is already the default

Note | Before | After
-- | -- | --
Here, to exacerbate the network gains, it was throttled. The total size is reduced by ~0.15Mb, due to the better chunking. And the DOM loading time is reduced by 30%, because we do not wait for all the scripts before starting the rendering. | Screenshot from 2023-08-18 16-45-30 | Screenshot from 2023-08-18 16-44-23
Without throttling, we can still observe a slight gain due to not executing 5Mb of scripts before rendering. The number is more volatile, but repetitions consistently show lower numbers for the lazy loaded build. | Screenshot from 2023-08-18 17-13-12 | Screenshot from 2023-08-18 17-12-00
More request, but the initial ones are smaller, and the remaining are lazy loaded | Screenshot from 2023-08-18 15-54-30 | Screenshot from 2023-08-18 16-18-26
Just to show the new build layout | Screenshot 2023-08-18 at 16-16-53 @nextcloud_text 18 Aug 2023 at 16 03 | Screenshot 2023-08-18 at 16-17-49 @nextcloud_text 18 Aug 2023 at 15 52

Current webpack config dates from: https://github.com/nextcloud/text/pull/1659

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -20,7 +20,6 @@
  *
  */
 
-import ViewerComponent from './components/ViewerComponent.vue'
 import { logger } from './helpers/logger.js'
 import { openMimetypesMarkdown, openMimetypesPlainText } from './helpers/mime.js'
 
@@ -33,8 +32,7 @@ if (typeof OCA.Viewer === 'undefined') {
 	OCA.Viewer.registerHandler({
 		id: 'text',
 		mimes: [...openMimetypesMarkdown, ...openMimetypesPlainText],
-		// Would be good to be able to lazyload that
-		component: ViewerComponent,
+		component: () => import('./components/ViewerComponent.vue'),
 		group: null,
 		theme: 'default',
 		canCompare: true,
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -33,6 +33,7 @@ if (typeof OCA.Viewer === 'undefined') {
 	OCA.Viewer.registerHandler({
 		id: 'text',
 		mimes: [...openMimetypesMarkdown, ...openMimetypesPlainText],
+		// Would be good to be able to lazyload that
 		component: ViewerComponent,
 		group: null,
 		theme: 'default',
--- a/src/files.js
+++ b/src/files.js
@@ -21,12 +21,9 @@
  */
 import { linkTo } from '@nextcloud/router'
 import { loadState } from '@nextcloud/initial-state'
-import Vue from 'vue'
+import { registerFileListHeaders } from '@nextcloud/files'
 
 import { logger } from './helpers/logger.js'
-import { registerFileActionFallback } from './helpers/files.js'
-import FilesSettings from './views/FilesSettings.vue'
-import store from './store/index.js'
 
 __webpack_nonce__ = window.btoa(OC.requestToken) // eslint-disable-line
 __webpack_public_path__ = linkTo('text', 'js/') // eslint-disable-line
@@ -34,13 +31,18 @@ __webpack_public_path__ = linkTo('text', 'js/') // eslint-disable-line
 const workspaceAvailable = loadState('text', 'workspace_available')
 const workspaceEnabled = loadState('text', 'workspace_enabled')
 
-document.addEventListener('DOMContentLoaded', () => {
+document.addEventListener('DOMContentLoaded', async () => {
 	if (typeof OCA.Viewer === 'undefined') {
+		const { registerFileActionFallback } = await import('./helpers/files.js')
 		logger.error('Viewer app is not installed')
 		registerFileActionFallback()
 	}
 
 	if (workspaceAvailable && OCA && OCA?.Files?.Settings) {
+		const { default: Vue } = await import('vue')
+		const { default: FilesSettings } = await import('./views/FilesSettings.vue')
+		const { default: store } = await import('./store/index.js')
+
 		Vue.prototype.t = window.t
 		Vue.prototype.n = window.n
 		Vue.prototype.OCA = window.OCA
@@ -54,6 +56,10 @@ document.addEventListener('DOMContentLoaded', () => {
 		}))
 	}
 
+	if (workspaceAvailable) {
+		const { FilesWorkspaceHeader } = await import('./helpers/files.js')
+		registerFileListHeaders(FilesWorkspaceHeader)
+	}
 })
 
 OCA.Text = {
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -27,7 +27,6 @@ import { loadState } from '@nextcloud/initial-state'
 
 import { getSharingToken } from './token.js'
 import { openMimetypes } from './mime.js'
-import RichWorkspace from '../views/RichWorkspace.vue'
 import store from '../store/index.js'
 import { getCurrentUser } from '@nextcloud/auth'
 import { showSuccess, showError } from '@nextcloud/dialogs'
@@ -211,7 +210,7 @@ export const FilesWorkspaceHeader = new Header({
 		return view.id === 'files'
 	},
 
-	render(el, folder, view) {
+	async render(el, folder, view) {
 		if (vm) {
 			// Enforce destroying of the old rendering and rerender as the FilesListHeader calls render on every folder change
 			vm.$destroy()
@@ -222,6 +221,8 @@ export const FilesWorkspaceHeader = new Header({
 		const content = newWorkspaceCreated ? '' : folder.attributes['rich-workspace']
 
 		newWorkspaceCreated = false
+
+		const { default: RichWorkspace } = await import('./../views/RichWorkspace.vue')
 
 		import('vue').then((module) => {
 			el.id = 'files-workspace-wrapper'
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -45,11 +45,6 @@ webpackConfig.optimization.splitChunks.cacheGroups = {
 	}
 }
 
-// webpackConfig.resolve.modules = [
-// 	path.resolve(__dirname, 'node_modules'),
-// 	'node_modules'
-// ]
-
 // Fix Buffer issues
 webpackConfig.plugins.push(new webpack.ProvidePlugin({
 	Buffer: ['buffer', 'Buffer'],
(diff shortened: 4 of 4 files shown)


src/viewer.js
```diff
@@ -20,22 +20,24 @@
...
-	OCA.Viewer.registerHandler({
-		id: 'text',
-		mimes: [...openMimetypesMarkdown, ...openMimetypesPlainText],
-		component: ViewerComponent,
-		group: null,
-		theme: 'default',
-		canCompare: true,
-	})
-}
+document.addEventListener('DOMContentLoaded', async () => {
```
juliushaertl, line 29: @skjnldsv I remember this used to be an issue with viewer, but the console log warning in viewer doesn't seem to be there anymore. Can you confirm this is fine to switch?

juliushaertl:
Thanks for looking into this. The change looks good to me, I just left one comment to ask @skjnldsv for confirmation.

src/viewer.js
skjnldsv: Good memory! We used to watch changes in registration with Viewer. But we stopped.
You need to register handlers and Viewer will wait for the DOMContentLoaded to be fired to process them, so you should not wait for it, [Viewer will](https://github.com/nextcloud/viewer/blob/1ef62842037915f5a82e9301c4dd2e4e94a32056/src/views/Viewer.vue#L508-L528).

It was before we started going for the global libraries and window._nc_xxxx variables which made it possible to register at any point in time :)

skjnldsv: If you wanna be safe, you might just add a php `addScript` dependency on Viewer? :thinking: 
But I think we mainly covered this already.

```php
Util::addScript('text', 'text-viewer', 'viewer');
```

skjnldsv (changes requested):
So, the idea is nice, but we're reaching another issue. This is too much split and will caue another issue: too many requests.
The request load overhead does have a huge impact.It's good to be on-demand, but I think we should group the chunks together to load 2-3 big async files instead of the new 50+ files that are in this PR :see_no_evil:

artonge:
> instead of the new 50+ files that are in this PR

Indeed, most of it was due to mermaid. I added a rule to merge those together. For the rest, not sure, it is what the default webpack config does at the moment, if this is not what we want, then we should fix that upstream instead of here, no ?

artonge:
Some are really small, maybe `splitChunk.minSize` could help here.

src/viewer.js
```diff
@@ -33,6 +33,7 @@ if (typeof OCA.Viewer === 'undefined') {
 	OCA.Viewer.registerHandler({
 		id: 'text',
 		mimes: [...openMimetypesMarkdown, ...openMimetypesPlainText],
+		// Would be good to be able to lazyload that
 		component: ViewerComponent,
```
juliushaertl, line 37: We cannot really lazy load a viewer component yet, so it would be nice if one could just pass a callback here for a lazy import

juliushaertl:
Might already work if we allow async components in https://github.com/nextcloud/viewer/blob/master/src/views/Viewer.vue#L814-L815

juliushaertl:
There is probably more that could be done with the file list plugin but this seems like a good first improvement, thanks for the help @pulsejet ❤️

juliushaertl:
Seems good from my perspective, the splitChunks.minSize does not seem to have an effect as there are still quite small chunks:

```
-rw-r--r--  1 julius  staff    33K Nov 28 09:48 js/data_image_svg_xml_base64_PCEtLSBUaGlzIGljb24gaXMgcGFydCBvZiBNYXRlcmlhbCBVSSBJY29ucy4gQ29weXJ-faaaea.js
-rw-r--r--  1 julius  staff   8.1K Nov 28 09:48 js/editor-collab.js
-rw-r--r--  1 julius  staff   5.3K Nov 28 09:48 js/editor-guest.js
-rw-r--r--  1 julius  staff   244K Nov 28 09:48 js/editor.js
-rw-r--r--  1 julius  staff   5.2K Nov 28 09:48 js/files-modal.js
-rw-r--r--  1 julius  staff   763K Nov 28 09:48 js/mermaid.js
-rw-r--r--  1 julius  staff   8.7K Nov 28 09:48 js/node_modules_d3-sankey_src_sankey_js-node_modules_d3-sankey_src_sankeyLinkHorizontal_js.js
-rw-r--r--  1 julius  staff   2.1K Nov 28 09:48 js/node_modules_dagre-d3-es_src_dagre-js_label_add-html-label_js-node_modules_dagre-d3-es_src_gr-77e053.js
-rw-r--r--  1 julius  staff   894B Nov 28 09:48 js/node_modules_dagre-d3-es_src_graphlib_json_js-_c4980.js
-rw-r--r--  1 julius  staff   894B Nov 28 09:48 js/node_modules_dagre-d3-es_src_graphlib_json_js-_c4981.js
-rw-r--r--  1 julius  staff    13K Nov 28 09:48 js/node_modules_dagre-d3-es_src_graphlib_json_js-node_modules_dagre-d3-es_src_index_js-node_modu-e535e5.js
-rw-r--r--  1 julius  staff   5.4K Nov 28 09:48 js/node_modules_dayjs_plugin_advancedFormat_js-node_modules_dayjs_plugin_customParseFormat_js-no-96543d.js
-rw-r--r--  1 julius  staff   690B Nov 28 09:48 js/node_modules_nextcloud_dialogs_dist_chunks_index-27e9ea0a_mjs.js
-rw-r--r--  1 julius  staff   2.9K Nov 28 09:48 js/node_modules_uuid_dist_esm-browser_v5_js.js
-rw-r--r--  1 julius  staff   131K Nov 28 09:48 js/src_extensions_index_js-src_components_Editor_EditorOutline_vue-data_image_svg_xml_base64_PCE-a8d943.js
-rw-r--r--  1 julius  staff    35K Nov 28 09:48 js/src_helpers_files_js-src_components_ViewerComponent_vue.js
-rw-r--r--  1 julius  staff   4.8K Nov 28 09:48 js/src_helpers_files_js.js
-rw-r--r--  1 julius  staff    16K Nov 28 09:48 js/src_store_index_js.js
-rw-r--r--  1 julius  staff   2.2K Nov 28 09:48 js/src_views_FilesSettings_vue.js
-rw-r--r--  1 julius  staff    65K Nov 28 09:48 js/src_views_RichWorkspace_vue-node_modules_nextcloud_vue_dist_Functions_emoji_mjs-node_modules_-0e99a0.js
-rw-r--r--  1 julius  staff   170K Nov 28 09:48 js/text-editors.js
-rw-r--r--  1 julius  staff   261K Nov 28 09:48 js/text-files.js
-rw-r--r--  1 julius  staff   587K Nov 28 09:48 js/text-init.js
-rw-r--r--  1 julius  staff   4.6M Nov 28 09:48 js/text-public.js
-rw-r--r--  1 julius  staff   114K Nov 28 09:48 js/text-text.js
-rw-r--r--  1 julius  staff    75K Nov 28 09:48 js/text-viewer.js
-rw-r--r--  1 julius  staff   912K Nov 28 09:48 js/vendors-node_modules_braintree_sanitize-url_dist_index_js-node_modules_quartzy_markdown-it-me-212f05.js
-rw-r--r--  1 julius  staff   427K Nov 28 09:48 js/vendors-node_modules_cytoscape-cose-bilkent_cytoscape-cose-bilkent_js-node_modules_cytoscape_-d92e99.js
-rw-r--r--  1 julius  staff    51K Nov 28 09:48 js/vendors-node_modules_dagre-d3-es_src_dagre_index_js.js
-rw-r--r--  1 julius  staff   1.3M Nov 28 09:48 js/vendors-node_modules_elkjs_lib_elk_bundled_js.js
-rw-r--r--  1 julius  staff    53K Nov 28 09:48 js/vendors-node_modules_mdast-util-from-markdown_lib_index_js.js
-rw-r--r--  1 julius  staff   141K Nov 28 09:48 js/vendors-node_modules_mdi_svg_svg_folder-move_svg_raw-node_modules_mdi_svg_svg_folder-multiple-c6f33e.js
-rw-r--r--  1 julius  staff    55K Nov 28 09:48 js/vendors-node_modules_nextcloud_axios_dist_index_es_mjs.js
-rw-r--r--  1 julius  staff    52K Nov 28 09:48 js/vendors-node_modules_nextcloud_browser-storage_dist_index_js-node_modules_debounce_index_js-n-906a21.js
-rw-r--r--  1 julius  staff   211K Nov 28 09:48 js/vendors-node_modules_nextcloud_browser-storage_dist_index_js-node_modules_debounce_index_js-n-984d74.js
-rw-r--r--  1 julius  staff   555K Nov 28 09:48 js/vendors-node_modules_nextcloud_moment_dist_index_js-node_modules_moment_locale_af_js-node_mod-a9a4c5.js
-rw-r--r--  1 julius  staff   196K Nov 28 09:48 js/vendors-node_modules_nextcloud_paths_dist_index_js-node_modules_nextcloud_auth_dist_index_es_-a26a98.js
-rw-r--r--  1 julius  staff   2.8M Nov 28 09:48 js/vendors-node_modules_nextcloud_vue_dist_index_mjs.js
-rw-r--r--  1 julius  staff    75K Nov 28 09:48 js/vendors-node_modules_vue_dist_vue_runtime_esm_js.js
```

juliushaertl:
Cypress failures need some work as now it seems our default timeout for the editor is enough to lazy load all required js

pulsejet:
I've no idea how cypress works so can't help with that. But I've a request -- can we backport this to 28? This is absolutely killing the massive loading time improvements we got with 27 => 28

juliushaertl:
Sure, I'll have a look into the cypress stuff and would also backport to 28 still.

- [ ] For now increase cypress timeouts for loading the editor
- [ ] Follow up: check if we can somehow preload the js files (maybe using https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules)

juliushaertl:
Actually timeouts are not the issue but outdated viewer js

https://github.com/nextcloud/viewer/pull/2083
https://github.com/nextcloud/viewer/pull/2084

juliushaertl (approved):

juliushaertl (approved):

PR #4721, opened 2023-08-24
fix: Use proper value for checking if editable

juliushaertl:
Use the right property for checking if the editor is read only


juliushaertl:
Let's wait for https://github.com/nextcloud/text/pull/4663 so I can extend with a test for this one

skjnldsv (approved):

mejo-:
/backport 468edd478c57cf2696b1ce2a9ef60507011151f0 to stable27

PR #4747, opened 2023-08-30
ci(node): do not install cypress as we do not need it.

max-nextcloud:
### 📝 Summary

Avoid installing an unnecessary dependency that sometimes even fails the build due to network issues.

### 🏁 Checklist

- no tests needed
- no docs needed


max-nextcloud:
Just testing this... might as well upstream it into the nextcloud workflow templates.

Here's the docs for the env variable: https://docs.cypress.io/guides/references/advanced-installation#Skipping-installation

skjnldsv:
Yes!! please upstream it! :rocket:

PR #4749, opened 2023-08-30
ci: Add workflow to update node dist files on main

juliushaertl:
Adapting https://github.com/nextcloud/server/pull/39674 for the text app to give it a test run on this branch

- [x] Remove compile check

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -66,14 +66,3 @@ jobs:
           webpackStatsFile: ./webpack-stats.json
           key: ${{ secrets.RELATIVE_CI_KEY }}
           token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Check webpack build changes
-        run: |
-          bash -c "[[ ! \"`git status --porcelain `\" ]] || (echo 'Please recompile and commit the assets, see the section \"Show changes on failure\" for details' && exit 1)"
-
-      - name: Show changes on failure
-        if: failure()
-        run: |
-          git status
-          git --no-pager diff
-          exit 1 # make it red to grab attention
--- a/.github/workflows/update-node-dist.yml
+++ b/.github/workflows/update-node-dist.yml
@@ -1,0 +1,61 @@
+name: Update Node dist
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      # implemented since 28, once branched off, add your stable branch here
+      - main
+      # - ci/node-dist-workflow
+      # - stable28
+
+concurrency:
+  group: update-node-dist-${{ github.head_ref || github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  update-node-dist:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+
+      - name: Read package.json node and npm engines version
+        uses: skjnldsv/read-package-engines-version-actions@8205673bab74a63eb9b8093402fd9e0e018663a1 # v2.1
+        id: versions
+        with:
+          fallbackNode: '^20'
+          fallbackNpm: '^9'
+
+      - name: Set up node ${{ steps.versions.outputs.nodeVersion }}
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3
+        with:
+          node-version: ${{ steps.versions.outputs.nodeVersion }}
+
+      - name: Set up npm ${{ steps.versions.outputs.npmVersion }}
+        run: npm i -g npm@"${{ steps.versions.outputs.npmVersion }}"
+
+      - name: Setup git
+        run: |
+          git config --local user.email "nextcloud-command@users.noreply.github.com"
+          git config --local user.name "nextcloud-command"
+      - name: Install dependencies & build
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+        run: |
+          npm ci
+          npm run build --if-present
+
+      - name: Check webpack build changes
+        id: changes
+        continue-on-error: true
+        run: |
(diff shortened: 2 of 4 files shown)


juliushaertl:
I was fearing some endless loop, but it seems that the git operation by default uses GITHUB_TOKEN which would block any follow up workflow triggered by that push. Meaning CI wouldn't run for the main branch on the compile commit, but i think that is fine since it runs on the previous and next one.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

max-nextcloud:
> CI wouldn't run for the main branch on the compile commit, but i think that is fine since it runs on the previous and next one.

Yes. For me that's even a plus as it will save a considerable amount of CI time.

max-nextcloud:
This looks very interesting - I'm all for trying it out.

I have some follow up questions / ideas:
* Is there any reason we cannot use this on stable27 and maybe older?
* Could we maybe drop this at some point for master and only have js assets in stable branches?

.github/workflows/update-node-dist.yml
```diff
@@ -0,0 +1,51 @@
+name: Update Node dist
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      # implemented since 28, once branched off, add you stable branch here
```
max-nextcloud, line 7:
```suggestion
      # implemented since 28, once branched off, add your stable branch here
```

.github/workflows/update-node-dist.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: npm i -g npm@"${{ steps.versions.outputs.npmVersion }}"
+
+      - name: Setup git
+        run: |
+          git config --local user.email "nextcloud-command@users.noreply.github.com"
+          git config --local user.name "nextcloud-command"
+      - name: Install dependencies & build
+        run: |
+          npm ci
+          npm run build --if-present
```
max-nextcloud, line 46:
```suggestion
      - name: Install dependencies & build
        env: 
          CYPRESS_INSTALL_BINARY: 0
        run: |
          npm ci
          npm run build --if-present
```

.github/workflows/update-node-dist.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Add and commit
+        run: |
+          git add --force js/
```
max-nextcloud, line 49: Currently this will fail the CI if there is nothing to commit. See first run on this branch.
Would be nice to handle that more gracefully. Either with an `if:` condition or by making the whole job depend on changes to `src`, `package.json` and so on.

juliushaertl: I pushed a commit to reuse the build changes check in here for that.

juliushaertl:
> Is there any reason we cannot use this on stable27 and maybe older?

Not from my perspective, maybe @skjnldsv had one as the original author of it in the server draft https://github.com/nextcloud/server/pull/39674?

> Could we maybe drop this at some point for master and only have js assets in stable branches?

Not until we build during release as there are also [nightly packages](https://download.nextcloud.com/server/daily/) built from master/main for the server release.

juliushaertl:
Also improved the concurrency block to work with pushes as head_ref is only there for prs

.github/workflows/update-node-dist.yml
```diff
@@ -0,0 +1,61 @@
...
+          CYPRESS_INSTALL_BINARY: 0
+        run: |
+          npm ci
+          npm run build --if-present
+
+      - name: Check webpack build changes
+        id: changes
+        continue-on-error: true
+        run: |
+          bash -c "[[ ! \"`git status --porcelain `\" ]] || (echo 'Please recompile and commit the assets, see the section \"Show changes on failure\" for details' && exit 1)"
```
max-nextcloud, line 54:
```suggestion
          bash -c "[[ ! \"`git status --porcelain `\" ]] || (echo 'Compile output changed. Will recompile and commit the assets.' && exit 1)"
```

max-nextcloud (approved):
One small change - other than that looks good to me.

skjnldsv:
Nice! Haven't found the time to keep looking deeper for https://github.com/nextcloud/server/pull/39674
Glad it was caught and of any help :rocket:

I don't see why that would be of any issue.
There could be a race condition if for any reason a releaser starts a release right after merge and before the dist files gets updated, but I think that would not be an issue if we just wait a bit :)

.gitignore
```diff
@@ -11,3 +11,6 @@ cypress/downloads/
 /tests/.phpunit.result.cache
 dist/
 webpack-stats.json
+
+# js folder, to be updated wth --force
+/js
```
skjnldsv, line 16: This was an issue, weirdly, if you have any ignored file at least ONCE in the git history, it will suggest to be committed again. Which defeated the purpose of not suggesting to add the changed files. That what made me stop any further progress on the server PR. I wanted to make sure only CI was updating the js folder, and no one mistakenly added their own changes in a future PR :thinking:

skjnldsv: Seems like a good idea yes :)
You can even probably use the GH path trigger property for this

juliushaertl: Added as an adjustment to the node.yml file

.github/workflows/node.yml
```diff
@@ -51,6 +51,17 @@ jobs:
...
+        run: |
+          bash -c "[[ \"`git status --porcelain `\" ]] || (echo 'Do NOT commit changes in js/ as those will be built automatically once merged' && exit 1)"
+
+      - name: Show changes on failure
+        if: failure()
+        run: |
+          git status
+          git --no-pager diff
+          exit 1 # make it red to grab attention
+
```
max-nextcloud, line 64: I don't think this does what you think it does.

`git status --porcelain` will not find anything as no compilation was run yet.

I see two alternatives:
* `git diff {name of base branch} js` and then fail if it has some output
* use the github mechanism to run tasks on changed files in js. (Maybe even in a separate task)

juliushaertl: Yes indeed. I moved to a separate workflow and used https://github.com/tj-actions/changed-files which uses the API instead of a local git clone to keep the action simple.

max-nextcloud (approved):
:rocket:

PR #4772, opened 2023-09-06
test(cypress): only run in parallel for pull requests

max-nextcloud:
Results from pushes to integration branches are usually not time critical.
Save some resources by avoiding the overhead of parallelization.

https://github.com/nextcloud/text/actions/runs/6089520881 is how pushes to integration branches would look in the future.


skjnldsv (approved):
I'm not sure it will save us that much time, but :+1:

juliushaertl (approved):

PR #4773, opened 2023-09-06
Use deployment environment for js build

juliushaertl:
Should fix the lack of permission to build dist assets on main


juliushaertl:
I can give a quick walk through next week :)

PR #4776, opened 2023-09-06
fix: Avoid layout jump with rich workspace

juliushaertl:
Alternative approach to #4254 which uses additional dav properties to already know on initial rendering if space for the rich workspace should be preserved.

In addition we can actually use the webdav markdown response to even show the full file content rendered before the text session is even started. As a follow up we could think about only initializing the session if the user clicks the rich workspace which would save us from lots of requests and problems with rich workspace always being locked.

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -187,6 +187,7 @@ export const FilesWorkspaceHeader = new Header({
 
 	render(el, folder, view) {
 		addMenuRichWorkspace()
+		const hasRichWorkspace = !!folder.attributes['rich-workspace-file']
 
 		import('vue').then((module) => {
 			el.id = 'files-workspace-wrapper'
@@ -201,6 +202,8 @@ export const FilesWorkspaceHeader = new Header({
 			vm = new View({
 				propsData: {
 					path: folder.path,
+					hasRichWorkspace,
+					content: folder.attributes['rich-workspace'],
 				},
 				store,
 			}).$mount(el)
@@ -208,7 +211,10 @@ export const FilesWorkspaceHeader = new Header({
 	},
 
 	updated(folder, view) {
+		const hasRichWorkspace = !!folder.attributes['rich-workspace-file']
 		vm.path = folder.path
+		vm.hasRichWorkspace = hasRichWorkspace
+		vm.content = folder.attributes['rich-workspace']
 	},
 })
 
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -22,10 +22,10 @@
 
 <template>
 	<div v-if="enabled"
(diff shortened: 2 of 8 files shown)


src/helpers/files.js
```diff
@@ -187,6 +188,8 @@ export const FilesWorkspaceHeader = new Header({
 
 	render(el, folder, view) {
 		addMenuRichWorkspace()
+		console.log(el, folder, view)
+		const hasRichWorkspace = folder.attributes['rich-workspace-file']
```
juliushaertl, line 192: @skjnldsv I'm having some trouble receiving the additionally registered properties on the initial page load.

Registration happens here: https://github.com/nextcloud/text/pull/4776/files#diff-6653fe3a8ac35dad2e28ef2134d921749c4334d4b81a593f80ae312cde95df4eR38

Is there anything in terms of loading order that can be done to fix that? When navigating to a sub folder this works fine.

src/helpers/files.js
```diff
@@ -201,14 +204,19 @@ export const FilesWorkspaceHeader = new Header({
...
+					hasRichWorkspace,
 				},
 				store,
 			}).$mount(el)
+			console.log('mounted => ' , vm)
 		})
 	},
 
 	updated(folder, view) {
+		const hasRichWorkspace = folder.attributes['rich-workspace-file']
```
juliushaertl, line 216: @skjnldsv Similar problem to the above, but when using bread crumb navigation the file list changes but neither render or updated is being called.

src/views/RichWorkspace.vue
```diff
@@ -134,9 +135,12 @@ export default {
...
 		getFileInfo(autofocus) {
-			this.loaded = false
-			this.autofocus = false
+			if (!this.hasRichWorkspace) {
+				return
+			}
 			this.ready = false
+			this.loaded = true
+			this.autofocus = false
 			const params = { path: this.path }
```
juliushaertl, line 144: FIXME: We still need to drop the following api request if we can get all info from the webdav response of the file list

src/helpers/files.js
juliushaertl: Fix in https://github.com/nextcloud/server/pull/40345

src/views/RichWorkspace.vue
juliushaertl: I kept this for now but instead made the loading more smooth as we have the markdown content already we can just preview that

juliushaertl:
Will bump OCP composer package once other CI is good :)

skjnldsv (approved):

juliushaertl:
Remaining failures unrelated

szaimen:
🎉🎉🎉🎉🎉

PR #4780, opened 2023-09-07
Fixes and additions to editor API

mejo-:
### 📝 Summary

This is required for Collectives to migrate to the editor API. Changes are separated by commits.

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -660,6 +660,10 @@ export default {
 			this.emit('delete-image-node', imageUrl)
 		},
 
+		async save() {
+			await this.$syncService.save()
+		},
+
 		async close() {
 			if (this.currentSession && this.$syncService) {
 				try {
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -43,13 +43,15 @@
 			<MainContainer v-if="hasEditor">
 				<!-- Readonly -->
 				<div v-if="readOnly" class="text-editor--readonly-bar">
-					<ReadonlyBar>
-						<Status :document="document"
-							:dirty="dirty"
-							:sessions="filteredSessions"
-							:sync-error="syncError"
-							:has-connection-issue="hasConnectionIssue" />
-					</ReadonlyBar>
+					<slot name="readonlyBar">
+						<ReadonlyBar>
+							<Status :document="document"
+								:dirty="dirty"
+								:sessions="filteredSessions"
+								:sync-error="syncError"
+								:has-connection-issue="hasConnectionIssue" />
+						</ReadonlyBar>
+					</slot>
 				</div>
 				<!-- Rich Menu -->
 				<template v-else>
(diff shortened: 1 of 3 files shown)


src/components/Editor.vue
```diff
@@ -726,6 +730,7 @@ export default {
 
 		outlineToggled(visible) {
 			this.emit('outline-toggled', visible)
+			emit('text:outline:toggle', visible)
```
juliushaertl, line 733: I'm wondering if we should scope the event by the file that is being edited, just in case there might be multiple editor instances at the same time?

Otherwise maybe you could also just add a callback handler to the editor api similar to https://github.com/nextcloud/text/blob/a0f9aa64c10200479c6c1643a0ce33de221b5a76/src/editor.js#L57-L59 which triggers the callback on the vue emit

src/editor.js
```diff
@@ -81,6 +81,15 @@ class TextEditorEmbed {
 		return this
 	}
 
+	async save() {
+		return this.#vm.$children[0].save?.()
```
juliushaertl, line 85: Maybe we could move `this.#vm.$children[0]` to a private getEditor method for more readability?

juliushaertl:
Not directly related, but maybe something to file a follow up issue to have tests covering the editor API. Not sure what the best approach would be there, maybe a small dummy app that just renders an editor?

mejo-:
> Not directly related, but maybe something to file a follow up issue to have tests covering the editor API. Not sure what the best approach would be there, maybe a small dummy app that just renders an editor?

I agree that we should test the API with dedicated tests. For now I'd say given that Collectives will test it anyways, it's not super urgent though. But definitely something to track. I opened https://github.com/nextcloud/text/issues/4790 to track this.

src/services/SyncService.js
```diff
@@ -268,6 +268,7 @@ class SyncService {
 			this.autosave.clear()
 		} catch (e) {
 			logger.error('Failed to save document.', { error: e })
+			throw e
```
juliushaertl, line 271: I thought I commented that already, but i feel this might be a risky change if the save fails during other code parts calling this method. Can we safeguard that other callers catch this error then and handle it gracefully?

juliushaertl (commented):
Looks good besides the one comment

max-nextcloud (commented):
reviewed and looks good to me.

Did not investigate julius' comment on the trow in `SyncService.js` yet. Is that needed / related?

src/services/SyncService.js
mejo-: If we can get this in and backported more quickly without the throw e change, I'm fine with postponing this one :grimacing:

mejo-: I had a look at all occasions of `.save()`, `.autosave()` and `.forceSave()`. Seems to be mostly places where throwing an error should be save. But maybe it's better to remove this potentially dangerous change from this PR that is supposed to be backported and move it to a seperate PR that doesn't get backported.

mejo-: Did this now.

max-nextcloud (approved):

Issue #4525, opened 2023-07-14
Loading error after deleting rich workspaces readme.md


dartcafe:
**Describe the bug**
See video

**To Reproduce**
See video

**Expected behavior**
* The folder description should disappear
* no error should raise, because it is expected that the deleted file can't be loaded

**Screenshots**

https://github.com/nextcloud/text/assets/26707476/9cd91a67-0f78-4c1d-9b99-70e2101a5ddf

**Server details:**
- Nextcloud version: 27.0.0
- PHP Version: 8.1
- Database: MariaDB 10.6

**Client details:**
 - OS: Windows
 - Browser: Firefox
 - Browser version: current
 - Device: desktop

I think it is reproducable, so no logs are necessary

dartcafe:
possibly related to #3719

juliushaertl:
With files to vue there should be an event that we can catch to hide the rich workspace accordingly

https://github.com/nextcloud/server/blob/7ee34a3fac802b1512fe3d12e9cc4d5b817a8ee1/apps/files/src/actions/deleteAction.ts#L52

PR #4782, opened 2023-09-07
fix: deleting rich workspaces readme.md

luka-nextcloud:
### 📝 Summary

* Resolves: #4525

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -53,6 +53,7 @@ import RichTextReader from '../components/RichTextReader.vue'
 
 const IS_PUBLIC = !!(document.getElementById('isPublic'))
 const WORKSPACE_URL = generateOcsUrl('apps/text' + (IS_PUBLIC ? '/public' : '') + '/workspace', 2)
+const SUPPORTED_STATIC_FILENAMES = ['Readme.md', 'README.md', 'readme.md']
 
 export default {
 	name: 'RichWorkspace',
@@ -112,6 +113,8 @@ export default {
 		}
 		subscribe('Text::showRichWorkspace', this.showRichWorkspace)
 		subscribe('Text::hideRichWorkspace', this.hideRichWorkspace)
+		subscribe('files:node:deleted', this.onFileDeleted)
+		subscribe('files:node:renamed', this.onFileRenamed)
 
 		this.listenKeydownEvents()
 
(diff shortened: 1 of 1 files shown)


src/views/RichWorkspace.vue
```diff
@@ -107,6 +107,7 @@ export default {
 		}
 		subscribe('Text::showRichWorkspace', this.showRichWorkspace)
 		subscribe('Text::hideRichWorkspace', this.hideRichWorkspace)
+		subscribe('files:node:deleted', this.onFileDeleted)
```
juliushaertl, line 110: Please also unsubscribe in beforeDestroy :) Otherwise this looks good 👍

juliushaertl: Ah and maybe we should also cover the `files:node:renamed` event to cover file names being changed from/to rich workspace filenames (reference for the possible filenames https://github.com/nextcloud/text/blob/23ce8f876852771accba68bad2c297271fb3aa22/lib/Service/WorkspaceService.php#L14-L43).

max-nextcloud (approved):

max-nextcloud:
Code looks good. Cypress test would be nice to avoid regressions in the future.

luka-nextcloud:
@max-nextcloud @juliushaertl I found that cypress tests for workspace was set to skip [here](https://github.com/nextcloud/text/blob/main/cypress/e2e/workspace.spec.js#L26)
Should we enable them again?
It seems like there are many errors while enabling them. Please check below screenshot. Should we fix them in this PR?
Test case for above issue already included in case "[Hides the workspace when switching to another view](https://github.com/nextcloud/text/blob/main/cypress/e2e/workspace.spec.js#L56)". But need small update.

mejo-:
@luka-nextcloud I agree that we certainly should re-enalble the Cypress tests for rich workspaces. Seems like skipping them was meant as temporary in https://github.com/nextcloud/text/commit/db7aa50c0377c54bc795fdc704c0ddabb7ad3b74 anyway. Maybe you can take a look at fixing the Cypress tests?

max-nextcloud:
I think a first step towards this would be running the current master branch locally and checking how rich workspaces work there. Maybe some of the functionality is still missing.

mejo-:
@luka-nextcloud I'm fine with this. Could you open the issue?

luka-nextcloud:
Here's a separate ticket: https://github.com/nextcloud/text/issues/4829

Issue #4823, opened 2023-09-25
Y.js fails to disconnect because session is already closed


max-nextcloud:
**Describe the bug**
When closing the editor or navigating to another page in collectives
A last `push` request is send that results in a 403.
This request originates in the yjs websocket handlers `disconnectBc` function.
It would remove the corresponding users session from the awareness state.

**To Reproduce**
1. Open a markdown document in text
2. Type some chars
3. Close the editor
4. See error in the network tab.

**Expected behavior**
The last yjs message passes fine and only then the connection is closed by text.

In addition texts client side session api prevents sending request through a closed connection.

**Server details:**
- Nextcloud version: > 26
- PHP Version: 8.2
- Database: MariaDB (nextcloud docker dev instance)

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
nextcloud_1          | 192.168.23.5 - - [25/Sep/2023:13:48:47 +0000] "POST /index.php/apps/text/session/push HTTP/1.1" 403 852 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36"

```

#### Browser log

![grafik](https://github.com/nextcloud/text/assets/97337118/5e3f6a8d-8443-402b-b103-f3ff8124184f)

![grafik](https://github.com/nextcloud/text/assets/97337118/c1264d67-c396-4330-996d-f5eb4392330b)

</details>

PR #4824, opened 2023-09-26
Fix 403 on close

max-nextcloud:
### 📝 Summary

* Resolves: #4823
* Prevent sending requests with closed session - raise an exception instead.
* Send remaining steps before closing connection so last yjs awareness update gets through.

### 🏁 Checklist

### DIFF
--- a/src/services/SessionApi.js
+++ b/src/services/SessionApi.js
@@ -22,6 +22,14 @@
 import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 
+export class ConnectionClosedError extends Error {
+
+	constructor(message = 'Close has already been called on the connection', ...rest) {
+		super(message, ...rest)
+	}
+
+}
+
 class SessionApi {
 
 	#options
@@ -50,6 +58,7 @@ class SessionApi {
 export class Connection {
 
 	#content
+	#closed
 	#documentState
 	#document
 	#session
@@ -66,6 +75,7 @@ export class Connection {
 		this.#content = content
 		this.#documentState = documentState
 		this.#options = options
+		this.closed = false
 	}
 
 	get session() {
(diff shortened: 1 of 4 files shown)


src/services/SessionApi.js
```diff
@@ -22,6 +22,14 @@
 import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 
+export class ConnectionClosed extends Error {
```
mejo-, line 25: Nitpicking, but wouldn't `ConnectionClosedError` be a better class name?

src/services/WebSocketPolyfill.js
```diff
@@ -112,7 +112,8 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 				.findLast(s => s > 'AQ') || ''
 		}
 
-		close() {
+		async close() {
+			await this.#sendRemainingSteps()
```
mejo-, line 116: Great idea to try to send remaining steps before closing. I just wonder whether this might have user-facing implications on the responsibility of closing the session (e.g. when closing a document)?

max-nextcloud: I tried and it was smooth... but that was with a fast network. Will try with throttling.

max-nextcloud: Tried with slow 3g throtteling. The viewer closes fast and the first request in the background is the save request which writes the document. Then we push the remaining steps and update the awareness state and finally we send the close request. Due to the throtteling each request took 2 sec. - so after a total of 6 seconds the session was closed. However these were no noticable for the user as the viewer window closed fast - and it does not delay the saving of the document as that happens first.

So all in all i'd say that's the behavior we want to see. I'll test with collectives though as switching pages is a different scenario.

max-nextcloud: Collectives also works great. The session `create` request for the next document happens in parallel with the `save` of the old. Then there's `push`, `sync` and `close`. Where `sync` is independent and happening in parallel.

juliushaertl (approved):

mejo- (approved):

mejo-:
/backport d17bc9a8db6524be40e8aec958ace79ec77e7211,d89971462e6233fa0d24be39bfa903c79f6cb6d2 to stable27

PR #4843, opened 2023-10-11
fix: CSP issue on loading RichWorkspace component

luka-nextcloud:
### 📝 Summary

* Resolves: Fix CSP issue on loading RichWorkspace component

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image
 | A

### 🏁 Checklist

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -1,6 +1,10 @@
 import { registerFileListHeaders, registerDavProperty } from '@nextcloud/files'
 import { loadState } from '@nextcloud/initial-state'
 import { FilesWorkspaceHeader } from './helpers/files.js'
+import { linkTo } from '@nextcloud/router'
+
+__webpack_nonce__ = window.btoa(OC.requestToken) // eslint-disable-line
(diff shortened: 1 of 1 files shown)


src/helpers/files.js
```diff
@@ -22,14 +22,17 @@
...
-import { imagePath } from '@nextcloud/router'
+import { imagePath, linkTo } from '@nextcloud/router'
 import { loadState } from '@nextcloud/initial-state'
 
 import { getSharingToken } from './token.js'
 import { openMimetypes } from './mime.js'
 import RichWorkspace from '../views/RichWorkspace.vue'
 import store from '../store/index.js'
 
+__webpack_nonce__ = window.btoa(OC.requestToken) // eslint-disable-line
```
juliushaertl, line 33: Usually we tend to have this in the actual entry point. I think we lack it in `src/init.js` which seems to be the only webpack entry point without it, so maybe we can also just move it there?

juliushaertl (approved):

PR #4847, opened 2023-10-11
ci(cypress): Use cypress-split for running tests in parallel

juliushaertl:
Not as equally distributed as with cypress cloud but much better than waiting for over an hour on one job to finish.

current distribution
Screenshot 2023-10-11 at 20 33 52

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -85,21 +85,6 @@ jobs:
         server-versions: [ 'master' ]
         run-in-parallel:
           - false # only for PRs: ${{ !!github.head_ref }}
-        exclude:
-          - run-in-parallel: false
-            containers: 2
-          - run-in-parallel: false
-            containers: 3
-          - run-in-parallel: false
-            containers: 4
-          - run-in-parallel: false
-            containers: 5
-          - run-in-parallel: false
-            containers: 6
-          - run-in-parallel: false
-            containers: 7
-          - run-in-parallel: false
-            containers: 8
 
     name: runner ${{ matrix.containers }}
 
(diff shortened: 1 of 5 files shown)


juliushaertl:
Job summary is not optimal likely due to our custom specPattern config (ref https://github.com/bahmutov/cypress-split/issues/69)

.github/workflows/cypress.yml
```diff
@@ -83,23 +83,6 @@ jobs:
         containers: [1, 2, 3, 4, 5, 6, 7, 8]
         php-versions: [ '8.1' ]
         server-versions: [ 'master' ]
-        run-in-parallel:
```
mejo-, line 86: `run-in-parallel` is used further down in the workflow, so I think it's worth keeping it for now. Maybe we can change this to `${{ !!github.head_ref }}` again, to make sure that CI runs on development+release branches don't register in the Cypress cloud - and only the PR CI jobs end up in Cypress cloud.

mejo- (changes requested):

cypress.config.js
```diff
@@ -1,11 +1,23 @@
...
+const cypressSplit = require('cypress-split')
+
+module.exports = defineConfig({
+  e2e: {
+    setupNodeEvents(on, config) {
+      // IMPORTANT: return the config object
+      return config
+    },
+  },
+})
```
skjnldsv, line 11:
```suggestion
const cypressSplit = require('cypress-split')

module.exports = defineConfig({
	e2e: {
		setupNodeEvents(on, config) {
			// IMPORTANT: return the config object
			return config
		},
	},
})
```

skjnldsv, line 11:
```suggestion
const cypressSplit = require('cypress-split')

module.exports = defineConfig({
	e2e: {
		setupNodeEvents(on, config) {
			// IMPORTANT: return the config object
			return config
		},
	},
})
```

skjnldsv, line 11:
```suggestion
const cypressSplit = require('cypress-split')

module.exports = defineConfig({
	e2e: {
		setupNodeEvents(on, config) {
			// IMPORTANT: return the config object
			return config
		},
	},
})
```

juliushaertl:
Good to go from my side.

The author has also already [fixed the github summary for our case](https://github.com/bahmutov/cypress-split/pull/118) already which now nicely displays spec stats:

https://github.com/bahmutov/cypress-split/actions/runs/6496437037#summary-17643354799

One more report remaining about the time but that doesn't seem too critical to me.  https://github.com/bahmutov/cypress-split/issues/121

max-nextcloud (approved):

Issue #4768, opened 2023-09-04
Pasting and editing multi line text removes line breaks


ChristophWurst:
**Describe the bug**

Pasted multi line text shows with line breaks until one of the lines is edited.

**To Reproduce**
1. Create a new text file
2. Copy-paste
```
Lorem
Ipsum
```
4. Add a *x* after *Lorem*

**Expected behavior**

```
Loremx
Ipsum
```

**Actual behavior**

```
Loremx Ipsum
```

**Screenshots**

**Server details:**
- Nextcloud version: 27.1
- PHP Version: [e.g. 8.2]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome]
 - Browser version: [e.g. 22]
 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

XueSheng-GIT:
The same happens if you create a markdown file locally with regular (single) line breaks and upload this file to nextcloud. Line breaks of the whole text are removed if you start editing in nextcloud. I assume this is the same issue.
Maybe related: https://github.com/nextcloud/text/pull/3568

mejo-:
@ChristophWurst what you describe is already reported here, no? https://github.com/nextcloud/text/issues/4602

ChristophWurst:
It sounds very similar but there are no unexpected linebreaks for me. I do expect the lines to break once and they do. If I modify the text there are no line breaks left at all

PR #4624, opened 2023-08-02
fix(NodeViewContent): Don't set `white-space: pre-wrap`

mejo-:
Fixes: #4602

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits

PR #4872, opened 2023-10-17
fix(paste): collapse whitespace before pasting

### LINKED ISSUE #4602: Pasting paragraphs sometimes adds unexpected linebreaks
**Describe the bug**
When pasting paragraphs to a document, sometimes unexpected linebreaks are added in between the paragraph. As soon as you do one change in the paragraph, all unexpected linebreaks get deleted at once.

As an example, paste the first 3-4 paragraphs of the [wikipedia article about Fairy Tales](https://en.wikipedia.org/wiki/Fairy_tale) into a document.

**To Reproduce**
1. Open a document, paste the example paragraph from above.
2. See unexpected linebreaks
3. Add one character anywhere in the paragraph
4. The the unexpected linebreaks disappearing.

max-nextcloud:
### 📝 Summary

By default html does not display whitespace but text does. Remove newlines when pasting to ensure consistency with the previous display.

Keep the newlines when they are also displayed in the source.

* Resolves: #4602
* Resolves: #4768
* Superseedes: #4624 

#### 🏚️ Before

grafik 

#### 🏡 After

grafik

### 🏁 Checklist

### DIFF
--- a/src/extensions/transformPastedHTML.js
+++ b/src/extensions/transformPastedHTML.js
@@ -1,0 +1,100 @@
+/**
+ * @copyright Copyright (c) 2023 Max <max@nextcloud.com>
+ *
+ * @author Max <max@nextcloud.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/**
+ *
+ * Prepare pasted html for insertion into tiptap
+ *
+ * We render paragraphs with `white-space: pre-wrap`
+ * so newlines are visible and preserved.
+ *
+ * Pasted html may contain newlines inside tags with a different `white-space` style.
+ * They are not visible in the source.
+ * Strip them so the pasted result wraps nicely.
+ *
+ * At the same time we need to preserve whitespace inside `<pre>` tags
+ * and the like.
+ *
+ * @param {string} html Pasted html content
+ */
+export default function(html) {
+	const parser = new DOMParser()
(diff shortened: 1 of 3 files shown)


src/extensions/transformPastedHTML.js
```diff
@@ -0,0 +1,97 @@
...
+	// Values of `white-space` css that will collapse newline whitespace
+	// See https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#values
+	const COLLAPSING_WHITE_SPACE_VALUES = ['normal', 'nowrap']
+	let ancestor = textNode.parentElement
+	while (ancestor) {
+		const style = getComputedStyle(ancestor)
+		const whiteSpace = style?.getPropertyValue('white-space')
+		if (whiteSpace) {
+			return COLLAPSING_WHITE_SPACE_VALUES.includes(whiteSpace)
+		}
```
max-nextcloud, line 87: While this is more precise as it also detects `<div style="white-space: pre-wrap">` and `<pre><em style="white-space: normal">` it's only supported in Firefox. So not sure if it's worthwhile the inconsistency and extra logic.

mejo-:
Probably also will fix https://github.com/nextcloud/text/issues/4768.

src/extensions/transformPastedHTML.js
```diff
@@ -0,0 +1,97 @@
...
+function collapseWhiteSpace(textNode) {
+	// Values of `white-space` css that will collapse newline whitespace
+	// See https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#values
+	const COLLAPSING_WHITE_SPACE_VALUES = ['normal', 'nowrap']
+	let ancestor = textNode.parentElement
+	while (ancestor) {
+		const style = getComputedStyle(ancestor)
+		const whiteSpace = style?.getPropertyValue('white-space')
+		if (whiteSpace) {
+			return COLLAPSING_WHITE_SPACE_VALUES.includes(whiteSpace)
```
mejo-, line 86: Maybe a short explanationary comment above this line: `// Returns false if white-space has a value not listed in COLLAPSING_WHITE_SPACE_VALUES`.

src/extensions/transformPastedHTML.js
mejo-: I'm in favour of keeping the extra logic nevertheless. I'd move the comment from below to the top of `while(ancestor)` though and add something like:
> // Therefore the following logic only works on Firefox

Then an empty line before the `ancestor.tagName === 'PRE'` part and another comment above:
> // Check for `tagName` as fallback

mejo- (approved):
Max gave me a walkthrough through the code. Very clean and thorough solution, thanks a lot! Only nitpicking comments.

mejo-:
Cypress failures are unrelated.

Issue #3863, opened 2023-03-01
[BITV] 9.2.4.3/3.9 - Toolbar focus handling


AndyScherzinger:
- [ ] Fix
- [ ] Backport

Note: When style is applied to selected text in the editing region (div with contenteditable=true) by activating the button for that style the focus is set to the selected text. There are pros and cons for moving the focus. 

In any case the controls in the div with role=menubar (role=toolbar would be more appropriate) should have ONE focus stop for the toolbar, implement arrow navigation for its elements, and support aria-activedescendant and scripted focus management so the focus can be set to the currently active style option (the last used of them when several styles have been applied) to facilitate editing for keyboard users. (2)

![5e2cbbfe0fd4258eab94ad65371fb1fa](https://user-images.githubusercontent.com/1315170/222250672-8c463058-775b-4c0a-b079-830ff7168593.png)

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-89806fe784-v3-n9

</details>

- [ ] 9.2.4.3/10.9 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-89806fe784-v10-n9

datenangebot:
Can confirm, no arrow navigation, but tabs work fine.

If you click on the expandable menu, the arrows works.
So atm it is mixed. 🙄

Issue #3911, opened 2023-03-07
[BITV] 9.4.1.2/3.16 - Custom editing menubar should be implemented using role=toolbar and not role=menubar. (2)


AndyScherzinger:
Currently selected styles on selected text in the editing region ("div" with "contenteditable=true") are shown visually (light blue background) but they are not programmatically determinable via aria-selected attribute. aria-activedescendant is not used on toolbar styling elements. Also missing is appropriate scripted focus management for the style toolbar and for its drop-down submenus like headings, insert image, insert emoi). 

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-bfbf07e5c3-v3-n16

</details>

- [ ] 9.4.1.2/10.16 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-bfbf07e5c3-v10-n16

juliushaertl:
Isn't this already covered by https://github.com/nextcloud/text/issues/3863?

AndyScherzinger:
> Isn't this already covered by https://github.com/nextcloud/text/issues/3863?

Cant tell @juliushaertl - best to clarify with @michaelnissenbaum

blizzz:
Any update on the state? @juliushaertl @michaelnissenbaum

susnux:
> Isn't this already covered by #3863?

@juliushaertl yes and no, the title of this issue is misleading.  That issue covers the toolbar vs menubar handling.
But this one is about:

> Currently selected styles on selected text in the editing region ("div" with "contenteditable=true") are shown visually (light blue background) but they are not programmatically determinable via aria-selected attribute. aria-activedescendant is not used on toolbar styling elements. Also missing is appropriate scripted focus management for the style toolbar and for its drop-down submenus like headings, insert image, insert emoi).

- [ ] Adding `aria-selected` to the currently active menu element (currently only the primary background is set, so only visual information is available)
- [ ] Adding `aria-activedescendant` to the menu item with the content being the ID of the currently active descendant (for submenus)
- [x] "Also missing is appropriate scripted focus management": Part of #3863

Pytal:
This issue seems to be the same as https://github.com/nextcloud/text/issues/3864 + `aria-selected` and can be done simultaneously, feel free to unassign me and assign yourself @susnux if you think it's within the same scope :)

ShGKme:
[...]

PR #4874, opened 2023-10-17
feat(MenuBar): Make the menu bar `role=toolbar` and add focus handling

susnux:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/3863
* Part of https://github.com/nextcloud/text/issues/3911

For accessibility the toolbar should be implemented using `role=toolbar`. This requires custom focus handling[1] this is done by making only one element tabable / only one tab stop.

I had to remove the ActionEntry component as this functional component did not forward the template ref, so replaced it with a conditional component.

[1] https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role#description

#### 🖼️ Screenshots

https://github.com/nextcloud/text/assets/1855448/f63367b8-439c-414c-9623-201123933cdb

As you can see there is only one tab stop in the toolbar (the last used formatting option).

### 🚧 TODO

- Not part of that issue but we need to keep the focus on the menu item rather than jumping to the content after click.

### 🏁 Checklist


Pytal (approved):
Tested and works!

skjnldsv (approved):

JuliaKirschenheuter (approved):

juliushaertl (approved):

juliushaertl:
Failures are unrelated, especially broken rich workspace seems to come from a recent server merge

PR #4886, opened 2023-10-20
fix(makefile): Disable parallel execution of targets

provokateurin:
### 📝 Summary

I have `export MAKEFLAGS="-j$(nproc)"` in my environment which allows make to spawn multiple jobs at the same time. This breaks running `make` because installing dependencies and compiling happens at the same time which ultimate fails.
Adding this line disables any parallel execution of any targets. See https://www.gnu.org/software/make/manual/html_node/Parallel-Disable.html

### 🏁 Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 # later. See the COPYING file.
 app_name=$(notdir $(CURDIR))
 
+.NOTPARALLEL:
+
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

provokateurin:
Failure is of course unrelated to this change.

PR #4904, opened 2023-10-26
Revert "tmp: skip known attachment failures"

juliushaertl:
This reverts commit 66d5c7dc07330669b0e7fd7a917648584a2d84d2

Further commit fixes the tests, isolates them a bit more and moves file operations from clicking through UI to actual API calls.


mejo- (approved):
Nice, much cleaner tests now! Probably worth being backported to stable28, no?

Issue #4600, opened 2023-07-28
Sync issues with missing Yjs steps/structs


mejo-:
**Describe the bug**
In a test session with four participants we were able to create a situation were changes of two participants were no longer applied to the document of the two others.

Some first insights after analysing the HAR file of three participants

* Synchronisation of client (a) ended at exactly the same Yjs client clock count for two other clients (c) and (d). Synchronisation of client (a) kept working on client (b). We thus assume that changes of client (a) were propagated to the server.
* Both clients (c) and (d) received Yjs steps with a higher clock count by client (a). They were not applied to the document state due to two missing steps.
* Apparently client (a) reconnected and changed its Yjs client ID just in the moment when synchronisation stopped working on clients (c) and (d).
* Client (a) sent a Yjs query request (starting with `AAA...`) that's filtered out by the server in the moment the synchronisation stopped for clients (c) and (d).

max-nextcloud:
I think we are missing a mechanism for resyncing clients once they get out of sync. yjs-websocket provides this in https://github.com/yjs/y-websocket/blob/master/src/y-websocket.js#L300-L311 - but that only works with a server that understands the y-js protocol and can answer the sync type 1 messages from the client.

All a shared resync with multiple clients would take would be:
1. One client (A) sends its full current yjs state incl. history and a matching sync type 1 message.
2. These are distributed to all clients.
3. All clients reply to the sync type 1 if they have additional steps
4. These replies are again distributed to all clients.

* `1.` and `2.` ensure all other clients are at least in sync with (A).
* `3.` and `4.` ensures all steps the other clients have in addition are also synced.

At the same time this sync scales fairly well:
1. grows linear with the compressed document size (some 10k)
2. grows linear with the number of clients (`n`) and the compressed document size ( some 10k per client)
3. will be small as it's only the diff to 1. - so at most the compressed steps since the last resync
4. grows with `n^2` - times the small diff.

For this to scale nicely it's important that only one client performs 1. Otherwise everything will be multiplied by `n`.

Since we already have an autosave request that is send by one client every 30 seconds (?) I propose the following:
a) in addition to the autosave content the saving client sends the content of 1.
b) the server adds these two messages to the distributed steps even though one of them is a sync type 1 which we normally drop.

This should be enough to provide the resync mechanism. We could rely on it and simplify other parts of the sync:

c) When autosaving we don't need to upload the doc state anymore as that's encoded in the first message in 1.
[...]

PR #4938, opened 2023-10-30
fix(sync): prevent race condition by relying on autoincrement

max-nextcloud:
Prevent a possible race condition when two clients add steps at the same time.

See #4600.

Rely on the autoincrementing id in order to provide a canonical order that steps can be retrieved in.

When two clients push steps at the same time
the entries receive destinct ids that increment.
So if another client fetches steps in between
it will see the smaller id as the version of the fetched step and fetch the other step later on.

Transition:
In the future we can drop the version column entirely but currently there are still steps stored in the database that make use of the old column.
So we need to transition away from that.

In order to find entries that are newer than version x we select those that have both a version and an id larger than x.

Entries of the new format are newer than any entry of the old format. So we set their version to the largest possible value. This way they will always fulfill the version condition and the condition on the id is more strict and therefore effective.

For the old format the version will be smaller than the id as it's incremented per document while the id is unique accross documents. Therefore the version condition is the more strict one and effective.

The only scenario where the version might be larger than the id would be if there's very few documents in the database and they have had a lot of steps stored in single database entries.

### 🏁 Checklist


max-nextcloud:
Obviously WIP - opened the PR to be able to discuss the approach.

lib/Service/DocumentService.php
```diff
@@ -263,15 +263,14 @@ private function insertSteps(Document $document, Session $session, array $steps,
...
-			$this->cache->set('document-version-' . $document->getId(), $newVersion);
+			$this->logger->debug("Adding steps to " . $document->getId() . ": bumping version from $stepsVersion");
 			$step = new Step();
 			$step->setData($stepsJson);
 			$step->setSessionId($session->getId());
 			$step->setDocumentId($document->getId());
-			$step->setVersion($newVersion);
-			$this->stepMapper->insert($step);
+			$step = $this->stepMapper->insert($step);
+			$newVersion = $step.id;
```
juliushaertl, line 272:
```suggestion
			$newVersion = $step->getId();
```

lib/Db/Step.php
```diff
@@ -37,9 +37,19 @@
...
+	 * In order to make sure that new steps always
+	 * have a larger version than those that used the version field
+	 * use the largest possible value for big int
+	 */
+	public const VERSION_STORED_IN_ID = 18446744073709551615;
+
 	public $id = null;
 	protected string $data = '';
-	protected int $version = 0;
+	protected int $version = VERSION_STORED_IN_ID;
```
juliushaertl, line 52:
```suggestion
	protected int $version = self::VERSION_STORED_IN_ID;
```

lib/Db/Step.php
```diff
@@ -55,10 +65,13 @@ public function jsonSerialize(): array {
 		if (\json_last_error() !== JSON_ERROR_NONE) {
 			throw new \InvalidArgumentException('Failed to parse step data');
 		}
+		$version = $this->getVersion() == VERSION_STORED_IN_ID
```
juliushaertl, line 68:
```suggestion
		$version = $this->getVersion() == self::VERSION_STORED_IN_ID
```

mejo-:
I like the approach, it's a way cleaner solution than my approach to make read+write of `version` atomic. I pushed some fixes (force-pushed code-style fixes into the original commit and added another commit for logical fixes).

mejo-:
Successfully tested the following scenario:


- main branch checked out
- client1 creates document and creates steps
- client2 joins document session and creates steps
- client2 leaves document session
- switch to PR branch
- client3 joins document session and creates steps
- client2 joins document session and creates steps
- client1 creates steps

All change are synced between all clients as expected.

mejo- (approved):
Ready for merge from my perspective.

max-nextcloud:
There are two things that i did not look into yet that might be worthwhile ensuring:
* I'm not sure how the database queries perform now. We're now limiting by both `version` and `id` and I think there's an index on each of them but no combined index. Since these endpoints are called very often would be good to ensure there's no performance regression. Maybe we can get hold of the sql query that's run right now and run `EXPLAIN` on it to see which indexes it would use.
* There's a theoretical possibility that the highest version in the steps table is larger than the largest id. Versions can increment by more than one so if i only have one document that has a few steps and some of them increase the version by more than one the resulting version will be higher than the max id. This is unlikely but i am unsure if we can completely ignore it. It could be mitigated by inserting a step with a higher id that does not serve any other purpose. Not sure that cornercase is worth the effort though.

mejo-:
> There's a theoretical possibility that the highest version in the steps table is larger than the largest id. Versions can increment by more than one so if i only have one document that has a few steps and some of them increase the version by more than one the resulting version will be higher than the max id. This is unlikely but i am unsure if we can completely ignore it. It could be mitigated by inserting a step with a higher id that does not serve any other purpose. Not sure that cornercase is worth the effort though.

I thought a bit about this and I think we should ignore this corner case. Nextcloud instances with just one (or very few) document sessions likely aren't used collaboratively. Sure, there's a theoretical possibility that people setup a Nextcloud instance to work together on one file, but that's not very likely. And even in this scenario, the autosave should have saved the last document state. So worst case in this very unlikely scenario people will loose a few editing steps. This is not worth adding complex migration logic in my opinion.

mejo-:
>     I'm not sure how the database queries perform now. We're now limiting by both version and id and I think there's an index on each of them but no combined index. Since these endpoints are called very often would be good to ensure there's no performance regression. Maybe we can get hold of the sql query that's run right now and run EXPLAIN on it to see which indexes it would use.

@juliushaertl and me looked into this. The index on `document_id` is always used first and the following where statements are only executed on its subset. In our tests, the new queries didn't look worse regarding performance than the old ones.

mejo-:
/backport ea208ee0d69f08241f7525352e9affd060ff34e9,308d64069f8903a113985d7d71d027b127a3b0ce to stable27

Issue #4944, opened 2023-10-31
JS error: `this.visibleEntries[this.activeMenuEntry] is undefined`


mejo-:
**Describe the bug**
In Collectives, Text regularly produces the following error when the editor is being loaded:

```
TypeError: this.visibleEntries[this.activeMenuEntry] is undefined
    visibleEntries ToolBarLogic.js:19
    VueJS 5
    visibleEntries ToolBarLogic.js:18
    VueJS 12
    onResize MenuBar.vue:210
    p index.mjs:2394
    VueJS 5
    c index.mjs:2390
    mounted MenuBar.vue:177
    VueJS 20
    updateSessions Editor.vue:436
    onChange Editor.vue:556
vue.runtime.esm.js:3049:16
```

I'm pretty sure it's due to a recent change in Text (main branch). I only see this since a few weeks.

PR #4945, opened 2023-11-01
fix(menubar): Code style fixes

mejo-:
### 📝 Summary

* Fixes: #4944

### DIFF
--- a/src/components/Menu/MenuBar.vue
+++ b/src/components/Menu/MenuBar.vue
@@ -49,7 +49,7 @@
 			@keydown.right.stop="handleToolbarNavigation">
 			<!-- The visible inline actions -->
 			<component :is="actionEntry.component ? actionEntry.component : (actionEntry.children ? 'ActionList' : 'ActionSingle')"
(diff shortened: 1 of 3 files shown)


src/components/Editor/EditorOutline.vue
```diff
@@ -61,10 +61,14 @@ export default {
 	// 204px = 50px nc header + 60px collectives titlebar + 44px menubar + 50px bottom margin
 	max-height: calc(100% - 204px);
 
+	border-radius: var(--border-radius-large);
+	box-shadow: 0 1px 10px var(--color-box-shadow);
+	background-color: var(--color-main-background);
```
juliushaertl, line 66: Can we drop this change and rather discuss it separately?

Having the outline floating without borders was part of the design requirements or feedback when we introduced it, so I'd not change that as part of a backport at least.

mejo-: Yeah, sorry for pushing two unrelated fixes into one PR. I changed this PR to only contain the menubar fix now.

juliushaertl (approved):

Issue #4937, opened 2023-10-30
Cypress: Bring back create rich workspace through UI


juliushaertl:
As stated in https://github.com/nextcloud/text/pull/4853/files#r1358150289 with the merge of https://github.com/nextcloud/text/pull/4873 we should now have a cypress test back that would:

- Create a rich workspace file by clicking the new file menu "Add description" (cover https://github.com/nextcloud/text/pull/4873)
  - Assert that the file was created
  - Assert that the workspace shows above the file list
  - Assert that the entry is no longer visible in the files menu
- Enter a folder with an existing Readme.md file 
  - Assert that the entry is no longer visible in the new files menu (cover https://github.com/nextcloud/text/pull/4873)
  - Delete the file
  - Assert that the entry is visibile in the new files menu (cover https://github.com/nextcloud/text/pull/4782)

@luka-nextcloud Can you look into that as a follow up of your pull requests?

PR #4965, opened 2023-11-06
feat: bring back cypress command createDescription through UI

luka-nextcloud:
### 📝 Summary

* Resolves: #4937

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -147,7 +147,7 @@ const registerFileActionFallback = () => {
 
 }
 
-const addMenuRichWorkspace = () => {
+export const addMenuRichWorkspace = () => {
 	const descriptionFile = t('text', 'Readme') + '.' + loadState('text', 'default_file_extension')
 	addNewFileMenuEntry({
 		id: 'rich-workspace-init',
@@ -177,7 +177,7 @@ const addMenuRichWorkspace = () => {
 			})
 			const fileid = parseInt(response.headers['oc-fileid'])
 			const file = new File({
-				source,
+				source: context.source + '/' + encodeURIComponent(descriptionFile),
 				id: fileid,
 				mtime: new Date(),
 				mime: 'text/markdown',
@@ -185,11 +185,13 @@ const addMenuRichWorkspace = () => {
 				permissions: Permission.ALL,
 				root: context?.root || '/files/' + getCurrentUser()?.uid,
 			})
-			context._children.push(file.fileid)
 
 			showSuccess(t('text', 'Created "{name}"', { name: descriptionFile }))
+
 			emit('files:node:created', file)
-			emit('Text::showRichWorkspace', { autofocus: true })
+			setTimeout(() => {
+				emit('Text::showRichWorkspace', { autofocus: true })
+			}, 200)
 		},
 	})
 }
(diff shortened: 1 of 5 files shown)


src/views/RichWorkspace.vue
```diff
@@ -144,9 +143,10 @@ export default {
 			})
 		},
 		getFileInfo(autofocus) {
-			if (!this.hasRichWorkspace) {
```
juliushaertl, line 147: Can you elaborate a bit why this change is needed? as far as I see this would no longer check on each folder change on if the folder has a rich workspace

src/helpers/files.js
```diff
@@ -177,19 +177,21 @@ const addMenuRichWorkspace = () => {
...
 			})
-			context._children.push(file.fileid)
 
 			showSuccess(t('text', 'Created "{name}"', { name: descriptionFile }))
+
 			emit('files:node:created', file)
-			emit('Text::showRichWorkspace', { autofocus: true })
+			setTimeout(() => {
+				emit('Text::showRichWorkspace', { autofocus: true })
+			}, 200)
```
juliushaertl, line 194: Can you tell why the previous one didn't work?

src/init.js
```diff
@@ -12,5 +12,6 @@ registerDavProperty('nc:rich-workspace', { nc: 'http://nextcloud.org/ns' })
 registerDavProperty('nc:rich-workspace-file', { nc: 'http://nextcloud.org/ns' })
 
 if (workspaceAvailable) {
+	addMenuRichWorkspace()
```
juliushaertl, line 15: Good idea to move that over here 👍

juliushaertl (changes requested):
Would be great to get a bit of further insights on the changes (see inline)

src/views/RichWorkspace.vue
luka-nextcloud: As I checked,
- when Readme.md file was created, it did not trigger updated function (at https://github.com/nextcloud/text/blob/main/src/helpers/files.js#L232) to update property hasRichWorkspace's value to true.
- So, the RichWorkspace component was not be shown even though Text::showRichWorkspace was triggered since the property hasRichWorkspace was still false (until page refreshed)

src/helpers/files.js
luka-nextcloud: When user directly accessed an empty folder, the render() (at https://github.com/nextcloud/text/blob/main/src/helpers/files.js#L207) did not run. So, RichWorkspace component was not created.
Then, triggering `Text::showRichWorkspace` right after creating Readme.md did not work.
The RichWorkspace component was created only after any file or folder was created.
So, we have to wait some time until RichWorkspace component is created.

juliushaertl: Ah i see, that seems tricky. Let's get this in then for now. We might want to have the files app adjusted so that the header component itself is also rendered for empty file lists mid term. I only pushed a commit to use the file created event instead which seemed more clean.

juliushaertl (approved):

mejo- (approved):
Code changes look good to me :ok_hand:

juliushaertl:
@luka-nextcloud just FYI I found a reasonable workaround for updating the view without the timestamp (see last commit)

Issue #4276, opened 2023-06-09
outline div is not scrollable if height bigger than parent


hugodu69:
**Describe the bug**
when the content of the ouline is higher than the page, you cannot access the overflowed titles because it's not scrollable

**To Reproduce**
1. create a document with a lot of titles
2. Click on 'show outline'
3. try to scroll down the outline section
4. you cannot

**Expected behavior**
the ouline div should be scrollable

hugodu69:
I tried to fix it inside the navigator, and it works like that :

the structure of the divs is like that :
```
<div class="text-editor__main">
    <div class="text-editor--readonly-bar"></div>
    <div class="text-editor__content-wrapper">
        <div class="text-editor__content-wrapper__left">
            <div class="editor--outline"></div>
        </div>
        <div class="text-editor__content"></div>
    </div>
</div>
```

that we can represent like this :
```
+--text-editor__main------------------------------------------------------+
|                                                                         |
| +--text-editor--readonly-bar------------------------------------------+ |
| |                                                                     | |
| +---------------------------------------------------------------------+ |
|                                                                         |
| +--text-editor__content-wrapper---------------------------------------+ |
| |                                                                     | |
| | +--text-editor__content-wrapper__left--+ +--text-editor__content--+ | |
| | |                                      | |                        | | |
| | | +--editor--outline-----------------+ | |                        | | |
| | | |                                  | | |                        | | |
| | | +----------------------------------+ | |                        | | |
| | |                                      | |                        | | |
| | +--------------------------------------+ +------------------------+ | |
| |                                                                     | |
| +---------------------------------------------------------------------+ |
|                                                                         |
+-------------------------------------------------------------------------+
```

the current css is something like that :
```
.text-editor__main {
    overflow: auto;
}

    .text-editor--readonly-bar {
        position: sticky;
    }

    .text-editor__content-wrapper {
        display: grid;
        grid-template-columns: 100px auto 100px;
    }

        .text-editor__content-wrapper__left {
        }

            .editor--outline {
                position: fixed;
                overflow: auto;
            }

        .text-editor__content {
        }
```

the problem is that the `text-editor__content-wrapper__left` aka `ouline section` cannot have a height of the section, because its parent don't have it

so I propose a solution where the height is pass through the childs, like that :
```
/*only addition and subtraction to the css are shown*/

.text-editor__main {
    display: flex;
    flex-direction: column;
}

    .text-editor--readonly-bar {
    }

    .text-editor__content-wrapper {
[...]

PR #4967, opened 2023-11-07
fix(outline): Make sure the outline is always in foreground

mejo-:
### 📝 Summary

Add a border shadow around the outline. Also, make it overlay the editor content (e.g. on small screens). Required e.g. for Collectives.
    
Fixes: #4276

Copied from https://github.com/nextcloud/text/pull/4945#pullrequestreview-1711284684:

> Having the outline floating without borders was part of the design requirements or feedback when we introduced it, so I'd not change that as part of a bugfix at least.

@nextcloud/designers for feedback from the design team.

#### 🖼️ Screenshots

🏚️ Before Text standalone | 🏡 After Text standalone
---|---
grafik | grafik

🏚️ Before Collectives | 🏡 After Collectives
---|---
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/EditorOutline.vue
+++ b/src/components/Editor/EditorOutline.vue
@@ -61,10 +61,14 @@ export default {
 	// 204px = 50px nc header + 60px collectives titlebar + 44px menubar + 50px bottom margin
 	max-height: calc(100% - 204px);
 
+	border-radius: var(--border-radius-large);
+	box-shadow: 0 1px 10px var(--color-box-shadow);
+	background-color: var(--color-main-background);
+	z-index: 10021;
+
 	&-mobile {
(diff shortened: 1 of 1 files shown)


marcoambrosini:
I think the outline should definitely go in the right sidebar. We cannot add yet another column to collectives (we have 4). 

Screenshot 2023-11-08 at 11 21 24

nimishavijay:
Agreed with Marco, in Collectives it could go in the sidebar.
In Text, I do still think that we could keep the outline without borders, it blends seamlessly with the rest of the document that way. Otherwise it feels a bit like losing the immersiveness of a full screen text editor, what do you think? :)

mejo-:
Let's discuss this separately for collectives. The text related parts are already addressed.

PR #4973, opened 2023-11-08
Revert "fix(deps): update dependency @nextcloud/dialogs to ^4.2.1 (stable26)"

juliushaertl:
Reverts nextcloud/text#4952 as this was causing silent failures that only occur if js is built which it wasn't on that PR 🙈


mejo- (approved):

mejo-:
Thanks for looking into this @juliushaertl!

PR #4975, opened 2023-11-09
ci(cy): has npm_package_name been fixed?

max-nextcloud:
Testing if https://github.com/nextcloud/text/commit/3fa26b0702c538cab11325c9a78f499cd18913b6 has been fixed with newer versions of the job.


max-nextcloud:
https://github.com/cypress-io/github-action/issues/524 is the cypress action issue in question.

max-nextcloud:
`npm_packege_name` still has not been fixed.

PR #4997, opened 2023-11-13
 fix: Limit asset/source loader rule to inline svg embeddings from @mdi/svg

juliushaertl:
Fix checkbox icons that were broken due to using the wrong rule:

Screenshot 2023-11-13 at 09 35 58


datenangebot (approved):
looks reasonable

mejo- (approved):
While I don't fully understand the logic around webpack raw-loader, webpack rules and loading the FilePlus icon as raw SVG, the PR indeed fixes the issue at hand in my local tests.

PR #4998, opened 2023-11-14
fix: Add migration to drop conflicting index ts_session

### LINKED ISSUE #4958: Cleanup potential duplicated index
In NC 27.0.1 RC2 we had an index which was colliding with one in the Talk app. This was fixed for 27.0.1 proper (#4553), nevertheless some people already ran into issues with the upgrade to 28.0.0 Beta 1, because a potentially existing index was not removed (cf. https://help.nextcloud.com/t/first-beta-of-nextcloud-28/173501/6 and https://help.nextcloud.com/t/first-beta-of-nextcloud-28/173501/12).

juliushaertl:
Fix #4958 for instances that had the broken beta release installed.

### DIFF
--- a/composer/composer/autoload_classmap.php
+++ b/composer/composer/autoload_classmap.php
@@ -53,6 +53,7 @@ return array(
     'OCA\\Text\\Migration\\Version030201Date20201116123153' => $baseDir . '/../lib/Migration/Version030201Date20201116123153.php',
     'OCA\\Text\\Migration\\Version030501Date20220202101853' => $baseDir . '/../lib/Migration/Version030501Date20220202101853.php',
     'OCA\\Text\\Migration\\Version030701Date20230207131313' => $baseDir . '/../lib/Migration/Version030701Date20230207131313.php',
+    'OCA\\Text\\Migration\\Version030901Date20231114150437' => $baseDir . '/../lib/Migration/Version030901Date20231114150437.php',
     'OCA\\Text\\Notification\\Notifier' => $baseDir . '/../lib/Notification/Notifier.php',
     'OCA\\Text\\Service\\ApiService' => $baseDir . '/../lib/Service/ApiService.php',
     'OCA\\Text\\Service\\AttachmentService' => $baseDir . '/../lib/Service/AttachmentService.php',
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -11,7 +11,7 @@
 - **💾 Open format:** Files are saved as [Markdown](https://en.wikipedia.org/wiki/Markdown), so you can edit them from any other text app too.
 - **✊ Strong foundation:** We use [🐈 tiptap](https://tiptap.scrumpy.io) which is based on [🦉 ProseMirror](https://prosemirror.net) – huge thanks to them!
 	]]></description>
-	<version>3.9.0</version>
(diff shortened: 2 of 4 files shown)


lib/Migration/Version03009Date20231114150437.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace OCA\Text\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version03009Date20231114150437 extends SimpleMigrationStep {
```
juliushaertl, line 34: Apparently the version schema is different from what occ migration:generate suggest. It should be 3009 starting with that now would cause the migrations on new setups to run in the wrong order.

@nickvergessen Do you see any issue with continuing with the 0 padded major version? Is there any possible migration path to e.g. merge existing migration steps into one matching the new schema?

mejo-: Shouldn't this version string actually start with `030901` given the version 3.9.1?

mejo- (approved):
Looks good apart from the question about the version string.

lib/Migration/Version03009Date20231114150437.php
juliushaertl: No just major/minor:

> We recommend to use the major and minor digits of your apps version for that. This allows you to introduce a new migration in your branch for a Nextcloud version if there is already a migration path for a newer one in another branch. Since you can’t change this retroactive, we recommend to leave enough space in between and therefore map the numbers to 3 digits: `1.0.x => 1000, 2.34.x => 2034, etc.

https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html

juliushaertl: Ah it only gets executed first because I also overseen the extra minor step that I added. However looking at https://github.com/nextcloud/server/blob/6f39d82031202a2a8d1fb81ea448a2ffdc52cc29/lib/private/DB/MigrationService.php#L223-L231 i think it should be fine to just continue with our current pattern

nickvergessen: Yeah should work

composer/composer/autoload_classmap.php
```diff
@@ -49,6 +49,7 @@
     'OCA\\Text\\Migration\\ResetSessionsBeforeYjs' => $baseDir . '/../lib/Migration/ResetSessionsBeforeYjs.php',
     'OCA\\Text\\Migration\\Version010000Date20190617184535' => $baseDir . '/../lib/Migration/Version010000Date20190617184535.php',
     'OCA\\Text\\Migration\\Version030001Date20200402075029' => $baseDir . '/../lib/Migration/Version030001Date20200402075029.php',
+    'OCA\\Text\\Migration\\Version03009Date20231114150437' => $baseDir . '/../lib/Migration/Version03009Date20231114150437.php',
```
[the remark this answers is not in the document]
juliushaertl: You were actually right that existing migrations followed AABBCC (major/minor/patch) format. Adjusted that.

blizzz (approved):

PR #5017, opened 2023-11-20
fix: Get file from share link instead of user directory in case of no access

### LINKED ISSUE #5001: Public share links broken for logged in users
It seems that the public API requests fail, to be further investigated on 28

juliushaertl:
If a user opens a public share link but does not have access to the file
in question on their filesystem, text failed to render the public share
link as we always tried to get the file directly.

Steps to reproduce:
- Create a file as userA
- Share the file as share link
- Login as userB
- open the share link as userB

Fixes #5001 

### 🏁 Checklist

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -436,7 +436,11 @@ class DocumentService {
 	 */
 	public function getFileForSession(Session $session, ?string $shareToken = null): File {
 		if (!$session->isGuest()) {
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
Code changes look good, there are 2 cypress tests failed but seems like they are not related to code changes.

juliushaertl:
Yep, let me restart those

Issue #3974, opened 2023-03-22
Use list of own attachment files in AttachmentResolver


mejo-:
## Summary

Currently [`AttachmentResolver.js`](https://github.com/nextcloud/text/blob/main/src/services/AttachmentResolver.js) misses important information about attachments, which leads to a lot of complexity in resolving attachments.

To make this more straight-forward (at least for "own attachment files" - a.k.a. the ones living in `.attachments.<fileId>`), we could do the following:

* Add an API endpoint to get the list of files in the attachment folder along with their metadata (`fileId`, `fileName`, `size`, `mimetype`, `timestamp`, `hasPreview`, maybe even `davUrl` already). This endpoint would need to allow authentication both via Text session and via logged in user.
* Always request this list when loading a Text document in [`Editor.vue`](https://github.com/nextcloud/text/blob/00b8b908ec601f3541533afb5f8addc06038423e/src/components/Editor.vue#L485) (asynchronously, non-blocking).
* When resolving attachments in [`AttachmentResolver.js`](https://github.com/nextcloud/text/blob/00b8b908ec601f3541533afb5f8addc06038423e/src/services/AttachmentResolver.js#L58), look up the attachment in the retained list and use all available information from there.

## Implementation details

* The list could either be stored in vuex store or the promise of the request could be passed to `AttachmentResolver` constructor.
* When a new attachment is uploaded in Text, we either have to add it to the list in frontend, or retrieve an updated list as response and replace the list. (That might be arguments to maintain the list in vuex store :thinking:)

## Advantages

* For "own attachments" we could return better first candidates from `AttachmentResolver`.
  * Non-image attachments would get a first candidate with `type: media`, resulting in fewer requests during `<img>` loads.
  * No extra requests required for getting metadata for attachments.
* We would have the `fileId` of attachments available.
  * Allows to link to attachments in Files app (useful for #3849).
  * Allows to pass the `fileId` when emitting events for deleted attachments, useful in Collectives.

## Disadvantages

* One more backend request (at least for documents without attachments).

@juliushaertl @julien-nc @max-nextcloud for feedback.

PR #5042, opened 2023-11-27
Refactor attachment loading

mejo-:
### 📝 Summary

Introduces an API endpoint to fetch a list of native attachments (all files in `.attachments.<fileId>` folder). Frontend fetches this list when `AttachmentResolver` is initialized and uses it when resolving attachments.

The main advantage is that all metadata about native attachments is available instantaneously. This obsoletes the logic to iterate through candidates in the frontend code for determining whether it's an image or non-image attachment.

Additionally, it allows to decide in frontend what to do with non-image attachments. [When viewer is available and not in use, supports the attachment mimetype and we're not in a public share](https://github.com/nextcloud/text/blob/1916845d1485f5fba24e043a8ecf7db8f1ed98ab/src/nodes/ImageView.vue#L274-L285), we open the attachment in viewer now. Otherwise we download the attachment.

The editor API now uses `MarkdownContentEditor.vue` when `fileId` is passed along with `useSession = false`. We need the `fileId` in `MarkdownContentEditor.vue` for using the `AttachmentResolver` in a meaningful way. Also required: the backend allows to get the attachment list without a document session.

Also:
* Fixes several issues with `ShowImageModal`
* Fixes setting the filename for returned attachments
* Removes support for obsolete `text://` syntax
* Removes support for `hasPreview=true` query parameter in attachment URLs
* Removes special-handling of attachment preview URLs (`/core/preview/`)
* Removes obsolete metadata API endpoint
* Resolves: #3974

### 🚧 TODO

* [x] fallback in frontend resolver: update list if attachment is not found
* [x] generate full image/media URL in backend
* [x] generate preview (image/mediaPreview) URL in backend
* [x] get rid of `getMediaMetadataUrl` in frontend
* [x] Move to RequireDocumentSessionUserOrShareToken in all read-only endpoints of
      AttachmentController.
* [ ] Check access of user/shareToken to document in SessionMiddleware

#### additional scenarios
* [x] In public share (Editor.vue)
* [x] Read-only in public share (Editor.vue)
* [x] Without session (MarkdownContentEditor.vue)
* [x] Without session in public share (MarkdownContentEditor.vue)
* [x] Attachment is added during session (Editor.vue)
* [x] Delete and re-add attachment in session, works for other clients?
* [x] Upload several attachments at once
* [x] Test direct URL (image & media) to full URL on same instance
* [x] Test URL with `.attachments/<wrongId>`
* [-] Test `data:...` image
  * [-] Text opening in modal
* [x] Test all scenarios with missing attachments file (image & media)
* [x] Test opening image (image modal)
  * [x] In Editor.vue with session and user in Text/Viewer
  * [x] In Editor.vue with session and user in Collectives
  * [x] In Editor.vue with session and public share in Text/Viewer
  * [x] In Editor.vue with session and public share in Text standalone (single-file share)
  * [x] In Editor.vue with session and public share in Collectives
  * [x] In MarkdownEditor.vue without session with user in Collectives
  * [x] In MarkdownEditor.vue without session in public share in Collectives
* [x] Test attachment with (pdf) & without viewer support (tar.gz)
  * [x] In Editor.vue with session and user in Text/Viewer
  * [x] In Editor.vue with session and user in Collectives
  * [x] In Editor.vue with session and public share in Text/Viewer
  * [x] In Editor.vue with session and public share in Text standalone (single-file share)
  * [x] In Editor.vue with session and public share in Collectives
  * [x] In MarkdownEditor.vue without session with user in Collectives
  * [x] In MarkdownEditor.vue without session in public share in Collectives

#### questions
* [x] what is this `hasPreview=true` URL parameter about?

#### further
* [x] Delete legacy `text://` syntax support
* [x] Delete `&hasPreview=true` support
* [x] Get rid of `imageFileId()`, `getQueryVariable()`, `isSupportedImage()` in `ImageView.vue`
* [x] Error message in caption on image load error
* [ ] ensure correct order (similar to in document) in ShowImageModal
* [x] open non-image attachments in viewer/new tab
  * [x] Get rid of `internalLinkOrImage()` in ImageView.vue
  * [x] Open in public share

#### tests
* [ ] mock AttachmentController response and test AttachmentService logic
* [ ] open image in Text in viewer (->image modal)
* [ ] open image in Text standalone (->image modal)
* [ ] open PDF attachment in Text in viewer (->download)
* [ ] open PDF attachment in Text standalone (->viewer)
* [ ] open tar.gz attachment in Text in viewer (->download)
* [ ] open tar.gz attachment in Text standalone (->download)

### 🏁 Checklist


lib/Controller/AttachmentController.php
```diff
@@ -81,6 +83,27 @@ public function __construct(
...
+	public function getAttachmentList(?string $shareToken = null): DataResponse {
+		$documentId = $this->getDocument()->getId();
+		try {
+			$session = $this->getSession();
+		} catch (InvalidSessionException) {
+			$session = null;
+		}
+
+		if ($shareToken) {
+			$attachments = $this->attachmentService->getAttachmentList($documentId, null, $session, $shareToken);
```
juliushaertl, line 98: Not critical: Can we not skip the session (which should be null in this case anyways? would make it a bit clearer to read i think, maybe even using named arguments:

```suggestion
			$attachments = $this->attachmentService->getAttachmentList($documentId, shareToken: $shareToken);
```

lib/Controller/AttachmentController.php
```diff
@@ -252,7 +275,7 @@ public function getMediaFilePreview(string $mediaFileName, ?string $shareToken =
 			if ($preview['type'] === 'file') {
 				return new DataDownloadResponse(
 					$preview['file']->getContent(),
-					(string) Http::STATUS_OK,
+					$mediaFileName,
```
juliushaertl, line 278: Good catch 🙈

lib/Middleware/Attribute/RequireDocumentSessionUserOrShareToken.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace OCA\Text\Middleware\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class RequireDocumentSessionUserOrShareToken {
```
juliushaertl, line 8: Not sure how to name it differently but maybe to make it clear that it is session or user and not SesssionUser
```suggestion
class RequireDocumentSessionOrUserOrShareToken {
```

lib/Service/AttachmentService.php
```diff
@@ -215,79 +203,89 @@ private function getMediaFilePreviewFile(string $mediaFileName, File $textFile):
...
+			$name = $node->getName();
+			$attachments[] = [
+				'fileId' => $node->getId(),
+				'name' => $name,
+				'size' => Util::humanFileSize($node->getSize()),
+				'mimetype' => $node->getMimeType(),
+				'mtime' => $node->getMTime(),
+				'isImage' => $isImage,
+				'shareToken' => $shareToken,
+				'davPath' => '/' . implode('/', array_slice(explode('/', $node->getPath()), 3)),
```
juliushaertl, line 255: Not entirely sure this will always get the right path, maybe rather use getRelativePath from the user folder?
Also to clarify would be what is dav path considered for a public share link

juliushaertl (commented):
Some first feedback :)

lib/Middleware/Attribute/RequireDocumentSessionUserOrShareToken.php
mejo-: Agreed. Changed now.

lib/Service/AttachmentService.php
mejo-: Good catch. Changed now to use `userFolder->getRelativePath()` if there's a user and set `davPath` to `null` in public shares. We most likely don't have DAV access to the attachments in public shares anyway.

src/components/ImageView/ShowImageModal.vue
```diff
@@ -11,7 +11,7 @@
 		@previous="showPreviousImage"
 		@close="$emit('close')">
 		<div class="modal__content">
-			<img :src="currentImage.source">
+			<img :src="currentImage.fullUrl || currentImage.previewUrl">
```
juliushaertl, line 14: Shall we rather prefer to load the previewUrl if present?

src/services/AttachmentResolver.js
```diff
@@ -43,190 +45,64 @@ export default class AttachmentResolver {
 		this.#user = user
 		this.#shareToken = shareToken
 		this.#currentDirectory = currentDirectory
-		fileId ||= session?.documentId
-		this.#attachmentDirectory = `.attachments.${fileId}`
+		this.#documentId = fileId ?? session?.documentId
```
juliushaertl, line 48: Since we do no longer initialize the resolver without fileId, i think we can skip the second part and rather throw if no file id was passed?

juliushaertl (changes requested):
Some more comments, but look great so far already.

lib/Controller/AttachmentController.php
mejo-: It's not said that session is null in this case. In public shares, when using `Editor.vue`, we still have a session. Only when using `MarkdownContentEditor.vue`, we don't have one.

src/components/ImageView/ShowImageModal.vue
mejo-: Well, given that this is a fullscreen view of the image I thought that `fullUrl` is the better option here, no? Or do you think that we should use `previewUrl` for the fullscreen modal view? In any case the alternative can be removed as both are always set.

src/services/AttachmentResolver.js
mejo-: So far, we don't pass the fileId from `Editor.vue` yet as we already pass the `session` there, which includes the `fileId`. We could change this though and always pass in the `fileId`. In any case we can remove the `?` after `session` because we need a `fileId` - either directly provided or from the provided session.

lib/Middleware/SessionMiddleware.php
```diff
@@ -54,9 +67,31 @@ private function assertDocumentSession(ISessionAwareController $controller): voi
...
+		}
+	}
+
+	private function assertUserOrShareToken(ISessionAwareController $controller): void {
+		$documentId = (int)$this->request->getParam('documentId');
+		if (null !== $userId = $this->userSession->getUser()?->getUID()) {
+			$controller->setUserId($userId);
+			// TODO: check if user has access to document
+		} elseif ('' !== $shareToken = (string)$this->request->getParam('shareToken')) {
+			// TODO: check if shareToken has access to document
```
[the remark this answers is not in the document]
mejo-: Full ack. Implemented now in 97a80d6ec9695d45666a7f460d84bf46357348db

src/services/AttachmentResolver.js
juliushaertl: All right, then let's stick with that

src/components/ImageView/ShowImageModal.vue
juliushaertl: I think the preview should be big enough and the viewer app also always uses the preview iirc.

lib/Controller/AttachmentController.php
juliushaertl: Right, missed that part. :)

src/components/ImageView/ShowImageModal.vue
mejo-: Alright. Changed now.

src/services/AttachmentResolver.js
mejo-: Also done now.

juliushaertl (approved):

juliushaertl:
xy

PR #5054, opened 2023-11-29
fix(css): Fix print view in viewer and single-file share

mejo-:
### 📝 Summary

Also adds screenshot comparison tests for print view in viewer and single-file share.

* Resolves: #4536

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [x] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [x] Documentation is not required

PR #5063, opened 2023-11-30
fix: Hide cursor in print output

juliushaertl:
Follow up to #5054 to make tests more stable by hiding the cursor that may be more or less transparent depending on the activity state.

Also it is not much useful in print view.


mejo-:
Thanks for the follow-up. Let me push two more commits in the hope they help:

* Set `caret-color` to transparent before doing the snapshot
* One more time update the screenshots

mejo-:
But looking at the screenshot diffs it rather seems like the whole rendering is not pixel-perfect and moves a few pixels up/down left/right between runs :thinking:

mejo- (approved):

PR #5081, opened 2023-12-04
fix: cypress error

luka-nextcloud:
### 📝 Summary

* Resolves: cypress failure on #5010, update dependency debounce to v2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "@tiptap/suggestion": "^2.1.13",
         "@tiptap/vue-2": "^2.1.13",
         "@vueuse/shared": "^10.6.1",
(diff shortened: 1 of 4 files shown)


max-nextcloud (commented):
Looks like the check was introduced here: https://github.com/sindresorhus/debounce/commit/95eef872be4d812c129fcc976caddc0a3446b5e6

@luka-nextcloud Did you investigate what is causing this error message? I suspect we are using debounce in the wrong way somewhere. I think it would be good to fix this rather than ignoring the error message.

max-nextcloud (approved):
Nice one! Thanks for looking into this. The one remaining cypress error seems unrelated.

juliushaertl (approved):

Issue #4943, opened 2023-10-31
Reconnect handling


juliushaertl:
I noticed when testing #4938 that there seems to be a case when a connection is offline long enough that the reconnect is not working as expected.

When trying to address the issue there by dropping the disconnected websocket provider with the following patch I noticed that after reconnecting the offline client would no longer send changes while still picking up the ones from the other clients.

```diff
diff --git a/src/components/Editor.vue b/src/components/Editor.vue
index 407004d99..879928d03 100644
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -423,6 +423,7 @@ export default {
                reconnect() {
                        this.contentLoaded = false
                        this.hasConnectionIssue = false
+                       this.$providers.forEach(p => p.destroy())
                        this.close().then(this.initSession)
                        this.idle = false
                },
```

juliushaertl:
Actually seems like after the reconnect the step is not pushed at all so we have a diverging state between both clients

Failing step:
<img width="1279" alt="Screenshot 2023-10-31 at 15 23 16" src="https://github.com/nextcloud/text/assets/3404133/f86f1ce0-da0f-4314-95b8-de0ba7711f25">

First successful push after reconnect:
<img width="1278" alt="Screenshot 2023-10-31 at 15 23 46" src="https://github.com/nextcloud/text/assets/3404133/652231f2-0686-42f4-92d8-7c85a6293d2e">

juliushaertl:
https://github.com/nextcloud/text/pull/5082 was an attempt but unclear how it could have an effect, needs further debugging.

- [ ] Clarify state behaviour of Editor.vue and related components (sync service, connection, websocket polyfill) in the lifecycle of editing especially with reconnect, what objects are reused, recreated, are there any leftovers still active?
- [ ] @max-nextcloud Looking into preserving the queue from one websocket polyfill instance to another
- [ ] Try commenting out the code part that makes the editor read only for easier debugging
- [ ] See if applying the document state again after reconnect causes troubles (maybe rather just sync the steps instead?)

juliushaertl:
Idea:

- for easier debugging try to add buttons locally to disconnect or reconnect in editor.vue

PR #5082, opened 2023-12-04
fix: Do not overwrite current local version on reconnect

juliushaertl:
Experimenting a bit with reconnect behaviour i noticed an issue that might cause disconnected session states.

Two separate sessions A and B
- A and B are editing
- A network goes offline (sometimes i needed to also quickly type something as A after setting the network offline)
- B continues typing
- A waits until the first auto close request gets blocked in the network console
- A network goes online again
- Reconnect happens

Now I observed that the last sync message had a different version to request steps compared to the first of the new session which turned out to be the case because we always overwrite the version with the docStateVersion on opening a session in the frontend, however in the reconnect case we already have a different state as we do not start from scratch.

Contributes to #4943 

- [ ] Check if there is a full recreate approach possible as a quick solution
- [ ] Just refetching the steps should not have any effect here as the new document state should already contain those
  - Maybe it is more about the local steps that then conflict with the new document state?
- [ ] check how applying the document state influences the existing y.js doc
- [ ] Figure out some visual indicator for beta versions if something goes wrong and one should report the browser console logs

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -377,6 +377,7 @@ export default {
 
 			this.listenSyncServiceEvents()
 
+			this.$providers.forEach(p => p?.destroy())
 			const syncServiceProvider = createSyncServiceProvider({
 				ydoc: this.$ydoc,
 				syncService: this.$syncService,
--- a/src/services/SyncService.js
+++ b/src/services/SyncService.js
@@ -111,7 +111,12 @@ class SyncService {
 			return
 		}
 		this.backend = new PollingBackend(this, this.connection)
-		this.version = this.connection.docStateVersion
+
+		// If we reconnect we may already have a version set so we should keep that
+		if (this.version === null) {
+			this.version = this.connection.docStateVersion
+		}
+
 		this.emit('opened', {
 			...this.connection.state,
 			version: this.version,


juliushaertl:
To check additionally if the local document state gets somehow set still on reconnect as at least the changes from other still online clients occur fine

juliushaertl:
> To check additionally if the local document state gets somehow set still on reconnect as at least the changes from other still online clients occur fine

Oh indeed, sync service emits loaded the document state is just applied in https://github.com/nextcloud/text/blob/0777cb1f9c10f87ca2987f672e159dff0ace569e/src/components/Editor.vue#L488

@max-nextcloud I'm wondering if that could cause additional troubles if we already have an existing YDoc? We may just want to skip this then and fully rely on catching up with steps during sync.

Additionally we may want to saveguard this to actually do a full reload if the yjs history has been reset in the meantime I guess.

max-nextcloud:
If I remember correctly my thinking was... the doc state will include all changes to the document up to the version send out during the (re-)connect. So if the client applies them it's up to sync until that version and can start from there.

If this is still the same session I don't see why this would not work. If it's a different session then we'd really need to clear the local state completely and then reconnect.

Issue #4784, opened 2023-09-08
Support for Markdown in Text templates


fabricemous:
**Is your feature request related to a problem? Please describe.**
When using Text templates consisting of Markdown, the formatting is not parsed and thus the template is lacking formatting support. 

**Describe the solution you'd like**
I would like to see Text templates use the applied Markdown of the Text template. If no Markdown is used you can assume it is plain text. 

**Describe alternatives you've considered**
I am using a little text expander solution which uses a shortcut. But I prefer the use of Text templates. 

**Additional context**
I have no further context. Discussed this with @juliushaertl and he is aware of this request.

juliushaertl:
Thinking more about it I'd say it is fine if we always consider smart picker output as markdown.

@luka-nextcloud Maybe you can check that to make sure that text inserted through the smart picker is always passed through markdownit? We do something similar already when pasting text in the Markdown tiptap extension.

https://github.com/nextcloud/text_templates is the app mentioned that can be useful for testing.

PR #5106, opened 2023-12-08
feat: check and apply markdown for smart picker output

luka-nextcloud:
### 📝 Summary

* Resolves: #4784

Always consider smart picker output as markdown.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Suggestion/LinkPicker/suggestions.js
+++ b/src/components/Suggestion/LinkPicker/suggestions.js
@@ -24,6 +24,7 @@ import LinkPickerList from './LinkPickerList.vue'
 import { searchProvider, getLinkWithPicker } from '@nextcloud/vue/dist/Components/NcRichText.js'
 import menuEntries from './../../Menu/entries.js'
 import { getIsActive } from '../../Menu/utils.js'
+import markdownit from '../../../markdownit/index.js'
 
 const suggestGroupFormat = t('text', 'Formatting')
 const suggestGroupPicker = t('text', 'Smart picker')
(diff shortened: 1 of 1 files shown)


src/components/Suggestion/LinkPicker/suggestions.js
```diff
@@ -34,6 +35,31 @@ const filterOut = (e) => {
 
 const important = ['task-list', 'table']
 
+const hasMarkdownSyntax = (content) => {
+	// Regular expressions for common Markdown patterns
+	const markdownPatterns = [
+		/\*\*.*?\*\*/, // Bold: **text**
+		/\*.*?\*/, // Italics: *text*
+		/\[.*?\(.*?\)/, // Links: [text](url)
+		/^#\s.*$/, // Headings: # text
```
juliushaertl, line 44: Would be good to also cover headings of other levels h2-h6

src/components/Suggestion/LinkPicker/suggestions.js
```diff
@@ -34,6 +35,31 @@ const filterOut = (e) => {
...
+		/- .*/, // Unordered list: - item
```
juliushaertl, line 45: Markdown also supports `+` and `*` as list entries

src/components/Suggestion/LinkPicker/suggestions.js
```diff
@@ -34,6 +35,31 @@ const filterOut = (e) => {
...
+		/\d\..*/, // Ordered list: 1. item
```
juliushaertl, line 46:
```suggestion
		/^\s\d\..*/m, // Ordered list: 1. item
```

src/components/Suggestion/LinkPicker/suggestions.js
```diff
@@ -34,6 +35,31 @@ const filterOut = (e) => {
...
+		/^>\s.*/, // Blockquote: > text
```
juliushaertl, line 47:
```suggestion
		/^>+\s.*/, // Blockquote: > text
```

src/components/Suggestion/LinkPicker/suggestions.js
juliushaertl, line 45:
```suggestion
		/^\s*[+-*]\s.*/m, // Unordered list: - item
```

juliushaertl (changes requested):
A few suggestions for detecting the markdown syntax (I know we can't cover everything there), but otherwise this seems fine and works well 👍

mejo-:
Maybe I'm missing something here, but why do we have to detect markdown manually at all, instead of just doing `isValidMarkdown()` and always sending the content through `markdownit.render()` if that succeeds?

juliushaertl:
Hm,  unsure about that. It might be fine to just always pass it to markdown it indeed.

luka-nextcloud:
I have tried this way and faced an issue. The markdown-it render function always adds some newline character to content. So, If input value is an url, the inserted content will be a normal string instead of a link (enclose it in an angle brackets).

juliushaertl (approved):

PR #5126, opened 2023-12-11
Preserve step queue during reconnect

max-nextcloud:
### 📝 Summary

During a network disconnect
Yjs-websocket notices missing awareness messages and closes the connection.
In this case the remaining steps can also not be send out and the request fails.
Preserve the queue so we can use it once the network is back up.

### 🏁 Checklist

### DIFF
--- a/src/services/WebSocketPolyfill.js
+++ b/src/services/WebSocketPolyfill.js
@@ -125,7 +125,9 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 
 		#sendRemainingSteps() {
 			if (this.#queue.length) {
+				let outbox = []
 				return syncService.sendStepsNow(() => {
+					outbox = [...this.#queue]
 					const data = {
 						steps: this.#steps,
 						awareness: this.#awareness,
@@ -136,6 +138,7 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 					return data
 				})?.catch(err => {
 					logger.error(err)
+					this.#queue = [...outbox, ...this.#queue]
 				})
 			}
 		}
--- a/src/services/WebSocketPolyfill.js
+++ b/src/services/WebSocketPolyfill.js
@@ -83,21 +83,20 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 			// data.forEach(logStep)
 
 			queue.push(...data)
-			let outbox = []
+			let outbox
 			return syncService.sendSteps(() => {
-				outbox = [...queue]
 				const data = {
 					steps: this.#steps,
 					awareness: this.#awareness,
 					version: this.#version,
 				}
-				queue = []
+				outbox = queue.splice(0, queue.length)
 				logger.debug('sending steps ', data)
 				return data
 			})?.catch(err => {
 				logger.error(err)
-				// try to send the steps again
-				queue = [...outbox, ...queue]
+				// Prefix the queue with the steps in outbox to send them again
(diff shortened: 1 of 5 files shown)


src/services/WebSocketPolyfill.js
```diff
@@ -124,18 +123,21 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
...
 						awareness: this.#awareness,
 						version: this.#version,
 					}
-					this.#queue = []
+					queue = []
 					logger.debug('sending final steps ', data)
 					return data
 				})?.catch(err => {
 					logger.error(err)
+					queue = [...outbox, ...queue]
```
mejo-, line 140: Interesting, this is an additional fix, right? It ensures that steps get reapplied in case of caught errors. Let's hope that this and the persistent queue solve synchronization issues :pray:

mejo- (approved):
Looks good to me, thanks!

src/services/WebSocketPolyfill.js
```diff
@@ -83,32 +82,32 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
...
 			return syncService.sendSteps(() => {
-				outbox = [...this.#queue]
+				outbox = [...queue]
 				const data = {
 					steps: this.#steps,
 					awareness: this.#awareness,
 					version: this.#version,
 				}
-				this.#queue = []
+				queue = []
```
juliushaertl, line 94: I don't think this works as expected yet. After a reconnect the first push only contains the initial steps. I suspect that due to assigning a new array the original one in Editor.vue is not adjusted but a local state is introduced instead.

This patch is quite helpful for testing:

(first change avoids the old provider leaking error events, second keeps the editor editable when disconnected to produce some steps)

```
diff --git a/src/components/Editor.vue b/src/components/Editor.vue
index 6b4cce517..cad543a92 100644
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -378,6 +378,8 @@ export default {

                        this.listenSyncServiceEvents()

+                       this.$providers.forEach(p => p?.destroy())
+                       this.$providers = []
                        const syncServiceProvider = createSyncServiceProvider({
                                ydoc: this.$ydoc,
                                syncService: this.$syncService,
@@ -572,7 +574,7 @@ export default {

                onError({ type, data }) {
                        this.$nextTick(() => {
-                               this.$editor?.setEditable(false)
+                               //this.$editor?.setEditable(false)
                                this.emit('sync-service:error')
                        })
```

juliushaertl (changes requested):

src/services/WebSocketPolyfill.js
max-nextcloud: Ahh... good catch! Thanks for testing.
This should be easy to fix.

src/services/WebSocketPolyfill.js
max-nextcloud: Yes... that's the idea. Found some additional issues with this approach:
* As julius pointed out the queue variable get's overwritten instead of changing the underlying array.
* The sending is async so between the clearing of the queue and refilling it it's empty for some time. Fixed that by using the opposite approach - everything stays in the queue until it was successfully send.

src/services/WebSocketPolyfill.js
max-nextcloud: Included the first part in the latest commit as discussed during todays team call.

juliushaertl (approved):
Tested and works

PR #5133, opened 2023-12-14
debug: run only attachment spec and debug spec

max-nextcloud:


juliushaertl:
Thanks for debugging. That was what lead to https://github.com/nextcloud/text/pull/5142 right? Shall we close this one then?

max-nextcloud:
> Thanks for debugging. That was what lead to #5142 right? Shall we close this one then?

yes. Let's close this one. I can reopen if i want to create further logs.

PR #5142, opened 2023-12-17
test(cy): skip test failing due to server issue

max-nextcloud:
https://github.com/nextcloud/server/issues/42306
causes the DELETE request to fail with 423 Locked.

### 🏁 Checklist

- fixes tests and does not need docs.


juliushaertl (approved):
Please open a revert PR right after merging so we keep track of the skipped test.

We probably could debug this further on CI with https://github.com/nextcloud/server/blob/master/config/config.sample.php#L2188-L2197

PR #5150, opened 2023-12-19
Configure renovate bot reviewers and auto merge

max-nextcloud:
* Make @mejo- and @max-nextcloud reviewer of renovate PRs
* Set them to automerge

### DIFF
--- a/.github/workflows/renovate-approve-merge.yml
+++ b/.github/workflows/renovate-approve-merge.yml
@@ -1,0 +1,49 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud/.github
+# https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
+
+name: Auto approve renovate PRs
+
+on:
+  pull_request_target:
+    branches:
+      - main
+      - master
+      - stable*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: renovate-approve-merge-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  auto-approve-merge:
+    if: github.actor == 'renovate[bot]'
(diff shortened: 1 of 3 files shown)


.github/workflows/renovate-approve-merge.yml
```diff
@@ -0,0 +1,49 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # GitHub actions bot approve
+      - uses: hmarr/auto-approve-action@v3
+        if: startsWith(steps.branchname.outputs.branch, 'renovate/')
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # Enable GitHub auto merge
+      - name: Auto merge
```
juliushaertl, line 45: Difference to the other action (for dependabot) is that it has a limit on only minor version updates https://github.com/nextcloud/text/blob/8a8d84c5c9b27e20aa7e2a97b6d09517019b1563/.github/workflows/dependabot-approve-merge.yml#L39

I would feel ok to have this for all in text due to our test coverage but not sure about making this a general action template then with that.

juliushaertl (approved):

.github/workflows/renovate-approve-merge.yml
max-nextcloud: I think major version bumps are really rare. But I might be wrong. How about we start with limited to minor and then remove the restriction if we they become an issue?

juliushaertl: I think the new actions don't know about the renovate version numbers so we cannot limit unless we find (or write an action) to check that, but fine as it is for now I'd say.

max-nextcloud: Ah.. now i got it. So now we don't have the limit. Fine with me aswell.

PR #5151, opened 2023-12-19
[stable27] preserve websocket provider queue during reconnects

max-nextcloud:
### 📝 Summary

* Backports #5126


max-nextcloud:
We will need to investigate the sync spec failure in cypress. That might be related.

juliushaertl:
I know one is also sometimes failing on other branches, but would be good to check, yes :)

max-nextcloud:
> I know one is also sometimes failing on other branches, but would be good to check, yes :)

grafik
Looks like it's been failing quite consistently on `stable27` for a while now.

max-nextcloud:
The failures look exactly the same as before. Opened #5154 to investigate further. Let's move on with this.

juliushaertl (approved):

juliushaertl:
@max-nextcloud Can you check the backport to 26?

Issue #3691, opened 2023-01-21
Links in Text files will open 2x


softweb-dev:
https://help.nextcloud.com/t/links-in-text-files-will-open-2x/149729/1

And also 

[linkissue.webm](https://user-images.githubusercontent.com/48855596/213891110-547f3b19-2d52-407a-bc48-7e4097ce77b0.webm)

luka-nextcloud:
@softweb-dev I am not able to reproduce the first issue. Please take a look at the video:

https://user-images.githubusercontent.com/89908051/214514751-21f5a704-2ac7-49be-a80d-22f752fee98c.mp4

softweb-dev:
html,body{background-color:#fff;color:#333;line-height:1.4;font-family:sans-serif,Arial,Verdana,Trebuchet MS;}Issue comes only on shared link. Bud share is now not more possible since latest version 25.0.3.

luka-nextcloud:
@softweb-dev  In that case, we have nothing to fix.

juliushaertl:
@luka-nextcloud Still needs to be fixed also if it is not working due to another bug ;)

@softweb-dev Please fill out the issue template so we know which version this happens with and which browser/OS you are using. Also knowing the exact steps taken for creating the share link would be good..

juliushaertl:
I could not reproduce this issue anymore also not on share links. @softweb-dev Can you check this again with the latest release?

softweb-dev:
https://eb12.de/2.webm

Issue #2198, opened 2022-02-23
Middle click on links must not insert text


PVince81:
**Describe the bug**
Middle clicking on links still inserts text

**To Reproduce**
1. Create a text document with a link
2. Middle click on the link to open to a new tab
3. Come back to the editor's tab

**Expected behavior**
New tab is opened, no text inserted.

**Screenshots**

**Client details:**
 - OS: openSUSE Tumbleweed
 - Browser: Firefox
 - Version: 97.0
 - Device: desktop

**Server details:**
c.nc.com

luka-nextcloud:
Middle click does nothing - neither insert a text nor open a new tab.
It should open a new tab.

mejo-:
@luka-nextcloud not sure what operating system you're using. The problem with middle-click inserting text is probably on Linux only. There, middle click inserts content from a buffer that contains the last selection. I can reproduce the issue and regularly run into it myself :see_no_evil:

mejo-:
In latest Text (both NC 25.0.2 and NC 26 development branch), middle clicking on links no longer opens the links on Linux. Instead, it only inserts text from the selection clipboard. I don't know which code change introduced that change of behaviour, though. Maybe some upstream changes in Tiptap?

I wonder whether opening links in new tabs also doesn't work anymore on non-Linux operating systems. Maybe people with Windows and MacOS X can comment.

To be honest, I would be fine with this behaviour. Middle-clicking for opening in new Tab is probably a niche featue anyway and those who use it will learn quickly that it doesn't work in Text. It's kindof useless in text anyway as we open external links in new tabs/windows per default.

For me, this solves the issue at hand (other than expected, admittedly :wink:). Waiting for others to confirm before closing the issue though.

juliushaertl:
On macOS middle click also does nothing when clicking a link.

Could be related to introducing the click preventer in 36699030e9

Without that custom behavior I could imagine an upstream issue like https://github.com/ueberdosis/tiptap/issues/3302

mejo-:
> Could be related to introducing the click preventer in https://github.com/nextcloud/text/commit/36699030e90422f7007887f78172116d936eb281

Reverting this commit doesn't change the behaviour for me.

In any case, given that the changed behaviour fixed the issue at hand, shall we close this issue?

blizzz:
I still happens to me all the time. When I intend to open links in new tabs, I always automatically hit middle-click.

[...]

PR #5158, opened 2023-12-23
Link bubble menu to preview and edit links

mejo-:
### 📝 Summary

Opens a link bubble when clicking on a link and removes in-document link previews.

* Opens link bubble when clicking on a link or navigating inside a link mark.
* Displays a link preview in the bubble. Loads a fallback if no preview can be loaded.
* Actions inside the bubble: copy link to clipboard, edit link URL, remove link mark.
* Anchor links (href starting with `#`) don't open a link bubble but navigate to the anchor directly.
* No longer display link previews in the document.
* <kbd>Ctrl</kbd> + click and middle click on a link doesn't open link bubble. Instead opens the link directly in `target="_blank"`.
* <kbd>Esc</kbd> closes the link bubble.
* Remove the possibility to customize link click handlers. Instead, we use the reference widgets now, and they should provide their own click handlers if desired.
* Resolves: nextcloud/collectives#845
* Resolves: #2198
* Resolves: #3691

#### 🖼️ Screenshots

| Overview |
|---|
| overview |

Copy link to clipboard | Edit link| Remove mark | Edit mode
--- | --- | --- | ---
clipboard | edit1 | remove | edit

| Screencast |
| --- |
| recording1 |

### 🚧 TODO

* [x] Display link URL if no widget is loaded
* [x] Display link href instead of preview when `!isLoggedIn`
* [x] Read-only support
* [x] Remove link preview from paragraph node view
* [x] Refactor link handling callbacks
* [x] Manual testing
  * [x] Check on mobile
  * [x] Check in public shares
  * [x] Check with references to collective page
* [x] Cypress tests (require #5281 to pass)
* [x] Optional: Display link text in header of link bubble
* [x] Optional: Allow to set text to link title

### 🏁 Checklist


juliushaertl:
Looks very nice already. I stumbled over the following two issues that would be worth to check and address in the same run:

- https://github.com/nextcloud/text/issues/2198
- https://github.com/nextcloud/text/issues/3691

src/marks/Link.js
```diff
@@ -63,30 +68,61 @@ const Link = TipTapLink.extend({
...
+						},
+						// Prevent paste on middle click in links (Linux only)
+						paste: (view, event) => {
+							if (event.target.closest('a') && this.storage.nopaste) {
+								event.stopPropagation()
+								event.preventDefault()
+								event.stopImmediatePropagation()
+								this.storage.nopaste = false
+							}
+						},
```
max-nextcloud, line 109: So the `preventDefault()` in `pointerup` is not enough?

That's a pitty as this seems brittle. For example what would happen if i use middle click to open a link in a new tab on windows and then later on try to paste something into a link? Wouldn't `this.storage.nopaste` still be true?

max-nextcloud (commented):
The overall approach looks good to me. I added some minor comments but nothing blocking.

src/marks/Link.js
mejo-: Unfortunately the `preventDefault()` in `pointerup` is not sufficient according to my tests. But you have a fair point. Mhh, have to think more about this.

src/components/Link/LinkBubbleView.vue
```diff
@@ -0,0 +1,308 @@
...
+		<!-- link edit form -->
+		<div v-if="isEditable && edit" class="link-view-bubble__edit">
+			<NcTextField name="newHref"
+				:label="t('text', 'URL')"
+				:value.sync="newHref"
+				@keypress.enter.prevent="updateLink" />
+			<NcTextField name="text"
+				:label="t('text', 'Text')"
+				:value.sync="newText"
+				@keypress.enter.prevent="updateLink" />
```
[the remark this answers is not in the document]
mejo-: Agreed, removed now.

src/marks/Link.js
mejo-: The best option that comes to my mind is to set `this.storage.nopaste = true` only on Linux (by checking the user agent). What do you think @max-nextcloud?

mejo-: I implemented this now: https://github.com/nextcloud/text/pull/5158/commits/d27d2fa91e421f95cff0abf7f43b5266f4b70b45#diff-90d862ff7cd51c7b18035dbb20c02bc177926bf7224193d0a58b5f870f7a8adbR96-R98

mejo-:
The PR is finally ready for review. Cc @nextcloud/designers.

Cypress tests should pass once #5281 got merged.

marcoambrosini:
Still think that it's a better to allow editing the name within this dialog than relying on editing the string. Mainly because there's ambiguity when it comes to add spaces to the end of the name.

juliushaertl:
I think this is something we should still address differently. We cannot easily just have a textbox for the link title as it may also contain other formatting marks like bold/italic which we could not preserve through a simple input

marcoambrosini:
> may also contain other formatting

I think it's totally fine to discard formatting within the link and use whatever the first character style is for the whole string. Still much less of a drawback than not being able to append a word to the link name IMO. 

Here's an example of google 

https://github.com/nextcloud/text/assets/26852655/64580635-978c-4a55-9eee-fcc006650a1c

max-nextcloud:
> Still much less of a drawback than not being able to append a word to the link name IMO.

I'm not sure why this is not working. I can append a word to an italic text fine - and from the tiptap perspective they are both marks.

max-nextcloud (approved):

juliushaertl:
The discussion about adding a description field should not block this PR imo, as it would be just another additional new feature. Maybe we can move that to an issue to discuss as a potential follow up.

juliushaertl:
> I'm not sure why this is not working. I can append a word to an italic text fine - and from the tiptap perspective they are both marks.
> update: looks like we inherit this behavior from both tiptap and prosemirror.

Just for reference, quoting prosemirror docs:

> By default, marks are inclusive, meaning that they get applied to content inserted at their end (as well as at their start when they start at the start of their parent node). For link-type marks, this is usually not the expected behavior, and the [inclusive](https://prosemirror.net/docs/ref/#model.MarkSpec.inclusive) property on the mark spec can be set to false to disable that behavior.

And tiptap seems to auto set that if autolink is enabeld
https://github.com/ueberdosis/tiptap/blob/main/packages/extension-link/src/link.ts#L88

nimishavijay:
Agreed with both that the link text can be edited in the popup, and that it can be a follow up PR :) This PR looks really great! Only some minor details for feedback:

- [ ] Indeed the image preview can be smaller, as at this size the text will likely be ellipsized most of the time. How about 64px? :) 
- [ ] There can also be more space between the image preview and the edges of the popup, so some margin can be added
- [ ] The border radius of the image preview should be nicely aligned with border radius of the popup, currently it's too less
- [ ] When you are editing the link, the right and left spacing seems pretty tight. I suggest some left and right margin be added to the text field 
- [ ] In the screenshots there is no text on the left of the buttons, but in the screencast it's there. Is that intentional or am I missing something? I suggest that there doesn't need to be text, but instead there can be an "Open link" secondary button (what do you think @marcoambrosini ?) :)

mejo-:
The failing Cypress test in workspaces.spec.js is because in CI we have `http://localhost:8080` or something alike as local URL and `NcReferenceList` doesn't regard this as a valid URL. The following PR aims to change this: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5176

mejo-:
Thanks for your feedback @marcoambrosini and @nimishavijay.

>  Indeed the image preview can be smaller, as at this size the text will likely be ellipsized most of the time. How about 64px? :)

Given that this would need to happen in the `NcReferenceWidget` component in `nextcloud-vue`, I'd suggest to make this a follow-up topic.

>     In the screenshots there is no text on the left of the buttons, but in the screencast it's there. Is that intentional or am I missing something? I suggest that there doesn't need to be text, but instead there can be an "Open link" secondary button (what do you think @marcoambrosini ?) :)

Addin an "open link" button is a bit more technically involved. The link preview from `NcReferenceWidget` implementations might have custom link handling implemented. Like opening files in the viewer, or opening other pages of a collective via re-routing in vue. Currently there is no easy way to "copy" this behaviour outside the widgets. So I'd rather move this to a follow-up as well.

marcoambrosini:
> what do you think @marcoambrosini

I think that there's no need for such button because there's a link element just below it

PR #3799, opened 2023-02-16
Refactor Polling Backend and Sync service

max-nextcloud:
### 📝 Summary

Refactor the Polling Backend and Synchronization service

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing ... let's see
- [x] Documentation is not required

PR #5188, opened 2024-01-02
cleanup(SyncService): this.sessions and this.steps

max-nextcloud:
### 📝 Summary

Both were kept up to date but never used as far as i can tell.

Extracted from #3799 

### 🏁 Checklist

- Documentation is not needed

### DIFF
--- a/src/services/SyncService.js
+++ b/src/services/SyncService.js
@@ -77,9 +77,6 @@ class SyncService {
 		this._api = new SessionApi(options)
 		this.connection = null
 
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
Good catch, also couldn't see any usage when double checking.

juliushaertl:
@max-nextcloud Labels and milestone would be good to set. That way it is much easier to collect changes for either a changelog or aggregating what's new on releases.

Issue #4233, opened 2023-06-01
Translate popup does not open in mobile apps


thgoebel:
**Describe the bug**

The Translate popup never appears in the Rich Text Editor (Nextcloud Text) in the mobile apps.

Both in the Files Android+iOS apps but also in the Nextcloud Notes Android+iOS apps.
I'm opening this here centrally, instead of creating 4 separate issues for (Files, Notes) x (Android, iOS). Feel free to move it.

**Steps To Reproduce**

1. Open a document with the Rich Text editor in Files or Nextcloud Notes in the Android or iOS app.
2. Click on ...
3. Click on Translate

**Expected behavior**

The Translate popup should open.

**Actual behavior**

Nothing happens.

**Screenshots**

n/a

**Client details:**

Device 1:
 - OS: iPadOS 16.5
 - Device: iPad Pro 2020
 - Files app: 4.8.0.5
 - Notes app: 4.1.7

Device 2:
 - OS: Android 13
 - Device: OnePlus 5
 - Files app: 3.24.1
 - Notes app: 4.0.0

<details>
<summary>Server details</summary>

**Text app version:** 3.7.2

**Translate app version:** 1.1.3

**Operating system:**

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** 26.0.2

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
n/a
```

#### Browser log
```
n/a
```

</details>

blizzz:
switching triage label as i do not have either android nor ios devices

juliushaertl:
Endpoints only became public with https://github.com/nextcloud/server/commit/e5d0ff0c19c7806129057856133c7183973882b5 so might be that this is only an issue on 26.

thgoebel:
I've upgraded my instance to 27.0.0 and this is still an issue.

Hephi2:
There seem to be problems with loading the state. 
The translation modal opens when adding the fallback value `false` in line https://github.com/nextcloud/text/blob/6f12481acc774f7d4e8cb70347ec2585d070dbbb/src/components/Modal/Translate.vue#L94
However, all `loadState` commands seem to fail and just return the fallback value so translation is not possible.
This needs to be further inverstigated.

PR #5197, opened 2024-01-05
fix: translate popup does not open in mobile apps

luka-nextcloud:
### 📝 Summary

* Resolves: #4233

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/DirectEditing/TextDirectEditor.php
+++ b/lib/DirectEditing/TextDirectEditor.php
@@ -25,10 +25,10 @@ namespace OCA\Text\DirectEditing;
 
 use OCA\Text\AppInfo\Application;
 use OCA\Text\Service\ApiService;
+use OCA\Text\Service\InitialStateProvider;
 use OCP\AppFramework\Http\NotFoundResponse;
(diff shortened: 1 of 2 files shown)


lib/DirectEditing/TextDirectEditor.php
```diff
@@ -159,6 +164,14 @@ public function open(IToken $token): Response {
 				'session' => \json_encode($session->getData())
 			]);
 			$this->initialStateService->provideInitialState('directEditingToken', $token->getToken());
+			$this->initialStateService->provideInitialState(
```
juliushaertl, line 167: Maybe we can just reuse the existing [InitialStateProvider](https://github.com/nextcloud/text/blob/130a4ec5fe5f24e0e27a63e459eeea22472770e5/lib/Service/InitialStateProvider.php#L20) class for this?

The textprocessing state would also be useful for direct editing.

juliushaertl (approved):

PR #5212, opened 2024-01-08
fix: Properly copy selection as markdown to the plaintext clipboard

juliushaertl:
Using clipboardTextSerializer we can copy the current selection as markdown and don't need the extra keyboard shortcut, which was broken anyways as it always selected the full text

Steps to reproduce:
- Select a text range
- Copy using Ctrl-C
- Paste to a plaintext editor, github issue or https://evercoder.github.io/clipboard-inspector/

Fix https://github.com/nextcloud/text/issues/5213

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -111,7 +111,7 @@ import { createEditor, serializePlainText, loadSyntaxHighlight } from './../Edit
 import { createMarkdownSerializer } from './../extensions/Markdown.js'
 import markdownit from './../markdownit/index.js'
 
-import { CollaborationCursor, Keymap } from '../extensions/index.js'
+import { CollaborationCursor } from '../extensions/index.js'
 import DocumentStatus from './Editor/DocumentStatus.vue'
 import isMobile from './../mixins/isMobile.js'
 import setContent from './../mixins/setContent.js'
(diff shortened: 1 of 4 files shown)


src/components/Editor.vue
```diff
@@ -371,8 +371,8 @@ export default {
 				filePath: this.relativePath,
 				forceRecreate: this.forceRecreate,
 				serialize: this.isRichEditor
-					? () => createMarkdownSerializer(this.$editor.schema).serialize(this.$editor.state.doc)
-					: () => serializePlainText(this.$editor),
+					? (content) => createMarkdownSerializer(this.$editor.schema).serialize(content ?? this.$editor.state.doc)
+					: (content) => serializePlainText(content ?? this.$editor.state.doc),
```
max-nextcloud, line 375: I think you are not making use of this (*anymore?*). Might be enough to keep the old less complex version - or use this in the `clipboardTextSerializer` below.

max-nextcloud (commented):
Have not tested it yet. Code changes look good. Nice to see more code getting removed than added. :)

I wonder what would happen with non-markdown files. I think the code in the editor factory might serve the purpose of extracting plaintext from the codeblock we currently wrap other files such as source code in.

src/components/Editor.vue
juliushaertl: We don't need to use it in the clipboardTextSerializer as we only overwrite it within the markdown extension. Otherwise tiptap will make use of this and i checked that both partial and full selection work as expected with plain text files.

Now we could just skip the parameter for the plain text serialize but I'd rather like to have consistent callback signatures.

max-nextcloud (approved):
Tested and works like a charm.

susnux:
While this works this creates a lot of UX issues, as even if you just copy parts of code it always adds the code tags.
You can no longer e.g. copy email addresses, passwords (you should not do this but thats a different story), shell commands etc.

Could we maybe somehow instead put both into the clipboard? `text/plain` and `text/markdown`?

Issue #5216, opened 2024-01-09
[BITV]: Remove unneeded roles `menuitem` and `menu`


JuliaKirschenheuter:
In most toolbar buttons wrong roles have been assigned. To correct the roles please pay attention on:

- [x] All buttons and div's which are direct children of the toolbar should not have a `role="menuitem"`
- [x] All buttons which contains menu inside (like heading) should not have any additional roles. All menu roles have been already implemented correct inside of `NcActions` component 
- [ ] All selected buttons should not have `aria-selected="true"` but should have instead `aria-pressed="true"` and `aria-pressed="false"` on unpressed state (requires `@nextcloud/vue` update) (note that `aria-pressed` can't be on a `div class="entry-action action-item"`) 
- [ ] All popovers should have correctly used `aria-haspopup` (may require `@nextcloud/vue` update)

![Screenshot from 2024-01-09 18-35-36](https://github.com/nextcloud/text/assets/6078378/fb0fffeb-ad85-40ea-8cfc-e9e7552d4aa6)

JuliaKirschenheuter:
Dear @juliushaertl 

could you please explain for us a difference between `aria-active="false"` on "B" or "I" and "Undo-button"? 

Does it means if `aria-active="false"` have "I" button then "I" button is not pressed right now, right? 

But what does it mean for button "Undo"? "Undo" button can't be active.

ShGKme:
The problem here is that boolean property `active` in the action button `state` in the mixin is used to define the `active` state for all the buttons, not only button that may have active state. 

We have buttons:
1. **Toggle buttons** like "Bold", "Italic". They can be pressed (active=true) and not pressed (active=false). And they should always have `aria-pressed` attribute with the current `active` state.
2. **Plain buttons** like "Undo" and "Redo". They should have neither `active` not `non-active` state.
3. **Menu trigger button** like "Heading" on a large screen and pickers. They also should not have active state.

Having boolean `active` state on all the action buttons is not enough to determine which button is not pressed (not active) and which cannot be active at all.

`active === false` at the same time may mean:
- This is an unpressed toggle button
- This is not a toggle button

The question is, is there a way to determine that a button is a toggle button and has an active state?

If there is no, we can add it to the action description.

ShGKme:
I guess this is `isActive` property.

juliushaertl:
Took me a bit to refresh my overview on the menu code parts.

[...]

PR #5218, opened 2024-01-09
fix(menubar): menubar and table menus a11y

ShGKme:
## 📝 Summary

* Resolves: #5216 
* Resolves: #3911

Currently `aria-*` and `role` attrs in the `MenuBar` are set incorrectly. These attributes are already covered by `@nextcloud/vue`. This PR removes setting attributes from the Text and adjust using `NcAction*` components.

There are quite a lot of changes. Separated with 💙 into small commits.

Before | After
---|---
image | image

### fix: apply a11y attrs for session list trigger

- Migrate from deprecated slot syntax
- Use a11y attrs from `NcPopover` on a custom button

image

**No visual changes**

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### fix(menubar): remove unneeded roles `menu`, `menuitem`

All these roles are correctly set by `NcAction*` components when needed.
Setting these roles in mixins and menu bar puts them on wrong elements.

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image
image | image | image

### fix(menubar): remove incorrect `aria-haspopup`

- It must have string `"true"/"menu"` value
- It must be placed on the `button` itself
- It is already correctly set by `NcActions` and `NcPopover`

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### fix(menubar): remove incorrect `aria-activedescendant`

- `aria-activedescendant` should identify a visually focused element when the real focus remains on this element.
  In the current implementation, it identifies the checked element (even when the menu is closed), not the focused.
- `aria-activedescendant` is not needed because NcAction has actual focus anyway.
  In case it will be actually needed, it should be implemented in the `NcActions`

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### fix(table): replace `InlineActionsContainer` by `NcActionButtonGroup`

It has the same styles and correct a11y attributes.

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### fix(table): make text align buttons radio

🏚️ Before | 🏡 After
---|---
image | image
image | image

### fix(menubar): replace `aria-pressed` with text description in `ActionList`

- `aria-pressed` is not valid for a menu trigger button. And 
- As an alternative solution - show that there is a selected value directly in the text
  - `Headings, "Heading 1" is selected`
- Remove incorrect `name` prop for `NcActions`

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### refactor: implement `ActionSingle` via template instead of render-function

Just to simplify further work and Vue 3 migration next release

**No visual changes**

### fix(menubar): set correct type for buttons instead of `aria-selected`

- Since `@nextcloud/vue@8.6.0` correct attributes are covered by `NcAction*` and `NcButton` components, including fixes:
  - Attribute should display not only active `attr="true"` state, but also non-active `attr="false"`
  - It should be `aria-pressed` for buttons and `aria-checked` for menu items instead of `aria-selected`
- Set correct `type` and active state
  - `type="radio"` for a list of options like `Heading`
  - `type="checkbox"` for toggle buttons like `Bold`
  - `type="button"` for general buttons without active state like `Undo`

**This doesn't fully work without the next one commit**

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image
image | image | image

### fix(menubar): separate `NcActionButton` from `ActionSingle`

- `ActionSingle` was used in 2 places:
  1. As a single button in the menu
  2. As an item in `NcActions`
- `NcActions` doesn't fully support non-direct `NcAction*` children
- Move `NcActionButton` usage from `ActionSingle` to a new component
- This new component is named `NcActionButton` so that `NcActions` will consider it to be a valid `NcAction*` component

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### fix(menubar): add separator between radio groups

- Separates different parts visually
- Required for a11y to group radio button in the menu

🏚️ Before | 🏡 After
---|---
image | image
image | image

### fix(menubar): name `ActionFormattingHelp` as `NcActionButton`

Same as with separating `NcActionButton` - for `NcActions` support

👀 Place | 🏚️ Before | 🏡 After
---|---|---
image | image | image

## 🏁 Checklist


JuliaKirschenheuter:
> Remove aria-selected

Or replace it with aria-pressed?

ShGKme:
> Or replace it with aria-pressed?

It should be replaced on toggle buttons and removed on general buttons. (Updated description)

src/components/Menu/MenuBar.vue
```diff
@@ -161,12 +162,19 @@ export default {
...
+			})
+			const entries = remainingEntries.reduce((acc, entry, index) => {
 				// If entry has children, merge them into list. Otherwise keep entry itself.
 				const children = entry.children ?? [entry]
+				// If this block has menu entries, it should be separated for better visibility and a11y (menu item radio grouping)
+				if (children.length > 1) {
+					const separatorBefore = acc.length ? [{ key: `separator-before-${entry.id}`, isSeparator: true }] : []
+					const separatorAfter = index !== remainingEntries.length - 1 ? [{ key: `separator-after-${entry.id}`, isSeparator: true }] : []
+					return [...acc, ...separatorBefore, ...children, ...separatorAfter]
+				}
```
max-nextcloud, line 177: If i parsed this correctly it may result in duplicate separators as soon as two entries with children are next to each other. I don't know if this can happen currently nor if it would be a problem. Will investigate some more.

max-nextcloud (commented):
Thanks a lot for splitting up the individual aspects. This was very helpful for understanding.

I only looked at the code so far. I'd like to play a bit with it before approving. I also have one minor comment I want to look into further.

src/components/Menu/MenuBar.vue
ShGKme: Nice catch

ShGKme: Added a fixup

max-nextcloud: Thanks! Looks good. I was not able to reproduce this as we never move the headings into the 'more' menu - but that might change.

max-nextcloud (approved):
Works like a charm.

emoral435 (approved):
The splitting of the issues within the description was really nice! No comments from me, just always great to learn how I should structure future problems with your best practices :)

JuliaKirschenheuter (approved):
wooow! That looks great and works right!

All good from my side besides:

image

But this is NcVue part

src/components/Editor/SessionList.vue
```diff
@@ -22,17 +22,20 @@
...
+			<div>
+				<button :title="label"
+					:aria-label="label"
+					class="avatar-list"
+					v-bind="attrs">
+					<div class="avatardiv icon-group" />
+					<AvatarWrapper v-for="session in sessionsVisible"
+						:key="session.id"
+						:session="session"
+						:size="40" />
```
Pytal, line 35: The div here should be switched to span and the divs inside AvatarWrapper also need to be span to not have block-level elements inside the inline button

Otherwise good!

Pytal (approved):

ShGKme:
Three `fixup!`s:
1. Fix potential issue @max-nextcloud found
2. To make `NcActions`' children named `NcActionButton` is was enough to change `name` property, without having to change the filename.
3. Fix 1/2 issues @JuliaKirschenheuter has found (the second one is on `@nextcloud/vue` and is less critical)

ShGKme:
Rebased onto master, squashed fixup commits (only small changes like renaming components were made, still works).

src/components/Editor/SessionList.vue
ShGKme: Thanks. I'll do it in another PR with fixing HTML validation issues.

Issue #5222, opened 2024-01-10
[BITV]: Gap of keyboard navigation through a toolbar


JuliaKirschenheuter:
1. Open a text file and navigate via keyboard
2. Focus on a "Close" button
3. Make 1 next navigation stop (via tab) and see: there is no element which is focused 
4. Push Enter
5. Dialog to open a file will be open -> completely unexpected behavior

Please set a right button and focus on it that there will be no gap in between

![Peek 2024-01-10 10-42](https://github.com/nextcloud/text/assets/6078378/c9e206d2-1a56-4295-adf6-637855f1cf3b)

Same behavior is on a preview page:
![Screenshot from 2024-01-10 11-32-57](https://github.com/nextcloud/text/assets/6078378/f2d2f0c0-9d2f-48f1-b1e0-4941a413d1b4)

JuliaKirschenheuter:
backport to 28 is still open

juliushaertl:
No https://github.com/nextcloud/text/pull/5227

szaimen:
I am not sure what is happening here but something seems to be very broken:
### cycling through with tab:

https://github.com/nextcloud/text/assets/42591237/a73ac051-6933-4382-8cfd-54797f1e075a


### cycling through with shift-tab:

https://github.com/nextcloud/text/assets/42591237/4be253e0-cc4e-4966-a589-44c83c416520

szaimen:
This doesnt seem to work correclty yet on stable28:

https://github.com/nextcloud/text/assets/42591237/ec77c856-0c2b-41fc-a38e-96a322344db6

(there is a focus-gap or it seems to focus elements that are not visible)

PR #5224, opened 2024-01-10
fix: Skip editor container and file upload input during keyboard navigation

juliushaertl:
Fix https://github.com/nextcloud/text/issues/5222

The hidden `input[type=file]` is used by the image menu action to trigger a file upload but does not need to be interacted with. I'm not entirely sure why the `#editor-container` is considered a focussable dom element, but having the tabindex disabled there fixes the tab navigation to be logical.

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -24,6 +24,7 @@
 	<div id="editor-container"
 		data-text-el="editor-container"
 		class="text-editor"
+		tabindex="-1"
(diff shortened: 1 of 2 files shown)


mejo- (approved):

juliushaertl:
Thanks for the insight, wasn't aware of that difference.

PR in https://github.com/nextcloud/text/pull/5230

Issue #5220, opened 2024-01-10
[BITV]: On small screens it is not possible to reach a submenu


JuliaKirschenheuter:
1. Open a text file
2. Set a viewport to the width of 320px
3. Lots of buttons are located inside of the "Remanining actions"
4. Try to select a button which is a `menu` itself like "headings" or "emojis"
5. "undefined" will be written in a text file:
![Screenshot from 2024-01-10 10-22-25](https://github.com/nextcloud/text/assets/6078378/5552f5bd-a0ca-4973-9810-59ece7615200)

juliushaertl:
Is the emoji picker usable anywhere with that screen size?

JuliaKirschenheuter:
@szaimen why it is not usable if scrollbar is there?

szaimen:
> @szaimen why it is not usable if scrollbar is there?

The scrollbar is outside the emojy picker if you look in detail

juliushaertl:
So the menubar in text is sticky, that is why scrolling the page does not move the position of the emoji picker modal. Though I don't see a good way to solve this as of now

JuliaKirschenheuter:
> So the menubar in text is sticky, that is why scrolling the page does not move the position of the emoji picker popover. Though I don't see a good way to solve this as of now

All right @juliushaertl, thank you! We can do what we can for now. @szaimen if you have no ideas how to fix it, we can close the issue.

szaimen:
> @szaimen if you have no ideas how to fix it, we can close the issue.

The only fix I see for now would be disabling the emojiy picker on mobile if the viewport (height or width) is too low for it to work correctly...

JuliaKirschenheuter:
I don't think there is something we can change for now. Thank you @juliushaertl, closing the issue

juliushaertl:
> Emoji picker should be fixed later separately (another issue). Idea from @ShGKme: on small screens a full width of emoji picker can be shown.

Yes, probably also in full height then but we'll need some kind of close button or close on insert. Probably also something to generally think about for the NcPopover component (ref https://github.com/nextcloud-libraries/nextcloud-vue/issues/3159)

PR #5226, opened 2024-01-10
fix: Prioritize un-collapsible menubar entries for mobile and aggregate children in the hidden entries

juliushaertl:
Fix https://github.com/nextcloud/text/issues/5220

There are some menubar entires that cannot be collapsed in the hidden entries list as they trigger another popover which comes with too much complexity to implement a nested popover now. However with 320px we have enough space to show all of them.

For others (like callout actions) where the menu is just a list of simple action buttons we can just merge them into the overflow menu.

This ensures that any action is usable on all screen sizes >= 320px.

Screenshot 2024-01-10 at 15 24 13

### DIFF
--- a/src/components/Menu/MenuBar.vue
+++ b/src/components/Menu/MenuBar.vue
@@ -161,14 +161,20 @@ export default {
 			return list
 		},
 		hiddenEntries() {
+			const entries = this.entries.filter(({ priority }) => {
+				// reverse logic from visibleEntries
+				return priority !== undefined && priority > this.iconsLimit
+			}).reduce((acc, entry) => {
+				// If entry has children, merge them into list. Otherwise keep entry itself.
+				const children = entry.children ?? [entry]
+				return [...acc, ...children]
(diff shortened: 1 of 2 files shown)


src/components/Menu/MenuBar.vue
```diff
@@ -161,14 +161,20 @@ export default {
 			return list
 		},
 		hiddenEntries() {
+			const entries = this.entries.filter(({ priority }) => {
+				// reverse logic from visibleEntries
+				return priority !== undefined && priority > this.iconsLimit
+			}).reduce((acc, entry) => {
+				// merge all entries children into the entries list
```
mejo-, line 168:
```suggestion
				// If entry has children, merge them into list. Otherwise keep entry itself.
```

I didn't get what the comment said straight away, so here's a suggestion for being more explicit.

mejo- (approved):
Didn't test the code, but looks sensible to me. Only one comment.

PR #5230, opened 2024-01-11
fix: Use proper way of hiding the attachment input

juliushaertl:
Hide the input field properly to make it not focusable and hidden for assistive technologies.

as described in https://github.com/nextcloud/text/pull/5224#issuecomment-1885255799

### DIFF
--- a/src/components/Editor/MediaHandler.vue
+++ b/src/components/Editor/MediaHandler.vue
@@ -28,13 +28,11 @@
 		@dragover.prevent.stop="setDraggedOver(true)"
 		@dragleave.prevent.stop="setDraggedOver(false)"
 		@file-drop="onEditorDrop">
-		<input ref="attachmentFileInput"
-			tabindex="-1"
+		<input v-show="false"
+			ref="attachmentFileInput"
 			data-text-el="attachment-file-input"
 			type="file"
 			accept="*/*"
-			aria-hidden="true"
-			class="hidden-visually"
 			multiple
 			@change="onAttachmentUploadFilePicked">
 		<slot />


src/components/Editor/MediaHandler.vue
```diff
@@ -28,13 +28,12 @@
...
 		@dragleave.prevent.stop="setDraggedOver(false)"
 		@file-drop="onEditorDrop">
-		<input ref="attachmentFileInput"
-			tabindex="-1"
+		<input v-show="false"
+			ref="attachmentFileInput"
 			data-text-el="attachment-file-input"
 			type="file"
 			accept="*/*"
 			aria-hidden="true"
```
juliushaertl, line 36: @ShGKme Also do we still need the aria-hidden then or is this already fine with just having display: none; through v-show

mejo- (approved):

src/components/Editor/MediaHandler.vue
ShGKme: No, `display: none` (and `visibility: hidden`) both make element completely hidden.

Let's say, be default element is visible for both monitor's and screen reader's users.

- `display: none`/`visibility: hidden` - hides element completely for everything
- `aria-hidden="true"` - hides element only for screen readers (meaning, it is not important element, like decorative image)
- `hidden-visually` - hides element only visually (meaning, this is only for screen reader or replaces some only visual element)

One specific case when `aria-hidden="true"` and `hidden-visually` used together - the prev and next slides in Viewer. They must be hidden, but for image/video pre-loading and because of a bug in FF we need then to not be `display: none`.

juliushaertl: Thanks, that makes sense ❤️

ShGKme (approved):

PR #5174, opened 2023-12-29
feat: switch to native browser tooltips

luka-nextcloud:
### 📝 Summary

* Resolves: #4321

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [ ] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [ ] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required

PR #5233, opened 2024-01-11
Fix migration from `v-tooltip` to native `title` in menu bar buttons `ActionEntry`

ShGKme:
#### Migration from `v-tooltip` to native `title` was not complete in `<ActionSingle>`

- Regression of: https://github.com/nextcloud/text/pull/5174/
  - Shows incorrect `title` on `NcAction`
  - Floods console with dozen errors per several seconds
- Removed using directive and fixed `title` text

image

##### 🖼️ Screenshots

Before | After
---|---
image | image

#### Duplication of `t` definition

In some components translation method `t` was added as a computed property `t: () => window.t` or a method. But method `t` is already added globally by prototype monkey-patching: `Vue.prototype.t = t`.

It caused error. Removed as not needed.

image

### 🏁 Checklist

### DIFF
--- a/src/components/Menu/ActionSingle.vue
+++ b/src/components/Menu/ActionSingle.vue
@@ -104,18 +104,10 @@ export default {
 
 		const { class: classes, ...attrs } = bindState
 
-		// do not use tooltip if is a item of action list
-		const directives = isItem
-			? []
-			: [{
-				name: 'tooltip',
-				value: tooltip,
-			}]
-
 		const children = [h(icon, { slot: 'icon' })]
 
 		// do not use title if is a item of action list
-		const title = isItem ? undefined : label
+		const title = isItem ? undefined : tooltip
 
 		if (isItem || actionEntry.forceLabel) {
 			// add label
@@ -123,7 +115,6 @@ export default {
 		}
 
 		return h(component, {
-			directives,
 			staticClass: 'entry-single-action entry-action',
 			class: classes,
 			attrs: {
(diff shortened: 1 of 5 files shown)


src/components/Menu/ActionSingle.vue
```diff
@@ -98,20 +98,11 @@ export default {
...
 
 		const { class: classes, ...attrs } = bindState
 
-		// do not use tooltip if is a item of action list
-		const directives = isItem
-			? []
-			: [{
-				name: 'tooltip',
-				value: tooltip,
-			}]
```
ShGKme, line 113: Currently `title` (that replaces the tooltip) has `isItem ? undefined : label`, not `tooltip` as a value.

https://github.com/nextcloud/text/blob/4a707d05643c182e114b9c1a44d718b9c26bc877/src/components/Menu/BaseActionEntry.js#L68-L73

Should here `title` also be `tooltip` and not `label`?

ShGKme: Hmm, I think, yes, it should. It shows the shortcut on desktop

ShGKme: PR Updated

juliushaertl:
No need for comiting compiled assets here, for text we automated this after the merge.

juliushaertl (approved):
Thanks for that ❤️

ShGKme:
This seems to be a regression of https://github.com/nextcloud/text/pull/5174

So should also be backported

ShGKme:
2 commits just squashed into one as they both actually fix migrating from `v-tooltip` to `title.

No changes in the code.

ShGKme:
Removing `t` from suggestions was a mistake. Suggestion lists are mounted as separated apps and don't reuse `t` from the text app init.

Returned it 🙈

Issue #5221, opened 2024-01-10
[BITV]: missing labels on textaria inside of translate modal dialog


JuliaKirschenheuter:
![Screenshot from 2024-01-10 10-19-03](https://github.com/nextcloud/text/assets/6078378/86fb61cc-0342-4cb0-a69f-64fbfe9abd9f)

Please adapt `label` for the textaria, `NcTextArea` component

Please check other places on existing `label`'s.

luka-nextcloud:
@JuliaKirschenheuter @juliushaertl Got it, what could be the texts for the labels of textarea elements?
"Input" and "Output"?

juliushaertl:
Maybe "Text to translate from" "Translated text result" to be a bit less technical

szaimen:
I cannot test this but the backport seems to be merged

PR #5234, opened 2024-01-11
feat: add labels to textarea elements in translate modal

luka-nextcloud:
### 📝 Summary

* Resolves: #5221

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Modal/Translate.vue
+++ b/src/components/Modal/Translate.vue
@@ -35,10 +35,12 @@
 							:disabled="disableFromLanguageSelect"
 							:append-to-body="false" />
 					</div>
-					<textarea ref="input"
-						v-model="input"
+					<NcTextArea ref="input"
+						:value.sync="input"
+						:label="t('text', 'Text to translate from')"
 						autofocus
-						@input="autosize"
+						input-class="translate-textarea"
+						resize="none"
 						@focus="onInputFocus" />
 				</div>
 				<div class="col">
@@ -51,9 +53,12 @@
 							:disabled="!fromLanguage"
 							:append-to-body="false" />
 					</div>
-					<textarea ref="result"
-						v-model="result"
+					<NcTextArea ref="result"
+						:value.sync="result"
+						:label="t('text', 'Translated text result')"
 						readonly
+						input-class="translate-textarea"
+						resize="none"
 						:class="{'icon-loading': loading }" />
 				</div>
 			</div>
@@ -80,7 +85,7 @@
 import axios from '@nextcloud/axios'
 import { loadState } from '@nextcloud/initial-state'
 import { generateOcsUrl } from '@nextcloud/router'
-import { NcModal, NcButton, NcSelect, NcLoadingIcon } from '@nextcloud/vue'
+import { NcModal, NcButton, NcSelect, NcLoadingIcon, NcTextArea } from '@nextcloud/vue'
 import { useIsMobileMixin } from '../Editor.provider.js'
 
 const detectLanguageEntry = {
@@ -95,6 +100,7 @@ export default {
 		NcButton,
 		NcSelect,
 		NcLoadingIcon,
+		NcTextArea,
 	},
 	mixins: [
 		useIsMobileMixin,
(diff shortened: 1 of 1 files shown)


juliushaertl (changes requested):
The autosize method seems to be broken due to that, you can check that by trying to type something in the input.

We probably can drop it and instead use resize from the component https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextarea just need to ensure that the max height is limited to not exceed the visible screen space in the modal

Screenshot 2024-01-11 at 14 10 58

emoral435:
Since this issue deals with accessibility, would you want the rest of the accessibility team to also help review? :)

juliushaertl:
@emoral435 Definitely, additional reviews are very welcome

src/components/Modal/Translate.vue
```diff
@@ -210,7 +225,7 @@ export default {
 }
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
```
emoral435, line 228: question: Any reason you removed this CSS being scoped? Just for best practices, I know this is only a text modal, but keeping it scoped will help ensure nothing funky happens in case there are similar classes

emoral435: especially if you are adding the `!important` elements in the CSS, this indicates that other pieces of css that may not be scoped are affecting this code block, then having this be scoped helps prevent similar things in the future

emoral435 (approved):
Other than these general questions, LGTM! Will ping other, more experiences a11y team members too :)

ShGKme (changes requested):
Agree with @emoral435, `scoped` should not be removed. These styles are not about something really global.

You can use `:deep`: https://vuejs.org/api/sfc-css-features.html#deep-selectors

Best to specify a selector on this component and then with deep selector on manually provided class.

```css
.translate-dialog :deep(.translate-textarea)

# or more specific, set class on the NcTextarea and then select on textarea

.translate-dialog__textarea :deep(.translate-textarea)
```

src/components/Modal/Translate.vue
```diff
@@ -35,10 +35,12 @@
...
 					</div>
-					<textarea ref="input"
-						v-model="input"
+					<NcTextArea ref="input"
+						v-model="inputValue"
+						:label="t('text', 'Text to translate from')"
 						autofocus
-						@input="autosize"
+						input-class="translate-textarea"
+						resize="none"
```
ShGKme, line 43: It works, but it is incorrect usage of the component. `NcTextArea`, same as other fields in `@nextcloud/vue`, doesn't have `v-model` - it has no defined `model` and doesn't emit `input` event.

It works, because here `v-model` directive catches native event `input` that bubbles with the native event. What it why it was needed to add computed proxy.

Instead, `.sync` modifier should be used on value binding: `:value.sync="input"`.

src/components/Modal/Translate.vue
```diff
@@ -119,6 +125,15 @@ export default {
...
 	computed: {
+		inputValue: {
+			get() {
+				return this.input
+			},
+			set(event) {
+				this.input = event.target.value
+				this.autosize()
+			},
+		},
```
ShGKme, line 136: If we want to resize on input, we can watch the `input` property, so it won't be important how the value has changed for the resize.

ShGKme:
Also, I cannot focus any input in the modal, but I cannot say if it isn't an issue on my side...

luka-nextcloud:
> Also, I cannot focus any input in the modal, but I cannot say if it isn't an issue on my side...

Sorry, I cannot reproduce the issue on my side.

ShGKme (approved):

emoral435:
With two reviews, should be good to merge now @luka-nextcloud 👍

juliushaertl (approved):

emoral435:
/backport d29c9670b8ffad91e6bab0093b9e31ed1100b5ca to stable28

PR #5264, opened 2024-01-15
fix(deps): stick to @nextcloud/files v2

max-nextcloud:
### 📝 Summary

* Resolves: failing build in #5251 

`@nextcloud/files` requires node 20 and npm 9. This dependency was introduced based on a beta version that did not have these requirements yet. Roll back to version 2 rather than updating the entire node stack.

Only difference here is that `formatFileSize` which we use will default to non-binary file sizes - i.e 1.024KB instead of 1KiB as this only became available with v3:
* https://github.com/nextcloud-libraries/nextcloud-files/pull/770/files#diff-237392127cb275fc6789cf6faec014d758950c87ccdab0329dde793d034190d8


max-nextcloud:
I'm not sure if we should do that same for stable27.

mejo- (approved):

Issue #4686, opened 2023-08-16
Assistant synchronous task handling and polishing


juliushaertl:
from https://github.com/nextcloud/text/pull/4627

- [ ] Implement synchronous tasks 
- [ ] Smaller polishing
  - [ ] use notifications for triggering fetch instead of polling
  - [ ] FAB on mobile
  - [ ] How to handle markdown vs plain text
  - [ ] Dismiss notifications
  - [ ] merge translate to it and just move the actions to popover directly
  - [ ] Think about the making it show/hide depending on the mouse cursor (gdocs)

julien-nc:
Additionally:
* [ ] Decide what to do when the assistant returns a completed task immediately (new [sync flow](https://github.com/nextcloud/assistant/pull/13)).

jancborchardt:
If there are any design questions regarding this, @nimishavijay is the expert since she worked on the current implementation. :)

luka-nextcloud:
> As discussed in the call
> 
> * Assistant will offer a flag to close the modal after a successful sync task
>   
>   * Can be used to directly insert and act with a sync result in text (e.g. when generating text on an empty line)
>   * If a task is switched from sync to async the promise resolves as usual with the task but the info about the scheduled task is shown no matter if the close modal flag is set or not
> * The main promise always resolves the task with its current status
> * There will be a list of action buttons that can be provided through the `openAssistantForm` parameters
>   
>   * Can be used in text to offer insert/replace actions on the result screen

@juliushaertl I have checked the repo [Assistant app](https://github.com/nextcloud/assistant) but seems like it still does not provide these functions:
- There is no event or public function `to close the modal after a successful sync task`
- There are no parameters for `a list of action buttons that can be provided through the openAssistantForm parameters`

juliushaertl:
I linked the wrong PR but the main branch should have both https://github.com/nextcloud/assistant/blob/main/src/assistant.js#L37-L44

Reference PR https://github.com/nextcloud/assistant/pull/18

PR #5269, opened 2024-01-16
feat: handle assistant sync task

luka-nextcloud:
### 📝 Summary

* Resolves: #4686

- update assistant name space
- add insert button for assistant sync task result
- use notifications for triggering fetch instead of polling

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -41,6 +41,13 @@ jobs:
           ref: ${{ matrix.server-versions }}
           path: apps/viewer
 
+      - name: Checkout assistant
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          repository: nextcloud/assistant
+          ref: 'main'
+          path: apps/assistant
+
       - name: Checkout app
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
@@ -68,6 +75,12 @@ jobs:
         run: |
           npm ci
           TESTING=true npm run build --if-present
+
+      - name: Install dependencies & build assistant
+        working-directory: apps/assistant
+        run: |
+          npm ci
+          npm run build
 
       - name: Save context
         uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
@@ -142,6 +155,8 @@ jobs:
           php occ user:add --password-from-env user2
           php occ app:enable viewer
           php occ app:enable text
+          php occ app:enable assistant
+          php occ app:enable testing
           php occ app:list
           php occ background:cron
           php occ config:system:set session_keepalive --value=false --type=boolean
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -79,6 +79,7 @@ jobs:
       - name: Install dependencies & build assistant
         working-directory: apps/assistant
         run: |
+          composer install --no-dev
           npm ci
           npm run build
 
--- a/cypress/entrypoint.sh
+++ b/cypress/entrypoint.sh
@@ -19,6 +19,8 @@ if [ $? -eq 0 ]; then
         php /var/www/html/occ user:add --password-from-env user2
         php /var/www/html/occ app:enable viewer
         php /var/www/html/occ app:enable text
+        php /var/www/html/occ app:enable assistant
+        php /var/www/html/occ app:enable testing
         php /var/www/html/occ app:list
     "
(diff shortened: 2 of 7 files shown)


lib/AppInfo/Application.php
```diff
@@ -41,7 +41,7 @@
 use OCA\Text\Middleware\SessionMiddleware;
 use OCA\Text\Notification\Notifier;
 use OCA\Text\Service\ConfigService;
-use OCA\TPAssistant\Event\BeforeAssistantNotificationEvent;
+use OCA\TpAssistant\Event\BeforeAssistantNotificationEvent;
```
juliushaertl, line 44: Can you add the class to the psalm stub file to fix the failing psalm action?

src/components/Assistant.vue
```diff
@@ -213,7 +214,7 @@ export default {
 	},
 	computed: {
 		showAssistant() {
-			return !this.$isRichWorkspace && !this.$isPublic && window?.OCA?.TPAssistant?.openAssistantForm
+			return !this.$isRichWorkspace && !this.$isPublic && window?.OCA?.TpAssistant?.openAssistantForm
```
juliushaertl, line 217: @julien-nc Was there a casing change in the API here? Is it bound to a Nextcloud release?

juliushaertl (commented):
Small comments, otherwise this looks good and works (with https://github.com/nextcloud/server/pull/43005 for testing)

juliushaertl:
@luka-nextcloud Could you also take the time to add cypress coverage for this part? We can enable the testing app for this in the github action.

juliushaertl:
Tested and works function wise 👍 

@julien-nc I noticed some odd behaviour with https://github.com/nextcloud/server/pull/43005 where having multiple text processing providers enabled leads to the async one alwas being used in the assistant modal. I haven't touched any configuration, though the AI settings show the synchronous one as a default:

Screenshot 2024-01-23 at 19 48 22

luka-nextcloud:
> @luka-nextcloud Could you also take the time to add cypress coverage for this part? We can enable the testing app for this in the github action.

I've added it, please check.

.github/workflows/cypress.yml
```diff
@@ -41,6 +41,13 @@ jobs:
           ref: ${{ matrix.server-versions }}
           path: apps/viewer
 
+      - name: Checkout assistant
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          repository: nextcloud/assistant
+          ref: 'main'
+          path: apps/assistant
```
juliushaertl, line 49: Need to ensure that both assistant and testing app are also enabled in line 151 when setting up the Nextcloud instance to be used in the cypress tests

cypress/entrypoint.sh
```diff
@@ -19,8 +19,10 @@ if [ $? -eq 0 ]; then
         php /var/www/html/occ user:add --password-from-env user2
         php /var/www/html/occ app:enable viewer
         php /var/www/html/occ app:enable text
+        php /var/www/html/occ app:enable assistant
```
juliushaertl, line 22: I think this was for some way of running isolated cypress tests locally, but never used that script myself. @max-nextcloud Do you happen to know if we still make use of that?

juliushaertl:
@luka-nextcloud Could you check why the cypress tests fail there?

src/components/Assistant.vue
julien-nc: Sorry for the late answer. This was and will stay OCA.TPAssistant.

julien-nc:
> I noticed some odd behaviour with https://github.com/nextcloud/server/pull/43005 where having multiple text processing providers enabled leads to the async one alwas being used in the assistant modal. I haven't touched any configuration, though the AI settings show the synchronous one as a default:

@juliushaertl Thanks, I'll check that. Maybe there's a mismatch between the default value displayed in the settings frontend and the actual default selected provider when creating a task.

juliushaertl (approved):

cypress/entrypoint.sh
max-nextcloud: I don't remember it. Looking at the log of that file indicates that it was introduced by @susnux in f62c29be.

juliushaertl:
Pushed a small adjustment for the github action that should make cypress pass.

cypress/entrypoint.sh
susnux: Was just for local tests, but nowadays we could use `@nextcloud/cypress/docker` which also works locally :)

PR #5273, opened 2024-01-17
[stable27] fix: Proper error message based on file permissions

juliushaertl:


mejo- (approved):

juliushaertl:
Cypress test is flaky and Node check is passing further down in the check list

PR #5279, opened 2024-01-17
fix(sync): push local changes on reconnect

max-nextcloud:
When an editing session is interrupted
steps that were already pushed to the server
may be cleared alongside the sync service session.

Make sure to push all local state
that is not part of the document state
when (re-)connecting.

### Todo

- [x] add `y-protocols` to package.json
- [x] keep `.yjs` file on server side even during session cleanup
- [x] try it out
- [ ] Add some actual e2e tests

### 🏁 Checklist


max-nextcloud:
This will break how we currently implemented `occ test:reset` without the `--force` option:
 * the occ job will remove the last steps from the database and clear all sessions
 * the clients with the local changes will reconnect
 * they will fetch the .yjs file, compute their local update and redistribute it.

We could clear the .yjs file in the cleanup job. We would need to detect this and reload the editor when there is no yjs file during a reconnect.

cypress.config.js
```diff
@@ -43,6 +34,12 @@ module.exports = defineConfig({
 		baseUrl: 'http://localhost:8081/index.php/',
 		specPattern: 'cypress/e2e/**/*.{js,jsx,ts,tsx}',
 	},
+	component: {
```
juliushaertl, line 37: We don't run component tests in CI yet I think

juliushaertl:
> This will break how we currently implemented occ test:reset without the --force option

That seems reasonable. For a manual test it would be good to try this before/after:
- Write a document
- Set the browser to offline
- Reset the document with occ
- Join in another browser and do some edits
- Reconnect the offline browser

src/components/Editor.vue
```diff
@@ -487,6 +487,11 @@ export default {
 		onLoaded({ documentSource, documentState }) {
 			if (documentState) {
 				applyDocumentState(this.$ydoc, documentState, this.$providers[0])
+				// distribute additional state that may exist locally
+				const updateMessage = getUpdateMessage(this.$ydoc, documentState)
+				if (updateMessage) {
+					this.$queue.push(updateMessage)
```
juliushaertl, line 493: Can we log this to the console so we have some trace that this happened?

src/helpers/yjs.js
```diff
@@ -21,27 +21,80 @@
...
 	const update = decodeArrayBuffer(documentState)
-	applyUpdate(ydoc, update, origin)
+	Y.applyUpdate(ydoc, update, origin)
+}
+
+/**
+ * Update message for everything in ydoc that is not in encodedBaseUpdate
+ *
+ * @param {Y.Doc} ydoc - encode state of this doc
+ * @param {string} encodedBaseUpdate - base64 encoded doc update to build upon
```
juliushaertl, line 56: Return type annotation would be good especially since undefined is an expected value

src/components/Editor.vue
```diff
@@ -487,6 +487,11 @@ export default {
 		onLoaded({ documentSource, documentState }) {
 			if (documentState) {
 				applyDocumentState(this.$ydoc, documentState, this.$providers[0])
+				// distribute additional state that may exist locally
```
juliushaertl, line 490: So we always distribute the local (possibly outdated) state? Can we ensure that this does not lead to unexpected reverting of newer changes from the server?

max-nextcloud: This state is yjs document state. So basically what we are distributing is our local knowledge of the history of the document. yjs will handle sorting those edits and turning them into a current state. In the end it will produce a consistent view for everyone - as long as all are on the same page. We need to share all local history so our future edits will be applied.

Let's say we worked together on a text. I got disconnected but made some changes before the editor noticed. Now I'm reconnecting.

If I distribute my local changes yjs will put all changes in an order and create a current content that may contain my or your changes depending on the order picked by yjs. There's no way of knowing what would be the correct order and we would have to live with that. At least we have a consistent  view on the content.

If I keep connect without distributing my local changes I will receive your changes - the doc would probably look the same as above for me. Your view would stay the same as it was when i was disconnected and changes i make later on will not show for you.

If I drop all local changes and basically reload the editor we'd also be on the same page and my changes would be lost.

mejo- (commented):
Thanks a lot for looking into this :heart: 

I tried to follow the logic of the code and your changes seem reasonable to me - also taking into account yours and Julius' comments. I didn't test myself and would give yours and Julius' assessment more weight than mine since I struggle to keep the big picture about yjs and document state in my head :exploding_head:

max-nextcloud:
> That seems reasonable. For a manual test it would be good to try this before/after:

## Before

Somewhat mixed results:
* occ reset while editor is connected - editor turned empty - not even a menu bar
* second attempt - editor showed reconnect button. continued editing on reconnect, after closing an opening doc was back to saved state
* third attempt - disconnected editor right after occ run, showed reconnect button, connected other editor, editing based on last changes, reconnected first editor - reconnected editor can see changes from the other but not the other way round.

None of this seems so good that we would want to keep it ;).

## After

Seeing lot's of 422 responses... Turns out I am mixing base64 encoded steps with Uint8Arrays in the queue. Will fix.

juliushaertl:
Sorry, I've had probably some confusion when testing yesterday. Did a reset now (probably with a bit more concentration) and it seems to work well. I couldn't see any trouble with changes being reverted.

juliushaertl (approved):

juliushaertl:
As the e2e test checkbox is still unchecked, any idea how you could test that? But would not consider that a blocker for the merge

PR #5281, opened 2024-01-18
fix: Fix tab focus when other elements are displayed next to text that are within a focus trap

juliushaertl:
Instead of manually messing with the focus trap which breaks the focus order when trying to tab out of the editor into the sidebar next to the viewer, we can just use the default viewer handling and instead prevent event propagation for cases where the Tab key issues a command within the editor.

Additional fix for https://github.com/nextcloud/text/issues/5222#issuecomment-1898486424

### DIFF
--- a/src/extensions/FocusTrap.js
+++ b/src/extensions/FocusTrap.js
@@ -1,0 +1,41 @@
+import { Extension } from '@tiptap/core'
+
+const toggleFocusTrap = ({ editor }) => {
+	const trapStack = window._nc_focus_trap ?? []
+	const activeTrap = trapStack[trapStack.length - 1]
+
+	const possibleEditorTabCommand = editor.can().sinkListItem('listItem')
+		|| editor.can().goToNextCell()
+		|| editor.can().goToPreviousCell()
+
+	if (possibleEditorTabCommand) {
+		activeTrap?.pause()
+	} else {
+		activeTrap?.unpause()
+	}
+}
+
+const unpauseFocusTrap = ({ editor }) => {
+	const trapStack = window._nc_focus_trap ?? []
+	const activeTrap = trapStack[trapStack.length - 1]
+
+	activeTrap?.unpause()
+}
+
+/**
+ * The viewer focus trap needs to be paused on the fly in order to properly handle tab commands in the editor,
+ * as we have no control over if a tab key event is reaching the editor otherwise. This is because the focus trap
+ * registeres a capture listener (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture), so whenever we reach a tab command in the editor the focus trap will already have captured the event.
+ *
+ * We also cannot work around this by pushing our own focus trap to the stack, as the focus trap package does not offer any reliable way to programmatically focus the next element of the parent trap if we allow tabbing out of the editor.
+ */
+const FocusTrap = Extension.create({
+	name: 'focustrap',
+	onFocus: toggleFocusTrap,
+	onBlur: unpauseFocusTrap,
+	onSelectionUpdate: toggleFocusTrap,
+	onTransaction: toggleFocusTrap,
+	onUpdate: toggleFocusTrap,
+})
+
+export default FocusTrap
--- a/src/EditorFactory.js
+++ b/src/EditorFactory.js
@@ -29,7 +29,7 @@ import { lowlight } from 'lowlight/lib/core.js'
 import hljs from 'highlight.js/lib/core'
 
 import { logger } from './helpers/logger.js'
-import { Mention, PlainText, RichText } from './extensions/index.js'
+import { FocusTrap, Mention, PlainText, RichText } from './extensions/index.js'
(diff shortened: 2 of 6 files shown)


juliushaertl:
Additional related question that came up while testing: Do we need a active focus indicator for the actual content editable? Currently that does not have any indicator except for the cursor position

Something like this would be easy to achieve but I'm unsure as this is quite a visual design change cc @nextcloud/designers 

Screenshot 2024-01-18 at 16 38 25

juliushaertl:
Checking the cypress failure

JuliaKirschenheuter:
Thank you @juliushaertl for taking over!

No, we don't need an active focus indicator for the actual content editable.

Just tested (i don't know if this PR is ready) but there seems to be 2 invisible focus stops between content editable and sidebar elements: same behavior for forward and backward navigation

juliushaertl:
> Just tested (i don't know if this PR is ready) but there seems to be 2 invisible focus stops between content editable and sidebar elements: same behavior for forward and backward navigation

Which browser did you try there? I haven't seen that happening yesterday.

JuliaKirschenheuter:
tested again, but still can reproduce in Chrome ;(

juliushaertl:
Ok, this turns out more tricky. Apparently we cannot just prevent the focus trap event listener form triggering as it is registered with https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture false so it will be executed even before we ever reach the event listeners from text.

pause/unpause of the focus trap could work but seems not very clean

Maybe we need to create our own trap for the content editable so we can control if and how it would be exited - just assuming for now that the parent trap would then just be picked up properly, but needs more work and investigation.

juliushaertl:
https://github.com/focus-trap/focus-trap/blob/master/docs/js/nested.js is probably a good sample for this if we handle the activation once the editor gets focus and the deactivation on the Tab/Shift-Tab keypress (if no native editor function like list indent is feasible)

max-nextcloud:
@juliushaertl sounds like this still needs some work, right? Turning into `2. to develop` and a draft and will review once it's ready.

juliushaertl:
All right, I went for a slightly different approach.

The viewer focus trap needs to be paused on the fly in order to properly handle tab commands in the editor, as we have no control over if a tab key event is reaching the editor otherwise. This is because the focus trap registers a capture listener (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture), so whenever we reach a tab command in the editor the focus trap will already have captured the event.

We also cannot work around this by pushing our own focus trap to the stack, as the focus trap package does not offer any reliable way to programmatically focus the next element of the parent trap if we allow tabbing out of the editor.

Therefore a new tiptap extension now pauses the focus trap whenever the tab key should be catched by editor actions (indent lists, navigate to the next table cell) and otherwise unpauses so that the actual next tabbable element in the trap containers is getting its focus.

@ShGKme Would appreciate your feedback here since you likely have better knowledge of focus trap handling. Does that approach make sense? At least function wise this seems to be the only way to achieve consistent tab handling I could find.

ShGKme:
@juliushaertl I have been just working on fixing the conflict between `NcActions` and `focus-trap`.

Maybe that's what you need?

https://github.com/nextcloud-libraries/nextcloud-vue/pull/4953

> focus trap needs to be paused on the fly

This is exactly what I have done. I'm pausing the focus trap stack and unpausing it back.

juliushaertl:
Ok, similar approach then. The fix here is unrelated to actions so we still need our own implementation, but then I'd set this back to review state

juliushaertl:
Ah just tests need some fixes

src/extensions/FocusTrap.js
```diff
@@ -0,0 +1,41 @@
+import { Extension } from '@tiptap/core'
+
+const toggleFocusTrap = ({ editor }) => {
+	const trapStack = window._nc_focus_trap ?? []
+	const activeTrap = trapStack[trapStack.length - 1]
+
+	const possibleEditorTabCommand = editor.can().sinkListItem('listItem')
```
mejo-, line 7: This will throw an exception when `listItem` is not a registered node type:

`Error: There is no node type named 'listItem'. Maybe you forgot to add the extension?`

We probably either have to catch the exception or check if the node type exists first.

mejo- (changes requested):
Thanks a lot for working on this! I tested this with my link bubble PR (#5158) and it fixs the tab navigation behaviour there :heart:

mejo-:
> Thanks a lot for working on this! I tested this with my link bubble PR (#5158) and it fixs the tab navigation behaviour there ❤️

In fact it also makes the Cypress tests for #5158 much more reliable :pray:

src/extensions/FocusTrap.js
juliushaertl: Fixed by only using it for the rich text editing. Plaintext has a different problem, but I'll address that separately.

mejo- (approved):

PR #5302, opened 2024-01-22
fix(mention): use new url schema

max-nextcloud:
### 📝 Summary

Fixes notifications for mentions in 28 and following.
* Resolves: Failing ci runs such as  https://github.com/nextcloud/text/actions/runs/7587197071?pr=5284

#### 🖼️ Screenshots

Test mentioning users -- Select a user will insert the mention (failed)

### 🏁 Checklist


juliushaertl:
Thanks for looking into that Max :)

PR #5310, opened 2024-01-24
Remove bottom margin for better warning blocks display

Jerome-Herbinet:
This CSS change removes the extra bottom padding (which is in fact a paragraph bottom margin).
It makes text more compact, which is IMO relevant because warning blocks aim to be compact.

### 🏚️ Before
Capture d’écran du 2024-01-24 09-21-10
### 🏡 After
Capture d’écran du 2024-01-24 09-21-17

Fixes: nextcloud/collectives#1175

### 🏁 Checklist

### DIFF
--- a/src/nodes/Callout.vue
+++ b/src/nodes/Callout.vue
@@ -86,7 +86,7 @@ export default {
 
 	.callout__content {
 		margin-left: 1em;
-		&:deep(p) {
+		&:deep(p .paragraph-content) {
 			&:last-child {
 				margin-bottom: 0;
 			}


src/css/prosemirror.scss
```diff
@@ -99,7 +99,7 @@ div.ProseMirror {
 	}
 
 	p .paragraph-content {
-		margin-bottom: 1em;
+		margin-bottom: 0;
```
juliushaertl, line 102: I think we should not generally remove this but only in scope of the callout boxes (warning/error/...)

juliushaertl (changes requested):
Thanks for the contribution. Small comment inside as I think the rule is applying to widely.

src/css/prosemirror.scss
mejo-: I agree. The `margin-bottom` is also what separates paragraphs from each other to my knowledge, so we have to keep it in general.

Jerome-Herbinet: I agree that the "margin-bottom" is used to separate paragraphs, except that in this case : 

- it creates a systematic, superfluous margin that visually unbalances the content within the block (it creates a kind of ugly extra bottom padding)
- warning blocks need to be compact; we don't put literature in them, but short informative content (so the paragraphs margins a bit useless in this context)

I'd also like to add that I've seen this type of block in the "Polls" app, for example, and they don't have this display issue. So there is a bug to be corrected in this context.

Jerome-Herbinet: Polls : 
Capture d’écran du 2024-01-25 08-58-52

juliushaertl: Yes, I think maybe my concern was not clear enough. Your current change also affects paragraphs outside of the warning blocks:

Screenshot 2024-01-25 at 09 10 38

For some code pointer we already have a rule for that but it does not seem to apply to the proper element https://github.com/nextcloud/text/blob/e806141296dd11dccfd0f4bb810d9c5b576bed8a/src/nodes/Callout.vue#L91

Jerome-Herbinet: OK, I understand @juliushaertl ; I'll make some changes.

Jerome-Herbinet:
@juliushaertl please check my new commit (old and new commit have been squashed)

Jerome-Herbinet:
Testing by your side is welcome

src/css/prosemirror.scss
```diff
@@ -103,6 +103,10 @@ div.ProseMirror {
 		line-height: 150%;
 	}
 
+	.callout p .paragraph-content {
```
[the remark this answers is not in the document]
Jerome-Herbinet: I'm sorry @juliushaertl, but what is the actual linked callout vue component you're talking about ?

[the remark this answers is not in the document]
mejo-: @Jerome-Herbinet https://github.com/nextcloud/text/blob/main/src/nodes/CodeBlockView.vue is the component @juliushaertl is talking about.

See the `<script>` block at the end of this file. There you can overwrite the behaviour for `p .paragraph-content`, but as @juliushaertl said it should only be done if it's the last `p` element.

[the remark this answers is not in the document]
Jerome-Herbinet: Thanks for your answer @mejo- but there is no `p .paragraph-content` css property to overwrite in https://github.com/nextcloud/text/blob/main/src/nodes/CodeBlockView.vue `<style>` block 🤔

mejo- (approved):
Turned out `Callout.vue` already had a rule to remove `margin-bottom` for the last child, but it was broken. I took the liberty to push a fix for this.

PR #5312, opened 2024-01-24
fix: register viewer by uniq name instead of "component"

ShGKme:
### 📝 Summary

- Anonymous function in object property has a name of the property - component
- This name is used for global viewer component registration
- `component` is not a valid name, because it is a built-in name for Vue and it registers a global component
- Using named function to have an explicit name

Probably should be handled by `Viewer`, but I haven't found any other app with async component registration. I guess it is not possible to use 2 viewers with dynamic anonymous components.

image

### 🏁 Checklist

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -26,13 +26,23 @@ import { openMimetypesMarkdown, openMimetypesPlainText } from './helpers/mime.js
 __webpack_nonce__ = btoa(OC.requestToken) // eslint-disable-line
 __webpack_public_path__ = OC.linkTo('text', 'js/') // eslint-disable-line
 
+/**
+ * Wrapper for async registration of ViewerComponent.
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

ShGKme:
@skjnldsv Do you think this should be handled by Viewer?
Currently Text is the only app that uses dynamic registration (according to the search in GitHub)

skjnldsv:
Just stumbled upon this.
We never claimed to support async components :see_no_evil: 

image
https://github.com/nextcloud/viewer/blob/c3e5c751b0decef5dbc103474208b6142f9deef9/src/services/Viewer.js#L11-L21

https://github.com/nextcloud/viewer/blob/d531abe602996292c8c0c45d0c525afcdc58bea6/src/views/Viewer.vue#L853-L854
I'm actually surprised this never blew up

PR #5327, opened 2024-01-31
fix: Only unpause focus traps that we paused ourselves

### LINKED ISSUE #5326: [BITV]: focus will be set on actions as far document will be saved
The problem is: even if the focus has been set on some NcActions button (like in a sidebar or avatar), after few seconds focus will be moved to one specific NcActions button on the top right. My assuming is that this happens when file will be saved:

Peek 2024-01-31 15-02

juliushaertl:
Fix #5326 

Still need to check if there is a more elegant approach, but this seems to fix it.

Also to double check, probably for a separate issue
- [ ] Having a link after a code block will allow tabbing out of the paused focus trap

### DIFF
--- a/src/extensions/FocusTrap.js
+++ b/src/extensions/FocusTrap.js
@@ -1,5 +1,7 @@
 import { Extension } from '@tiptap/core'
 
+let ownPaused = false
+
 const toggleFocusTrap = ({ editor }) => {
 	const trapStack = window._nc_focus_trap ?? []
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter (approved):
Seems to be fixed, thank you Julius!

Solution should not be perfect for now. Let's merge if it solve a problem ;)

juliushaertl:
Needs https://github.com/nextcloud/text/pull/5318 first

PR #5367, opened 2024-02-06
feat: Move to vite for bundling

juliushaertl:
- [x] Remove webpack config
- [x] Clean up dependencies

Improves js assets quite a lot, while more individual files are loaded the overall size by ~80%

### Before 17.43MB

Screenshot 2024-02-06 at 18 47 09

### After 3.94MB

Screenshot 2024-02-06 at 18 45 45


max-nextcloud:
Build is also failing for me locally with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` when `rendering chunks (466)`. Looks like one of the last chunks is too big or so.

`--sourcemap false` let's the build succeed.

max-nextcloud:
`NODE_OPTIONS: --max-old-space-size=4096` also works.

max-nextcloud:
in public read only shares prosemirror.scss is not getting applied (did not look at editable shares yet)

max-nextcloud (approved):
I think this is ready to go now. Would love to get another pair of eyes on the changes I made myself though.

max-nextcloud:
@juliushaertl I cannot ask you for review because it's your PR. Would you take a look and merge if you think it's ready?

PR #5378, opened 2024-02-11
Cleanup `_oc_webroot` stubs where possible

max-nextcloud:
- test(jest): remove internal _oc_webroot stub
- test(cy): introduce `cy.prepareWindowForSessionApi`
- test(cy): refactor to use aliases
- test(cy): introduce `cy.sessionUsers()` for users endpoint
- test(api): split test and use `.its` and `.should`

### DIFF
--- a/cypress/e2e/api/SessionApi.spec.js
+++ b/cypress/e2e/api/SessionApi.spec.js
@@ -34,11 +34,7 @@ describe('The session Api', function() {
 
 	before(function() {
 		cy.createUser(user)
-		window.OC = {
-			config: { modRewriteWorking: false },
-			webroot: '',
-		}
-		window._oc_webroot = ''
+		cy.prepareWindowForSessionApi()
 	})
 
 	beforeEach(function() {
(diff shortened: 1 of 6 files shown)


juliushaertl (approved):
Looks good

just one failure remaining

```

FAIL src/tests/services/AttachmentResolver.spec.js
  Image resolver
    ✓ is a class with one constructor argument (29 ms)
    ✓ handles .attachments urls to own fileId via Text API (2 ms)
    ✓ leaves urls unchanged if they can be loaded directly (1 ms)
    ✓ handles .attachments urls to different fileId via webdav with text API fallback
    ✕ handles non-native urls wia webdav (3 ms)

  ● Image resolver › handles non-native urls wia webdav

    expect(received).toBe(expected) // Object.is equality

    Expected: "http://localhost/remote.php/dav/files/user-uid/parentDir/path/to/some%20image.png"
    Received: "http://localhost//remote.php/dav/files/user-uid/parentDir/path/to/some%20image.png"

      88 | 		const attachment = await resolver.resolve(src)
      89 | 		expect(attachment.isImage).toBe(true)
    > 90 | 		expect(attachment.previewUrl).toBe('http://localhost/remote.php/dav/files/user-uid/parentDir/path/to/some%20image.png')
         | 		                              ^
      91 | 	})
      92 |
      93 | })

      at Object.toBe (src/tests/services/AttachmentResolver.spec.js:90:33)
```

PR #5393, opened 2024-02-19
fix: Add alt attribute to image nodes

juliushaertl:
### 📝 Summary

It seems otherwise browser would read out the full URL with screen readers. We already have the alt text as user defined description for each image so we should make use of that.

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -41,6 +41,7 @@
 							<div class="media__wrapper">
 								<img v-show="loaded"
 									:src="imageUrl"
+									:alt="alt"
 									class="image__main"
 									@load="onLoaded">
 								<div class="metadata">


mejo- (approved):

AndyScherzinger (approved):

src/nodes/ImageView.vue
```diff
@@ -41,6 +41,7 @@
 							<div class="media__wrapper">
 								<img v-show="loaded"
 									:src="imageUrl"
+									:alt="alt"
 									class="image__main"
 									@load="onLoaded">
 								<div class="metadata">
```
susnux, line 47: @juliushaertl maybe we should hide the span below (`class="name"`) for a11y? meaning making it `aria-hidden="true"`. Because it duplicates the alt tag.

Issue #4750, opened 2023-08-31
Make code type fields copyable


gbertolottiKS:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to manually copy the content of code blocks, especially when they are huge.

**Describe the solution you'd like**
A simple copy button in the top right corner of every code blocks

**Additional context**
That feature is already present in many sites/editors, like GitHub:

![immagine](https://github.com/nextcloud/text/assets/115481064/c9cef2ca-6bb4-4e10-968a-b9e635631da6)

juliushaertl:
Nice idea :) Pull requests would be very welcome for this

XueSheng-GIT:
With NC 28.0.2 copying code blocks got even more complicated. The whole markdown syntax is now copied (see https://github.com/nextcloud/text/pull/5212) so that you cannot just paste the command itself to the shell anymore.

Issue #5331, opened 2024-02-02
Can no longer copy content from code blocks


nickvergessen:
**Describe the bug**
Ref https://github.com/nextcloud/text/pull/5212

**To Reproduce**
1. Have a collective page with a list and inside the list there is a code block:
```
* List item 1
* List item 2
    ```
    ### Workarounds
    
    
    
    * Disable app X
    * Change config setting Y
    * No workaround available
    ```
```

2. Copy bits inside the code block
3. Paste any where

**Expected behavior**
You paste:
```
### Workarounds

<!-- Remove not applicable sections -->

* Disable app X
* Change config setting Y
* No workaround available
```

**Actual behavior**
You paste with list item and code block wrap
![grafik](https://github.com/nextcloud/text/assets/213943/e97ba26d-56af-4ab8-9048-6462ba70e350)

**Screenshots**

juliushaertl:
One further related topic I've noticed is that copying a link that was pasted earlier results in a link wrapped in brackets `<https://github.com/nextcloud/text/issues/5331>`

juliushaertl:
@elzody Another suggestion from @susnux in https://github.com/nextcloud/text/pull/5212#issuecomment-1992150485 might also be interesting to check if that is feasible

elzody:
Closing since a workaround was implemented via #5395

susnux:
> Closing since a workaround was implemented via https://github.com/nextcloud/text/pull/5395.

I do not think this is a workaround, this is in most cases not usable, because often the code block contains multiple entries and you only want to copy one. E.g. email addresses, shell commands etc.

```shell
ls -ahlp
cat foo.bar
``` 

Copying `ls -ahlp` should not result in ```` ```ls -ahlp``` ```` you can not use that directly but again need an editor to paste and copy only the **selected** text.

elzody:
> I do not think this is a workaround, this is in most cases not usable, because often the code block contains multiple entries and you only want to copy one. E.g. email addresses, shell commands etc.

Makes sense. I will reopen, we're trying to get #5487 merged as @max-nextcloud made some changes today and we can further iterate on it in the future. This specific issue and the one you described should be fixed once this happens.

Sorry for the confusion, just trying to get things a bit more organized around this

PR #5395, opened 2024-02-19
Feat/4750 Make code type fields copyable

elzody:
### 📝 Summary

* Resolves: #4750 

This change introduces a code copy button on code blocks in applications that use Text (Notes, Collectives, Tables, ...). The button utilizes a method which was already present in the application for copying links from the link bubble view (src/mixins/CopyToClipboard.js), and makes changes to this function in order to make it more generalized so that it may be used for both purposes. The LinkBubbleView.vue component remains unaffected and should still work normally.

The button only appears when the code block is actually populated with text/code, and in Collectives it will appear in both Preview and Edit mode. I figured even those who do not have edit access may still need/want to copy the content of the code block.

### 🖼️ Screenshots
<details>
<summary>Collectives preview mode</summary>

grafik
</details>

<details>
<summary>Collectives edit mode</summary>
The button makes way for code block options when in Edit mode

grafik
</details>

<details>
<summary>Copy toast text</summary>
This text is now displayed when code from a code block or a link from a link bubble view is copied.

grafik
</details>

### 🚧 TODO

- [x] ...

### 🏁 Checklist

### DIFF
--- a/src/mixins/CopyToClipboardMixin.js
+++ b/src/mixins/CopyToClipboardMixin.js
@@ -10,23 +10,23 @@ export default {
 	},
 
 	methods: {
-		async copyToClipboard(url) {
+		async copyToClipboard(content) {
 			// change to loading status
 			this.copyLoading = true
 
 			// copy link to clipboard
 			try {
-				await navigator.clipboard.writeText(url)
+				await navigator.clipboard.writeText(content)
 				this.copySuccess = true
 				this.copied = true
 
 				// Notify success
-				showSuccess(t('collectives', 'Link copied'))
+				showSuccess(t('text', 'Copied to the clipboard'))
 			} catch (error) {
 				this.copySuccess = false
 				this.copied = true
 				showError(
-					`<div>${t('collectives', 'Could not copy link to the clipboard:')}</div><div>${url}</div>`,
+					`<div>${t('text', 'Could not copy to the clipboard')}</div>`,
 					{ isHTML: true })
 			} finally {
 				this.copyLoading = false
--- a/src/nodes/CodeBlockView.vue
+++ b/src/nodes/CodeBlockView.vue
@@ -19,8 +19,7 @@
 				<NcActions v-if="isEditable"
 					data-cy="code-action-group"
 					:aria-label="t('text', 'Code block options')">
-					<NcActionInput
-						:label="t('text', 'Code block language')"
+					<NcActionInput :label="t('text', 'Code block language')"
 						:value="type"
 						:show-trailing-button="false"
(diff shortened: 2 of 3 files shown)


src/nodes/CodeBlockView.vue
```diff
@@ -104,6 +125,9 @@ export default {
 		}
 	},
 	computed: {
+		hasCode() {
+			return this.node.textContent
```
juliushaertl, line 129: As this is safeguarded further down, do we need this here as well?

```suggestion
			return this.node?.textContent
```

juliushaertl (commented):
Tested, works like a charm 👍 

Just a small nitpick, and the cypress tests seem to need an adjustment as we have two buttons now

src/nodes/CodeBlockView.vue
elzody: My initial thought process was that, since it's just checking to see if there is any text content in the node (the code block has something written in it), it wouldn't matter if we used the optional chaining `?.` or not, but I think for consistency you're right, it should be added there as well. As for the cypress tests, I was just about to look into those and give it a fix. Thank you :)

src/mixins/CopyToClipboardMixin.js
```diff
@@ -10,23 +10,23 @@ export default {
...
 				this.copied = true
 
 				// Notify success
-				showSuccess(t('collectives', 'Link copied'))
+				showSuccess(t('text', 'Copied to the clipboard'))
 			} catch (error) {
 				this.copySuccess = false
 				this.copied = true
 				showError(
-					`<div>${t('collectives', 'Could not copy link to the clipboard:')}</div><div>${url}</div>`,
```
max-nextcloud, line 29: If I remember correctly the url was in here on purpose. The idea was that if your browser cannot copy the url for you its still displaying it so you can copy it yourself.

I'm not sure how often people would need to do that. I remember that back in the days I could not copy share links because of my browser configuration and did not find a way around that.

However in this case I think it's different. I could open the link in a tab and copy the link from there - or I right click the link and select 'copy link' from the context menu.

So I'm fine with removing the url - just wanted to explain the context.

max-nextcloud (approved):
Thanks a lot for the well prepared PR :tada: - looks good to me.

I have one small comment fyi and will approve any way as I think this is good to go as is. The things Julius brought up have been addressed as far as i can tell.

juliushaertl (approved):
Small comment on the test selector, feel free to fix and merge then

juliushaertl:
Backporting even though it is an enhancement as it helps users to work around https://github.com/nextcloud/text/issues/5331

Issue #5392, opened 2024-02-19
dev console of main flooded with errors `$attrs is readonly`


max-nextcloud:
**Describe the bug**

Current text main branch produces tons of errors in the console.

**To Reproduce**
1. Open current markdown file with text main branch currently.

**Expected behavior**
No errors in the console.

**Screenshots**
![grafik](https://github.com/nextcloud/text/assets/97337118/4f4c0816-5e00-44a9-8057-150006c8f3db)

max-nextcloud:
https://github.com/nextcloud-libraries/nextcloud-vue/issues/3281 might fix it.

max-nextcloud:
> [nextcloud-libraries/nextcloud-vue#3281](https://github.com/nextcloud-libraries/nextcloud-vue/issues/3281) might fix it.

As far as i can tell we are already using that fix.

max-nextcloud:
https://github.com/nextcloud-libraries/webpack-vue-config/pull/533/files should be fixed as well but i am still seeing the error:
![grafik](https://github.com/nextcloud/text/assets/97337118/9b7d7dea-c6fb-4ac8-8c47-741840b7c77b)

I wonder if we somehow messed up our inheritance from `webpack-vue-config`

max-nextcloud:
I was able to narrow this down to the `LinkBubbleView`. If i replace it with a Vue component with a simple static template the errors are gone.

max-nextcloud:
There's another warning drowning in the sea of errors: `It looks like you are using a custom button as a <NcPopover> or other popover #trigger. If you are not using <NcButton> as a trigger, you need to bind attrs from the #trigger slot props to your custom button. See <NcPopover> docs for an example.`

max-nextcloud:
Looks like `VueRenderer` will [create a new vue instance by default](https://github.com/ueberdosis/tiptap/blob/a863e1c49a0531ddfe06c4e73a427c109a4757db/packages/vue-2/src/VueRenderer.ts#L8). In contrast `VueNodeViewRenderer` has [a fix to ensure the same vue instance is used](https://github.com/ueberdosis/tiptap/commit/5bc77081b7af182dfbbb8547e5934f071ee7aeac)

max-nextcloud:
I tried to use the same approach with no success:

```diff
diff --git a/src/extensions/LinkBubblePluginView.js b/src/extensions/LinkBubblePluginView.js
index bddb60f8b..08c8b7145 100644
[...]

PR #5402, opened 2024-02-20
fix(viewer): share Vue constructor from Viewer

ShGKme:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/5392

Briefly: `<ViewerComponent>` from Text is rendered in Viewer's so everything that it uses must use the same Vue constructor as Viewer. Otherwise, there is a mix of Viewer's Vue constructor and Text's one in a single Virtual DOM, which is not fully supported by Vue.

More details: https://github.com/nextcloud-libraries/nextcloud-vue/issues/5267#issuecomment-1955229436

I don't know much about `tiptap`. Probably there is a cleaner way to provide Vue constructor to the plugin than using shared `src/ViewerVue.js`.

Also, the best to use this everywhere instead of `import Vue from 'vue'` in modules that can be used in Viewer.

### 🏁 Checklist

### DIFF
--- a/src/ViewerVue.js
+++ b/src/ViewerVue.js
@@ -1,0 +1,16 @@
+/**
+ * @type {import('Vue').VueConstructor|null}
+ */
+let ViewerVue = null
+
+/**
+ * @param {import('Vue').VueConstructor} VueConstructor - Vue constructor from Viewer
+ */
+export const setViewerVue = (VueConstructor) => {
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):
Tested and works. Thanks for your help here @ShGKme

max-nextcloud (approved):
Awesome! Thanks a lot for tackling this!
# :tada:

Issue #5324, opened 2024-01-30
"Link to website" entry keeps resetting when trying to add/change link


the-dipsy:
**Describe the bug**
When trying to change or add a link, the "Link to website" text field keeps resetting to its original value (or an empty string respectively).

**To Reproduce**
1. Open a file in Nextcloud Text
2. Click on the "Insert link" tool
3. Hover the "Link to website" input
4. Enter some text
5. See the input being reset while you type

**Expected behavior**
Being able to type, insert or edit links consistently.

**Screenshots**
![nextcloud-text-bug](https://github.com/nextcloud/text/assets/155256047/032311bb-1b3e-4b8e-b95a-d6c3660c7299)

**Server details:**
- Nextcloud version: 28.0.1
- PHP Version: 8.2.15
- Database: sqlite

**Client details:**
 - OS: Debian 12 (on the host and in the container from docker.io/nextcloud)
 - Browser: Firefox (also tested in iOS Safari)
 - Browser version: 115.6.0esr
 - Device: Desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
{"reqId":"PPW8VwliDky5Be53xY2m","level":2,"time":"2024-01-30T20:44:42+00:00","remoteAddr":"10.0.2.100","user":"--","app":"no app in context","method":"GET","url":"/","message":"Host localhost was not connected to because it violates local access rules","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0","version":"","data":[]}
{"reqId":"PPW8VwliDky5Be53xY2m","level":2,"time":"2024-01-30T20:44:42+00:00","remoteAddr":"10.0.2.100","user":"--","app":"no app in context","method":"GET","url":"/","message":"Host localhost was not connected to because it violates local access rules","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0","version":"","data":[]}
{"reqId":"clSOyCFbuMemnCk7CbHx","level":2,"time":"2024-01-30T20:44:52+00:00","remoteAddr":"10.0.2.100","user":"--","app":"no app in context","method":"POST","url":"/index.php","message":"Host localhost was not connected to because it violates local access rules","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0","version":"","data":[]}
[...]

Issue #5003, opened 2023-11-17
mailto links are not possible


q-wertz:
**Describe the bug**
When trying to add a mailto link the form field starts to get sluggish, the cursor jumps to random positions, text gets deleted and a `https://` is added. 
I think it happens as soon as one adds the `@` symbol...

**To Reproduce**
1. write e.g. kontakt@randomcompany.com
2. Mark it
3. Click on "Insert Link"
4. Try to add mailto:kontakt@randomcompany.com (which should open to opening the OS default mail client)

**Expected behavior**
When clicking on the link a new mail window of the OS default mail client opens and one can start writing an email. 

**Screenshots**
-

**Server details:**
- Nextcloud version: 27.1.3
- PHP Version: 8.1.25
- Database: PostgreSQL 13.10

**Client details:**
 - OS: Manjaro
 - Browser: Firefox
 - Browser version: 119.0.1
 - Device: Desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
Nothing relevant

#### Browser log
Not sure if related:
```
You need to fill either the text or the ariaLabel props in the button component. 
Object { text: undefined, ariaLabel: null }
​
ariaLabel: null
​
text: undefined
​
<prototype>: Object { … }
 
Object { _uid: 1955, _isVue: true, __v_skip: true, _scope: {…}, "$options": {…}, _renderProxy: {…}, _self: {…}, "$parent": {…}, "$root": {…}, "$children": (1) […], … }
index.module.js:2:812000
```

</details>

PR #5422, opened 2024-02-28
Several link handling fixes

mejo-:
### 📝 Summary

* fix(LinkBubble): Remove custom blur/focus handling
* fix(LinkBubble): Use debounce instead of custom debounce implementation
* fix(link): Insert full URLs for links to Nextcloud files
* fix(link): Don't rewrite links with fileId inside Collectives
* fix(link): Prevent middle-click from opening two tabs on Firefox
* test(cypress): Don't hide LinkBubble on scroll event in Cypress
* fix(ActionInsertLink): Sync NcActionInput value property (Fixes: #5324, Fixes: #5003)

### 🏁 Checklist

### DIFF
--- a/src/extensions/LinkBubblePluginView.js
+++ b/src/extensions/LinkBubblePluginView.js
@@ -10,7 +10,6 @@ const updateDelay = 250
 class LinkBubblePluginView {
 
 	#component = null
-	#preventHide = false
 	#hadUpdateFromClick = false
 	#updateDebounceTimer = undefined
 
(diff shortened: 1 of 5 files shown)


src/extensions/LinkBubblePluginView.js
```diff
@@ -132,13 +103,7 @@ class LinkBubblePluginView {
...
 		}
 
-		if (this.#updateDebounceTimer) {
-			clearTimeout(this.#updateDebounceTimer)
-		}
-
-		this.#updateDebounceTimer = window.setTimeout(() => {
-			this.updateFromSelection(view)
-		}, updateDelay)
+		debounce(() => this.updateFromSelection.bind(this)(view), updateDelay, { immediate: true })
```
juliushaertl, line 106: Does this debounce actually have an effect? As far as I know it would still call every time, unless you would call the returned function

juliushaertl (changes requested):
Looks good .. just one concern

src/extensions/LinkBubblePluginView.js
mejo-: @max-nextcloud to the rescue :grimacing: I still struggle to understand the different scopes of JS functions :see_no_evil:

mejo-: Ah sorry, now I got what you mean. Indeed, debounce just returns a function that needs to be called. I thought it would be about the scope of  `this` in `updateFromSelection()`. Anyways, should be fixed now :relaxed:

mejo-:
Commit "fix(ActionInsertLink): Sync NcActionInput value property" should get backported to stable28, probably also to stable27.

juliushaertl (approved):

PR #5427, opened 2024-03-01
fix: Expose css var for maximum full size widgets

juliushaertl:
Expose a css var to be used for full size widgets like https://github.com/nextcloud/deck/pull/5630


blizzz (approved):
:elephant:

PR #5439, opened 2024-03-06
fix: Disable nested embedding fully

juliushaertl:
We want to prevent nesting interactive widgets, so this will ensure that we know about that text is rendered as an embedding with https://github.com/nextcloud/server/pull/44012

## Follow up to check

- [ ] Check if we can just use inject in the paragraph component if we use the viewer vue instance

### DIFF
--- a/src/EditorFactory.js
+++ b/src/EditorFactory.js
@@ -49,12 +49,13 @@ const loadSyntaxHighlight = async (language) => {
 	}
 }
 
-const createEditor = ({ language, onCreate, onUpdate = () => {}, extensions, enableRichEditing, session, relativePath }) => {
+const createEditor = ({ language, onCreate, onUpdate = () => {}, extensions, enableRichEditing, session, relativePath, isEmbedded = false }) => {
 	let defaultExtensions
 	if (enableRichEditing) {
 		defaultExtensions = [
 			RichText.configure({
 				relativePath,
+				isEmbedded,
 				component: this,
 				extensions: [
 					Mention.configure({
--- a/src/nodes/Preview.vue
+++ b/src/nodes/Preview.vue
@@ -29,7 +29,8 @@
 			:value.sync="value"
 			@update:value="convertToParagraph" />
 		<NcReferenceList :text="node.attrs.href"
-			:limit="1" />
+			:limit="1"
+			:interactive="!extension.options.isEmbedded" />
 	</NodeViewWrapper>
 </template>
 
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -181,6 +181,7 @@ export default {
 
 		return val
 	},
+	inject: ['isEmbedded'],
 	props: {
 		richWorkspace: {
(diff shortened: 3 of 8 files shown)


skjnldsv:
One would argue that this is not an official Viewer prop yet :see_no_evil: 
But it make sense, we'll need to consider this for Viewer 3

src/nodes/Preview.vue
```diff
@@ -28,8 +28,10 @@
 		<PreviewOptions v-if="editor.isEditable"
 			:value.sync="value"
 			@update:value="convertToParagraph" />
-		<NcReferenceList :text="node.attrs.href"
-			:limit="1" />
+		<NcReferenceList v-if="!extension.options.isEmbedded"
+			:text="node.attrs.href"
+			:limit="1"
+			:interactive="!extension.options.isEmbedded" />
```
max-nextcloud, line 34: The conditions on the `v-if` and the `interactive` are the same.
If we don't display the element at all there's no point in setting interactive to false.
I think we need to pick what we want here.

src/EditorFactory.js
```diff
@@ -49,12 +49,13 @@ const loadSyntaxHighlight = async (language) => {
...
 }
 
-const createEditor = ({ language, onCreate, onUpdate = () => {}, extensions, enableRichEditing, session, relativePath }) => {
+const createEditor = ({ language, onCreate, onUpdate = () => {}, extensions, enableRichEditing, session, relativePath, isEmbedded = false }) => {
 	let defaultExtensions
 	if (enableRichEditing) {
 		defaultExtensions = [
 			RichText.configure({
 				relativePath,
+				isEmbedded,
```
max-nextcloud, line 58: I don't think we call `creditEditor` with `isEmbedded` set to true anywhere. `isEmbedded` also defaults to false. So i think the changes to this file are not needed unless you are preparing something else on top of the changes already in this PR.

max-nextcloud (commented):
Just two minor comments. One of them might point at a conceptual issue though... Do we want to render the previews inside embedded content non-interactive or not at all?

If we don't render the Preview nothing will be shown - not even the embedded link. So maybe we leave out the `Preview` nodes all together then and at least show the link?

src/EditorFactory.js
juliushaertl: We do in https://github.com/nextcloud/text/pull/5439/files#diff-9a196164961cc525c2362569d39ae48b643c944dc108fc5843cd52c31600074eR546

src/nodes/Preview.vue
juliushaertl: Yes, we needed https://github.com/nextcloud/server/pull/44012#discussion_r1515875561 first, but let me remove the v-if now that this is in

max-nextcloud (approved):

src/nodes/Preview.vue
skjnldsv: merged

Issue #5442, opened 2024-03-07
Link preview follow up


juliushaertl:
Follow up from https://github.com/nextcloud/text/pull/5399

- [x] Do not render interactive widgets in link bubble (set interactive prop to false on the NcReferenceList component)
- [x] #5443 
- [x] #5444 
- [x] #5445 
- [x] #5518
- [x] #5522
- [x] #5524
- [x] #5539
- [x] #5723 
- [x] test for backspace (follow up on #5444)
- [ ] ~~edit link target from menu~~

ChristophWurst:
![image](https://github.com/nextcloud/text/assets/1374172/bdd3178c-d04d-4037-bdfb-8d70456dcd07)

^ I've noticed that when the link preview fails with the github integration there is no way to click your way to the link. You have to use the *Copy link to clipboard* action and open a new tab yourself.

I can file a separate issue if this isn't the right place to post this.

juliushaertl:
@max-nextcloud Editing the link target from the modal might be tricky as the mark might span multiple nodes or also overlap with other marks. I'd say something we should split in a separate ticket as it might be larger work.

max-nextcloud:
Postponing the editing of link targets until the next iteration on links.

PR #5445, opened 2024-03-07
fix(Preview): hide link bubble when preview menu opens

max-nextcloud:
### 📝 Summary

* Only show link bubble when cursor is *inside* the link. This way it does not show when opening the preview menu (which moves the cursor to the start of the line).
* Part of #5442 .

### 🏁 Checklist

### DIFF
--- a/src/marks/Link.js
+++ b/src/marks/Link.js
@@ -21,8 +21,8 @@
  */
 
 import TipTapLink from '@tiptap/extension-link'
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):

max-nextcloud:
failing tests are related. They expect the bubble to show when cursor is on first char.

PR #5464, opened 2024-03-12
test(ci): Revert some cypress workflow changes

mejo-:
Apparently Cypress tests still fail on our own self-hosted runners for some reason.

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -13,13 +13,13 @@ env:
 
 jobs:
   init:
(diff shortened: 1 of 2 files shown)


mejo-:
>  Also when just checking the diff again we should also revert the change from https://github.com/nextcloud/text/pull/5449/files#diff-b0f20e5dcba1fb11d1e8ae4441aa555561c825279506cb3151d3827cd076ca2aR24 or upstream it to the .github repo

Good point! Done in https://github.com/nextcloud/.github/pull/324

Issue #5368, opened 2024-02-07
403 on public share access


KarlKeu00:
**Describe the bug**
There is no way to deactivate the collaboration feature. When I share a folder and the end user is not authenticated, requests are still sent to `/apps/text/public/session/push`. These are rejected with 403 and lead to the user being blocked.

**To Reproduce**
1. Publish a folder in read-only mode for public access.
2. Open the link in an incognito window.
3. Go to the browser console and search for "push" in the requests.

**Expected behavior**
No request to push or the option to deactivate the collaboration feature.

**Screenshots**

**Server details:**
- Nextcloud version: 27.1.6
- PHP Version: Docker-Container (stable-apache)
- Database: MariaDB 10.6

**Client details:**
 - OS: macOS
 - Browser: Brave
 - Browser version: v1.62.156
 - Device: desktop

KarlKeu00:
The following error also occurs:
```
TypeError: Cannot read properties of undefined (reading 'userId')
    at s.<anonymous> (Status.vue:1:857)
    at t._render (vue.runtime.esm.js:2700:28)
    at s.r (vue.runtime.esm.js:3891:27)
    at t.get (vue.runtime.esm.js:3462:33)
    at new t (vue.runtime.esm.js:3452:51)
    at t (vue.runtime.esm.js:3908:5)
    at li.$mount (vue.runtime.esm.js:8797:12)
    at init (vue.runtime.esm.js:4427:19)
    at vue.runtime.esm.js:6591:17
    at h (vue.runtime.esm.js:6545:13)
```

Issue #5366, opened 2024-02-06
403 errors in NGINX


jameskimmel:
**Describe the bug**
Since the new nextcloud update to 28.0.2, there are mutliple 403 errors in the NGINX log regarding the text app. 
This causes apps like Crowdsec to block users. 

**To Reproduce**
 Install nextcloud behind a NGINX proxy. Let users use text in browser. See errors in NGINX logs. 

**Expected behavior**
No 403 errors. 

**Server details:**
- Nextcloud version: 28.0.2
- PHP Version: 8.1

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
Nothing in Nextcloud logs.

#### NGINX log
```
ClientIP - - [05/Feb/2024:20:42:54 +0100] "POST /apps/text/session/174648/push HTTP/2.0" 200 111 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Ge>
ClientIP - - [05/Feb/2024:20:42:55 +0100] "POST /apps/text/api/v1/users HTTP/2.0" 403 0 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chro>
ClientIP - - [05/Feb/2024:20:42:55 +0100] "POST /apps/text/session/174648/push HTTP/2.0" 200 130 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Ge>

```

</details>

I know this isn't much, please let me know if I can provide you any further logs or something else to track down the issue.

Issue #5223, opened 2024-01-10
Failed to write to document errors in readonly sessions


mejo-:
**Describe the bug**
When opening Text documents in read-only mode (e.g. shared documents), Text regularly tries to send push messages to the backend that get declined:

```
POST https://nextcloud.local/index.php/apps/text/public/session/17142/push [HTTP/2 403  66ms]
[ERROR] text: failed to write to document - not allowed
[ERROR] text: Unexpected Error "Failed to apply steps. Retry!" 
```

**To Reproduce**
1. Open Text document in read-only view (e.g. by sharing the document read-only)
2. Open the developer console
3. Maybe you have to click around in the document, not sure about that.

**Screenshots**
![2024-01-10T12:13:11,375500483+01:00](https://github.com/nextcloud/text/assets/3582805/4d93fdc8-bbbf-4301-bf0e-43e6162e9960)

**Server details:**
- Nextcloud version: development branch (29.0.0 dev)

juliushaertl:
Not sure what is best but I think we have a few options:
- Do not send out awareness messages on the client side
- Accept read only users awareness messages on the push endpoint (even with read only)
  - Either ignore any pushed messages
  - Or filter out and only allow pushing awareness messages

mejo-:
I don't think we should send awareness messages on a connection that got initiated as read-only (e.g. due to missing write privileges).

My understanding is that we currently don't differentiate between "on purpose read-only" (e.g. when connecting to a document session without write privileges) and "temporary read-only" (e.g due to connectivity issues). In the former case I think we should stop the client from sending any awareness messages. In the latter case we should keep the current logic of throwing an error and retrying later.

PR #5466, opened 2024-03-12
fix(backend): Accept pushs with only step1 messages by read-only clients

mejo-:
### 📝 Summary

This fixes read-only clients getting 403 responses on push requests that only contain questions for updates.
    
* Fixes: #5223
* Fixes: #5366
* Fixes: #5368

### 🏁 Checklist

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -206,10 +206,12 @@ class DocumentService {
 	}
 
 	/**
-	 * @throws DoesNotExistException
 	 * @throws InvalidArgumentException
+	 * @throws NotFoundException
+	 * @throws NotPermittedException
+	 * @throws DoesNotExistException
 	 */
-	public function addStep(Document $document, Session $session, array $steps, int $version): array {
+	public function addStep(Document $document, Session $session, array $steps, int $version, ?string $shareToken): array {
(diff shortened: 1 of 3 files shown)


lib/Service/DocumentService.php
```diff
@@ -206,10 +206,12 @@ public function writeDocumentState(int $documentId, string $content): void {
...
 
 	/**
-	 * @throws DoesNotExistException
 	 * @throws InvalidArgumentException
+	 * @throws NotFoundException
+	 * @throws NotPermittedException
+	 * @throws DoesNotExistException
 	 */
-	public function addStep(Document $document, Session $session, array $steps, int $version): array {
+	public function addStep(Document $document, Session $session, array $steps, int $version, ?string $token): array {
```
juliushaertl, line 214:
```suggestion
	public function addStep(Document $document, Session $session, array $steps, int $version, ?string $shareToken): array {
```

Maybe we call it like that to make it more clear that this is not the text session token

juliushaertl (approved):
Small naming nitpick, otherwise looks sane 👍

lib/Service/DocumentService.php
mejo-: Agreed and changed :relaxed:

PR #5470, opened 2024-03-13
fix: Avoid race condition that may initialize a document twice on the clients

juliushaertl:
This fix may help with duplicate content of documents.

While staring at code and reading https://github.com/yjs/yjs-demos/issues/16 I noticed we may have a race condition where two browsers would attempt to create a new document, while that is catched the session receiving the unique constraint exception would still continue with the freshSession value set to true and handing over the file content and using that to initialize the document again locally.


mejo- (approved):
Looks sensible to me :ok_hand:

max-nextcloud (approved):

Issue #5476, opened 2024-03-13
Cleanup document table entries and y.js state file handling


juliushaertl:
From our discussion:

### close

- delete session table entry
- delete inactive sessions (without steps)
- document is kept
- yjs file is kept
- steps are kept

### cron cleanup

- foreach document without sessions and unsaved changes
  - reset document
    - delete all steps
    - delete all sessions
    - delete document
    - keep y.js file
  
- delete inactive sessions (without steps) -> can trigger document cleanup in the next cron run
- sessions with steps and the document are kept

## Goal: support both keeping and removing y.js file

- [x] do not drop document db entry, only if y.js file is also removed (Julius)
- [ ] ensure reconnects can detect an underlying y.js file has been recreated
  - [x] backend: write unique id on insert to oc_text_documents for the base_version_etag (Julius)
  - [ ] frontend: on reconnect compare client base_version_etag with server one, full editor restart if no match (Max)
- [x] drop y.js state only on external file writes if no sessions are active (Jonas)
  - this will ensure we don't have a conflict dialog
- [x] check node deletion event for cleanup (follow up)

mejo-:
Scenarios (to be complemented):

* Session got removed in backend but document+yjs stayed (`baseVersionEtag` still the same) -> Frontend reconnect sufficient
* Sessions and document+yjs got removed in backend: -> Frontend needs a full reset (page reload) to load new yjs state

mejo-:
stable28 backport: https://github.com/nextcloud/text/pull/5541

Issue #5420, opened 2024-02-27
Content duplicates


juliushaertl:
I have gotten a report about a file content being duplicated. Checking the file and when this happend I've seen the following access log

Save highlighted and some possibly suspicious large pushs

- [26/Feb/2024:17:37:35 +0000] "PUT /apps/text/session/8319994/create HTTP/1.1" 200 2072 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 326482
- [26/Feb/2024:17:37:36 +0000] "POST /apps/text/session/8319994/push HTTP/1.1" 200 **4561** "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 298094
- [26/Feb/2024:17:37:36 +0000] "PUT /apps/text/session/8319994/create HTTP/1.1" 200 5425 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 262602
- [26/Feb/2024:17:37:37 +0000] "POST /apps/text/session/8319994/sync HTTP/1.1" 200 **4755** "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 240192
- [26/Feb/2024:17:37:37 +0000] "POST /apps/text/session/8319994/close HTTP/1.1" 200 804 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 258176
- [26/Feb/2024:17:37:37 +0000] "POST /apps/text/session/8319994/sync HTTP/1.1" 200 4757 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 196571
- [26/Feb/2024:17:37:38 +0000] "POST /apps/text/session/8319994/push HTTP/1.1" 200 4523 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 306216
- [26/Feb/2024:17:37:38 +0000] "PUT /apps/text/session/8319994/create HTTP/1.1" 200 5464 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0" 81013
[...]

PR #5477, opened 2024-03-13
Ensure document stays consistent with y.js state file

juliushaertl:
Contributes to #5476 

- fix: Clean up logic to return document state file or file content
- fix: Set base version etag to a unique id per document creation
- fix: Only cleanup earlier steps that haven't been persisted yet to the state file
- fix: Keep document table entry and only cleanup when we force it

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -136,25 +136,19 @@ class ApiService {
 
 		$session = $this->sessionService->initSession($document->getId(), $guestName);
 
+		$documentState = null;
+		$content = null;
 		if ($freshSession) {
 			$this->logger->debug('Starting a fresh editing session for ' . $file->getId());
-			$documentState = null;
 			$content = $this->loadContent($file);
 		} else {
 			$this->logger->debug('Loading existing session for ' . $file->getId());
-			$content = null;
 			try {
 				$stateFile = $this->documentService->getStateFile($document->getId());
 				$documentState = $stateFile->getContent();
+				$this->logger->debug('Existing document, state file loaded ' . $file->getId());
 			} catch (NotFoundException $e) {
-				$this->logger->debug('State file not found for ' . $file->getId());
-				$documentState = ''; // no state saved yet.
-				// If there are no steps yet we might still need the content.
-				$steps = $this->documentService->getSteps($document->getId(), 0);
-				if (empty($steps)) {
-					$this->logger->debug('Empty steps, loading content for ' . $file->getId());
-					$content = $this->loadContent($file);
-				}
+				$this->logger->debug('Existing document, but state file not found for ' . $file->getId());
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


lib/Service/ApiService.php
```diff
@@ -136,25 +136,19 @@ public function create(?int $fileId = null, ?string $filePath = null, ?string $t
...
 			$content = $this->loadContent($file);
 		} else {
 			$this->logger->debug('Loading existing session for ' . $file->getId());
-			$content = null;
 			try {
 				$stateFile = $this->documentService->getStateFile($document->getId());
 				$documentState = $stateFile->getContent();
+				$this->logger->debug('Existing document, state file loaded ' . $file->getId());
 			} catch (NotFoundException $e) {
-				$this->logger->debug('State file not found for ' . $file->getId());
```
juliushaertl, line 150: This might be another cause of #5420 and is not needed as we only want to load from the file content on the initial session creation

lib/Service/ApiService.php
```diff
@@ -136,25 +136,19 @@ public function create(?int $fileId = null, ?string $filePath = null, ?string $t
...
-				$documentState = ''; // no state saved yet.
-				// If there are no steps yet we might still need the content.
-				$steps = $this->documentService->getSteps($document->getId(), 0);
-				if (empty($steps)) {
-					$this->logger->debug('Empty steps, loading content for ' . $file->getId());
-					$content = $this->loadContent($file);
```
juliushaertl, line 156: This might be another cause of https://github.com/nextcloud/text/issues/5420 and is not needed as we only want to load from the file content on the initial session creation

lib/Service/DocumentService.php
```diff
@@ -422,11 +422,12 @@ public function resetDocument(int $documentId, bool $force = false): void {
 
 			$this->unlock($documentId);
 
-			$this->stepMapper->deleteAll($documentId);
+			$this->stepMapper->deleteBeforeVersion($documentId, $document->getLastSavedVersion());
```
juliushaertl, line 425: This should still be fine as a cleanup as all steps in the saved version are also persisted in the y.js state, right @max-nextcloud ?

juliushaertl:
Need to check the failing test:

> 1) The session Api
>       race conditions
>         sends initial content if other session is alive but did not push any steps:

I feel we might want to adjust that to account for the changed behaviour.

lib/Service/DocumentService.php
mejo-: I actually changed the logic of `resetDocument()` in #5481 to always delete all session data: document, sessions and steps from database as well as yjs document state file. And that will happen in the following scenarios:

* A document gets changed from outside text and there's no unsaved steps
* Cleanup background job runs for a stale document session (and no unsaved steps)
* occ command `text:reset` is run

I wonder whether the second case should just remove sessions and steps instead.

mejo-:
> Need to check the failing test:
> 
>         The session Api
>         race conditions
>         sends initial content if other session is alive but did not push any steps:
> 
> I feel we might want to adjust that to account for the changed behaviour.

I changed the cypress test to expect empty file content on `/create` requests if there's already a document session.

I also did some manual testing with removing the session connections (from `oc_text_sessions`) while having the document open.

mejo- (approved):

PR #5481, opened 2024-03-13
Reset document session on filesystem change

mejo-:
Contributes to #5476

### 📝 Summary

1. **fix(backend): Reset document session and yjs file when file is deleted**
2. **fix(backend): Reset document session when updated from outside editor**
   When a text file is updated via other means than from the editor (e.g. when uploaded/synced via webdav) and there is no unsaved steps in the document session, reset the document session  This will prevent conflict resolution dialogs in this case. Client frontends will have to reload the document afterwards though.
3. **fix(backend): Remove yjs file and all steps when resetting document session**
   Instead of just deleting the newest steps, always remove all session data: document, sessions and steps from the database as well as the yjs (document state) file.
   Without the `--force` option, don't reset document sessions with unsaved steps.

### 🏁 Checklist

### DIFF
--- a/lib/Listeners/BeforeNodeWrittenListener.php
+++ b/lib/Listeners/BeforeNodeWrittenListener.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2024 Jonas <jonas@freesources.org>
+ *
+ * @author Jonas <jonas@freesources.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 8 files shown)


juliushaertl (approved):
Code looks good, will give it a spin tomorrow

lib/Listeners/BeforeNodeWrittenListener.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021 Julien Veyssier <eneiluj@posteo.net>
+ *
+ * @author Julien Veyssier <eneiluj@posteo.net>
```
max-nextcloud, line 7: I doubt that ;)

mejo-: Haha, indeed :laughing:

PR #5484, opened 2024-03-14
fix(attachments): Don't use currentSession for fetching attachments

mejo-:
When switching pages in Collectives, `currentSession` might still be the session from the last page. The readonly view (`MarkdownContentEditor`) doesn't get a session, so the outdated session is used.

Instead, always pass the session from `AttachmentResolver` to `setAttachmentList`.

Fixes: nextcloud/collectives#1096
Fixes: nextcloud/collectives#1112

### 🏁 Checklist


grnd-alt (approved):
went through the changes and tested locally. Working as expected 💯

PR #5487, opened 2024-03-14
fix: copy-paste markdown/raw text inconsistencies

### LINKED ISSUE #5331: Can no longer copy content from code blocks

elzody:
### 📝 Summary

Fix #5331

As described in #5331 and #5212, there are some issues with the copy-paste behavior and some inconsistencies regarding how markdown is copied vs. raw text. I think it would be helpful to define what the expected behavior would be in certain situations so that this can be refined.

This draft PR currently contains some additions I made which traverses ProseMirror Nodes upon copying text, and tries to determine if markdown should be copied or if the raw text should be copied. In my opinion, it works a bit better than before and behaves more consistently, but I did still notice some issues, such as losing block nesting sometimes (discovered when copying a code block nested in a list).

### Currently defined behavior

- [x] Copying from a code block should just paste the code in question
- [x] Copying multiple list items should result in something that resembles a list
- [x] Copying a larger fragment of a doc should result in markdown as this can be useful for creating github issues
- [x] Copying an address from inside a info block should only result in the address
- [x] Copying version numbers should not result in an escaped character (28\\.0.3 e.g.)

All testing and feedback is welcome.

### DIFF
--- a/src/extensions/Markdown.js
+++ b/src/extensions/Markdown.js
@@ -114,7 +114,7 @@ const Markdown = Extension.create({
 							} else if (slice.isLeaf) {
 								return slice.textContent
 							} else {
-								traverseNodes(slice.content.firstChild)
+								return traverseNodes(slice.content.firstChild)
 							}
 						}
 
--- a/src/tests/extensions/Markdown.spec.js
+++ b/src/tests/extensions/Markdown.spec.js
@@ -1,5 +1,7 @@
 import { Markdown } from './../../extensions/index.js'
 import { createMarkdownSerializer } from './../../extensions/Markdown.js'
+import CodeBlock from '@tiptap/extension-code-block'
+import Blockquote from '@tiptap/extension-blockquote'
 import Image from './../../nodes/Image.js'
 import ImageInline from './../../nodes/ImageInline.js'
 import TaskList from './../../nodes/TaskList.js'
@@ -7,6 +9,7 @@ import TaskItem from './../../nodes/TaskItem.js'
 import Underline from './../../marks/Underline.js'
 import TiptapImage from '@tiptap/extension-image'
 import { getExtensionField } from '@tiptap/core'
+import { __serializeForClipboard as serializeForClipboard } from '@tiptap/pm/view'
 import { createCustomEditor } from '../helpers.js'
 
 describe('Markdown extension unit', () => {
@@ -75,6 +78,39 @@ describe('Markdown extension integrated in the editor', () => {
 		})
 		const serializer = createMarkdownSerializer(editor.schema)
 		expect(serializer.serialize(editor.state.doc)).toBe('inline image ![Hello](test) inside text')
+	})
+
+	it('copies task lists to plaintext like markdown', () => {
+		const editor = createCustomEditor({
+			content: '<p><ul class="contains-task-list"><li><input type="checkbox">Hello</li></ul></p>',
+			extensions: [Markdown, TaskList, TaskItem],
+		})
+		editor.commands.selectAll()
+		const slice = editor.state.selection.content()
+		const { text } = serializeForClipboard(editor.view, slice)
+		expect(text).toBe('\n- [ ] Hello')
+	})
+
+	it('copies code block content to plaintext according to their spec', () => {
+		const editor = createCustomEditor({
+			content: '<pre><code>Hello</code></pre>',
+			extensions: [Markdown, CodeBlock],
+		})
+		editor.commands.selectAll()
+		const slice = editor.state.selection.content()
+		const { text } = serializeForClipboard(editor.view, slice)
+		expect(text).toBe('Hello')
+	})
+
+	it('copies nested task list nodes to markdown like syntax', () => {
+		const editor = createCustomEditor({
+			content: '<blockquote><p><ul class="contains-task-list"><li><input type="checkbox">Hello</li></ul></blockquote>',
+			extensions: [Markdown, Blockquote, TaskList, TaskItem],
+		})
+		editor.commands.selectAll()
+		const slice = editor.state.selection.content()
+		const { text } = serializeForClipboard(editor.view, slice)
+		expect(text).toBe('\n- [ ] Hello')
 	})
 
 })


src/extensions/Markdown.js
```diff
@@ -108,7 +108,17 @@ const Markdown = Extension.create({
...
 					},
 					clipboardTextSerializer: (slice) => {
-						return createMarkdownSerializer(this.editor.schema).serialize(slice.content)
+						const traverseNodes = (slice) => {
+							if (slice.content.childCount > 1) {
+								return createMarkdownSerializer(this.editor.schema).serialize(slice.content)
+							} else if (slice.isLeaf) {
+								return slice.textContent
+							} else {
+								traverseNodes(slice.content.firstChild)
```
juliushaertl, line 117: What scenario are we hitting with that, from a quick look it seems odd that there is no return and we just return the result of `traverseNodes(slice)` in this case still

juliushaertl (commented):
Gave this a quick test and think the behaviour is quite nice now, left a question inline

src/extensions/Markdown.js
susnux: I also think a return is missing here?

elzody: The intended goal was that, it there is more than one node selected, odds are the user is copying a group of things and would want to retain the markdown. That's why, in the first if block, the markdown serializer is returned. Otherwise, if that is not the case, but the slice is a leaf (the final node in the tree), then it's just a single item and we can copy the raw text and probably don't need the markdown. The else statement is basically like, if neither of those are true, then recurse back into the traverseNodes function to try again on a deeper level of the slice until we find something that can be returned. I hope that makes sense in some way?

susnux: Yes but you still need to return the output of the recursive `traverseNodes` otherwise in this case `undefined` is returned

elzody: > Yes but you still need to return the output of the recursive `traverseNodes` otherwise in this case `undefined` is returned

I see, okay, thank you

mejo-:
Nice, I just gave it a try and indeed it improves the copy and paste situation quite a lot! When adding the missing `return` to line 117 it is even better :laughing: 

Still, I want to share some feedback that I got from different users after Nextcloud instances got upgraded to 28: The fact that the markdown syntax is copied into the `text/plain` clipboard creates a confusing user experience for them. They use Text/Collectives as a log and do a lot of copy+paste from it into different other systems. For them, copying to the Signal Desktop client and to any mail client is literally broken the way it works now. And while this PR improves the situation a bit (when the copied content stays in one paragraph or block node), the annoyance stays when copying larger parts of a document.

In essence I think we should revert the behaviour to copy markdown syntax in to the `text/plain` clipboard, like @susnux already commented [here](https://github.com/nextcloud/text/pull/5212#issuecomment-1992150485).

To my understanding, the only use-case where having markdown content in the `text/plain` clipboard is helpful is when you copy+paste to Github. In most cases where you copy into a rich text editor, the `text/html` clipboard is used anyway. And in the majority of cases where you copy into a plaintext only editor, having markdown pasted is not what you expect from a user perspective.

Some concrete examples with a broken user experience:

* Copy a list of email address links that spans over several blocks (e.g. `[mail@example.org](mailto:mail@example.org)\n\n[mail2@example.org](mailto:mail2@example.org)` into the recipient field of any mail client (Thunderbird, Roundcube)
* Copy two paragraphs of text with bold, italic, etc formatting into the Signal Desktop Client or the text-only body field of your mail client

juliushaertl:
Results from the call discussing this:

- Bring back the old Ctrl-Shift-C shortcut to copy the markdown with the new traverseNodes logic
- Change back default to copy text
- For later (file as separate issue)
  - Traverse nodes when copying plain text, Lists/Blockquotes could still be preserved for example
  - There may be further cases that need some thinking on how they should be mapped to plaintext (e.g. links with label, mailto links)

max-nextcloud:
> Results from the call discussing this:
> 
>     * Bring back the old Ctrl-Shift-C shortcut to copy the markdown with the new traverseNodes logic
> 
>     * Change back default to copy text

My impression is that we are going back and forth between two coarse grained approaches. Once we run into the downsides of one, we switch back to the other... until we run into the downsides of the previous.

Let me just illustrate what the change to markdown improved:

| source | before | after |
|------ | ------- | ------- |
|grafik|grafik|grafik|

I'd clearly prefer the latter even when pasting into a mail or signal.

I'm not convinced we need two different copy and paste modes. After all the actual markdown source can be downloaded and I think there is limited use in the pure text content of the cells. Rather than offering two bad options I'd propose to iterate on a solution that produces a good enough output in most cases.

susnux:
I think for most users having "random" characters inside the copy result is unexpected behavior. I really have a lot complains from my users that can not copy anything from code blocks. Especially as codeblocks are used to preserve the formatting.

When copying multiple elements the behavior is often not defined, so it would be fine to have the markdown e.g. for a table.
But for e.g. code blocks - especially if only selected parts from it - then you expect to also only copy that part.

Another common solution is to populate the clipboard with multiple mimetypes, e.g. office programs often provide HTML besides plain text.

max-nextcloud:
> I think for most users having "random" characters inside the copy result is unexpected behavior. I really have a lot complains from my users that can not copy anything from code blocks. Especially as codeblocks are used to preserve the formatting.

I totally agree. This PR fixes these issues as far as i can tell.

> Thats why I suggested to either provide multiple mime types so the application pasted to can select the content it supports.

We do provide html and plain mime types. I'm not aware of any app that uses the plain/markdown mime type when pasting. So while this would be a good approach in general it would not solve the problems in the `before` column when pasting into plain text fields.

susnux:
> We do provide html and plain mime types. I'm not aware of any app that uses the plain/markdown mime type when pasting. So while this would be a good approach in general it would not solve the problems in the `before` column when pasting into plain text fields.

Yes makes sense

mejo-:
> My impression is that we are going back and forth between two coarse grained approaches. Once we run into the downsides of one, we switch back to the other... until we run into the downsides of the previous.
> 
> Let me just illustrate what the change to markdown improved:

Thanks for looking into this @max-nextcloud. While I agree with most of your analysis, the concrete pain points that [I lined out](https://github.com/nextcloud/text/pull/5487#issuecomment-2016445857) that users (regularly) report to me with the current implementation remain with this pull requests. I think two things need to change at least to satisfy them:

* links/URLs should be plain text, not in markdown syntax
* marks like bold, italic, underline have to be stripped off

Without these changes, copying from Text documents to any plain text processors will confuse people, and - most importantly - breaks their current workflows.

max-nextcloud (approved):

mejo- (approved):
After discussing this with @max-nextcloud I'm confident that you have a good path forward in mind. Makes sense to get this PR merged first and improve the situation in a folllow-up afterwards :blush:

PR #5523, opened 2024-03-18
fix(sync): If `baseVersionEtag` changed, reset frontend

mejo-:
### 📝 Summary

The PR builds on #5477 

* fix(sync): If `baseVersionEtag` changed, reset frontend
   `baseVersionEtag` changes when a new document session got initialized, e.g. after an old document session without session clients got cleaned up, or because the markdown file got changed via webdav.
   Detect this in the client and ask the user to reload the page for resetting the session.
* fix(Middleware): Response with 412 if `baseVersionEtag` doesn't match
* fix(DocumentStatus): Refactor and migrate to `NcNoteCard`

### Screenshot

| warning |
| --- |
| 2024-03-18T20:06:02,244342045+01:00 |

### 🚧 TODO

- [x] Testing

### 🏁 Checklist


cypress/e2e/sync.spec.js
```diff
@@ -126,6 +126,25 @@ describe('Sync', () => {
 			.should('include', 'after the lost connection')
 	})
 
+	it.only('shows warning when document session got cleaned up', () => {
```
max-nextcloud, line 129:
```suggestion
	it('shows warning when document session got cleaned up', () => {
```

cypress/support/sessions.js
```diff
@@ -50,16 +50,40 @@ Cypress.Commands.add('pushSteps', ({ connection, steps, version, awareness = ''
 		.then(response => response.data)
 })
 
+Cypress.Commands.add('failToPushSteps', ({ connection, steps, version, awareness = '' }) => {
+	return connection.push({ steps, version, awareness })
+		.then((response) => {
+			throw new Error('Expected request to fail - but it succeeded!')
+		})
+		.catch((err) => err.response)
```
max-nextcloud, line 58: I think the `throw` in `then` will be caught in the catch and cause a `undefined` to be returned as `err.response` is undefined for the exception thrown.
You probably want to handle both cases in one [`then(onFullfilled, onRejected)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#syntax) call:
```suggestion
		.then(
			() => throw new Error('Expected request to fail - but it succeeded!'),
			err => err.response
		)
```

max-nextcloud (approved):
Looks good. Just some minor nitpicks.

cypress/support/sessions.js
mejo-: Indeed, sounds logical. Changed now.

PR #5537, opened 2024-03-20
fix: Force visibility for tippy

### LINKED ISSUE #5448: Assistant button is currently broken
As CI shows the assistant button does not work on main. 

When checking out earlier it seems to have been caused by https://github.com/nextcloud/viewer/pull/2181

juliushaertl:
This seems to have changed with viewer changing from `@nextcloud/vue` from 8.4.0 to 8.8.1 but I could not figure out the difference there, however this brings back the assistant button (consider it a hotfix)

Also noted as a workaround in https://github.com/ueberdosis/tiptap/issues/4851

Fix #5448

### DIFF
--- a/cypress/e2e/Assistant.spec.js
+++ b/cypress/e2e/Assistant.spec.js
@@ -4,7 +4,7 @@ const currentUser = randUser()
 
 const fileName = 'empty.md'
 
-describe.skip('Assistant', () => {
+describe('Assistant', () => {
 	before(() => {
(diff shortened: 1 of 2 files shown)


juliushaertl:
php lint failure is unrelated (CI issue)

mejo- (approved):
Wow, that looks violent, but if it works :see_no_evil:

juliushaertl:
Yeah spent too much hours already into that, so hopefully can figure out the root cause when things are a bit more quiet

PR #5538, opened 2024-03-20
fix(LinkBubble): Display link bubble view only if href is set

mejo-:
This prevents requests to `/resolve?reference=.../null` when a document gets opened without a link being selected.

Contributes to nextcloud/collectives#1167

### 🏁 Checklist


max-nextcloud (approved):
lgtm

Issue #5535, opened 2024-03-20
Update icons in "New" menu


juliushaertl:
Contributing to https://github.com/nextcloud/server/issues/44340 we need to update the icons for new text files and the folder description accordingly in text.

we aim to have the icons updated for v29, the remaining parts can be ignored or implemented given the 29-timebox

  - [ ] Text file: [article](https://fonts.google.com/icons?selected=Material%20Icons%3Aarticle%3A)
  - [ ] Description: [notes](https://fonts.google.com/icons?selected=Material%20Icons%3Anotes%3A)

| Before | After |
|---|---|
| ![image](https://github.com/nextcloud/server/assets/52440189/9051a0af-28f9-42b0-91f1-554eb379f7c1) | ![image](https://github.com/nextcloud/server/assets/52440189/1eb9d878-e385-4b34-9a98-501c217ff723) |

PR #5540, opened 2024-03-20
fix(files): Change icons+wording for adding text doc and folder description

mejo-:
This is only a partial fix so far. We'll have to wait for the icon renames in server and the introduction of `icon-add-folder-description` in https://github.com/nextcloud/server/issues/44340.

Fixes: #5535


Pytal:
Pushed a few updates!

> [!NOTE]
> `#969696` is used for the article svg instead of `--color-text-maxcontrast` to be consistent with the mimetype icons rendered in files which uses a background-image url that does not support CSS variables

juliushaertl:
Rebased again to take the nextcloud/ocp update into account

Pytal (approved):

juliushaertl (approved):

PR #5547, opened 2024-03-21
fix: Use our own vue instance rendering anything but the viewer component

### LINKED ISSUE #5531: Interactive widgets are broken in text
Just noticed on current main branch that interactive widgets don't render in text, while they work in talk.

The custom rendering just stays empty:

Screenshot 2024-03-19 at 16 03 33

I've bisected and it seems to come from [ea8ed3c801800db3b1a0cd5721fcd1a9fce9742b] chore(dependencies): Bump @nextcloud/vue to 8.10.0

Likely https://github.com/nextcloud-libraries/nextcloud-vue/pull/5394

Putting a debugger in place it looks like updating the ref value in https://github.com/nextcloud-libraries/nextcloud-vue/pull/5394/files#diff-08ec8752de5ca8ae39557deea144fd24e7fd409850c22adb3ea3b231afbb1730R69 does not trigger the [watcher of the isVisible value](https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcRichText/NcReferenceWidget.vue#L158).

Steps to reproduce:

- Copy an internal file link to an image
- Paste it into a new text document
- Switch the link type to preview
  Screenshot 2024-03-19 at 15 58 34

I've tried building viewer+text with the same `@nextcloud/vue` version but that does not seem to make a difference.

@ShGKme @susnux I don have a real clue on how to debug that further, so any hint from you as the vue experts would be much appreciated.

juliushaertl:
Fix #5531 

This is a slightly different approach from #5402 which i haven't managed to propagate through all layers of tiptaps vue renderer. As far as I could see this seems to work fine without any logged errors and kind of isolates any Vue usage in text to use our own Vue constructor.

I'm not sure this is the best approach and technically also not as clean, but given we want to fix the broken interactive previews the best I could come up for now. Ideally I would love to actually fully isolate text from the viewer itself in terms of their vue components, but that is likely a larger task and not feasible to fix the bug for the 29 release for now.


ShGKme (approved):
Nice

PR #5571, opened 2024-03-26
fix(links): set default for `loadState('core', 'active-app')`

mejo-:
Without, it threw an error in public shares, which broke loading pages with links.

### DIFF
--- a/src/helpers/links.js
+++ b/src/helpers/links.js
@@ -38,7 +38,7 @@ const domHref = function(node, relativePath) {
 		return ref
 	}
 	// Don't rewrite links in collectives app context
-	if (loadState('core', 'active-app') === 'collectives') {
+	if (loadState('core', 'active-app', '') === 'collectives') {
 		return ref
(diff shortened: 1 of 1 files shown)


src/helpers/links.js
```diff
@@ -38,8 +38,12 @@ const domHref = function(node, relativePath) {
...
 	}
 	// Don't rewrite links in collectives app context
-	if (loadState('core', 'active-app') === 'collectives') {
-		return ref
+	try {
+		if (loadState('core', 'active-app') === 'collectives') {
+			return ref
+		}
+	} catch {
+		// throws error in public shares
```
juliushaertl, line 46: loadState also supports a default which looks a bit cleaner

```suggestion
	if (loadState('core', 'active-app', '') === 'collectives') {
		return ref
```

mejo-: Makes much more sense :see_no_evil:

Thanks, changed now.

PR #5589, opened 2024-03-30
fix/yjs state initial

juliushaertl:
- **fix: Create idempotent y.js doc for initial content**
- **tests: Add tests for loading documents from different preconditions**
- **fix: Always return initial content when needed**

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/components/Editor.vue
```diff
@@ -497,6 +497,8 @@ export default {
 					logger.debug('onLoaded: Pushing local changes to server')
 					this.$queue.push(updateMessage)
 				}
+			} else {
+				this.setInitialYjsState(documentSource, { isRich: this.isRichEditor })
```
mejo-, line 501: Nitpicking, but any reason to use `isRich` instead of `isRichEditor` here? I always find it a bit confusing if variable names are changed when passed on, and in this case `isRichEditor` becomes `isRich` and later becomes `enableRichEditing`.

src/mixins/setContent.js
```diff
@@ -38,5 +43,38 @@ export default {
...
+				enableRichEditing: isRich,
+			})
+			const json = generateJSON(html, editor.extensionManager.extensions)
+
+			const doc = Node.fromJSON(editor.schema, json)
+			const getBaseDoc = (doc) => {
+				const ydoc = new Doc()
+				// In order to make the initial document state idempotent, we need to reset the clientID
+				// While this is not recommended, we cannot avoid it here as we lack another mechanism
+				// generate the initial state on the server side
```
mejo-, line 61: Missing "to" and dot at the end: `to generate the initial state on the server side.`

src/mixins/setContent.js
```diff
@@ -38,5 +43,38 @@ export default {
...
+				// The only other option to avoid this could be to generate the initial state once and push
+				// it to the server immediately, however this would require read only sessions to be able
+				// to still push a state
+				ydoc.clientID = 0
+				const type = /** @type {XmlFragment} */ (ydoc.get('default', XmlFragment))
+				if (!type.doc) {
```
mejo-, line 67: I don't understand this part.

cypress/e2e/initial.spec.js
```diff
@@ -0,0 +1,157 @@
...
+		let readToken = null
+		let writeToken = null
+		cy.shareFile('/test.md', { edit: true })
+			.then((token) => {
+				writeToken = token
+				cy.logout()
+				cy.visit(`/s/${writeToken}`)
+			})
+			.then(() => {
+				// Open read only for the first time
```
mejo-, line 124: This opens read-write, as the share token is read-write.

cypress/e2e/initial.spec.js
```diff
@@ -0,0 +1,157 @@
...
+				cy.getEditor().should('be.visible')
+				cy.getContent()
+					.should('contain', 'Hello world')
+					.find('h2').should('contain', 'Hello world')
+				cy.getContent()
+					.type('Something new {end}')
+				cy.get('.save-status button').click()
+				cy.wait('@save')
+
+				// Open read only for the second time
```
mejo-, line 136: Same here.

cypress/e2e/initial.spec.js
```diff
@@ -0,0 +1,157 @@
...
+						readToken = token
+						cy.logout()
+						cy.visit(`/s/${readToken}`)
+					})
+					.then(() => {
+						// Open read only for the first time
+						cy.getEditor().should('be.visible')
+						cy.getContent()
+							.should('contain', 'Hello world')
+							.find('h2').should('contain', 'Hello world')
```
mejo-, line 155: Shouldn't this check for `Something new Hello world`?

mejo- (commented):
Thanks a lot for diving into this. The explanations seem reasonable to me and sound like a good approach to address the underlying problem.

For now I only have minor comments.

cypress/e2e/initial.spec.js
juliushaertl: TODO: Double check that previous sessions are closed

mejo- (approved):
Thanks for the code walkthrough in the call.

src/mixins/setContent.js
juliushaertl: Clarified also with a comment in the code

Issue #5458, opened 2024-03-11
table in workspaces not expanding correctly


szaimen:
Just tested on 29.0.0b1
the table in workspaces does not seem to expand correctly:
![image](https://github.com/nextcloud/text/assets/42591237/79ad13ec-0dfc-4c2f-95a2-1762cfa331f5)

juliushaertl:
Can you share the markdown file as an example? It looks like the table body is longer then the header

szaimen:
Additional screenshot:
![image](https://github.com/nextcloud/text/assets/42591237/66b918af-2e85-480c-9cad-7542554a0798)
file: 
[Readme (1).md](https://github.com/nextcloud/text/files/14635676/Readme.1.md)

juliushaertl:
@max-nextcloud Could you check that?

juliushaertl:
As Max is out, maybe a quick one for you @elzody ?

elzody:
@szaimen It seems like after investigating a bit that the styles from [here](https://github.com/nextcloud/server/blob/6bced389b20e676babbe193c7fbdf77f66433c3a/apps/files/src/components/FilesListVirtual.vue#L409) are affecting it, not if this is something we can override on our end here or if it should rather be changes somehow in Files?

szaimen:
I guess this is.something you would need to judge where to fix it in the code

PR #5591, opened 2024-04-01
fix: workspace tables expand correctly

elzody:
### 📝 Summary

* Resolves: #5458 

#### 🖼️ Screenshots

<details>
<summary>Before</summary>

grafik
</details>

<details>
<summary>After</summary>

grafik
</details>

### 🏁 Checklist

### DIFF
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -291,24 +291,24 @@ export default {
 		margin: 0;
 	}
 
-  #rich-workspace:deep(.table-wrapper .content) {
-    tr {
-      th, td {
-        flex-grow: 1;
-      }
+	#rich-workspace:deep(.table-wrapper .content) {
+		tr {
+			th, td {
+				flex-grow: 1;
+			}
 
-      th {
-        div {
-          width: 100%;
-        }
+			th {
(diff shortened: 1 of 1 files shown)


src/views/RichWorkspace.vue
```diff
@@ -291,6 +291,25 @@ export default {
 		margin: 0;
 	}
 
+	#rich-workspace:deep(.table-wrapper .content) {
```
juliushaertl, line 294: Do you know why this only affects the rich workspaces?

elzody: > Do you know why this only affects the rich workspaces?

Looks like it's actually some styling from Files affecting it - particularly this part: https://github.com/nextcloud/server/blob/6bced389b20e676babbe193c7fbdf77f66433c3a/apps/files/src/components/FilesListVirtual.vue#L409

juliushaertl: All right, that seems reasonable then. While ideally the files app would not have such aggressive rules I also cannot think of an elegant approach to limit the scope properly. Fine with this one here then from my side 👍

juliushaertl (approved):

Issue #5576, opened 2024-03-28
Translation modal language selection is not possible


juliushaertl:
As reported by @come-nc in the chat, on 29 beta 6 selecting the language in the translate dialog is not possible.

The select input closes immediately:

- Open the translate modal
- Click either source or target language
- Either nothing happens or the select options show up and hide very quickly

juliushaertl:
@luka-nextcloud Can oyu check since you worked on the translation modal somewhat recently as well?

PR #5630, opened 2024-04-05
fix: focus issue on translate modal, remove duplicate code

luka-nextcloud:
### 📝 Summary

* Resolves: #5576
* Fix focus issue on translate modal
* Reduce duplicate code

### 🏁 Checklist

### DIFF
--- a/src/components/Assistant.vue
+++ b/src/components/Assistant.vue
@@ -43,7 +43,7 @@
 					</template>
 					{{ provider.name }}
 				</NcActionButton>
-				<NcActionButton close-after-click @click="openTranslateDialog">
+				<NcActionButton data-cy="open-translate" close-after-click @click="openTranslateDialog">
 					<template #icon>
 						<TranslateVariant :size="20" />
 					</template>
@@ -295,6 +295,9 @@ export default {
 			await this.fetchTasks()
 		},
 		openTranslateDialog() {
+			if (!this.selection.trim().length) {
+				this.$editor.commands.selectAll()
+			}
 			emit('text:translate-modal:show', { content: this.selection || '' })
 		},
 		async openResult(task) {
--- a/src/components/Assistant.vue
+++ b/src/components/Assistant.vue
@@ -63,12 +63,6 @@
 				class="floating-menu--badge" />
 		</FloatingMenu>
 
-		<Translate v-if="displayTranslate !== false"
-			:content="displayTranslate"
-			@insert-content="translateInsert"
-			@replace-content="translateReplace"
-			@close="hideTranslate" />
-
 		<NcModal :show.sync="showTaskList">
 			<div class="task-list">
 				<h4 v-if="tasks.length > 0">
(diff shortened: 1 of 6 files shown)


src/components/Modal/Translate.vue
```diff
@@ -20,7 +20,7 @@
   -->
 
 <template>
-	<NcModal size="large" @close="$emit('close')">
+	<NcModal :show.sync="isShow" size="large" @close="$emit('close')">
```
grnd-alt, line 23: what do we need the computed ```isShow``` for? Can't the ```show``` prop be used directly?

luka-nextcloud: It is because `show` is a prop. So, it is one-way-down binding data. It should not be used with `v-bind.sync`

max-nextcloud: My understanding is that [sync](https://v2.vuejs.org/v2/guide/components-custom-events.html#sync-Modifier) works like this:
```vue
<text-document v-bind:title.sync="doc.title"></text-document>
```
is the same as:
```vue
<text-document
  :title="doc.title"
  @update:title="doc.title = $event"
></text-document>
```

So it's two way binding as you said @luka-nextcloud - but you are not using the up binding from what i can tell. `isShow` as a computed cannot be assigned a new value either. `this.isShow = $event` won't work.

So from my point of view you could either use a one way bind without `.sync` or you'd need a data attribute to hold the show state and be able to assign to it.

src/components/Editor.vue
```diff
@@ -77,6 +77,11 @@
 				:is-rich-editor="isRichEditor" />
 		</Wrapper>
 		<Assistant v-if="$editor" />
+		<Translate :show.sync="translateModal"
```
max-nextcloud, line 80: In order for `:show.sync` to properly work you'd need to emit `update:show` from the `Translate` modal. I think that does not happen yet. Once you have that you don't need to handle `@close` with `hideTranslate` below anymore as the `translateModal` state would be synced already via `:show.sync` already.

luka-nextcloud:
@max-nextcloud Could you please review again?

src/components/Assistant.vue
```diff
@@ -304,27 +295,7 @@ export default {
...
-			this.$editor.commands.command(({ tr, commands }) => {
-				const selection = tr.selection
-				const range = {
-					from: selection.from,
-					to: selection.to,
-				}
-				return commands.insertContentAt(range, content)
-			})
-			this.displayTranslate = false
+			emit('text:translate-modal:show', { content: this.selection || '' })
```
max-nextcloud, line 298: Does this even make sense if the selection is empty?
The previous code looks as if it did not open the modal on an empty selection:
```js
			this.displayTranslate = this.selection
```

However I just tried it on cloud.nextcloud.com and there it opened anyway but did not work (due to the broken input but also as it had no content).

max-nextcloud (approved):
Code looks good. One comment on the scenario with no selected text. Also tests would be nice as this was a regression and we might break it again.

juliushaertl:
Restarted failing cypress test which seemed unrelated

src/components/Assistant.vue
luka-nextcloud: @max-nextcloud I didn't find any side effects. Yes, the text is selected fully.

PR #5655, opened 2024-04-11
fix(imageview): Don't try to load unresolved images

nickvergessen:
### 📝 Summary

Currently a lot of people on our instance get bruteforce protected in Talk regularly. After some log reading it seems to always happen when interactive widgets are involved.
After some trying the smallest possible test is:

1. Create a test.md file with the following content: (on our instance it's `.attachment` links with wrong file ids)
    ```md
    ![](C.png)
    ```
2. Copy the `/f/...` link and post it into a chat
3. Check the browser console for calls to `index.php/call/undefined`

After some debugging I found this trace. The problem is the "default" resolver responds:
https://github.com/nextcloud/text/blob/648cf2036f229eaa5b00e8d4f83c763da1d5d564/src/components/Editor.provider.js#L91-L94
This means that
https://github.com/nextcloud/text/blob/99e7a7a1b842516c1b29ac4574f9ceee49e47a55/src/nodes/ImageView.vue#L245
makes `this.attachment = [this.src]`
in the Promise it sets `img.src = this.attachment.previewUrl` but `previewUrl` is undefined as only `0` exists.
This will make the browser load `undefined` relative to the current page which results in `https://localhost/index.php/call/undefined`
This however looks like you are trying to open a conversation with the token `undefined`, but you have no access to it => bruteforce registered.

by simply checking if `previewUrl` exists and otherwise jumping into the fail mode, it prevents the request to load `index.php/call/undefined` and so also no bruteforce attempt it registered.

### 🏁 Checklist

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -243,6 +243,9 @@ export default {
 	methods: {
 		async loadPreview() {
 			this.attachment = await this.$attachmentResolver.resolve(this.src)
+			if (!this.attachment.previewUrl) {
+				throw new Error('Attachment source was not resolved')
+			}
 			return new Promise((resolve, reject) => {
 				const img = new Image()
 				img.onload = async () => {


nickvergessen:
/backport 99e7a7a1b842516c1b29ac4574f9ceee49e47a55 to stable29

src/nodes/ImageView.vue
```diff
@@ -243,6 +243,11 @@ export default {
 	methods: {
 		async loadPreview() {
 			this.attachment = await this.$attachmentResolver.resolve(this.src)
+			if (!this.attachment.previewUrl) {
+				this.loaded = true
+				this.failed = true
+				return
```
Antreesy, line 249: Maybe that could be the better approach, so `this.loadPreview().catch(this.onImageLoadFailure)` could handle it?

```suggestion
				throw new Error('Attachment source was not resolved')
```

nickvergessen: Also works and sounds better, yeah

Antreesy (approved):
Should be enough for a fix. Waiting for maintainers input

PR #5659, opened 2024-04-11
fix: Avoid throwing when a workspace file cannot be found due to a failed storage

juliushaertl:
This fixes an uncaught exception found in some log where a federated share has been throwing `"message": "Sabre\\HTTP\\ClientHttpException: Unauthorized",` while trying to iterate over rich workspace file candidates which lead to the files list not loading.


blizzz (approved):

blizzz:
Does it need a backport?

juliushaertl:
Doesn't hurt, yes

PR #5662, opened 2024-04-12
Request the native app to reload the direct editing view on 403 errors

nicofrand:
### 📝 Summary

* Resolves: https://github.com/nextcloud/android/issues/12138

When it receives a 403 error it triggers a reload event, catchable by the android app.

### Warning

I don't know how this app works/should behave in some cases, and this PR needs discussion.

1. Right now it keeps retrying the requests even though the 403 error is detected. Can I just stop the retries (they will fail anyway)? What about Direct edition outside the android app (which will not catch the reload event)?
2. The comment says "the session is invalid or the document is read only". Can I make sure the session is invalid (I should not reload the document in case it is read only I guess?)


max-nextcloud:
Dear @nicofrand - I'm sorry for taking so long to get back. Will prioritize this PR this week.

juliushaertl:
@max-nextcloud As discussed in our call here is the command for testing in the browser:

```bash
curl --request POST \
  --url 'https://admin:admin@nextcloud.local/ocs/v2.php/apps/files/api/v1/directEditing/open?editorId=text&creatorId=textdocument&format=json' \
  --header 'Content-Type: multipart/form-data' \
  --header 'OCS-APIRequest: true' --form path=/test.md
```

max-nextcloud:
@nicofrand Again I'm sorry for taking so long. Your code changes look good and I talked through the different scenarios today witth @juliushaertl . Direct editing is also used by the desktop client and ios devices and as you said they won't support the `reload` message - however we can tackle that separately.

I will rebase your pull request and see if I can get the CI to work on it.

Please sign your commit off to make the DCO check pass - this means you submit your code under the same license as the rest of the text app. (The contributing guidelines have [more info](https://github.com/nextcloud/.github/blob/3c6493036645e6a2a80bffaab6a91434e7ed2368/CONTRIBUTING.md#sign-your-work).)

You can do so by running `git commit --amend --signoff --no-edit` - this will add a signoff line to your commit message.

max-nextcloud (approved):

max-nextcloud:
merged as the CI failures were not related.

PR #5690, opened 2024-04-15
remove old link when editing

### LINKED ISSUE #5516: Updating the URL of a link via toolbar menu splits the link
**Describe the bug**
When updating the URL of a link via toolbar menu and not having the whole link selected, a new link is inserted at cursor position instead.

**To Reproduce**
1. Add a link and place the cursor somewhere in the link
2. Go to "Update link" in the link options of the menubar
3. Change the URL and submit
4. See the link being split and a new link being inserted.

**Expected behavior**
The URL of the link at cursor position should be updated

**Screenshots**

| Screencast |
|---|
| recording1 |

**Server details:**
- Nextcloud version: 29

grnd-alt:
Fixes #5516


max-nextcloud:
The current change works well when only changing a single link in a paragraph.
There are two more scenarios that would be good to account for:

## Link in a paragraph

Find more information on https://nextcloud.com as usual.

-> should only change the link.

## Link with custom text

[More informarion](https://nextcloud.com)

-> should only change the href.

max-nextcloud (changes requested):
As I said earlier... this will not work if the link is not the only thing in a paragraph.

@grnd-alt would you be up for picking this up again?

src/tests/marks/Link.spec.js
```diff
@@ -0,0 +1,93 @@
...
+							text: 'updated.de',
+							type: 'text',
+						},
+					],
+					type: 'paragraph',
+				},
+			],
+			type: 'doc',
+		})
+	})
```
max-nextcloud, line 66: This test doesn't really do what the title says.
The editor joins the two `<a>` tags because they have the same attrs. Then it replaces both of them.

I'd propose to use to different hrefs and test for only one of the links being replaced.

max-nextcloud (commented):
Looks good. Just one small remark about the test code.

juliusknorr:
> Warning: Failed to restore: getCacheEntry failed: Cache service responded with 500

Looked odd, restarted cypress

src/tests/marks/Link.spec.js
```diff
@@ -0,0 +1,106 @@
...
+							type: 'text',
+						},
+						{ text: ' HELLO WORLD', type: 'text' },
+					],
+					type: 'paragraph',
+				},
+			],
+			type: 'doc',
+		})
+	})
```
max-nextcloud, line 105: This test seems incomplete. It does not change anything but just asserts the initial state.

max-nextcloud (commented):
Just a minor comment. Other than that seems good code wise. Have not tested it though.

max-nextcloud:
Failing psalm test is unrelated. I'd be okay with force merging.

PR #5693, opened 2024-04-17
update(ci): cypress workflow to use latest dependencies with node 20

max-nextcloud:
### 📝 Summary

* Resolves: Warnings during cypress runs such as

grafik

### 🏁 Checklist

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -23,33 +23,33 @@ jobs:
 
     steps:
       - name: Checkout server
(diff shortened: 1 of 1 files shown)


max-nextcloud:
> Maybe we should have dependabot/renovate configured to bump the versions for github actions that we don't pull from the templates?

If we can limit it to those - would be nice.

max-nextcloud:
For time comparison

| before | after |
| ----- | ----- |
|  grafik  | grafik |

Some faster, some slower - i guess flaky tests have a bigger effect here than the change.

But all in all looks like we could use it.

juliushaertl (approved):

PR #5717, opened 2024-04-22
perf(bootstrap): Lazy register template creator through event

juliushaertl:
Requires https://github.com/nextcloud/server/pull/44952

This will avoid triggering dependency injection queries during app bootstrap for every request.

### DIFF
--- a/lib/Listeners/RegisterTemplateCreatorListener.php
+++ b/lib/Listeners/RegisterTemplateCreatorListener.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2024 Julius Härtl <jus@bitgrid.net>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


lib/Listeners/RegisterTemplateCreatorListener.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2020 Morris Jobke <hey@morrisjobke.de>
+ *
+ * @author Morris Jobke <hey@morrisjobke.de>
```
max-nextcloud, line 7: Probably outdated, right?

We might also start using the SPDX headers right away when adding new files.

max-nextcloud (commented):
Approach seems good - for what i understand. Linter and Static check still failing.

lib/Listeners/RegisterTemplateCreatorListener.php
juliushaertl: Adjusted the copyright, but would rather keep moving to SPDX separate and do it with the migration script from Joas then for all files

hweihwang:
Looks good, smart take on deferring the template creator setup until it is actually needed

hweihwang (approved):

Issue #5460, opened 2024-03-11
Mark callout button as disabled when all callout variants are disabled


mejo-:
**Describe the bug**
Currently, when callouts are disabled (e.g. when inside a node type that doesn't support block nodes as children), the callout menu button still appears as enabled. We should change this:

![2024-03-11T15:33:01,519259717+01:00](https://github.com/nextcloud/text/assets/3582805/08b8614f-60a4-466d-a3d0-ca33001d0004)

PR #5720, opened 2024-04-22
fix: disable callouts button when all callout variants are disabled

luka-nextcloud:
### 📝 Summary

* Resolves: #5460

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/.github/workflows/block-merge-freeze.yml
+++ b/.github/workflows/block-merge-freeze.yml
@@ -19,11 +19,14 @@ concurrency:
 jobs:
   block-merges-during-freeze:
     name: Block merges during freezes
-
-    if: github.event.pull_request.draft == false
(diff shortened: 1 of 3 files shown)


max-nextcloud (changes requested):
Failing `Jest` and `Cypress component tests` are not related I think.

Cypress is failing for an interesting reason:
Test the rich text editor menu bar -- word count -- empty file (failed)


Word count cannot be seen as all actions in the `more` menu are not possible right now. So in a way your changes are working - you will need to account for the special case of the word count entry though.

max-nextcloud (approved):

PR #5721, opened 2024-04-23
test(ci): use node-test template rather than our own jest

max-nextcloud:
* It will not download cypress.
* It will only run on changes to the relevant files and still pass

### 🏁 Checklist

- Tests are tweaked with this in CI


max-nextcloud:
This is a follow up for https://github.com/nextcloud/text/pull/5719#issuecomment-2070511399

mejo- (approved):

Issue #5325, opened 2024-01-31
Dirty table reads to check


juliushaertl:
## Sync

```
{
  "reqId": "OAKatiu7A2595Iw8pMUK",
  "level": 0,
  "time": "2024-01-31T07:22:32+00:00",
  "remoteAddr": "192.168.65.1",
  "user": "admin",
  "app": "no app in context",
  "method": "POST",
  "url": "/index.php/apps/text/session/6/push",
  "message": "dirty table reads: SELECT * FROM `*PREFIX*text_steps` WHERE (`document_id` = :dcValue1) AND (`version` > :dcValue2) AND (`id` > :dcValue3) ORDER BY `id` ASC LIMIT 1000",
  "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:122.0) Gecko/20100101 Firefox/122.0",
  "version": "29.0.0.5",
  "exception": {
    "Exception": "Exception",
    "Message": "",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Query/QueryBuilder.php",
        "line": 341,
        "function": "executeQuery",
        "class": "OC\\DB\\Connection",
        "type": "->",
        "args": [
          "SELECT * FROM `*PREFIX*text_steps` WHERE (`document_id` = :dcValue1) AND (`version` > :dcValue2) AND (`id` > :dcValue3) ORDER BY `id` ASC LIMIT 1000",
          [
            6,
            "*** sensitive parameters replaced ***",
            "*** sensitive parameters replaced ***"
          ],
          [
            2,
            2,
            2
          ],
          null
        ]
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Query/QueryBuilder.php",
        "line": 377,
        "function": "executeQuery",
        "class": "Doctrine\\DBAL\\Query\\QueryBuilder",
        "type": "->",
        "args": []
      },
      {
        "file": "/var/www/html/lib/private/DB/QueryBuilder/QueryBuilder.php",
        "line": 280,
        "function": "execute",
        "class": "Doctrine\\DBAL\\Query\\QueryBuilder",
        "type": "->",
        "args": []
      },
      {
        "file": "/var/www/html/lib/private/DB/QueryBuilder/QueryBuilder.php",
        "line": 293,
        "function": "execute",
        "class": "OC\\DB\\QueryBuilder\\QueryBuilder",
        "type": "->",
[...]

PR #5773, opened 2024-05-07
refactor(session):remove dup queries,refactorcode

hweihwang:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/5325#issuecomment-2095799143

### 🏁 Checklist
- [ ] Remove duplicated query when calling /save endpoint
- [ ] Refactor the code (Remove duplicate code, duplicated call to function, more solid)

### DIFF
--- a/lib/Middleware/SessionMiddleware.php
+++ b/lib/Middleware/SessionMiddleware.php
@@ -2,8 +2,11 @@
 
 namespace OCA\Text\Middleware;
 
+use Exception;
 use OC\User\NoUserException;
 use OCA\Text\Controller\ISessionAwareController;
+use OCA\Text\Db\Document;
+use OCA\Text\Db\Session;
 use OCA\Text\Exception\InvalidDocumentBaseVersionEtagException;
 use OCA\Text\Exception\InvalidSessionException;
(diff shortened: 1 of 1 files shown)


lib/Middleware/SessionMiddleware.php
```diff
@@ -48,108 +55,185 @@ public function beforeController(Controller $controller, string $methodName): vo
 			return;
 		}
 
-		$reflectionMethod = new \ReflectionMethod($controller, $methodName);
+		//ASSERTION
+		$documentId = $this->getDocumentId();
+		$this->document = $this->documentService->getDocument($documentId);
```
max-nextcloud, line 60: This always runs - even if non of the `Require...` attributes were set.

Before we'd only get the document if any of the attributes was there.
Is this intentional?

I don't think it matters in practice as I assume all `ISessionAwareController`s require at least one of them. Just want to point out a change in behavior that i am seeing here.

max-nextcloud (approved):
First of all... I am sorry for taking so long to review this @hweihwang . I needed some quite time that really allowed me to focus on it.

I appreciate your contribution :tada: .

Issue #5816, opened 2024-05-20
Search highlighting


elzody:
This issue is for organizational purposes and is related to [#648](https://github.com/nextcloud/collectives/issues/648) from Collectives. Most of the implementation is done in Text so far, and this issue tracks the related pull requests for accomplishing this.

📡 Tracking
- #5817

elzody:
Issue completed

PR #5817, opened 2024-05-20
Content search highlighting

elzody:
### 💬 Summary
Part of #5816 
This PR implements search highlighting in Text. It is mainly a TipTap extension which subscribes to an event emitted with a search query, and will add highlighting styles to the editor document if matches to the search query are found.

<details>
<summary>How to test this PR</summary>
From the browser console, emit an event with the event bus:

```js
_nc_event_bus.emit('text:editor:search', { query: 'Test' })
```

Replace 'Test' with the string you want to search for.
</details>

<details>
<summary>Screenshot</summary>

image
</details>

### 📝 TODO
- [x] search highlighting

### ☑️ Checklist
- [x] Tests
- [x] Formatting + Linting
- [x] Signed-off commits

### DIFF
--- a/src/plugins/searchDecorations.js
+++ b/src/plugins/searchDecorations.js
@@ -1,0 +1,92 @@
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+import { Decoration, DecorationSet } from '@tiptap/pm/view'
+
+/**
+ * Search decorations ProseMirror plugin
+ * Handles highlighting search matches for the search TipTap extension
+ *
+ * @return {Plugin<DecorationSet>}
+ */
+export default function searchDecorations() {
+	return new Plugin({
+		key: new PluginKey('searchDecorations'),
+		state: {
+			init(_, { doc }) {
+				const searchResults = runSearch(doc, '')
+				return highlightResults(doc, searchResults)
+			},
+			apply(transaction, _oldState) {
+				const query = transaction.getMeta('searchQuery') ?? ''
+
+				if (transaction.docChanged || query) {
+					const searchResults = runSearch(transaction.doc, query)
+					return highlightResults(transaction.doc, searchResults)
+				}
+			},
+		},
+		props: {
+			decorations(state) {
+				return this.getState(state)
+			},
+		},
+	})
+}
+
+/**
+ * Runs the search on the editor to find matches
+ *
+ * @param {Node} doc - Editor document
+ * @param {string} query - Search query
+ *
+ * @return {Array}
+ */
+export function runSearch(doc, query) {
+	const results = []
+
+	if (!query || query === '') {
+		return results
+	}
+
+	doc.descendants((node, offset, _position) => {
+		if (!node.isText) {
+			return
+		}
+
+		const matches = node.text.matchAll(new RegExp(query, 'gi'))
+
+		for (const match of matches) {
+			results.push({
+				from: match.index + offset,
+				to: match.index + offset + query.length,
+			})
+		}
+	})
+
+	return results
+}
+
+/**
+ * Creates decorations for the matches searches
+ *
+ * @param {Node} doc - Editor document
+ * @param {Array} results - Search results
+ *
+ * @return {DecorationSet}
+ */
+export function highlightResults(doc, results) {
+	const decorations = []
+
+	if (results.length < 1) {
+		return DecorationSet.empty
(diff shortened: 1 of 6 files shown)


juliushaertl:
For easier testing just paste this in the js console:

```
_nc_event_bus.emit('text:editor:search', { query: 'Test' })
```

juliushaertl:
Some odd behaviour seems that the search is reset when a sync request comes in, that would still be good to fix, otherwise this looks really good 👍

elzody:
> Some odd behaviour seems that the search is reset when a sync request comes in...

Interestingly, it doesn't happen for me in Collectives, but when I test it by making a plain markdown document in Files and doing a search, it does this. Will look into it further.

@juliushaertl

max-nextcloud (commented):
How would i clear the highlight now?

elzody:
@max-nextcloud Sending an empty string as a search query should clear the highlighting.

This PR is mostly to get the foundation of the search highlighting into Text, and other apps can then use it in their own way and provide the user with a way to clear the highlighting, and it will then just send an empty search query to Text to clear the highlighting

src/plugins/searchDecorations.js
```diff
@@ -15,10 +15,12 @@ export default function searchDecorations() {
...
 			},
-			apply(transaction, _oldState) {
-				const query = transaction.getMeta('searchQuery') ?? ''
+			apply(transaction, oldState) {
+				const query = transaction.getMeta('searchQuery')
 
-				if (transaction.docChanged || query) {
+				if (query === undefined) {
+					return oldState
+				} else if (transaction.docChanged || query) {
```
max-nextcloud, line 23: What will happen if `transaction.docChanged` is set and there is no `searchQuery` meta set?

Is that the intended behavior?

If you now highlight a word and then type this word into the doc - will it be highlighted?

I suspect not as the query did not change thus there was no searchQuery meta so `query` is undefined here and the state stays the same. But I might be wrong.

elzody: You make a good point. Originally it was not like this, but then I was looking into the bug Julius described in his comment above, and determined that when a sync request comes in, the editor state is reset (or, the plugin state at least). So I was trying to find a way to keep the highlighting decorations in that case. But based on your logic it would have unintended effects elsewhere.

Will have to look into it further

max-nextcloud: I think you will need the state of the `searchQuery` plugin so you can know what the current search query is even if it has not changed.
If the doc changed or the search query changed (including turning to '') you need to recompute the decoration with the current search query (the one from `searchQuery` plugin if it did not change).

src/plugins/searchDecorations.js
```diff
@@ -0,0 +1,100 @@
...
+			apply(tr, _, oldState, newState) {
+				const { query: oldQuery } = searchQueryPluginKey.getState(oldState)
+				const { query: newQuery } = searchQueryPluginKey.getState(newState)
+
+				let query = ''
+
+				if (tr.docChanged || (newQuery !== oldQuery)) {
+					query = newQuery
+				} else {
+					query = oldQuery
```
max-nextcloud, line 28: I think you can just return `oldState` here. The doc did not change and the query also stayed the same.

That way you also don't need the `query` variable and can just use `newQuery`.

max-nextcloud (approved):
One tweak still. Approving anyway so you can move on.

src/plugins/searchDecorations.js
max-nextcloud: Oh... the way this is defined in prosemirror is slightly confusing. I think you will need to use all four args of the function:
```js
apply(tr, value, oldState, newState) {
	...
	if (tr.docChanged || (newQuery !== oldQuery)) {
		...
	} else {
		return value // same as before
	}
```

elzody: So, I think you mean something like this?

```js
apply(tr, value, oldState, newState) {
    const { query: oldQuery } = searchQueryPluginKey.getState(oldState)
    const { query: newQuery } = searchQueryPluginKey.getState(newState)

    if (tr.docChanged || (newQuery !== oldQuery)) {
        const searchResults = runSearch(tr.doc, newQuery)
        return highlightResults(tr.doc, searchResults)
    } else {
        return value
    }
},
```

Issue #5838, opened 2024-05-27
Display links inserted from smart picker as preview per default 


mejo-:
**Describe the bug**
Right now, inserting links inserts them as text links per default, which makes sense. When inserting via smart picker, we should change the default to display the link as preview though, as that's what people expect after seeing the preview in the smart picker already. Even more so for options in the smart picker where you can select between link and content preview (like for references to Nextcloud Tables).

This was originally suggested in issue https://github.com/nextcloud/collectives/issues/1298

@max-nextcloud @nextcloud/designers what do you think?

Aveyron-RetD:
My 2 cents.
=> On a new line smart picker = preview automatically
=> inside a line = link 

Well this is what i do for formatting.

PR #5846, opened 2024-05-29
fix(SmartPicker): Insert smart picker links as preview per default

mejo-:
### Summary

* Fixes: #5838

Behaviour:
* Inserts a link preview with smart picker from the toolbar if editor selection is empty
* Inserts a text link with smart picker from the toolbar if editor selection is *not* empty
* Always inserts a link preview with smart picker from `/` command as it's impossible to have ane editor selection when typing `/`

#### 🖼️ Screenshots

| 🏡 Screencast |
|--- |
| recording1 |

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/nodes/Preview.spec.js
+++ b/cypress/e2e/nodes/Preview.spec.js
@@ -129,6 +129,21 @@ describe('Preview extension', { retries: 0 }, () => {
 
 	})
 
+	describe('insertPreview command', { retries: 0 }, () => {
+
+		it('is available in commands', () => {
+			expect(editor.commands).to.have.property('insertPreview')
+		})
+
+		it('inserts a preview', () => {
+			editor.commands.clearContent()
+			editor.commands.insertPreview('https://nextcloud.com')
+			editor.commands.setTextSelection(1)
+			expectPreview()
+		})
+
+	})
+
 	/**
 	 *  Expect a preview in the editor.
 	 */
--- a/src/components/Menu/ActionInsertLink.vue
+++ b/src/components/Menu/ActionInsertLink.vue
@@ -219,7 +219,7 @@ export default {
 				.then(link => {
 					const chain = this.$editor.chain()
 					if (this.$editor.view.state?.selection.empty) {
-						chain.focus().insertContent(link + ' ').run()
+						chain.focus().insertPreview(link).run()
 					} else {
 						chain.setLink({ href: link }).focus().run()
 					}
--- a/src/nodes/Preview.js
+++ b/src/nodes/Preview.js
@@ -62,6 +62,7 @@ export default Node.create({
 		state.write('[')
 		state.text(node.textContent, false)
 		state.write(`](${node.attrs.href} (${node.attrs.title}))`)
+		state.closeBlock(node)
 	},
 
 	addCommands() {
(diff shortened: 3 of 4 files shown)


src/components/Menu/ActionInsertLink.vue
```diff
@@ -232,11 +232,7 @@ export default {
 			getLinkWithPicker(null, true)
 				.then(link => {
 					const chain = this.$editor.chain()
-					if (this.$editor.view.state?.selection.empty) {
-						chain.focus().insertContent(link + ' ').run()
-					} else {
-						chain.setLink({ href: link }).focus().run()
-					}
+					chain.focus().insertPreview(link).run()
```
juliushaertl, line 235: I'd vote to only do this when the selection is still empty

juliushaertl:
Can we also do this when using the smart picker with typing `/` ?

src/components/Menu/ActionInsertLink.vue
mejo-: Not sure if that's what users expect. If you insert a link via smart picker, you already see the preview in the picker. So it's weird to then only insert a text link in my opinion.

An alternative I've considered is to always insert the preview link *after* the paragraph if the cursor is inside a paragraph (and not at beginning or end).

mejo-: After discussing different scenarios with @max-nextcloud I agree with your suggestion @juliushaertl and implemented it this way now. See the described behaviour above in the PR description.

mejo-:
> Can we also do this when using the smart picker with typing / ?

Yep, implemented now.

cypress/e2e/nodes/Preview.spec.js
```diff
@@ -129,6 +129,21 @@ describe('Preview extension', { retries: 0 }, () => {
...
+	describe('insertPreview command', { retries: 0 }, () => {
+
+		it('is available in commands', () => {
+			expect(editor.commands).to.have.property('insertPreview')
+		})
+
+		it('inserts a preview', () => {
+			editor.commands.clearContent()
+			editor.commands.insertPreview('https://nextcloud.com')
+			editor.commands.setTextSelection(1)
```
juliushaertl, line 141: Can we have a comment on why this is needed? Just from reading I don't get why ;)

src/nodes/Preview.js
```diff
@@ -84,13 +85,32 @@ export default Node.create({
...
+			 *
+			 */
+			insertPreview: (link) => ({ state, chain }) => {
+				return chain()
+					.insertContent({
+						type: 'preview',
+						attrs: { href: link, title: 'preview' },
+						content: [{
+							type: 'text',
+							marks: [{
```
juliushaertl, line 105: Seems a bit unexpected that the preview node still has an additional mark, can you tell why?

juliushaertl (approved):
Very nice, just two questions, not blocking

cypress/e2e/nodes/Preview.spec.js
mejo-: The `expectPreview()` command expects the selection to be set. For all other tests, this is done in `prepareEditor()`, but we cannot use that because we want an empty editor to start with.

src/nodes/Preview.js
max-nextcloud: I think it matches the data structure created otherwise. We basically change the wrapping block node between paragraph and preview when switching between the two:

The `setPreview` function above converts the surrounding paragraph of a link into a preview node:
grafik

If you look at `unsetPreview` above it also only unwraps the link inside the preview node:
grafik

max-nextcloud (approved):
Looks great!

PR #5887, opened 2024-06-11
Reset all sessions (in occ command and on upgrade)

mejo-:
### 📝 Summary

* feat(reset): Add `--all` option to `text:reset` occ command
* fix(reset): Reset all document sessions on upgrades from 4.0.0 or below
* Fixes: #5420
* Fixes: nextcloud/collectives#1270

:warning: Reminder: Versions and version comparison need adjustments when backporting to stable branches.

### 🏁 Checklist

### DIFF
--- a/lib/Command/ResetDocument.php
+++ b/lib/Command/ResetDocument.php
@@ -23,6 +23,7 @@
 
 namespace OCA\Text\Command;
 
+use OCA\Text\Db\Document;
 use OCA\Text\Exception\DocumentHasUnsavedChangesException;
 use OCA\Text\Service\DocumentService;
 use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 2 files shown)


lib/Command/ResetDocument.php
```diff
@@ -56,30 +63,44 @@ protected function configure(): void {
...
-		try {
-			$this->documentService->resetDocument($fileId);
-		} catch (DocumentHasUnsavedChangesException) {
-			$output->writeln('Not resetting due to unsaved changes');
-			return 1;
+		$rc = 0;
+		foreach ($fileIds as $id) {
+			if ($fullReset) {
+				$output->writeln('Force-reset the document session for file ' . $id);
+				$this->documentService->resetDocument($id, true);
```
juliushaertl, line 92: I guess we can skip the rest with continue?

lib/Command/ResetDocument.php
```diff
@@ -56,30 +63,44 @@ protected function configure(): void {
...
+			}
+
+			$output->writeln('Reset the document session for file ' . $id);
+			try {
+				// $this->documentService->resetDocument($id);
```
juliushaertl, line 97: Commented out on purpose?

mejo-: Ouch, debugging leftover.

lib/Command/ResetDocument.php
mejo-: Indeed, good catch :fishing_pole_and_fish:

juliushaertl (approved):

PR #5888, opened 2024-06-11
feat(db): Add timestamp field to steps database

mejo-:
This should help with debugging document session issues in the future.

* Related: https://github.com/nextcloud/text/issues/5420#issuecomment-2158055662

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -11,7 +11,7 @@
 - **💾 Open format:** Files are saved as [Markdown](https://en.wikipedia.org/wiki/Markdown), so you can edit them from any other text app too.
 - **✊ Strong foundation:** We use [🐈 tiptap](https://tiptap.scrumpy.io) which is based on [🦉 ProseMirror](https://prosemirror.net) – huge thanks to them!
 	]]></description>
-	<version>4.0.0</version>
+	<version>4.1.0</version>
 	<licence>agpl</licence>
 	<author mail="jus@bitgrid.net">Julius Härtl</author>
 	<namespace>Text</namespace>
--- a/lib/Migration/Version040100Date20240611165300.php
+++ b/lib/Migration/Version040100Date20240611165300.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2024 Jonas <jonas@freesources.org>
+ *
+ * @author Jonas <jonas@freesources.org>
+ *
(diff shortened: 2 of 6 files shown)


appinfo/info.xml
```diff
@@ -11,7 +11,7 @@
 - **💾 Open format:** Files are saved as [Markdown](https://en.wikipedia.org/wiki/Markdown), so you can edit them from any other text app too.
 - **✊ Strong foundation:** We use [🐈 tiptap](https://tiptap.scrumpy.io) which is based on [🦉 ProseMirror](https://prosemirror.net) – huge thanks to them!
 	]]></description>
-	<version>4.0.0</version>
+	<version>4.1.0</version>
```
juliushaertl, line 14: We only bumped for main do far, but since it's an internal version anyways fine with me

lib/Migration/Version040100Date20240611165300.php
```diff
@@ -0,0 +1,53 @@
...
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('text_steps')) {
+			$table = $schema->getTable('text_steps');
+			if (!$table->hasColumn('timestamp')) {
+				$table->addColumn('timestamp', Types::BIGINT, [
+					'notnull' => true,
+					'length' => 6,
+					'default' => 'CURRENT_TIMESTAMP',
```
juliushaertl, line 46: Do you know if this works on all our possible database backends? Or does doctrine maybe handle that?

appinfo/info.xml
mejo-: Mh, not sure I understand. You mean to bump it to `5.0.0` instead?

lib/Migration/Version040100Date20240611165300.php
mejo-: Good point. I looked it up now and MySQL, MariaDB, PostgreSQL, SQLite and Oracle all support `DEFAULT CURRENT_TIMESTAMP` since a long time. But I also searched our code and only the health app uses it. Circles [used it](https://github.com/nextcloud/circles/pull/14) for a short time but [migrated away shortly afterwards](https://github.com/nextcloud/circles/pull/55) due to [a bugreport](https://github.com/nextcloud/circles/issues/54).

mejo-: Changed now to `DEFAULT 0` and setting the timestamp when inserting the step.

Issue #5861, opened 2024-06-06
Initializing the table of contents creates a transaction even in read only mode


max-nextcloud:
**Describe the bug**

1. Document opens
2. Table of contents is initialized
3. Headings are parsed and get an id
4. ids are added to heading nodes creating a transaction
5. transaction updates document state leading to steps pushed even in read only mode.

**To Reproduce**

1. Create a markdown file with headings and a public read only share of it
2. observe network traffic
3. Notice step send in one of the push requests

**Expected behavior**
Opening the document read only should not alter the state.

**Screenshots**
*will attach later*

max-nextcloud:
This is related to https://github.com/nextcloud/text/issues/5800#issuecomment-2145591803

mejo-:
Probably the way forward would be to migrate `extractHeadings()` to use [ProseMirror Decorations](https://prosemirror.net/docs/ref/#view.DecorationAttrs) for setting the `id` attribute locally, instead of altering the document state. See [this example](https://github.com/nextcloud/text/blob/a66e6df7cd49d3c1b9c4658422471fe17bc01a7c/src/plugins/searchDecorations.js#L85-L91).

juliushaertl:
Ah sounds like a good idea, haven't thought of that with my comment in https://github.com/nextcloud/text/issues/5800#issuecomment-2154775155

PR #5893, opened 2024-06-12
Fix/toc without transaction

max-nextcloud:
### 📝 Summary

* Resolves: #5861

### 🚧 TODO

* [ ] implement click function
* [ ] sync plugin state to vuex
* [ ] cleanup file structure

### 🏁 Checklist


src/components/Editor/TableOfContents.vue
```diff
@@ -40,12 +40,7 @@ export default {
 	},
 	methods: {
 		goto(heading) {
-			this.$editor
-				.chain()
-				.focus()
-				.setTextSelection(heading.position)
-				.scrollIntoView()
-				.run()
+			document.getElementById(heading.id).scrollIntoView()
```
max-nextcloud, line 43: This is slightly different from what we used to do. In particular we do not set the cursor anymore.

We could also use the  old way replacing `heading.position` with `heading.offset + 1`. I changed it as relying on the id seemed more robust to me than the offset at that time - but both should be fine now as we recompute and update the offset whenever the doc changes.

juliushaertl:
Very nice, unfortunately this does not seem to fix the initial update step stent for read only sessions.

juliushaertl:
Also happens without headings for files that have not been opened by a write session before:

Screenshot 2024-06-13 at 07 25 53

juliushaertl:
Additional backend attempt at https://github.com/nextcloud/text/pull/5895

max-nextcloud:
> Very nice, unfortunately this does not seem to fix the initial update step stent for read only sessions.

Yes... that's exactly what I wondered this morning - whether the toc transaction was really the cause.

Thanks for coming up with the backend fix. I'll take a look at it.

max-nextcloud:
Since this does not fix the 403s fully, there's no rush to get it into todays RC anymore.

I'll fix the tests and see if it fixes https://github.com/nextcloud/collectives/issues/1114 .

max-nextcloud:
I fixed quite a few of the cypress tests and the failing ones are also failing for me locally.

One major issue remains though. When typing at the beginning of the heading nothing happens. I suspect this is due to the slightly different markup:
previously:
grafik
now:
grafik
(or without the span when the renderHTML function is left untouched.)

It's fairly hard to bring back the anchor in front of the span inside the heading tag.

max-nextcloud:
> One major issue remains though. When typing at the beginning of the heading nothing happens. I suspect this is due to the slightly different markup:

I figured it out. By default decorations are put behind the cursor in the given position. Adding a `{ side: -1 }` fixed that and now the cursor is between the anchor and the first letter of the heading and typing works just fine.

mejo- (approved):

PR #5895, opened 2024-06-13
fix: Ignore pushes of update steps for read only sessions

juliushaertl:
My suggestion for now would be to block any update steps for read only sessions according to what is described in the y.js protocol docs.

https://github.com/yjs/y-protocols/blob/master/PROTOCOL.md#handling-read-only-users

Ideally we'd of course also prevent sending those in the frontend code, but for now this seems the most reasonable quick fix.


max-nextcloud (approved):
Did not test - but code looks good.

Issue #5904, opened 2024-06-15
Session cleanup


juliushaertl:
From chat:

the Text app needs fixing. The session clean up that was added is eating the memory:
```
Force-reset all Text document sessions

 Starting ...
    0/4256 [>---------------------------]   0%
 Starting ...
    2/4256 [>---------------------------]   0%
 Starting ...
   14/4256 [>---------------------------]   0%
 Starting ...
   21/4256 [>---------------------------]   0%
 Starting ...
   26/4256 [>---------------------------]   0%
 Starting ...
   27/4256 [>---------------------------]   0%
 Starting ...
   29/4256 [>---------------------------]   0%
 Starting ...
   30/4256 [>---------------------------]   0%
 Starting ...
   31/4256 [>---------------------------]   0%
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /var/www/cloud.nextcloud.com/nextcloud/lib/private/Files/Cache/Cache.php on line 188

```

juliushaertl:
Maybe it could have been 10 minutes if I always watched the screen closely when it was done with the memory again

We should also speed this up

PR #5917, opened 2024-06-17
perf: Use generator/count query on cleanup tasks

juliushaertl:
Fix https://github.com/nextcloud/text/issues/5904

- [x] Avoid running out of memory when iterating over documents
- [ ] Think about a way to do the document cleanup later on as iterating over all documents and deleting the y.js state files is not feasible on larger instances
  - [x] Implemented this now for testing, in general this approach should work fine as we do set this value also for new file creation if no steps were pushed.
  - [ ] Would still need some testing and thinking if the reset on create could again cause problems

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -25,6 +25,7 @@ use OCP\Files\InvalidPathException;
 use OCP\Files\Lock\ILock;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
+use OCP\IAppConfig;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\Lock\LockedException;
@@ -44,7 +45,8 @@ class ApiService {
 		DocumentService $documentService,
 		EncodingService $encodingService,
 		LoggerInterface $logger,
-		IL10N $l10n
+		IL10N $l10n,
+		private IAppConfig $config
 	) {
 		$this->request = $request;
 		$this->sessionService = $sessionService;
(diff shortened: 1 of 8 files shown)


lib/Service/ApiService.php
```diff
@@ -97,6 +99,11 @@ public function create(?int $fileId = null, ?string $filePath = null, ?string $b
 
 			$this->sessionService->removeInactiveSessionsWithoutSteps($file->getId());
 			$document = $this->documentService->getDocument($file->getId());
+			$resetAfterUpdateTimestamp = $this->config->getValueInt(Application::APP_NAME, 'update_reset_before', 0);
+			if ($document && $document->getLastSavedVersionTime() < $resetAfterUpdateTimestamp) {
+				$this->documentService->resetDocument($document->getId());
+			}
```
max-nextcloud, line 105: Will the `resetDocument` bump the value of `getLastSavedVersionTime`? Otherwise this will be run for every session created until the next (auto-)save.
In particular for folder descriptions this can take a loong time if they are never changed and people randomly create sessions when visiting the folder.
What about read only sessiions? Can they reset the document? Will that update the last saved version time?

max-nextcloud: Ah... i see... `resetDocument` will delete the entry from the documents table. So `$document` will be empty for the next request coming in.

One thing that might still get in the way would be locks - `resetDocument` tries to unlock the file. I don't know what would happen if the file is not locked by text.

PR #5918, opened 2024-06-17
Truncate tables and rename documents folder on reset

mejo-:
### 📝 Summary

* Resolves: #5904

### 🏁 Checklist

### DIFF
--- a/lib/Command/ResetDocument.php
+++ b/lib/Command/ResetDocument.php
@@ -60,6 +60,12 @@ class ResetDocument extends Command {
 			return 1;
 		}
 
+		if ($all && $fullReset) {
+			// Truncate tables and clear document directory
+			$this->documentService->clearAll();
+			return 0;
+		}
+
 		if ($all) {
 			$fileIds = [];
 			foreach ($this->documentService->getAll() as $document) {
--- a/lib/Command/ResetDocument.php
+++ b/lib/Command/ResetDocument.php
@@ -61,9 +61,10 @@ class ResetDocument extends Command {
 		}
 
 		if ($all) {
-			$fileIds = array_map(static function (Document $document) {
-				return $document->getId();
-			}, $this->documentService->getAll());
+			$fileIds = [];
+			foreach ($this->documentService->getAll() as $document) {
+				$fileIds[] = $document->getId();
+			}
 		} else {
 			$fileIds = [$fileId];
 		}
--- a/lib/Db/DocumentMapper.php
+++ b/lib/Db/DocumentMapper.php
@@ -6,6 +6,7 @@
 
 namespace OCA\Text\Db;
 
+use Generator;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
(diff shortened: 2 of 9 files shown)


juliushaertl (approved):
👍 For the truncate part, would appreciate a review for my first commit then

lib/Command/ResetDocument.php
```diff
@@ -61,9 +60,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		}
 
 		if ($all) {
-			$fileIds = array_map(static function (Document $document) {
-				return $document->getId();
-			}, $this->documentService->getAll());
+			$fileIds = $this->documentService->getAll();
```
mejo-, line 69: @juliushaertl I don't understand this change. `DocumentService->getAll()` returns an array of `Document` objects, not the file IDs. How is this supposed to work?

mejo-: Tested and indeed it broke with your current approach. The old approach didn't work either as `array_map()` cannot be run against a `Generator`. I pushed a fix:

```php
$fileIds = [];
$documents = $this->documentService->getAll();
foreach ($documents as $document) {
	$fileIds[] = $document->getId();
}
```

mejo-:
> would appreciate a review for my first commit then

Apart from the commented part, your changes look good to me @juliushaertl

Could you review my changes another time, given that I implemented the logic to rename the directory and clean up old directories after your last review?

lib/Db/DocumentMapper.php
```diff
@@ -38,12 +39,33 @@ public function find(int $documentId): Document {
...
+			->from($this->getTableName());
+		$result = $qb->executeQuery();
+		$count = (int)$result->fetchOne();
+		$result->closeCursor();
+		return $count;
+	}
+
+	public function clearAll(): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName())
```
blizzz, line 68: truncate and delete are different things. Truncate might be faster than delete, but not all DB engines support it afaik. Not sure what is the current state in doctrine. OK for now.

lib/Db/SessionMapper.php
```diff
@@ -137,6 +148,12 @@ public function deleteByDocumentId(int $documentId): int {
 		return $qb->executeStatement();
 	}
 
+	public function clearAll(): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName())
```
blizzz, line 153: as above

lib/Db/StepMapper.php
```diff
@@ -62,6 +62,12 @@ public function deleteAll(int $documentId): void {
 			->executeStatement();
 	}
 
+	public function clearAll(): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName())
```
blizzz, line 67: as above

blizzz (approved):

lib/Db/DocumentMapper.php
mejo-: I tested with 120.000 entries and it was super fast. So I guess it's ok for now :grimacing:

blizzz: https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Migration/Version1130Date20211102154716.php#L52-L54

blizzz:
Runner 9 was green before, Runner 1 is a known Flaky test → Merging.

Issue #5855, opened 2024-06-03
Link bubble should not open when opening a document


mejo-:
**Describe the bug**
When opening a document and the cursor gets placed inside a link (e.g. because the document ends with a link), the link bubble will get opened automatically.

This is particularly annoying when opening documents in a read-only editor (e.g. in Collectives).

Related: https://github.com/nextcloud/collectives/issues/1296

**To Reproduce**
1. Create a document that ends with a link
2. Close and open the document
3. See the link bubble appearing

**Expected behavior**
We should make sure that no link bubble gets opened when the document gets opened. Only when manually clicking on a link, the link bubble should open.

PR #5924, opened 2024-06-18
fix(LinkBubble): Don't open bubble at initialization and on remote changes

mejo-:
Fixes: #5855
Fixes: nextcloud/collectives#1296

### 🏁 Checklist

### DIFF
--- a/src/plugins/links.js
+++ b/src/plugins/links.js
@@ -71,7 +71,9 @@ export function linkBubble(options) {
 		appendTransaction: (transactions, oldState, state) => {
 			const sameSelection = oldState?.selection.eq(state.selection)
 			const sameDoc = oldState?.doc.eq(state.doc)
-			if (sameSelection && sameDoc) {
+			// Prevent bubble from opening at editor initialisation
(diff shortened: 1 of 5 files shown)


src/plugins/links.js
```diff
@@ -69,9 +69,17 @@ export function linkBubble(options) {
 		}),
 
 		appendTransaction: (transactions, oldState, state) => {
+			// Don't open bubble at editor initialisation
+			if (oldState?.doc.content.size === 2) {
```
max-nextcloud, line 73: This will also be in effect if you paste into an empty document. I can't think of any case where that's a problem. Just wanted to mention.

max-nextcloud (approved):
Looks good to me.

src/plugins/links.js
max-nextcloud: Uuhh... what if you insert a link via the link picker on an empty document. Actually would be nice if that behaved exactly as in a doc that already contains some characters. Not so important - but still a bit inconsistent.

Oh... doesn't the initial load happen through the `ydoc`? That would be caught by the other conditions later on - wouldn't it?

PR #5930, opened 2024-06-20
perf: Reduce polling interval for read only users

juliushaertl:
Reduce the polling interval for sync requests on read only sessions to 30s.

This is a quick fix, a proper way of doing that would be to actually add the read only state to the session data for each session joined. That way we could dynamically adjust the sync interval depending on if there are other users with editing permission in the document or not.

Especially useful if you have documents that have high viewing traffic as our app tutorials

### DIFF
--- a/src/services/PollingBackend.js
+++ b/src/services/PollingBackend.js
@@ -28,6 +28,12 @@ const FETCH_INTERVAL_MAX = 5000
 const FETCH_INTERVAL_SINGLE_EDITOR = 5000
 
 /**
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Code changes look sensible, I didn't test though.

How about creating an issue with some description how a proper implementation would look like and adding it to our board so we can tackle this soon and it doesn't get lost?

juliushaertl:
Definitely, will file an issue

PR #5933, opened 2024-06-20
perf: Use getFirstNodeById and pass userId for our internal helper to avoid fetching mountpoints

juliushaertl:
If we don't care which node we get by id we can use the optimized/cached getFirstNodeById instead. In addition I noticed we can skip getting mounts for a given file id to determine the user if we pass it upfront to our internal helper to get the file with the highest permission.


max-nextcloud (approved):
looks good to me.

PR #5938, opened 2024-06-23
[main] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (3 unfixable)


max-nextcloud:
wait... there are no fixable audits... but let's have a pull request anyway? I guess this happened because there were fixable audits before that now have been resolved.

Issue #5724, opened 2024-04-23
Detect inconsistent initial steps in yjs and require page reload


max-nextcloud:
**Is your feature request related to a problem? Please describe.**
In case two users start a new yjs session at the same time they need to initialize it with the same step to create a consistent world view. If this fails we end up with a "split brain" situation - where both work on a similar yet different document and fail to sync fully.

**Describe the solution you'd like**
* Compare the first steps on the server as they are submitted. If the later submission is different respond with an error response such as 409 - conflict.
* Log the information this happened.
* On the client handle the error response with an appropriate message and ask the user to reload the page.

max-nextcloud:
I think this issue just changed from being theoretical to having been practically observed:
![grafik](https://github.com/nextcloud/text/assets/97337118/ea790ae2-3834-4a3c-a9dc-9d848274223f)


I used this script: https://github.com/nextcloud/text/blob/investigate/inconsistencies-in-document/src/tests/investigate-corruption.spec.js#L41-51

So what the screenshot is printing is `tr.afterState` - which seems to be the [state vector](https://github.com/yjs/yjs?tab=readme-ov-file#State-Vector) of all clients after processing that transaction.

Two things jump out here:
* The clock for client 0 increased. This special client ID is used for the initial step. The initial step should be idempotent and only be applied once.
* The number of clients increased a lot in a single step. Many of the newly connected clients had  quite a few steps as well.

So it looks like two editing sessions ran in parallel for a while here before they were 'mixed'. On possible cause for this would be different initial state steps - which would lead to a split brain situation - where one client does not see the other clients changes and vice versa.

PR #5941, opened 2024-06-24
Keep base version etag during reload

### LINKED ISSUE #5726: Close has already been called on the connection
While I thought we had resolved this we still seem to try to close a connection that has already been closed regularly. I'm not sure if this is avoidable, but if not we can maybe handle that more gracefully instead of throwing and erroring out on the console

Screenshot 2024-04-24 at 16 33 45
Screenshot 2024-04-24 at 16 33 50

https://nextcloud-gmbh.sentry.io/share/issue/4eff0f8ac50b49d2a481d35e2f718ab0/

@max-nextcloud Any idea?

max-nextcloud:
### 📝 Summary

`this.$syncService` is cleared during the `close` method.

However we need the `baseVersionEtag` to ensure the editing session
on the server
is still in sync with our local ydoc.

See https://github.com/nextcloud/text/issues/5724#issuecomment-2180699808 for a trace of the effect of this.

Also fixes #5726 

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -664,14 +664,11 @@ export default {
 
 		async close() {
 			if (this.currentSession && this.$syncService) {
-				try {
-					await this.$syncService.close()
-					this.unlistenSyncServiceEvents()
-					this.currentSession = null
-					this.$syncService = null
-				} catch (e) {
-					// Ignore issues closing the session since those might happen due to network issues
-				}
+				await this.$syncService.close()
+				this.unlistenSyncServiceEvents()
+				this.$syncService = null
+				// disallow editing while still showing the content
+				this.readOnly = true
 			}
 			if (this.$editor) {
 				try {
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -373,7 +373,7 @@ export default {
 				guestName,
 				shareToken: this.shareToken,
 				filePath: this.relativePath,
-				baseVersionEtag: this.$syncService?.baseVersionEtag,
+				baseVersionEtag: this.$baseVersionEtag,
 				forceRecreate: this.forceRecreate,
 				serialize: this.isRichEditor
 					? (content) => createMarkdownSerializer(this.$editor.schema).serialize(content ?? this.$editor.state.doc)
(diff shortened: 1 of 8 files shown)


max-nextcloud:
I'm looking into the flaky sync tests and fixing some issues with lost network connectivity on the way.
In particular when clicking reconnect the editor and the status message go away now. If the reconnect fails there's nothing one can do anymore.
For now I'll focus on just keeping the status message so one can try to reconnect multiple times. Keeping the editor would be great but seems more involved.

src/services/SyncService.js
```diff
@@ -270,6 +271,29 @@ class SyncService {
...
 
+	async sendRemainingSteps(queue) {
+		if (queue.length === 0) {
+			return
+		}
+		let outbox = []
+		const steps = queue.map(s => encodeArrayBuffer(s))
+			.filter(s => s < 'AQ')
+		const awareness = queue.map(s => encodeArrayBuffer(s))
+			.findLast(s => s > 'AQ') || ''
```
juliushaertl, line 282: We we have this now in two places with services/WebSocketPolyfill.js maybe we put in a helper function? Could probably then also use a y.js method if available to filter (I'm not a fan of the magic strings there)...

But not a blocker

juliushaertl (approved):
Very nice, thanks also for the nicely split commits. Makes reviewing a very straight forward ❤️

mejo- (commented):
Did some testing and seems to work well. Thanks so much for looking into it.

When cutting the network and pressing "Reconnect", we run into

`sendStepsNow(): this.connection is undefined` because the connection got closed but `sendRemainingSteps()` still tries to run `sendStepsNow()`.

In the same scenario `Error: Close has already been called on the connection` is also still logged.

max-nextcloud:
> When cutting the network and pressing "Reconnect", we run into
> 
> `sendStepsNow(): this.connection is undefined` because the connection got closed but `sendRemainingSteps()` still tries to run `sendStepsNow()`.
> 
> In the same scenario `Error: Close has already been called on the connection` is also still logged.

I tried but failed to reproduce these two.

I also don't see how that would happen as reconnect should not send the remaining steps anymore. But I will take a look at the code to confirm.

src/components/Editor.vue
```diff
@@ -432,7 +430,7 @@ export default {
 		reconnect() {
 			this.contentLoaded = false
 			this.hasConnectionIssue = false
-			this.close().then(this.initSession)
+			this.disconnect().then(this.initSession)
```
max-nextcloud, line 433: `reconnect` calling `disconnect` rather than close now.

max-nextcloud:
Merged as i could not reproduce the remaining issue. @mejo- Please open a new issue based on the comment if this still happens for you.

PR #5945, opened 2024-06-25
test(cypress): Fix initial content undo test

mejo-:
Undo button should be disabled after initial content got loaded and only get enabled after typing something.

### 🏁 Checklist


mejo-:
Will be part of #5924

PR #5948, opened 2024-06-25
perf: Cache workspace content for propfinds

juliushaertl:
Reading a file from disk or possibly remote storage can be costly. We can store the content in cache with the (fileId, etag) as key so we only return the result as long as the file content is unchanged.

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -14,9 +14,13 @@ use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Files\FilesHome;
 use OCA\Text\AppInfo\Application;
 use OCA\Text\Service\WorkspaceService;
+use OCP\Files\GenericFileException;
 use OCP\Files\IRootFolder;
+use OCP\Files\NotPermittedException;
 use OCP\Files\StorageNotAvailableException;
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Code changes look good to me.

I just wonder how often it happens that the same folder gets opened again and again in the files app within one hour.

juliushaertl:
It could also be mobile clients sending a propfind with the rich workspace property and for shared folders, group folders it will also help to cache between users. Likely not a huge improvement but might still contribute to less time spent on storage operations.

lib/DAV/WorkspacePlugin.php
```diff
@@ -97,10 +91,21 @@ public function propFind(PropFind $propFind, INode $node) {
...
+				$cacheKey = $file->getFileInfo()->getId() . '_' . $file->getFileInfo()->getEtag();
+				if ($cachedContent = $cache->get($cacheKey)) {
+					return $cachedContent;
+				}
+
+				try {
+					$cachedContent = $file->getContent();
+					$cache->set($cacheKey, $cachedContent, 3600);
+				} catch (GenericFileException|NotPermittedException|LockedException $e) {
+				}
```
come-nc, line 106: This should be logged at least in debug, no?

PR #5951, opened 2024-06-26
test(cypress): Fix flakyness of file links test in workspace

mejo-:
### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [x] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests

PR #5952, opened 2024-06-26
ci: Skip cypress test crashing electron right now

juliushaertl:
To open a revert right afterwards to not forget, but for now this should not block us from passing CI for easier merges


juliushaertl:
Planned to skip the other on as well but noticed @mejo- is looking into that right now in #5951

mejo- (approved):
Good idea. Only needs backport to stable29 though.

mejo-:
The failing workspace test should be fixed instead, it's not related to the electron crashes.

PR #5960, opened 2024-06-27
[draft] [stable28] perf(rich_workspace): only add property for parent

ShGKme:
Draft on 28, just because I was investigating on 28. I'll recreate on main soon

#### Steps to reproduce

1. Add 300-500 s3 external storages and mount to root subfolder
2. Enable text
3. See 10-15s loading time on `/apps/files/files` for PROPFIND request. On real instance it could be up to 90 seconds for 300 storages

| s3  | 28   | 27       |
|-----|------|----------|
| 0   | 0.1s | 0.1s     |
| 100 | 3s   | 0.2s     |
| 200 | 6s   | 0.5s     |
| 300 | 9s   | 0.9s     |
| 400 | 12s  | 1-2s     | 
| 500 | 15s  | 1.1-2.4s |

#### A problem

There is a new `<nc:rich_workspace>` property for each Node in the response.

After disabling Text or disabling `rich_workspace` the problem disappears.

### 🚧 TODO

- [x] only add property for parent ignoring deep nodes as it was in past

### 🏁 Checklist

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -100,6 +100,11 @@ class WorkspacePlugin extends ServerPlugin {
 			return;
 		}
 
+		// Only return the property for the parent node and ignore it for further in depth nodes
+		if ($propFind->getDepth() !== $this->server->getHTTPDepth()) {
+			return;
+		}
+
 		$file = null;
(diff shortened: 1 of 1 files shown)


mejo-:
@ShGKme this PR probably can be closed in favour of #5963, no?

Issue #5839, opened 2024-05-27
Ctrl+Backspace inserts new item into unordered list with Firefox


PatrickJosh:
**Describe the bug**
Since the update to Nextcloud 29, pressing Ctrl+Backspace in an unordered list behaves weirdly. When pressing it with more than one word left in the current line/item, it behaves normally and deletes the word before the cursor. However, when pressing it with only one word left in the current line, it inserts a new item below and focusses this item. However, it only happens with Firefox, not with Chromium; there, it behaves normally.
Have a look at the following video to get a better understanding:

https://github.com/nextcloud/text/assets/32456639/3c010557-5768-4067-8b5e-3ccc27e177ab

In the video, I only press Ctrl+Backspace, nothing else.

**To Reproduce**
1. Have an unordered list
1. Press Ctrl+Backspace with only one item left in the line

**Expected behavior**
The last remaining word is deleted as well.

**Screenshots**
See video above.

**Server details:**
- Nextcloud version: 29.0.1
- PHP Version: 8.3
- Database: mariadb Ver 15.1 Distrib 10.5.23-MariaDB, for Linux (x86_64) using EditLine wrapper

**Client details:**
 - OS: Fedora Linux 40
 - Browser: Firefox 125
 - Device: Laptop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
Probably not necessary; if it is, just tell me.

#### Browser log
[console-export-2024-5-27_18-3-41.txt](https://github.com/nextcloud/text/files/15457899/console-export-2024-5-27_18-3-41.txt)

</details>

mejo-:
I can reproduce this on latest Text main branch, but not in a Tiptap sandbox.

juliushaertl:
From #5962 :

Let me put some references here as well:
https://github.com/ProseMirror/prosemirror-view/blob/master/CHANGELOG.md#1337-2024-05-31
https://github.com/ProseMirror/prosemirror-view/commit/8c9a7d25021ce8e99d03a63c7fb7033a524fd4db

Was bumped in tiptap upstream for the upcoming 2.5.0 release with https://github.com/ueberdosis/tiptap/commit/747985b3a7617e35702a6dab4ecc87c76610204b

PatrickJosh:
This has indeed been resolved by the library update. Closing.

PR #5962, opened 2024-06-27
fix: ctrl+backspace in list item on firefox

luka-nextcloud:
### 📝 Summary

* Resolves: #5839

The issue came from old version of package `prosemirror-view`. It triggered "Enter" key from wrong detection logic. This issue was fixed on new `prosemirror-view` release.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,6 +74,7 @@
         "mermaid": "^10.9.1",
         "mitt": "^3.0.1",
         "path-normalize": "^6.0.13",
+        "prosemirror-view": "^1.33.8",
         "proxy-polyfill": "^0.3.2",
         "slug": "^9.1.0",
         "tippy.js": "^6.3.7",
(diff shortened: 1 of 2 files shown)


src/nodes/Paragraph.js
```diff
@@ -19,6 +19,20 @@ const Paragraph = TiptapParagraph.extend({
 
 	addKeyboardShortcuts() {
 		return {
+			Enter: ({ editor }) => {
+				// Fix for firefox
+				if (window.triggeredModBackspace) {
```
juliushaertl, line 24: This seems more like a dirty hack, could we rather please try to figure out the root cause? If it is not a problem in tiptap, there must be some difference in why our editor triggers a new line.

juliushaertl (changes requested):

juliushaertl:
Thanks for checking that.

Let me put some references here as well:
https://github.com/ProseMirror/prosemirror-view/blob/master/CHANGELOG.md#1337-2024-05-31
https://github.com/ProseMirror/prosemirror-view/commit/8c9a7d25021ce8e99d03a63c7fb7033a524fd4db

Was bumped in tiptap upstream for the upcoming 2.5.0 release with https://github.com/ueberdosis/tiptap/commit/747985b3a7617e35702a6dab4ecc87c76610204b

juliushaertl:
@luka-nextcloud @max-nextcloud Any oppinion on having direct prosemirror dependencies again in our package.json? I feel this could cause issues in the future if tiptap ships another version. Shall we just wait for the upstream update? Or merge and revert fixed in upstream?

luka-nextcloud:
I think if this issue is not urgent, we'd better wait for the new tiptap release.

juliushaertl:
Left a comment about this in the issue

PR #5963, opened 2024-06-28
perf(rich_workspace): only add property for parent

ShGKme:
### 📝 Summary

Fixes a performance regression for instances with many external storages.

#### Steps to reproduce

1. Add 300-500 s3 external storages and mount to the root subfolder
2. Enable `text` app
3. See 10-15s loading time on `/apps/files/files` for PROPFIND request for local s3.
   With external s3, like Amazon, it could be 10 times slower.

That was a regression in 28.

| s3  | 28   | 27       |
|-----|------|----------|
| 0   | 0.1s | 0.1s     |
| 100 | 3s   | 0.2s     |
| 200 | 6s   | 0.5s     |
| 300 | 9s   | 0.9s     |
| 400 | 12s  | 1-2s     | 
| 500 | 15s  | 1.1-2.4s |

#### The problem

There is a new `<nc:rich_workspace>` property in each Node in the response. To add this property (even empty), `text` checks for a file in every child folder, which could be an external remote storage.

Having a hundred external storages mounted in one parent, it results in a hundred requests to remotes for the data that is not really required to have in advance.

### 🚧 TODO

- [x] Only add property for parent ignoring deep nodes as it was in the past

### 💥 Breaking changes

This breaks the Android Files client a bit—it uses a response from a parent to know about description in children.

Broken scenario:
1. Open root
2. Open a child without README
3. Go back to parent
4. Open a child with README
   - Before: README is shown as a description
   - After: README is not shown and a swipe-reload is required

### Alternative solution

Only ignore external storages, which has the most performance impact.

### 🏁 Checklist


juliushaertl:
So the test actually has a comment on this that proofs what we discussed in the call:

https://github.com/nextcloud/text/blob/3a9c4f6f7a0ce3e6d148314311730b35250e34fc/cypress/e2e/propfind.spec.js#L43-L56

@alperozturk96 Can you keep us posted here on if that is possible to address on Android?

juliushaertl:
Other failure is unrelated.

juliushaertl:
Discussed with @tobiasKaminsky  as we couldn't think of a proper way to make the root only result working with android:

- [ ] Check if we can just cache the existence of a rich workspace
- [ ] alternative would be to use a special property for web

ShGKme:
> * alternative would be to use a special property for web

Then performance regression is not fixed on Android, having ~10-100 times slower root update than it was in 27

juliushaertl:
> Then performance regression is not fixed on Android, having ~10-100 times slower root update than it was in 27

True, but this should already have been the case before 28. Though only considered as an alternative, caching is probably the most reasonable

ShGKme:
> True, but this should already have been the case before 28. Though only considered as an alternative, caching is probably the most reasonable

Why? It was added only in 28, or I got it wrong?

juliushaertl:
No, the clients already request this property since a long time. 28 only introduced a change where the web frontend now also fetches those, so the performance issue is also there earlier, just was never noticed.

ShGKme:
> * Check if we can just cache the existence of a rich workspace

Working on this solution

juliusknorr:
Initially added a commit here but split it now into https://github.com/nextcloud/text/pull/6891

lib/Service/WorkspaceService.php
```diff
@@ -30,7 +30,10 @@ public function __construct(IL10N $l10n) {
 	public function getFile(Folder $folder): ?File {
 		foreach ($this->getSupportedFilenames() as $filename) {
 			try {
-				$file = $folder->get($filename);
+				$exists = $folder->getStorage()->getCache('')->get($filename);
```
icewind1991, line 33: why the separate check here? if the file doesn't exist `get` throws a `NotFoundException` which is catched below.

Thus just duplicated the number of times we're querying the filecache

icewind1991 (changes requested):

lib/Service/WorkspaceService.php
juliusknorr: The reason for this issue was slow performance with federated shares or external webdav storages where the additional check on the cache turned out faster then getting the file from the remote, but agree that we probably should not do that extra.

juliusknorr:
After this came up again as a performance bottleneck also for local folders, I pushed a change as an additional commit to introduce a separate property that only returns the workspace on the root node.

This is mainly to keep backward compatibility for the mobile clients that as far as I understood @tobiasKaminsky do not have a good way to work with only the parent workspace result. Performance wise the clients should also switch to the flat property at some point, so maybe you can check if that is still the case.

juliusknorr:
Blackfire proof of saving queries on a folder with 100 subfolders that have a readme of random casing each (Readme/README/readme)
Screenshot 2025-04-09 at 22 56 31

tobiasKaminsky:
@juliusknorr if there is no "regression", then client will work.
For future: to which property should the clients switch?

juliusknorr:
Can you test the clients with this specific commit https://github.com/nextcloud/text/pull/5963/commits/e5efd6477f689d69d4cdbb00d65c46af9dd3fe40 ? That would help to decide if we actually need the extra one.

I remember in the past the clients were confused when navigating around folders as the property was then intitially not reported on the child folders, only once you did a propfind directly to it.

icewind1991:
failing cypress test seems related

icewind1991 (approved):

icewind1991:
Can this be finished as it's preventing workspace from being enabled on c.nc.c

max-nextcloud:
> Can you test the clients with this specific commit [e5efd64](https://github.com/nextcloud/text/commit/e5efd6477f689d69d4cdbb00d65c46af9dd3fe40) ? That would help to decide if we actually need the extra one.

@tobiasKaminsky Could you test? (see https://github.com/nextcloud/text/pull/5963#issuecomment-2792124269)

icewind1991:
This is still relevant

max-nextcloud:
#8497 superseeds this. Please reopen if i am mistaking there.

PR #5990, opened 2024-07-04
fix: Use css vars to adapt to new clickable area size

juliushaertl:
We require some adjustments for https://github.com/nextcloud/server/issues/45657

So far I've only reduced the different sizes so that the menubar is also just 34px in height.

Feedback would be welcome @marcoambrosini @jancborchardt 

Screenshot 2024-07-04 at 14 31 10

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -837,7 +837,7 @@ export default {
 	top: 0;
 	opacity: 0;
 	visibility: hidden;
-	height: 44px; // important for mobile so that the buttons are always inside the container
+	height: var(--default-clickable-area); // important for mobile so that the buttons are always inside the container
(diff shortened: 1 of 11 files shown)


juliushaertl:
Also this is now more in line with the office UI:

Screenshot 2024-07-04 at 14 35 46

marcoambrosini (approved):
Looks good @juliushaertl :) 
If you link the vue lib master you likely spot more things coming up

mejo- (approved):
Thanks for starting to tackle this @juliushaertl <3

jancborchardt (approved):
Looks good, and yep @juliushaertl that's why I said in the Office call that the toolbar there basically is already in our new style. :D

Issue #5958, opened 2024-06-26
Custom reference widgets (link previews) broken


mejo-:
**Describe the bug**
After upgrading `@nextcloud/vue` from 8.11.3 to 8.13.0, previews of URLs that are provided by a custom reference widget (like files, collectives pages, tables, ...) are broken.

The request to the resolve endpoint still happens, but for some reason, the [custom widget Vue reference](https://github.com/nextcloud-libraries/nextcloud-vue/blob/b0a378de61235eb0c38ccf9166b783f869d74195/src/components/NcRichText/NcReferenceWidget.vue#L9) is not replaced.

Trying to git bisect failed so far. Weirdly, downgrading `@nextcloud/vue` to 8.11.3 via `npm i @nextcloud/vue@8.11.3` fixes the problem, but building against a locally linked `@nextcloud/vue` with `v8.11.3` tag checked out doesn't fix it.

**To Reproduce**

Insert a link to a file and see the missing reference widget in the link bubble

mejo-:
Luckily this got magically solved with an upgrade to @nextcloud/vue 8.14.0 :tada:

PR #6024, opened 2024-07-10
fix(vite): dedupe vue to only use one version of it

mejo-:
When building locally against linked nextcloud-vue, custom reference widgets didn't load since we migrated to vite as builder and module bundler.

Apparently this was because webpack before didn't use modules but vite now does. So apparently the two vue dependencies from nextcloud-vue and text are suddenly treated as separate modules and have separate states.

Adding `vue` to `resolve.dedupe` in the vite config solves, as it forces vite to only use vue from the root project (text).

Fixes: #5958

### 🏁 Checklist

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -21,6 +21,9 @@ const config = createAppConfig({
 		}
 	},
 	config: {
+		resolve: {
+			dedupe: ['vue'],
+		},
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
🎉

PR #6026, opened 2024-07-10
chore: Extract CSS instead of injecting

susnux:
### 📝 Summary

This will create a CSS file per entry point instead of injecting the CSS. Dynamic chunks will still import their CSS by them self.

Alternative for #6025 

### 🏁 Checklist


juliushaertl:
Is there any benefit or downside comparing the two options form here and #6025?

susnux:
Extracting:
- Smaller JS assets as the CSS is not included, this will significantly reduce the browsers JS parsing time
- Let the browser handle styles and not rely on the injection logic (this means we utilize `addStyle` so we support nonces on them e.g. if we in the future forbid unsafe styles (CSP)

Injecting:
- No need to care where the scripts are used the styles are always injected automatically

juliushaertl:
Thanks, then I'd say extracting is the most sane.

juliushaertl (approved):

Issue #6050, opened 2024-07-15
Cannot read properties of undefined (reading 'id')


juliushaertl:
Sentry reported a quite frequent console error, that only seems there since "29.0.4 RC1"

<img width="264" alt="Screenshot 2024-07-15 at 09 29 55" src="https://github.com/user-attachments/assets/6b1fc89e-d591-4a95-9ecd-7db3c26010d0">

https://nextcloud-gmbh.sentry.io/share/issue/0d2e881a7a144773a4460df9cb99acd1/

<img width="794" alt="Screenshot 2024-07-15 at 09 29 02" src="https://github.com/user-attachments/assets/bd1fe4a1-3d78-4964-8f81-3ff0cd087236">

max-nextcloud:
This is odd. I have not been able to reproduce it locally nor on the instance in question yet.
I wonder if this is related to our cleanup of the old sessions. But missing sessions should result in a 403 and then the message to the user to please reload the page.

max-nextcloud:
I tried to reproduce this in various ways:
* Editing with two users in the 29.0.4 version locally.
* On cloud.nextcloud.com trying various things.
* Setting up 29.0.2 locally and migrating to 29.0.4

I still failed to reproduce it. What I did run into was a situation where `sessions` was empty because during the update Nextcloud responded with a 200 maintenance page on push and sync. Even though this caused exceptions in the same function it does not seem related.

max-nextcloud:
Looking at the code this is probably caused by the list of sessions not including any with the current id.
In this case `sessions.find` will return `undefined` and `this.currentSession` is set to `undefined`.

I see two possible causes:
1) The server responds to a sync request with a list of sessions that does not include a session with the same id as the one that was used to authenticate the request.
2) The currentSession id changed in the meantime. Maybe this even is a response to an old request that gets handled after the document changed or so. (Which might point to collectives as there one can change pages fairly quickly).

juliushaertl:
URLs seem mostly from files, collectives is just a low amount, top most is a read only share (from the app tutorials), but could just be due to the amout of people browsing it.

juliushaertl:
Interestingly browsers are just chromium based ones on macOS, let me see if I can catch that somewhere locally

max-nextcloud:
uuuhhh... we have this routine to clean up old sessions during the request. That happens after auth. So maybe one can still authenticate with a session and then have it cleaned up.

max-nextcloud:
[...]

PR #6054, opened 2024-07-16
fix(sync): reuse open connection

max-nextcloud:
### 📝 Summary

Do not attempt to create a new connection
if there already is one and it is not closed.

If no messages are received for 30 seconds
yjs will open a new websocket.

Since we do not close the connection anymore from the websocket polyfill we also do not need to open it.

If the network connection has gone down
creating a new connection will fail anyway.

Once it comes back we will know if the session is still valid. Then we can either continue using it or reconnect.

This is part of #6050.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2024-07-16 11-40-43 | No error message anymore

### 🏁 Checklist

- No tests. We'd have to wait > 30sec. for yjs to consider the websocket closed and open it again.
- No documentation needed.

### DIFF
--- a/cypress/e2e/sync.spec.js
+++ b/cypress/e2e/sync.spec.js
@@ -94,6 +94,7 @@ describe('Sync', () => {
 
 	it('recovers from a lost and closed connection', () => {
 		let reconnect = false
+		// block all requests until the session is closed and reopened
 		cy.intercept('**/apps/text/session/*/*', (req) => {
 			if (req.url.includes('close') || req.url.includes('create') || reconnect) {
 				req.continue()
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):

mejo- (approved):

max-nextcloud:
Chances are this will actually fix #6050 as the newly created sessions come with polling backends and the old polling backends keep running in parallel. So once the current session is closed the other polling backends will continue running and hand the session list to the sync service which still has the closed session and then fails to find it in the session list. So the currentSession ends up getting cleared.

max-nextcloud:
fixed a related failing cypress test and now the REUSE  compliance test is failing due to some css files. Not sure what to make of that.

max-nextcloud:
cypress failures and compliance check are unrelated.

PR #6056, opened 2024-07-17
ci: Also commit css files during build

juliushaertl:
### 📝 Summary

When extracting css files in https://github.com/nextcloud/text/pull/6026 we missed that those would also need to be committed to the repo as text is a shipped app.

Otherwise cloning text will not be enough to run it.

@mejo- This could also explain some failures in collectives as @enjeck has seen this in tables


mejo- (approved):

max-nextcloud:
This breaks the reuse compliance check right now. See https://github.com/nextcloud/text/actions/runs/9990903976/job/27612952021 for an example

PR #6098, opened 2024-07-24
fix: Remove superfluous import

mejo-:


juliushaertl:
@mejo- I does this need a backport still? If not please just drop the label :)

mejo-:
This already got backported to stable28 and stable29. I wonder how the backport-request label was added in the first place? :thinking:

PR #6122, opened 2024-07-25
fix(css): Use single quotes

mejo-:
This got caught by stylelint on stable29 but not on main branch.


juliushaertl:
> This got caught by stylelint on stable29 but not on main branch.

Do you know why that is not failing on main? 🙈

juliushaertl (approved):

mejo-:
> Do you know why that is not failing on main? 🙈

Max' assumption was that it's related to the migration to vite.

PR #6123, opened 2024-07-25
fix(search): Add data attribute to search decoration

mejo-:
This allows to get the element and scroll it into view, e.g. by Collectives.


elzody (approved):
Was wondering how to best achieve this, this seems like a good solution

PR #6155, opened 2024-07-31
[stable29] fix: Avoid throwing exceptions during propfind

juliushaertl:
Fix https://github.com/nextcloud/server/issues/42663

Was already fixed on main with #5948

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -31,9 +31,12 @@ use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Files\FilesHome;
 use OCA\Text\AppInfo\Application;
 use OCA\Text\Service\WorkspaceService;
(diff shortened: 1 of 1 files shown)


blizzz (approved):

elzody (approved):

lib/DAV/WorkspacePlugin.php
```diff
@@ -115,9 +118,13 @@ public function propFind(PropFind $propFind, INode $node) {
...
 		// Only return the property for the parent node and ignore it for further in depth nodes
 		$propFind->handle(self::WORKSPACE_PROPERTY, function () use ($file) {
-			if ($file instanceof File) {
-				return $file->getContent();
+			try {
+				if ($file instanceof File) {
+					return $file->getContent();
+				}
+			} catch (GenericFileException|NotPermittedException|LockedException) {
 			}
```
come-nc, line 126: The exception should be logged, at least on debug level

PR #6158, opened 2024-08-02
ci(cypress): Adapt to file name link being a button now

juliushaertl:
Another run for green CI

Upstream fix:
- [x] https://github.com/nextcloud/server/pull/47015

Additionally change selector and scroll file list entry into view before trying to assert it


mejo- (approved):
:clap: green CI :heart_eyes_cat:

PR #6167, opened 2024-08-05
fix: Avoid throwing in BeforeNodeWrittenListener on new files

juliushaertl:
Found in sentry.

For new files we get a BeforeNodeWrittenEvent with a NonExistingFile object that might not have a file id yet. We should silently ignore this case. Debug logging added to still be able to trace this down if needed.


mejo- (approved):
Nice catch!

PR #6178, opened 2024-08-06
fix: Move settings to use NcCheckboxRadioSwitch

juliushaertl:
contributes to https://github.com/nextcloud/server/issues/40283

We should use the proper component for the checkbox so that it is in line with the other settings options.


elzody (approved):
Works great now, Cypress failures are not related I think

PR #6181, opened 2024-08-09
fix: Only insert smart picker result as preview if it is a valid URL

juliushaertl:
Otherwise if the result from an assistant smart picker action is not considered markdown it would insert it as a broken link. We should only use the link node type instead if we actually have a valid URL.


elzody (approved):
Looks reasonable

elzody:
Not actually sure if the Cypress failure is related -- the tests seem to be really inconsistent right now in text overall

Issue #6133, opened 2024-07-26
Move search highlighting to editor API


elzody:
Right now, search highlighting uses the event bus to listen for events from other apps which tell it what query to highlight, and also when to move to the next search, toggle matching all occurrences, etc.

It would be a bit cleaner to move this to the text / editor API rather than use the event bus, since it would avoid listening for events with some "undocumented" API functionality.

PR #6199, opened 2024-08-11
feat: Move search highlighting to editor API

elzody:
### 📝 Summary

* Resolves: #6133 

Previously, Text would emit and subscribe to events using [@nextcloud/event-bus](https://github.com/nextcloud-libraries/nextcloud-event-bus) and other apps which wanted to use search highlighting would need to subscribe and emit their own corresponding events to trigger it. This was not ideal since it required apps to use some unknown / undocumented API for triggering search highlighting. This PR moves away from using the event bus for cross-app communication in this case, and allows apps to just call some search methods on the editor itself.

Please note, the event bus is still used in one area, but it is not used for cross-app communication in that case. It is used only for communicating with deeply-nested Prosemirror plugins and the editor class itself, and is thusly not exposed as something other apps can use. They must still use the editor API.

### 🏁 Checklist

### DIFF
--- a/cypress/component/editor/search.cy.js
+++ b/cypress/component/editor/search.cy.js
@@ -24,7 +24,7 @@ describe('editor search highlighting', () => {
 
 		const highlightedElements = document.querySelectorAll('span[data-text-el]')
 		expect(highlightedElements.length).to.equal(1)
-		expect(highlightedElements[0].innerText).to.equal(searchQuery)
+		verifyHighlights(highlightedElements, searchQuery)
 	})
 
 	it('can highlight multiple matches', () => {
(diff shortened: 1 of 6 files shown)


cypress/component/editor/search.cy.js
```diff
@@ -0,0 +1,92 @@
...
+				extensions: [Document, Text, Search, Paragraph, HardBreak],
+			})
+		})
+	})
+
+	it('can highlight a match', () => {
+		const searchQuery = 'Lorem ipsum dolor sit amet'
+		editor.commands.setSearchQuery(searchQuery)
+
+		const highlightedElements = document.querySelectorAll('span[data-text-el]')
```
mejo-, line 30:
```suggestion
		const highlightedElements = document.querySelectorAll('span[data-text-el="search-decoration"]')
```

Nitpicking, but just to make sure to only catch search decoration elements, and not other `data-text-el` elements :wink:

mejo- (approved):
Really nice, looks great @elzody! Also extra kudos for the cypress component tests :heart:

I didn't test, but I guess you already tested this with corresponding code changes in Collectives?

I only have one nitpicking comment, otherwise good to be merged from my side.

juliushaertl (approved):
Nice and clean 🚀

PR #6200, opened 2024-08-12
fix/sync bugs

juliushaertl:
- **fix: Track push errors and do not reset error state if push fails but sync passes**
- **fix: Do not reset read only state if sync errors occurred**
- **fix: emit onerror for websocket and reopen connection on reconnect attempts of y-websocket**
- **fix: Emit error if push fails in other unhandled cases**

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


src/services/SyncService.js
```diff
@@ -175,6 +182,7 @@ class SyncService {
 			}).catch(err => {
 				const { response, code } = err
 				this.sending = false
+				this.pushError++
```
mejo-, line 185: Treating push errors in `hasConnectionIssues` means the editor turns readonly as soon as there a connection issue, right? I can imagine that usability will suffer in flaky networks (e.g. in trains) with this change. We could think about displaying a document status that warns about unpushed changes instead. But I'm also fine with merging this as is and seeing how it will perform in reality.

src/components/Editor.vue
```diff
@@ -559,14 +559,14 @@ export default {
...
-			const editable = !this.readOnly
+			const editable = !this.readOnly && !this.hasConnectionIssue
 			if (this.$editor.isEditable !== editable) {
 				this.$editor.setEditable(editable)
 			}
 		},
 
 		onSync({ steps, document }) {
-			this.hasConnectionIssue = false
+			this.hasConnectionIssue = !this.$providers[0].wsconnected || this.$syncService.pushError > 0
```
mejo-, line 569: I think I'm missing something obvious, but I don't understand why we should evaluate `this.$syncService.pushError > 0` here. `onSync()` is called when `SyncService` emits `'sync'`, which in my understanding happen either if we pushed steps successfully (which means `pushError = 0`), or if we received steps from the server (which is not related to `pushError`).

mejo- (commented):
I tried to follow the logic by stepping through with a debugger and review, but have more questions than answers :grimacing:

Maybe we can schedule a call later today or tomorrow to go through the changes together?

src/services/SyncService.js
juliushaertl: After going through the code again I still feel this is more sane as a quick fix then skipping the commit. Without it we currently have even worse behaviour on flaky networks where the editor randomly switches between editable and non-editable depending if a sync request passes or not.

My proposal would be to get this in and follow up with a PR to improve the overall behaviour with error handling.

juliushaertl:
@mejo- After testing the behaviour again with just the two commits as discussed yesterday i noticed that this is quite flaky that way:

- There is no indication that the push failed
- The editor "randomly" changes between editable and not editable depending if a sync succeeded or a push failed.

I would feel more confident in merging this as it is and follow up to improve:
- [ ] Refactor the general sync state into a defined data strcuture by splitting push/sync/save state and also remembering last success/failure time, number of failures. That way we have more control to show errors only if the issue persists and not immediately on flaky connections
- [ ] Indicate that local changes were not pushed to the server with a proper error message on the save indicator
  - [ ] Figure out if we can auto retry the push or if we want to let the user manually trigger it
- [ ] Avoid immediately setting the editor to read only
- [ ] Avoid saving the file as the user if there are unpushed steps

Regardless, I pushed another commit to warn users when they attempt to close the browser window that there have been unpushed changes.

mejo-:
Thanks for going through this again @juliushaertl. Your assessment sounds reasonable. And the list of follow-up todos is also a nice starting point! So let's get this merged as is. And probably we should also backport until `stable28` at least, no?

juliushaertl:
Follow up started in https://github.com/nextcloud/text/pull/6202

PR #6202, opened 2024-08-13
fix/sync error cleanup

juliushaertl:
- [ ] Refactor the general sync state into a defined data strcuture by splitting push/sync/save state and also remembering last success/failure time, number of failures. That way we have more control to show errors only if the issue persists and not immediately on flaky connections
- [ ] Indicate that local changes were not pushed to the server with a proper error message on the save indicator
  - [ ] Figure out if we can auto retry the push or if we want to let the user manually trigger it
- [ ] Avoid immediately setting the editor to read only
- [ ] Avoid saving the file as the user if there are unpushed steps
- [ ] Warn users when navigating away with unpushed steps
- [ ] File follow up issue for proper offline support, keeping changes in the browser storage and pushing later

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -566,7 +566,7 @@ export default {
 		},
 
 		onSync({ steps, document }) {
-			this.hasConnectionIssue = false
+			this.hasConnectionIssue = !this.$providers[0].wsconnected || this.$syncService.pushError > 0
 			this.$nextTick(() => {
 				this.emit('sync-service:sync')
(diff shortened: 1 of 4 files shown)


juliushaertl:
Closing until further discussed during the contributor week

PR #6203, opened 2024-08-13
fix(test): increase timeout

elzody:
Should fix `ImageView.spec.js`. The issue seemed to be that the elements were loading, but too slowly and the Cypress commands timed out. Adding a longer timeout should be reasonable enough.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -324,8 +324,8 @@ Cypress.Commands.add('getModal', () => {
 
 Cypress.Commands.add('getEditor', { prevSubject: 'optional' }, (subject) => {
 	return subject
-		? cy.wrap(subject).find('[data-text-el="editor-container"]')
-		: cy.get('[data-text-el="editor-container"]')
+		? cy.wrap(subject).find('[data-text-el="editor-container"]', { timeout: 20000 })
+		: cy.get('[data-text-el="editor-container"]', { timeout: 20000 })
 })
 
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):

cypress/support/commands.js
```diff
@@ -324,8 +324,8 @@ Cypress.Commands.add('getModal', () => {
 
 Cypress.Commands.add('getEditor', { prevSubject: 'optional' }, (subject) => {
 	return subject
-		? cy.wrap(subject).find('[data-text-el="editor-container"]')
-		: cy.get('[data-text-el="editor-container"]')
+		? cy.wrap(subject).find('[data-text-el="editor-container"]', { timeout: 20000 })
+		: cy.get('[data-text-el="editor-container"]', { timeout: 20000 })
```
juliushaertl, line 328: Looking at the replay it seems to be not the timeout but somehow viewer is not opening but the file is downloaded instead:

Screenshot 2024-08-14 at 08 44 43

juliushaertl:
New attempt use chrome instead of electron, now downgrading to 114 due to the running out of memory error.

If that passes generally we may also want to try to bump cypress and keep chrome pinned for now

Next could be the fix from https://github.com/cypress-io/cypress/issues/27415#issuecomment-2169155274

juliushaertl:
Let me close this as it seems a timeout is not the problem and my attempts also failed 😢

PR #6217, opened 2024-08-16
fix: Ensure WebsocketPolyfill always has the latest session state and version

juliushaertl:
## Steps to reproduce

- Have an active sync session
- Block requests to /sync with 401
- Wait for push to fail and have no version

## What happened

If the sync request was failing and the websocket polyfill was reinitialized there was no event emitted by the syncService to set the current session state, version on the new instance again. This lead to push requests missing the current version in the request data.

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -566,7 +566,7 @@ export default {
 		},
 
 		onSync({ steps, document }) {
-			this.hasConnectionIssue = !this.$providers[0].wsconnected || this.$syncService.pushError > 0
+			this.hasConnectionIssue = this.$syncService.backend.fetcher === 0 || !this.$providers[0].wsconnected || this.$syncService.pushError > 0
 			this.$nextTick(() => {
 				this.emit('sync-service:sync')
 			})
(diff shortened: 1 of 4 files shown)


src/components/Editor.vue
```diff
@@ -566,7 +566,7 @@ export default {
 		},
 
 		onSync({ steps, document }) {
-			this.hasConnectionIssue = !this.$providers[0].wsconnected || this.$syncService.pushError > 0
+			this.hasConnectionIssue = this.$syncService.backend.fetcher === 0 || !this.$providers[0].wsconnected || this.$syncService.pushError > 0
```
juliushaertl, line 569: This callback is triggering for either a successful sync or push request so we need this as a workaround way to check if the other request currently has issues.

src/services/WebSocketPolyfill.js
```diff
@@ -60,7 +56,16 @@ export default function initWebSocketPolyfill(syncService, fileId, initialSessio
 					}
 				},
 			})
-			syncService.open({ fileId, initialSession })
+
+			syncService.open({ fileId, initialSession }).then((data) => {
+				if (syncService.hasActiveConnection) {
```
juliushaertl, line 61: I decided to refactor this slightly as we then can just always rely on the open method to return the relevant data, in both first connects and reconnects

juliushaertl (commented):
A few comments to hopefully make the changes more clear

blizzz (approved):

PR #6243, opened 2024-08-21
fix: catch all errors when getting rich workspace content

icewind1991:
Don't let errors to load a single file become errors for loading the entire directory.

### DIFF
--- a/lib/DAV/WorkspacePlugin.php
+++ b/lib/DAV/WorkspacePlugin.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace OCA\Text\DAV;
 
+use Exception;
 use OC\Files\Node\File;
 use OCA\DAV\Connector\Sabre\Directory;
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

lib/DAV/WorkspacePlugin.php
```diff
@@ -96,7 +96,7 @@ public function propFind(PropFind $propFind, INode $node) {
 				try {
 					$cachedContent = $file->getContent();
 					$cache->set($cacheKey, $cachedContent, 3600);
-				} catch (GenericFileException|NotPermittedException|LockedException $e) {
+				} catch (\Exception $e) {
```
ChristophWurst, line 99: nit: shall we log this, at least with debug level?

ChristophWurst: Changed to still ignore file, permission and locking exception but log anything else as error

Issue #3646, opened 2023-01-07
Collapsible sections


rocramer:
**Is your feature request related to a problem? Please describe.**
We are working with pretty large documents. Often we would like to hide some parts of the document to have a better overview of the document. Collapsible sections could be a good addition for the text editor. A wide variety of markdown editors are supporting this. For example [GitHub](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections).

Are there any plans on adding this feature?

juliushaertl:
Copying discussion from #1841 @luka-nextcloud wrote:

1. Add markdown collapsible section (https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) :
 - steps:
   - Add toolbar item to insert collapsible section
   - Add styling to editing area to make user understand that cursor is being in or out side collapsible section
   - Add fold/unfold function on section title

juliushaertl:
Copying discussion from https://github.com/nextcloud/text/issues/1841 @juliushaertl wrote:

A few more things to consider from my perspective:

- Which node types for formattings can be supported within the section title and content (especially when representing this in markdown)?
- Implementation wise this could reuse https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details but would need to work with contenteditable inline editing, this is something to check upfront, otherwise we may need custom implementation that we should avoid if possible
- The menu button could insert a new details element, but how would it behave if
  - Clicking it with selected text
  - Should there be a way to undo the details section but keep the content?
- This should be covered by  jest tests for the markdown syntax parts and cypress for any user iteractions possible

juliushaertl:
Copying discussion from #1841 @mejo- wrote:

Another question would be whether spoilers (that's another way how these expandable sections are often called) can be nested. Seems like HTML allows this, and so does GFM (Github flavoured markdown: https://gist.github.com/scmx/eca72d44afee0113ceb0349dd54a84a2?permalink_comment_id=4196669#gistcomment-4196669. But it seems rather useless to me and brings extra complexity, so I don't think that we have to support it.

> The menu button could insert a new details element, but how would it behave if -> Clicking it with selected text

I'd expect the editor to ask me for a summary text and move the selected content into the details section of the spoiler.

> Should there be a way to undo the details section but keep the content?

I'd expect this, definitely. When inside a details section, the menu button should remove the details section but preserve its content.

juliushaertl:
[...]

Issue #2836, opened 2022-08-26
Menu bar redesign


juliushaertl:
Notes from the design review call at 23\.08.2022

## Overall

* [ ] Formatting bar too wide, especially on mobile.
  * [ ] Maybe show only a "Formatting" button on the bottom and on clicking have a bottom sheet style with all the formatting options like on iOS notes on mobile.
  * [ ] Grouping:
    * [ ] **Undo**
    * [ ] **Redo**
    * [ ] **Formatting**
      * [ ] Title, heading, subheading, body
      * [ ] bold, italic, underline, strikethrough
      * [ ] Lists: ordered, unordered, checklist (in future: indent & un-indent)
    * [ ] **Link**
    * [ ] **Add:** Upload file, add from Nextcloud, Link to file, possibly Emoji
    * [ ] "**Insert block**" --> Callouts, blockquote, table, code block
    * [ ] Emojis: **Either separate entry**, or put into "Insert images"
    * [ ] Formatting help: Either put on top, or separate entry, or into formatting
  * [ ] On desktop:
    * [ ] Blockquote and code block go into the callouts menu
    * [ ] Heading menu left of bold
    * [ ] Space between groups (Undo/redo SPACE headings & formatting SPACE lists SPACE all the rest)
* [ ] Replace "Saved" indicator with an icon, PR at <https://github.com/nextcloud/text/pull/2734>
* [ ] Check if emoji picker can be put in a submenu (of the "Add" menu)
* [ ] Add a separate item in the heading menu "Body text" as the first item (but still only show dot indicator when any heading level is selected)
* [ ] Change icon of headings option to [https://fonts.google.com/icons?selected=Material Icons%3Aformat_size%3A](https://fonts.google.com/icons?selected=Material%20Icons%3Aformat_size%3A)

## Accessibility

* [ ] Contrast for selected formatting option is not enough
  * [ ] Go back to dot or a rounded bar on the bottom
  * [ ] Icon needs to be black as well

mejo-:
Discussion with @marcoambrosini:

* Always group block elements (blockquote, code block, callouts), use CardText MDI icon
* Group text marks (bold, italic, underscore, strikethrough) on mobile, use FormatTextVariant (or FormatText?) icon

PR #6251, opened 2024-08-23
feat(editor): Add support for collapsible sections

mejo-:
### 📝 Summary

Uses `<details>` and `<summary>` summary both for markdown and HTML serialization.

Fixes: #3646

#### 🖼️ Screenshots

Details expanded | Details collapsed | Option
---|--- | ---
grafik | grafik | grafik

### 🚧 TODO

- [x] Fix enter in summary (jump to details content, but don't add newline)
- [x] Discuss icon in menubar
- [x] Support marks in summary (markdown-it plugin supports it already, but `toMarkdown()` in Tiptap plugin not)
- [x] Do we want to persist expand/collapse state in document? If not, it will not be synchronized between view/edit mode in Collectives either
- [x] Jest tests for markdown through editor
- [x] Cypress tests

@nextcloud/designers for feedback

### 🏁 Checklist

### DIFF
--- a/src/components/Menu/entries.js
+++ b/src/components/Menu/entries.js
@@ -28,6 +28,7 @@ import {
 	Paperclip,
 	Positive,
 	Table,
+	UnfoldMoreHorizontal,
 	Warn,
 } from '../icons.js'
 import EmojiPickerAction from './EmojiPickerAction.vue'
@@ -321,6 +322,16 @@ export default [
 			return command.insertTable()
 		},
 		priority: 17,
+	},
+	{
+		key: 'details',
+		label: t('text', 'Details'),
+		isActive: 'details',
+		icon: UnfoldMoreHorizontal,
+		action: (command) => {
+			return command.toggleDetails()
+		},
+		priority: 18,
 	},
 	{
 		key: 'emoji-picker',
--- a/cypress/e2e/nodes/Details.spec.js
+++ b/cypress/e2e/nodes/Details.spec.js
@@ -1,0 +1,58 @@
+/**
+ * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { initUserAndFiles, randUser } from '../../utils/index.js'
+
+const user = randUser()
+const fileName = 'empty.md'
+
+describe('Details plugin', () => {
+	before(() => {
+		initUserAndFiles(user)
+	})
+
+	beforeEach(() => {
+		cy.login(user)
+
+		cy.isolateTest({
+			sourceFile: fileName,
+		})
+
+		return cy.openFile(fileName, { force: true })
+	})
+
+	it('inserts and removes details', () => {
+		cy.getContent()
+			.type('content{selectAll}')
+
+		cy.getMenuEntry('details').click()
+
+		cy.getContent()
+			.find('div[data-text-el="details"]')
+			.should('exist')
+
+		cy.getContent()
+			.type('summary')
+
+		cy.getContent()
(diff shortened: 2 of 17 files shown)


juliushaertl:
Very nice ❤️

> Do we want to persist expand/collapse state in document? If not, it will not be synchronized between view/edit mode in Collectives either

Just a thought here, persisting would likely also mean that toggling would propagate to all people during collaboration with I would not expect.

mejo-:
>  Disallow nested details in Tiptap schema

After some more testing I managed to support nested details in the markdown-it plugin, and I don't see a reason any longer to not support nested details.

jancborchardt (changes requested):
Very nice! :) Just some details:
- [x] A better icon for the actions entry would be https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%235f6368&selected=Material+Symbols+Outlined:unfold_more:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=collapse
- [ ] There is more padding on the bottom than at the top when opened, making it look a bit unbalanced
- [ ] The code block does not have rounded corners here?

mejo-:
Thanks for the feedback @jancborchardt :heart: 

>  The code block does not have rounded corners here?

That's a separate issue unrelated to this PR.

src/components/Menu/entries.js
```diff
@@ -322,6 +323,16 @@ export default [
 		},
 		priority: 17,
 	},
+	{
```
juliushaertl, line 326: @jancborchardt Keeping https://github.com/nextcloud/text/issues/2836 into consideration should we take action for this one right now and instead of adding this as yet another menu bar entry try to group them in the callout menu as listed in the issue? 

>  Blockquote and code block go into the callouts menu

In the end the details is just another "block style"

juliushaertl: Though one argument against would be that you could still have a callout nested in a details block which is then not very obvious from the active indication in the menu bar

mejo-:
@jancborchardt @nimishavijay do you have a good idea how to support nested details UX-wise? Usually the menu items toggle the feature on/off depending on whether it's currently active or not. If we just always make the menu item create a new details section, there would be no way to turn the details off.

I'm also fine with not supporting nested details in the UI for now as I don't expect this to be used this a lot. People who really need nested details will have to manually add it in their markdown source then.

src/components/Menu/entries.js
jancborchardt: > In the end the details is just another "block style"

Yeah, good point, makes sense to directly group it properly.

jancborchardt:
> I'm also fine with not supporting nested details in the UI for now as I don't expect this to be used this a lot. People who really need nested details will have to manually add it in their markdown source then.

@mejo- I agree it’s best to skip nested details for now. :)

src/components/Menu/entries.js
mejo-: I agree, but then we probably should move block quote and code blocks into this submenu as well straight away, right? What icon should be used for the "block style" submenu?

jancborchardt (approved):

juliushaertl:
Very nice, code looks clean, has tests and works like a charm. 👍 

Small UI issue with content that seems to be artificially limited in width, but otherwise good to go from my side.

Screenshot 2024-08-27 at 15 09 08

mejo-:
> Small UI issue with content that seems to be artificially limited in width, but otherwise good to go from my side.

Good catch, fixed now.

juliushaertl (approved):

bthnakkurt:
is it added now? can i use it by cloning?

mejo-:
> is it added now? can i use it by cloning?

This will become available in Nextcloud 31 (currently in development).

Issue #6267, opened 2024-08-27
Embedded editor in preview mode via `NcReferenceWidget` doesn't support attachments


mejo-:
**Describe the bug**
The embedded editor in preview mode via `NcReferenceWidget` doesn't know about a `fileId` and therefore has not `AttachmentsResolver`. That results in attachments/images not being loaded.

**To Reproduce**
1. Create a Text document with an image
2. Add the internal link to this document to another Text document or a Talk conversation
3. If added to anoter Text document, turn the link into a preview
4. See error "Attachment source was not resolved"

**Expected behavior**

**Screenshots**

![grafik](https://github.com/user-attachments/assets/5933b19b-79e5-46ca-8a2a-9f5e3021456b)

**Server details:**
- Nextcloud version: 30.0.0rc2

juliushaertl:
We should have the file id available as the file reference widget passes it to our viewer component 
https://github.com/nextcloud/server/blob/5b7dcc1427e996a3154761877d1e11463ed4bef4/apps/files/src/views/ReferenceFileWidget.vue#L149

So it might be just a matter of passing that along to the read only view in https://github.com/nextcloud/text/blob/3a9c4f6f7a0ce3e6d148314311730b35250e34fc/src/components/ViewerComponent.vue#L20

mejo-:
Nice. We have to migrate from `RichTextReader` to `MarkdownContentEditor` in the `ViewerComponent`, but a quick text revealed it works. I'll open a PR.

PR #6268, opened 2024-08-27
fix(ViewerComponent): use MarkdownContentEditor for readonly views

mejo-:
Also pass the `fileId` to the component. This adds support for displaying images and attachments in the read-only view.

Fixes: #6267

I tested this PR with
* Text in Text (via preview link) in Viewer
* Text in Talk (via preview link)
* Text in Collectives (both view and edit mode)

The editor CSS includes (`prosemirror.css` and `print.css`) were missing in `MarkdownContentEditor`, which resulted in broken view in Talk - probably because that was the only scenario where no second editor was loaded that the same time.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/versions.spec.js
+++ b/cypress/e2e/versions.spec.js
@@ -34,12 +34,12 @@ describe('Versions', () => {
 			cy.get('[data-files-versions-versions-list] li a').should('have.length', 3)
 
 			cy.get('[data-files-versions-versions-list] li a').eq(1).click()
-			cy.get('.viewer__content #read-only-editor')
+			cy.getContent()
 				.find('h1')
 				.should('contain.text', 'V2')
 
 			cy.get('[data-files-versions-versions-list] li a').eq(2).click()
-			cy.get('.viewer__content #read-only-editor')
+			cy.getContent()
 				.find('h1')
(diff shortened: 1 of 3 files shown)


juliushaertl:
cypress failure seems related:

```
  1) Versions
       View versions with close timestamps:
     AssertionError: Timed out retrying after 10000ms: Expected to find element: `.viewer__content #read-only-editor`, but never found it.
      at Context.eval (http://localhost:8081/__cypress/tests?p=cypress/e2e/versions.spec.js:104:12)

  2) Versions
       View versions:
     AssertionError: Timed out retrying after 10000ms: Expected to find element: `.viewer__content #read-only-editor`, but never found it.
      at Context.eval (http://localhost:8081/__cypress/tests?p=cypress/e2e/versions.spec.js:123:12)

  3) Versions
       Compare versions:
     AssertionError: Timed out retrying after 10000ms: Expected to find element: `.viewer__content #read-only-editor`, but never found it.
      at Context.eval (http://localhost:8081/__cypress/tests?p=cypress/e2e/versions.spec.js:144:12)

Issue #6269, opened 2024-08-07
Windows compatibiliy: upload a file with wrong filename


tobiasKaminsky:
1. enable windows compatibility
2. upload a not-allowed file to an collectives page, e.g. CON.jpg
3. see 
![image](https://github.com/user-attachments/assets/15018b0f-ad63-4a99-8b85-a98adc7f542a)

mejo-:
Transfered to Text as this is a Text issue.

What would be the expected behaviour here? Convert the filename to a compatible name? Is there APIs available in frontend and/or backend to sanitize the name?

juliushaertl:
I think we should just provide a proper error message there to tell the user what was wrong.

Those would be some pointers around that:

https://github.com/nextcloud/files-clients/issues/58#issuecomment-2200130583
https://github.com/nextcloud/spreed/pull/12919/files

PR #6271, opened 2024-08-28
fix(attachments): Show proper error message at upload error

mejo-:
Fixes: #6269

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/lib/Controller/AttachmentController.php
+++ b/lib/Controller/AttachmentController.php
@@ -23,6 +23,7 @@ use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\RedirectResponse;
 use OCP\Files\IMimeTypeDetector;
+use OCP\Files\InvalidPathException;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\Util;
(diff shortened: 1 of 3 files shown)


lib/Controller/AttachmentController.php
```diff
@@ -135,7 +135,8 @@ public function uploadAttachment(string $token = ''): DataResponse {
 			return new DataResponse(['error' => 'No uploaded file'], Http::STATUS_BAD_REQUEST);
 		} catch (Exception $e) {
 			$this->logger->error('Upload error', ['exception' => $e]);
-			return new DataResponse(['error' => 'Upload error'], Http::STATUS_BAD_REQUEST);
+			$error = $e->getMessage() ?: 'Upload error';
```
juliushaertl, line 138: Always returning the exception message is not ideal as it may leak internal data to end users, maybe we can limit it to InvalidPathException instances?

Ideally the server would extend HintException and have a error message explicitly specified for the end user, but that is another story I guess 🙈

juliushaertl (approved):
Small comment, but otherwise 👍

lib/Controller/AttachmentController.php
mejo-: Good catch! I adjusted this now.

juliushaertl (approved):

Issue #6031, opened 2024-07-12
Backport read only session fix to 27


juliushaertl:
https://github.com/nextcloud/text/pull/5895

PR #6287, opened 2024-08-29
[stable27] Ignore pushes of update steps for read only sessions

luka-nextcloud:
### 📝 Summary

* Resolves: #6031

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -36,6 +36,7 @@ use OCA\Text\Db\Step;
 use OCA\Text\Db\StepMapper;
 use OCA\Text\Exception\DocumentHasUnsavedChangesException;
 use OCA\Text\Exception\DocumentSaveConflictException;
+use OCA\Text\YjsMessage;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\Constants;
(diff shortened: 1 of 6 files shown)


lib/Service/DocumentService.php
```diff
@@ -230,11 +231,16 @@ public function writeDocumentState(int $documentId, string $content): void {
 	public function addStep(Document $document, Session $session, $steps, $version, $shareToken): array {
 		$sessionId = $session->getId();
 		$documentId = $session->getDocumentId();
+		$readOnly = $this->isReadOnly($this->getFileForSession($session, $shareToken), $shareToken);
 		$stepsToInsert = [];
 		$querySteps = [];
 		$getStepsSinceVersion = null;
 		$newVersion = $version;
 		foreach ($steps as $step) {
+      $message = YjsMessage::fromBase64($step);
```
juliushaertl, line 240: Indentation is wrong. ;)

juliushaertl (changes requested):
Can you also fix the commit message to be the same as the original commit? Otherwise this will be very painful in case anyone needs to go through the git history

max-nextcloud (approved):
looks good to me.
Thanks for taking care of this!

Issue #6324, opened 2024-09-06
[Nextcloud 31] Text does not work on public shares


susnux:
**Describe the bug**

With Nextcloud 31 public shares also use the same UI as the files app, meaning the legacy UI is dropped in favor of the new (Nextcloud 28) Vue UI.
Sadly the text app does not work out-of-the-box, meaning if you open a public share and then open a markdown file (any file that text can open) the app fails to show the file.
The problem seems that Text does not detect the public share correctly and thus calls the wrong entry point to create a session.

Maybe you can use this functions for public share detection / share token retrieval if needed:
https://nextcloud-libraries.github.io/nextcloud-sharing/modules/public.html

**To Reproduce**
1. Use the master branch (Nextcloud 31)
2. Create a public share
3. Put a MD file inside
4. Open public share
5. Try to view the markdown file

**Expected behavior**
Text can open and view / edit the file as with Nextcloud 30 and before.

**Screenshots**

[Bildschirmaufnahme_20240906_164408.webm](https://github.com/user-attachments/assets/a59702f7-2f04-4833-a1d7-3467f44aab4e)

```
XHRPUT
http://localhost:8081/index.php/apps/text/session/251/create
[HTTP/1.1 401 Unauthorized 28ms]

	
fileId	251
filePath	"/Neue Textdatei.md"
forceRecreate	false
guestName	""
token	null
```

**Server details:**
- Nextcloud version: 31
- PHP Version: 8.3

susnux:
Thank you @mejo- for awesomely fast fixing this :rocket:

PR #6339, opened 2024-09-09
fix(public): Use `@nextcloud/sharing` to detect public shares

mejo-:
Fixes: #6324

Depends on:
* [x] https://github.com/nextcloud/server/pull/47841
* [x] https://github.com/nextcloud/viewer/pull/2486
### 🏁 Checklist


max-nextcloud (approved):
Looks good to me. Happy that we are using `@nextcloud/sharing` rather than extracting the info from the page ourselves.

max-nextcloud:
image
ImageView native attachment specs are failing. Looks like this might also be related.

mejo-:
>  ImageView native attachment specs are failing. Looks like this might also be related.

This is another unrelated failure that already existed before the changes on server 31. These tests are also failing on stable30.

Issue #5736, opened 2024-04-30
Content lost after file renaming from .md to .txt


cybermyber:
**Describe the bug**
If you rename an .md file to a .txt file, the content will be lost after opening it in the built-in text editor.
If you download the renamed file before opening it in the built-in text editor, the content is retained.

**To Reproduce**
1. Create or upload a new .md-File with some content
2. Open the file with the text editor and check the content
3. Close the file and rename it to example_filename.txt
4. Open the example_filename.txt-File with the text editor

**Expected behavior**
After renaming from .md to .txt, the former content should be retained after opening with the text editor.

**Screenshots**
![Screenshot from 2024-04-30 13-40-26](https://github.com/nextcloud/text/assets/67204073/0db7d375-a26f-4ce6-873b-0a08e7ac208f)
![Screenshot from 2024-04-30 13-40-50](https://github.com/nextcloud/text/assets/67204073/a0623dc0-faed-4cc7-b970-7fed6c5bb903)

**Server details:**
- Nextcloud version: 28.0.2
- PHP Version: 8.1
- Database: MariaDB 10.5.23

**Client details:**
 - OS: Fedora 39
 - Browser: Chrome
 - Browser version: 123.0.6312.122 
 - Device: Desktop

mejo-:
I was able to reproduce this. Definitely something we should look into. @juliushaertl @max-nextcloud any idea what is happening here? I would have expected the markdown source to be displayed in the editor instead of all content being deleted :thinking:

max-nextcloud:
I suspect the file id stays the same after the rename but the file is now opened with the plaintext editor rather than the markdown editor. At the same time it tries to recover the y.js state from the previous editing session which is invalid in the plaintext editor context.

Nice catch! We did not account for that yet.

max-nextcloud:
I see two ways to fix this:
1) Clear the document state and all related database entries when a rename changes the mime type between markdown and something else.
2) add a mimetype or editor type to the `key` for finding the document state for a given document
3) both

PR #6345, opened 2024-09-11
fix: Reset session if file rename changes mimetype from/to markdown

mejo-:
For the event target file, we cannot use `getMimetype()` or `getExtension()` as it's a node of type `NonExistingFile`.

Fixes: #5736

### 🏁 Checklist


max-nextcloud (approved):
Code changes look good after a review. Did not test it though.

Would it be possible to have an integration or e2e test for this?

mejo-:
> Would it be possible to have an integration or e2e test for this?

Good point, I added e2e tests now.

Issue #5695, opened 2024-04-18
Editing in code block does not work in firefox 


Aveyron-RetD:
**Describe the bug**
In text try to edit code blocks. Cursor is going everywhere, input is jumping from line to line

**To Reproduce**
Steps to reproduce the behavior: reproduced on firefox not on edge.
1. Insert some code 
2. Click on the code => Blue borders appears
3. ![image](https://github.com/nextcloud/text/assets/53221471/9c21431e-4b7d-4708-9af3-d96d557ec252)
4. Put the cursor to edit delete or add some stuff. Text is put in different sections of the code

**Expected behavior**
Cursor is put and follow normal behaviour.

**Screenshots**

**Server details:**
- Nextcloud version: 28.0.4
- PHP Version: [e.g. 8.2]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome] Firefox 125
 - Browser version: [e.g. 22]
 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

Aveyron-RetD:
Solved in 29.0.1

XueSheng-GIT:
I'm also on NC 29.0.1 and for me the issue is only partly fixed. Those blue lines still appear (not sure what the meaning of this marking is) and adding or removing text by typing seems to work again.
But pasting text still does not work as expected (text is placed at different places) and the cursor jumps around.

@Aveyron-RetD Could you double check whether pasting text does work while the lines are marked blue?

@mejo- Should I create a new issue or would it be better to reopen this one?

Aveyron-RetD:
@XueSheng-GIT you are right, and i can still re-trigger the original bug report /-(
sorry for the false report !!
re-opening as not fixed in 29.0.1

mejo-:
Thanks for the reports everyone. So far I'm unable to reproduce this with latest Firefox (127.0) on Linux with Wayland. Could you give a few more details about your setup?

* Are you only able to reproduce this in Firefox, or also in Chrome/Chromium or other browsers?
* If it's only in Firefox, could you try to reproduce this in safe browsing mode with all add-ons disabled?
[...]

PR #6346, opened 2024-09-11
fix(CodeBlock): Fix visibleFocus and pasting in code blocks

mejo-:
Fixes: #5695

The `tabindex` was introduced with commit 2c9e7a0, but I was not able to reproduce a problem with tab navigation and/or focus trap after removing it.

### 🏁 Checklist


juliushaertl:
I cannot remember the exact reason, but double checked the focus order and seems all fine. 👍

juliushaertl (approved):

Issue #2438, opened 2022-05-23
Add buttons to change list indentation level


flueders:
**Is your feature request related to a problem? Please describe.**
When I tried to create (un-)ordered lists with more than one level, I couldn't find an intuitive way to do this (since there is also no way to change to code view). My first workaround was to edit the file in an external editor, but this is no good solution, I think.

After some research, I discovered the "official" way of indenting list entries: pressing the tab key. However, there was [this issue ](https://github.com/nextcloud/text/issues/601) which added additional problems.

And, by the way, I still don't know how to decrease the list indention level using the keyboard.

**Describe the solution you'd like**
I would like to have a button to easily change a list entry's level. This could also be a drop-down underneath the list buttons.

**Describe alternatives you've considered**
As written before: using external editors or the tab key. There is also the [markdown editor app](https://github.com/icewind1991/files_markdown), which handles this problem by offering a code view.

**Additional context**
Maybe related to https://github.com/nextcloud/text/issues/234

juliushaertl:
I think especially on mobile such buttons would generally make sense, where I also run into this from time to time. Would be interesting to see how other editors and mobile apps handle this, as adding just more icons to the menubar is nothing we can do any longer, as is is already really crowded now that we added more general formatting options.

One option might be a contextual toolbar that only displays relevant actions based on the content type, maybe at the bottom of the screen? Or a floating menu before or after the content https://tiptap.dev/api/extensions/floating-menu

cc @nimishavijay @jancborchardt @max-nextcloud @vinicius73

nimishavijay:
> One option might be a contextual toolbar that only displays relevant actions based on the content type

This sounds good! I like the idea of having formatting options in the bottom of the screen since the top is getting crowded now.

> Or a floating menu before or after the content

Floating menu can cover the content of the document which can be a bit frustrating in my experience. 

This is how Google does it: When you select a bullet list, some of the formatting options (text and highlight color) are hidden and new options to increase and decrease indent as well as a check list toggle are shown. We could do something similar with the increase and decrease indent buttons
| No formatting | Bullet list |
|--|--|
|![image](https://user-images.githubusercontent.com/52440189/169866914-ebc14743-bada-4355-ac12-6b543845b690.png) | ![image](https://user-images.githubusercontent.com/52440189/169867004-12ffcd8f-ff4a-4eee-ab0e-4410841a1382.png) |

flueders:
Thank you for your reactions!

[...]

PR #6353, opened 2024-09-13
feat(menu): Put list items in submenu, add indentation options

mejo-:
Fixes: #2438
Contributes to: #2836

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

| Screencast |
| --- |
| recording1 |

### 🏁 Checklist


jancborchardt (approved):
Very nice! Basically good to go, only 2 remarks:
- [x] A separator line between the list options and indentation options would be useful (as per the suggestion)

Sidenote: This change makes it more obvious that once you decide for ordered/unordered list or checklist, you can not change between them. Any specific reason for that? (But that's a different issue.

jancborchardt (changes requested):
Sorry, actually missed something: It is "indent**ation**", not "indention". :)

cypress/e2e/workspace.spec.js
```diff
@@ -104,15 +104,15 @@ describe('Workspace', function() {
 		})
 	})
 
-	it('creates lists', function() {
+	it.only('creates lists', function() {
```
juliushaertl, line 107:
```suggestion
	it('creates lists', function() {
```

mejo-:
> Sidenote: This change makes it more obvious that once you decide for ordered/unordered list or checklist, you can not change between them. Any specific reason for that? (But that's a different issue.

True, but that's another issue tracked in its own issue already. No idea so far, why this is the case.

mejo-:
> - A separator line between the list options and indentation options would be useful (as per the suggestion)
> - (EDIT: Actually both seem to exist, but "indentation" is the more commonly used one. If we want to be short, then we can use "Increase indent" / "Decrease indent".)

Thanks for the review @jancborchardt. Both added now, see the updated screenshot (I didn't update the screencast) :blush:

juliushaertl (approved):

Issue #6185, opened 2024-08-09
pressing UP on newline goes to line 1, not up one line


gerbrent:
**Describe the bug**
When the cursor is on an empty line (i.e. newline), then the user pushes `up` arrow on the keyboard, the cursor jumps up to line 1 instead of the line immediately above.

**To Reproduce**
1. create new text document
2. create several newlines
3. Click on an empty line in your document - i.e. line 3, cursor must appear at character 1 position
4. press `up` to go to the line above

**Expected behavior**
cursor should move 1 line up

ACTUAL BEHAVIOR

cursor moves to line 1, regardless of which line it started on.

**Screenshots**
[Screencast_20240809_130421.webm](https://github.com/user-attachments/assets/71ef15e8-b5d8-42ff-9063-9f0ae5dbf030)

**Server details:**
- Nextcloud version: 30.0.0 beta 5
- PHP Version: 
- Database: 

**Client details:**
 - OS: linux / NixOS
 - Browser: Firefox
 - Browser version: 128.0.3
 - Device: desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

XueSheng-GIT:
Actually, I wanted to comment to this issue but did to another one by accident. Thus, now adding my intended comment as quote:

> Just tried this on Win11 with Firefox 129 and NC29.0.4. Following your steps, cursor does not jump to line 1, it doesn't move at all (it just stays at its position). I need to press left or use the mouse to move the cursor up again.
> Thus, there seems to be an issue with different symptoms.

I double checked with NC29.0.6 and Firefox 130 on Win11 and the behaviour is still the same (as quoted).

mejo-:
Some more findings:

* Using the upstream `Paragraph` extension fixes this. So it's related to our custom `Paragraph` implementation.
* It's not reproducible in upsteams Tiptap example editors.
* Removing the `PreviewOptions` element from `ParagraphView` fixes it as well.
* Removing everything from `PreviewOptions` and making it a simple `<div />` doesn't fix it.
* Maybe somehow Tiptap doesn't like an extra element in the paragraph structure?

@max-nextcloud

s-ailus2:
[...]

PR #6425, opened 2024-09-18
fix(paragraph): migrate preview options to decorations

mejo-:
### 📝 Summary

* Resolves: #6185

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


cypress/e2e/nodes/PreviewOptions.spec.js
```diff
@@ -42,8 +42,8 @@ describe('Preview Options', function() {
 	})
 
 	it('should Remove link', function() {
-		cy.get('.paragraph-content').should('have.text', 'nextcloud.com')
+		cy.get('p').should('have.text', 'nextcloud.com')
```
max-nextcloud, line 45: hmm... That selector looks super generic. Doesn't it match any `p` element on the page?

max-nextcloud: Maybe we can bring back the `paragraph-content` class - the other `p` selectors in css are also quite generic.

max-nextcloud (approved):
Just a minor nitpick.

max-nextcloud:
Oh... did you manage to fix that function that checked if the decorations can be mapped or need to be recomputed?

cypress/e2e/nodes/PreviewOptions.spec.js
mejo-: I changed it to `p > a` now, which also fixes the test.

Issue #4054, opened 2023-04-14
Migrate to files:node:updated


juliusknorr:
After https://github.com/nextcloud/server/pull/37726 we need to adjust the event that a file has been updated, however we should actually make sure to send a full Node object there not just the fileid.

> Note: For the other apps that already started using the old files:file:updated syntax, we should migrate.
As those apps are shipped and have stablexx branches, I think it's safe to change it for master in both server and those apps. Only Notes would be an issue:

[nextcloud/notes@f635c5c/src/components/NoteRich.vue#L63](https://github.com/nextcloud/notes/blob/f635c5c94d30130a5eaea317a05c7f558b78aaa4/src/components/NoteRich.vue#L63) @juliushaertl
[nextcloud/text@01ccc7c/src/components/Editor.vue#L637](https://github.com/nextcloud/text/blob/01ccc7c17eb21d39fd8f4077538b97aa36dd4f0d/src/components/Editor.vue#L637) @juliushaertl

PR #6427, opened 2024-09-18
feat: Migrate to files:node:updated

luka-nextcloud:
### 📝 Summary

* Resolves: #4054

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -76,6 +76,7 @@ import { getCurrentUser } from '@nextcloud/auth'
 import { loadState } from '@nextcloud/initial-state'
 import { isPublicShare } from '@nextcloud/sharing/public'
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
+import { File } from '@nextcloud/files'
 import { Collaboration } from '@tiptap/extension-collaboration'
 import Autofocus from '../extensions/Autofocus.js'
 import { Doc } from 'yjs'
@@ -121,6 +122,8 @@ import Wrapper from './Editor/Wrapper.vue'
 import SkeletonLoading from './SkeletonLoading.vue'
 import Assistant from './Assistant.vue'
 import Translate from './Modal/Translate.vue'
+import { generateRemoteUrl } from '@nextcloud/router'
+import { fetchNode } from '../services/WebdavClient.ts'
 
 export default {
 	name: 'Editor',
@@ -246,6 +249,7 @@ export default {
 			document: null,
 			sessions: [],
 			currentSession: null,
+			fileNode: null,
 
 			filteredSessions: {},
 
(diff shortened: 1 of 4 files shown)


luka-nextcloud:
> @luka-nextcloud what's the background for this change? Does it fix anything? Is it a technical dept thing?

Sorry for not mentioning. This is a technical debt, I've added the issue on the description.

mejo-:
Thanks for the additional context @luka-nextcloud. Do you know what the emitted event is supposed to do? I.e. is there a reproducer of how to test whether this works?

luka-nextcloud:
@mejo-  The emitted event is supposed to notify the other apps that a file was updated. Install text app and notes app (with this PR https://github.com/nextcloud/notes/pull/1377) then open notes app to edit a note.

src/components/Editor.vue
```diff
@@ -650,7 +652,14 @@ export default {
 		},
 
 		onSave() {
-			emit('files:file:updated', { fileid: this.fileId })
+			const user = getCurrentUser()
+			const node = new File({
+				id: this.fileId,
+				source: generateRemoteUrl(`dav/files/${user.uid}${this.relativePath}`),
+				mtime: new Date(),
+				mime: 'text/markdown',
```
juliushaertl, line 660: The text app might also handle other mime types, so we should take the one that the viewer passes along to us

juliushaertl (changes requested):
Small comment

src/components/Editor.vue
juliusknorr: @luka-nextcloud Can you check that one please?

juliusknorr (approved):

juliusknorr:
Was just reported as an issue so this likely fixes https://github.com/nextcloud/text/issues/6671

juliusknorr:
- [ ] Double check and test if this resolves https://github.com/nextcloud/text/issues/6671

src/components/Editor.vue
```diff
@@ -667,7 +669,13 @@ export default {
 		},
 
 		onSave() {
-			emit('files:file:updated', { fileid: this.fileId })
+			const node = new File({
+				id: this.fileId,
+				source: generateRemoteUrl(`dav/files/${this.currentSession.userId}${this.relativePath}`),
+				mtime: new Date(),
+				mime: this.mime,
```
juliusknorr, line 676: @luka-nextcloud Can you explain why we need to also emit source and mime here? I fear source could break that way if we are using text on a public share link.

luka-nextcloud: @juliusknorr I have just checked again, source and mime are mandatory.

luka-nextcloud: @juliusknorr according to
https://github.com/nextcloud-libraries/nextcloud-files/blob/333c1fc5f64904ec73dff627118b9872992c0d28/lib/files/nodeData.ts#L80-L82
https://github.com/nextcloud-libraries/nextcloud-files/blob/333c1fc5f64904ec73dff627118b9872992c0d28/lib/files/nodeData.ts#L107-L110

juliusknorr:
So many cypress failures seem odd, will do another rebase.

susnux:
> [x] Double check and test if this resolves

Can we backport to 30 and 29?

PR #6479, opened 2024-09-30
fix(sync): Disable browser broadcast

max-nextcloud:
Ensure the browser broadcast does not lead to out of sync docs.

If a tab fails to send a step over the network
it may still distributes it to a different tab / window
via the browser broadcast.

That other window will include the step in the yjs history
without forwarding it to other parties.

From that point on all changes in the second tab
will depend on the step that other clients did not receive.

Therefore these changes will not be applied at other parties.

To prevent this scenario disable browser broadcast completely.


mejo- (approved):

max-nextcloud:

PR #6484, opened 2024-10-01
fix(deps): update typedoc to support typescript version

max-nextcloud:
Also add a renovate rule to always update them together.

Before:
grafik

After:
grafik


blizzz (approved):
~~is renovate added intentionally?~~ reading helps

GretaD (approved):

PR #6486, opened 2024-10-01
Check link protocol

max-nextcloud:
Only preview and link known protocols such as http and https.

Links will still be added to the dom for mailto: and tel: as well.
So they can still be opened via the browser context menu.


blizzz (approved):

max-nextcloud:

PR #6490, opened 2024-10-02
fix(sync): handle 502 (Bad Gateway) gracefully

max-nextcloud:
This is the proxies response when the request to the server fails.
It is often temporary - so it makes sense to keep trying.

Without this text will stop fetching changes from the server
on the first 502 it encounters.


blizzz (approved):

max-nextcloud:

juliushaertl:
Nice one :)

PR #6505, opened 2024-10-08
refactor(EditorFactory): split createEditor in rich and plain variant

max-nextcloud:
Use two functions that do one thing
rather than one with a boolean flag.

Simplifies the functions use and signature a lot
in particular since plain and rich editor need different arguments.

### 🏁 Checklist

- [x] tests are not needed as this is a refactoring.
- [x] Documentation is not required.


max-nextcloud:
Cypress failures in assistant and version tests are known and unrelated.

max-nextcloud:
Superseeded by #6507

Issue #6449, opened 2024-09-23
stylelint broken after migrating from webpack to vite


mejo-:
**Describe the bug**
For some reason `npm run stylelint` doesn't catch errors since the migration to vite. That's true on main and stable30 branches.

PR #6510, opened 2024-10-08
chore: install @nextcloud/prettier-config, eslint-config-prettier

luka-nextcloud:
### 📝 Summary

* Resolves: #6449

- Install @nextcloud/prettier-config
- Install eslint-config-prettier
- Format code with prettier

### DIFF
--- a/src/components/Assistant.vue
+++ b/src/components/Assistant.vue
@@ -4,7 +4,8 @@
 -->
 <template>
 	<div v-if="showAssistant" class="text-assistant">
-		<FloatingMenu v-if="$editor"
+		<FloatingMenu
+			v-if="$editor"
 			:editor="$editor"
 			:tippy-options="floatingOptions()"
 			:should-show="floatingShow"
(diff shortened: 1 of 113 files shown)


src/tests/plaintext.spec.js
```diff
@@ -3,85 +3,92 @@
...
+		.split('&')
+		.join('&amp;')
+		.split('<')
+		.join('&lt;')
+		.split('>')
+		.join('&gt;')
+		.split('"')
+		.join('&quot;')
+		.split("'")
+		.join('&#039;')
```
max-nextcloud, line 22: I prefer the previous version as these calls come im pairs of `split(..).join(...)` and that's easier to see in the original.

Normally I would agree that it's a good idea to put every function call on a new line... But here I don't. So I wonder if the rule that enforces this is really a good idea.

max-nextcloud (commented):
Thanks for looking into this @luka-nextcloud 

I'm all for standardizing things and linting them. I'll point out a few things that i am not so sure about though.

src/components/Assistant.vue
```diff
@@ -14,34 +15,45 @@
 				<template #icon>
 					<CreationIcon :size="20" class="icon" />
 				</template>
-				<NcActionButton v-for="type in taskTypes"
+				<NcActionButton
+					v-for="type in taskTypes"
```
max-nextcloud, line 19: I think pretty much everywhere else in nextcloud we have the first attribute on the line with the tag name just like this was before.

I actually like the separation - but I care about consistency more. So not sure what best to do here.

luka-nextcloud:
@max-nextcloud I have checked through prettier config options. The only way to avoid line break is rising the value of `printWidth`. The current value is `85`. I think it will work?
BTW, we can use // prettier-ignore to ignore formatting for any code block that we don't want to format.

max-nextcloud:
I like the fact that prettier has so little options.
Maybe we can briefly discuss on Monday before moving the code over?

juliusknorr:
@luka-nextcloud What are the next steps here? Do we want to continue or close this?

luka-nextcloud:
@max-nextcloud @juliusknorr 
I will close this PR and open a new one.

- Install `@nextcloud/prettier-config` and `eslint-config-prettier`
- Format code with prettier
- Add flow `.github/workflows/lint-prettier.yml`
- Remove flow `.github/workflows/lint-stylelint.yml`

PR #6539, opened 2024-10-17
fix(build): without non deterministic vendor-LICENSE.txt

max-nextcloud:
The content of `vendor.LICENSE.txt` is not deterministic.

If two versions of one library are imported
one of them will be picked and listed in the file.

This results in arbitrary changes to the file
which in turn create npm fix audit PRs such as #6534.

See also: https://nextcloud-libraries.github.io/nextcloud-vite-config/interfaces/BaseOptions.html#thirdPartyLicense


juliushaertl:
I think we added that on purpose for compliance reasons, but I'm unsure what the current state there is.

activity for example also has this disabled and a separate flag but commented out: https://github.com/nextcloud/activity/blame/master/vite.config.ts#L16-L18

@susnux Any hint there? Is this fine to turn off for now?

susnux:
It is ok to do so, but instead enabled `extractLicenseInformation` to keep us compliant.

juliusknorr:
@max-nextcloud Can you adjust that?

juliusknorr (changes requested):
Needs changes

max-nextcloud:
> @max-nextcloud Can you adjust that?

Done. Added `extractLicenseInformation: true`.

juliusknorr (approved):

juliusknorr:
@max-nextcloud I think this broke reuse compliance, can you check that? Seems it only happend on main as the committed assets were only updated there.

max-nextcloud:
@juliusknorr ahh... I was wondering what broke it. I'll take a look

juliusknorr:
Please remove the backport-request label if you manually backported :)

PR #6540, opened 2024-10-17
feat(table): add row, column buttons outside of table

luka-nextcloud:
### 📝 Summary

* Resolves: #2267

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] check if rows and columns can be added / removed and disable buttons accordingly
- [ ] 

### 🏁 Checklist

### DIFF
--- a/src/css/prosemirror.scss
+++ b/src/css/prosemirror.scss
@@ -322,10 +322,10 @@ div.ProseMirror {
 
 	table {
 		border-spacing: 0;
-		width: calc(100% - 50px);
+		width: calc(100% - 84px);
 		table-layout: auto;
 		white-space: normal; // force text to wrapping
-		margin-bottom: 1em;
+		margin-bottom: 4em;
 		&+ * {
 			margin-top: 1em;
 		}
--- a/src/nodes/Table/TableCell.js
+++ b/src/nodes/Table/TableCell.js
@@ -6,6 +6,7 @@
 import { TableCell } from '@tiptap/extension-table-cell'
 import { Plugin } from '@tiptap/pm/state'
 import { Fragment } from '@tiptap/pm/model'
+import { mergeAttributes } from '@tiptap/core'
 
 export default TableCell.extend({
 	content: 'inline*',
@@ -56,6 +57,17 @@ export default TableCell.extend({
 				parseHTML: (element) => element.style.textAlign || null,
 			},
 		}
+	},
+
+	renderHTML({ HTMLAttributes }) {
+		const attributes = mergeAttributes(this.options.HTMLAttributes, HTMLAttributes)
+		if (attributes.colspan === 1) {
+			delete attributes.colspan
+		}
+		if (attributes.rowspan === 1) {
+			delete attributes.rowspan
+		}
+		return ['td', attributes, 0]
 	},
 
 	addProseMirrorPlugins() {
--- a/src/nodes/Table/TableView.vue
+++ b/src/nodes/Table/TableView.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<NodeViewWrapper data-text-el="table-view" class="table-wrapper">
+	<NodeViewWrapper data-text-el="table-view" class="table-wrapper" :class="{ 'focused': isFocused }">
 		<NcActions v-if="isEditable"
 			force-menu
 			data-text-table-actions="settings"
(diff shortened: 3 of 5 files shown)


juliushaertl:
Seems like a user interface change so please provide some screenshots ;)

luka-nextcloud:
Screenshot & video
image

[demo.webm](https://github.com/user-attachments/assets/0e56e582-61b2-44bf-9e49-6cabb5c184c3)

jancborchardt (commented):
Looks very nice!

When a new row or column gets added, it should also get focused. The first relevant cell, so leftmost for new rows and topmost for new columns.

src/nodes/Table/TableView.vue
```diff
@@ -22,20 +22,41 @@
...
+			:aria-label="t('text', 'Add column after')"
+			:title="t('text', 'Add column after')"
+			@click="addColumnAfter">
+			<template #icon>
+				<TableAddColumnAfter />
+			</template>
+		</NcButton>
+		<NcButton v-if="isEditable"
+			class="table-add-row"
+			:aria-label="t('text', 'Add row after')"
```
mejo-, line 36:
```suggestion
			:aria-label="t('text', 'Add row below')"
```

Maybe better language-wise?

src/nodes/Table/TableView.vue
```diff
@@ -50,25 +71,60 @@ export default {
...
+				.focus()
+				.setTextSelection(this.getPos() + headerRowNode.nodeSize - 1)
+				.addColumnAfter()
+				.run()
+		},
+		addRowAfter() {
+			const lastRowNode = this.node.lastChild
+			this.editor.chain()
+				.focus()
+				.setTextSelection(this.getPos() + this.node?.nodeSize - lastRowNode.nodeSize + 1)
```
mejo-, line 110: The `node?` seems error-prone here. If `this.node` was faulty here, initialization of `lastRowNode` above would already fail, and it would probably lead to unwanted side-effects, no? If `this.node` can really be faulty here, I'd suggest to catch this properly and not do anything in this case.

src/nodes/Table/TableView.vue
```diff
@@ -50,25 +71,60 @@ export default {
...
+	},
 }
 </script>
 
 <style scoped lang="scss">
 .table-wrapper {
 	position: relative;
 	overflow-x: auto;
+
+	&.focused {
```
mejo-, line 123: I'd expect the `&.focused` ruleset to be inside the `.table-add-column` and `.table-add-row` rulesets below, but that's rather nitpicking.

src/nodes/Table/TableCell.js
```diff
@@ -58,6 +59,17 @@ export default TableCell.extend({
 		}
 	},
 
+	renderHTML({ HTMLAttributes }) {
+		const attributes = mergeAttributes(this.options.HTMLAttributes, HTMLAttributes)
```
mejo-, line 63: Could you elaborate a bit on why this is necessary? I didn't grasp what the `colspan` and `rowspan` attributes are used for before.

src/nodes/Table/TableView.vue
```diff
@@ -84,10 +140,41 @@ table {
...
+	opacity: .5;
+	position: absolute;
+	top: var(--default-clickable-area);
+	right: 0;
+	bottom: 60px;
+	margin-top: 0 !important;
+
+	&:hover {
+		opacity: 1;
+		visibility: unset;
```
mejo-, line 162: In my understanding the `visibility: unset` is superfluous. The button already becomes visible on table focus, and it cannot be hovered when `invisible: hidden` anyway.

src/nodes/Table/TableView.vue
```diff
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<NodeViewWrapper data-text-el="table-view" class="table-wrapper">
+	<NodeViewWrapper data-text-el="table-view" class="table-wrapper" :class="{ 'focused': isFocused }">
```
mejo-, line 7: So far, the buttons only become visible when the table is focussed (i.e. the cursor is inside it). How about also making them visible when the table is hovered? I'd expect this from a user experience perspective.

mejo- (changes requested):
In general works really well, thanks for tackling this. I have left some comments in the code. Also I think a cypress test that adds a row or column using one of the buttons would be great to avoid regressions later on.

src/css/prosemirror.scss
```diff
@@ -322,10 +322,10 @@ div.ProseMirror {
 
 	table {
 		border-spacing: 0;
-		width: calc(100% - 50px);
+		width: calc(100% - 84px);
```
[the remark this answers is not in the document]
luka-nextcloud: I have updated to use css variable to make it more meaningful.

src/nodes/Table/TableView.vue
luka-nextcloud: The node should not be null. So, I removed `?`

src/nodes/Table/TableView.vue
luka-nextcloud: This is the focused state of table. So, I think it should be inside `.table-wrapper`.

src/nodes/Table/TableCell.js
luka-nextcloud: The colspan and rowspan will be added automatically by tiptap.

luka-nextcloud:
[demo.webm](https://github.com/user-attachments/assets/01a21826-4503-4518-93e0-df70fe56d6c9)

mejo- (approved):
Thanks, works really well now. Only one remaining question because I don't understand why it belongs into this PR. But not a blocker :relaxed:

src/nodes/Table/TableCell.js
luka-nextcloud: @mejo- I think this comment should answer you.
https://github.com/nextcloud/text/issues/2267#issuecomment-2410600211

PR #6563, opened 2024-10-24
fix/stable27 multi line breaks

max-nextcloud:
- **test(plaintext) add regression test for newlines**
- **fix: only load CodeBlock extension without CodeBlockLowlight**


elzody (approved):
Works well

PR #6564, opened 2024-10-26
feat: Add `allowed_view_extensions` config node

Koc:
### 📝 Summary
It isn't possible to view md files if they were shared without allow downloading. Related to https://github.com/nextcloud/server/pull/48903

<details><summary>:mag: Repoducer</summary>
<p>
Share file without download possibility

image

Try to view file as share recipient

image

</p>
</details> 

To fix that we can introduce a new `allowed_view_extensions` config node.

```php
  'allowed_view_extensions' => [
    'md',
    // ...
  ]
```

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


max-nextcloud:
@Koc Thanks a lot for addressing this. The current state is really confusing.

I'm not sure your proposed solution is really what we want. Before diving into the code I think we should clarify that with @nextcloud/designers .

We're always hesitant to add config options. However this config would not be exposed in the UI I assume. But the problematic scenario would remain for files which are not part of the `allowed_view_extensions` config.

I know the permission setting to prevent downloads is messy. I personally think that this should not exist for the filetypes you list as one can never prevent a screenshot or a copy and paste download. But I'll wait to hear what the designers say.

jancborchardt (commented):
Generally I think it should just work and not necessitate another setting or parameter.

What exactly "just work" means for the "Prevent download" vs being able to view though, it is up to @AndyScherzinger @juliusknorr @sorbaugh to define as it is a technical and security question related to both Files and Office.

Koc:
@jancborchardt any suggestions how to make it "just works". In this case I guess we should hardcode list of extensions somewhere

max-nextcloud:
I had another look at this today. When originally introduced the 'prevent download' option was only available for office mime types:
https://github.com/nextcloud/server/pull/32482/files#diff-a767a3e36d22b92db810b48aae05bc3cad43c31de6afb17e926ad1dee6f4da31R425-R439

I just tried and now i can also apply it to markdown documents. Might make sense to investigate when and why this change happened.

max-nextcloud:
I discussed this again with @juliusknorr today. Looks like removing `allow download` always made it impossible to view the file.

We came up with a plan. As a first step it would be good to have a setting in server that changes the `allow download` option into a `hide download` option with the same behavior as for external shares.

This setting would first need to land in server and then needs to be respected in text. I'll close this PR as it's a different approach and I think it would make sense to work on it in a separate PR.

Thanks a lot for your work @Koc - sorry we can't include it as is.

PR #6598, opened 2024-11-04
feat(debug): Expose yjs debug function in editor API

mejo-:
Should make it easier to debug sync issues.

### Screenshot

Screenshot Firefox | Screenshot Chromium
--- | ---
2024-11-06T12:32:38,426154186+01:00 |2024-11-06T12:38:10,598435570+01:00

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -343,6 +343,7 @@ export default {
 		subscribe('text:image-node:delete', this.onDeleteImageNode)
 		this.emit('update:loaded', true)
 		subscribe('text:translate-modal:show', this.showTranslateModal)
+		this.setupEditorDebug()
 	},
 	created() {
 		this.$ydoc = new Doc()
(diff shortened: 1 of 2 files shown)


src/components/Editor.vue
```diff
@@ -761,6 +761,22 @@ export default {
 			console.debug(editor.getHTML())
 		},
 
+		/**
+		 * Helper method to debug yjs issues
+		 */
+		debugYjs() {
```
juliusknorr, line 767: Do we have this somewhat accessible when text open in viewer so we can also ask people editing to dump the output when editing outside of collectives?

juliusknorr: Having an array of editors sounds reasonable. We can add the filename maybe to the debug output so that the individual ones can be identified (e.g. a readme is open and a viewer instance).

I think we don't need to make it too complex, as long as we can somehow get the information and identify the relevant instance.

mejo-: `OCA.Text.debugYjs` now prints the debugging data of all registered editor instances along with instructions, see updated screenshot above.

juliusknorr (approved):

PR #6602, opened 2024-11-05
fix(addContent): only process addExtensions once

max-nextcloud:
Use `editor.options.extensions`.
This only includes the extensions that were configured
when instantiating the editor.
It does not include the extensions these extensions added.

`generateJSON` will process the provided extensions
and add extensions according to their `addExtensions` field.

Before we used `editor.editorManager.extensions`
which includes all extensions used in the editor
including those added by other extensions.
This lead to `generateJSON` adding the same extensions again
which resulted in a warning.

### DIFF
--- a/src/mixins/setContent.js
+++ b/src/mixins/setContent.js
@@ -35,7 +35,7 @@ export default {
 				? createRichEditor()
 				: createPlainEditor()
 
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Thanks for the swift fix :heart:

juliusknorr (approved):
Good catch 👍

Issue #6152, opened 2024-07-30
Design review issues 2024-07-30


juliusknorr:
**Bold items are considered high priority**

- [x] nextcloud/text#6153
- [x] User element is still rounded, and should be 34px through css variables
- [x] **Viewer: needs to be updated to latest** 
- [x] nextcloud/text#6157
- [ ] #6974
- [x] Heading anchor is not fully aligned with the heading text
- [ ] #6973
- [ ] Assistant bugs
  - [x] **Select text and pick summarize does not go to the right section and no prefilled text**
  - [x] Component: Translate dialog: Component known issue needs to be double checked
  - [x] Text: More minimum height in the translate dialog modal
  - [ ] Components: Select input height and jumping 
  - [ ] Components: Disabled style
  - [x] Show assistant results does not have any results shown (emptycontent), https://github.com/nextcloud/text/pull/6799
- [x] After inserting an attachment, focus should go after the the image -> to the image description, https://github.com/nextcloud/text/pull/6609
- [x] Cursor seems to be sometimes missing for the image caption (second time clicking into it) Maybe only on Firefox
- [x] Reduce spacing between image and description a bit, https://github.com/nextcloud/text/pull/6613
- [x] Code block should have a border radius (large border radius due to the buttons) https://github.com/nextcloud/text/pull/6453
- [x] Table should also have the large border radius -> table has already radius of `--border-radius-large`
- [x] **Table should have a 2px spacing in columns/headers, breathing room for buttons [https://github.com/nextcloud/text/pull/6150](https://github.com/nextcloud/text/pull/6150)**
  - [x] Can use "tertiary no background" as type for the buttons, https://github.com/nextcloud/text/pull/6800
- [x] Code block should have reduced minimum height to align with the buttons, https://github.com/nextcloud/text/pull/6614

---

## Separate topics

- [ ] nextcloud/server#48608

JuliaKirschenheuter:
@nimishavijay could you please remind me, what was meant under

> Cursors in the heading are too small in headings

Thank you!

juliusknorr:
> Cursors in the heading are too small in headings

Have two browser tabs open on the same document, then you can see that the colorful cursors of other users are smaller in height then the text in headings:

![Image](https://github.com/user-attachments/assets/c33c6f13-0a4b-42f8-a121-c4690a2101dd)
![Image](https://github.com/user-attachments/assets/6695eee6-2d2a-4faf-aeba-18e4970a112c)

[...]

PR #6614, opened 2024-11-07
fix(files): Align vertically bottons in code block

JuliaKirschenheuter:
### 📝 Summary

* Part of https://github.com/nextcloud/text/issues/6152
Code block should have reduced minimum height to align with the buttons

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-11-07 13-33-13 | Screenshot from 2024-11-08 10-58-09

### 🏁 Checklist

### DIFF
--- a/src/nodes/CodeBlockView.vue
+++ b/src/nodes/CodeBlockView.vue
@@ -239,7 +239,7 @@ export default {
 	align-items: stretch;
 	position: absolute;
 	right: 12px;
-	margin-top: 4px;
+	margin-top: 9px;
 }
 
 .split-view {


src/nodes/CodeBlockView.vue
```diff
@@ -239,7 +239,9 @@ export default {
 	align-items: stretch;
 	position: absolute;
 	right: 12px;
-	margin-top: 4px;
+	margin: 0;
+	top: 50%;
+	transform: translateY(-50%);
```
juliusknorr, line 244: Not to sure about this, if the code block has multiple lines i think it should still top align. Maybe we can just reduce the height so that it is vertically centered if it is empty/one line?

JuliaKirschenheuter: why then not to change `margin-top` to 8px instead of 4px?

juliusknorr: Sure; sounds good

elzody (approved):
Looks better and seems to do the trick when I tried it

max-nextcloud (approved):
Thanks for tackling this! :heart:

PR #6650, opened 2024-11-14
Send one full update from y-websocket

max-nextcloud:
Always send full diff to server state.

* Inline y-websocket, so we can adjust it to our needs.
* Track the server state by updating a ydoc with all the updates
  received from the server.
* Compute updates based on that server state.
* Only send the last update from the queue (as it includes all previous).

## Todo:

* [ ] Check why it does not work with document state provided.
* [ ] Confirm it works for new files with content.


juliusknorr (commented):
Gave this a first try and basic syncing is still working fine, also some offline online switching on two browser sessions worked as expected.

One small difference I noticed (without reading the code yet in detail):

- When loading the file it seems we no longer wait until the content is loaded to hide the loading indicator, we start with an empty document and can type before any document state is propagated to us. @max-nextcloud is this expected or can it cause any side effects?

src/services/y-websocket.js
```diff
@@ -0,0 +1,539 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
```
juliusknorr, line 3: The file license is probably still MIT then, but not entirely sure, probably something to clarify

I didin't diff how much it is based on it (https://github.com/yjs/y-websocket/blob/master/LICENSE)

lib/Service/DocumentService.php
```diff
@@ -228,8 +228,10 @@ public function addStep(Document $document, Session $session, array $steps, int
 			}
 			$newVersion = $this->insertSteps($document, $session, $stepsToInsert);
 		}
-		// If there were any queries in the steps send the entire history
-		$getStepsSinceVersion = count($querySteps) > 0 ? 0 : $version;
+		// If there were any queries in the steps send all steps since last save.
+		$getStepsSinceVersion = count($querySteps) > 0
+			? $document->getLastSavedVersion()
```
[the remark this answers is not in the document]
max-nextcloud: addressed in the next commit by including the document state that matches `$document->getLastSavedVersion`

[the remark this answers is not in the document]
max-nextcloud: Still using the smaller would mean we have a redundancy in case the client that did the save was out of sync.

max-nextcloud:
What's missing:
* [ ] license for y-websocket
* [ ] loading spinner
* [ ] remove `$queue` from Editor.vue
* [ ] ensure to only store remote state in documentState.

src/services/SyncService.js
```diff
@@ -298,25 +305,12 @@ class SyncService {
 		})
 	}
 
-	async sendRemainingSteps(queue) {
-		if (queue.length === 0) {
+	sendRemainingSteps() {
+		if (!this.outbox.hasUpdate()) {
```
mejo-, line 310: Should probably be `!this.#outbox.hasUpdate()`.

src/services/Outbox.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default class Outbox {
+	#awarenessUpdate = ''
+	#syncUpdate = ''
+	#syncQuery = ''
+
+	storeStep(step: Uint8Array) {
+		const encoded = encodeArrayBuffer(step)
+		if (encoded < 'AAA') {
```
mejo-, line 20: This condition seems a bit superfluous to me given that we log the same message at the end of the function if no other condition matched.

src/services/Outbox.ts
```diff
@@ -0,0 +1,68 @@
...
+			awareness: this.#awarenessUpdate,
+		}
+	}
+
+	get hasUpdate(): boolean {
+		return !!this.#syncUpdate
+	}
+
+	/*
+	 * Clear data that has just been
```
mejo-, line 51: Probably missing a ' sent' at the send of the comment :wink:

mejo- (commented):
Some comments.

max-nextcloud:
> * When loading the file it seems we no longer wait until the content is loaded to hide the loading indicator, we start with an empty document and can type before any document state is propagated to us. @max-nextcloud is this expected or can it cause any side effects?

I cannot reproduce this right now.

src/services/Outbox.ts
max-nextcloud: But then we'd always need to check if encoded is bigger 'AAA' in the other conditions.

We could have one condition in the beginning that checks if it's valid (i.e. > AAA and < 'Ag') and then do the rest.

src/services/SyncService.js
max-nextcloud: Thanks! Plus the hasUpdate is a getter - so need to also remove the `()`.

Interesting sideeffect of this was that `sendRemainingSteps` erred out and `disconnect` never got called in the editor so the connection stayed up even though the viewer was closed. Made that more robust as well.

src/services/Outbox.ts
mejo-: I guess the size comparison is very fast and don't know if this is really relevant from a performance perspective. Maybe it's also relevant how often we actually get messages that are smaller than `AAA` (and bigger than `Ag`) in practice. In any case no blocker, fine to keep it as is :blush:

juliusknorr (approved):

PR #6674, opened 2024-11-21
feat(editor): add "Open in new tab" option for link previews

pbirrer:
This update aims to improve accessibility to files added as previews in the text editor.

### 📝 Summary

- Enhanced `PreviewOptions.vue` to include a new "Open in new tab" button for link previews.
  - Introduced the `openLink` method to open the link in a new browser tab.
  - Added `OpenIcon` for the button to improve UI consistency.
  - New `href` prop added to handle link URLs dynamically.

- Updated `extractLinkParagraphs.js` to extract `href` attributes from nodes.
  - Ensures link previews include the necessary `href` for the "Open in new tab" action.

#### 🖼️ Screenshots

text-open-option-link-preview

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/PreviewOptions.vue
+++ b/src/components/Editor/PreviewOptions.vue
@@ -27,6 +27,14 @@
 				{{ t('text', 'Show link preview') }}
 			</NcActionRadio>
 			<NcActionSeparator />
+			<NcActionButton v-if="href"
+				close-after-click
+				@click="openLink">
+				<template #icon>
+					<OpenIcon :size="20" />
+				</template>
+				{{ t('text','Open in new tab') }}
+			</NcActionButton>
 			<NcActionButton close-after-click @click="deleteNode">
 				<template #icon>
 					<DeleteIcon :size="20" />
@@ -41,6 +49,7 @@
 import { NcActions, NcActionButton, NcActionRadio, NcActionCaption, NcActionSeparator } from '@nextcloud/vue'
 import DotsVerticalIcon from 'vue-material-design-icons/DotsVertical.vue'
 import DeleteIcon from 'vue-material-design-icons/Delete.vue'
+import OpenIcon from 'vue-material-design-icons/OpenInNew.vue'
 
 export default {
 	name: 'PreviewOptions',
@@ -53,12 +62,18 @@ export default {
 		NcActionRadio,
 		NcActionSeparator,
 		DeleteIcon,
+		OpenIcon,
 	},
 
 	props: {
 		type: {
 			type: String,
 			required: true,
+		},
+		href: {
+			type: String,
+			required: false,
+			default: '',
 		},
 	},
 
@@ -78,6 +93,10 @@ export default {
 		},
 		deleteNode() {
 			this.$emit('delete')
+		},
+		openLink() {
+			if (!this.href) return
+			window.open(this.href, '_blank').focus()
 		},
 	},
 }
--- a/src/plugins/extractLinkParagraphs.js
+++ b/src/plugins/extractLinkParagraphs.js
@@ -20,16 +20,17 @@ export default function extractLinkParagraphs(doc) {
 				pos,
 				nodeSize: node.nodeSize,
 				type: 'text-only',
+				href: extractHref(node.firstChild),
 			}))
 		} else if (node.type.name === 'preview') {
 			paragraphs.push(Object.freeze({
 				pos,
 				nodeSize: node.nodeSize,
 				type: 'link-preview',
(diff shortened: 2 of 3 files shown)


max-nextcloud:
Dear @pbirrer 

Thanks a lot for your contribution. I really like the idea and have been missing this in the past.
I wonder if it would make sense to get this out of the three dot menu as a separate button left of it. That would make it more visible and also distinguish it from the preview settings in the dropdown.

However this could also be a follow up.

I will add this PR to my review list and do a code review as soon as possible.

rvjr:
From a usability point of view I would also prefer a direct button; if this is okay with the NC team. It would make the function mich more obvious.

max-nextcloud (commented):
Code looks good. There are a few test failures because of the additional `href` property in `extractLinkParagraph`. They are fairly easy to spot in the Files changed view here on github.

max-nextcloud:
@pbirrer could you try and fix the tests?
You should be able to run them locally with
```
npx vitest src/tests/plugins/extractLinkParagraphs.spec.js

```

Please let me know if you need any help with this.

max-nextcloud:
@pbirrer I revisited the current behavior again and i think you are trying to work around a bug here. The link preview itself should open the link in a new tab but currently does not.
Would fixing that also address the accessibility concern this PR is about

I tried to capture how it's currently working with the plain text link and how it's supposed to be working with the preview as well:
https://github.com/user-attachments/assets/6ec301b7-e0c0-4c00-9e21-4533c659bd76

Does the plain text link version of clicking the link and then clicking the preview work for you?
Would it help if we also made the caption of the link bubble clickable?

juliusknorr:
I would say a button is still great to have in addition to the clickable link preview. Especially since the link preview might be interactive or rendered by external apps, so I'd always prefer to still have a common way of just opening the link

mejo-:
Thanks, I also agree that a button to open the link is still a good idea. But I wonder whether the button should be in the link bubble instead of the three-dot-menu. I perceive the three dot menu more like a "config menu", also something normal users probably won't spot that easily. Most people will just click on the link and see the link bubble that opens. I would expect a "Open in new tab" button there.

pbirrer:
@mejo-: From what I can tell, the link bubble becomes inaccessible in preview mode. To resolve this, the button could be placed nearby—perhaps beside or below the triple-dot button?

juliusknorr:
> From what I can tell, the link bubble becomes inaccessible in preview mode. To resolve this, the button could be placed nearby—perhaps beside or below the triple-dot button?

Preview/read only documents should also show the preview bubble, so I'd agree to always place it in there. It also will make it available for inline links that are not a standalone paragraph, as then we do not show the menu.

Screenshot 2024-12-18 at 23 25 07

max-nextcloud:
@juliusknorr 

> Edit: Though I would also be ok with keeping the action in the three dots as well. Might be more accessible for users if the menu is shown anyways

Just to  clarify - that would mean the same action in two places, right?

max-nextcloud:
Ahh... I think now i get the three dot menu - you're looking into things that have an embedded preview like whiteboards and so on, right?

grafik
(Using a different text document here for illustration as currently whiteboards have 0 hight)

There's no link here right now that takes one to the actual whiteboard.
In this case I think it would be better to put the link next to the ":pencil2: Edit" - button.

max-nextcloud:
All that said... I will fix the linter and prepare this for merging as it's already an improvement over the current state. We can move the link to more meaningful places (i.e. into the link bubble and the preview) as a follow up.

mejo-:
Thanks to everyone who looks into this. I just want to stress the fact that non-techy users expect a link to be accessible through clicking the link. So I still think that the link bubble that opens when clicking the link (the text one, not the preview) should have a "open in new tab" button in my opinion. That's also been requested here: https://github.com/nextcloud/collectives/issues/1226

For link previews (where the link bubble doesn't show) I would prefer to have the button as an overly button next to the "edit" button, like @max-nextcloud suggested.

When I watch people using Text/Collectives, my impression is that the three-dots menu left to links often remains unnoticed. Plus, as @juliusknorr already pointed out, it's only displayed for links that span over a separate paragraph, not for inline links.

rvjr:
@mejo- , @max-nextcloud , what's the final decision here so we can move on?

max-nextcloud (approved):
Definitely an improvement over the status quo. We can still move the link into the link bubble and the preview in a follow up.

max-nextcloud:
> @mejo- , @max-nextcloud , what's the final decision here so we can move on?

I just approved. Could you rebase and push again?

PR #6676, opened 2024-11-21
feat(attachments): add support for creating new attachments

pbirrer:
These changes enable users to create and insert new attachment files directly from within the editor.

### 📝 Summary

- Introduced a new endpoint (`Attachment#createAttachment`) to create attachment files via POST requests.
- Added `createAttachment` method in `AttachmentService` to handle file creation logic with permission checks.
- Updated `MediaHandler.vue` and `MediaHandler.provider.js` to integrate the new attachment creation functionality in the editor.
- Enhanced `ActionAttachmentUpload.vue` to support dynamic attachment creation from file app templates.
- Extended `SessionApi.js` with a `createAttachment` method to interface with the new API endpoint.

#### 🖼️ Screenshots

text-create-attachment

### :ballot_box_with_check: Todo

* [x] The [attachment cleanup](https://github.com/nextcloud/text/blob/main/lib/Service/AttachmentService.php#L544) does not wipe the files because it thinks the attachments were removed from the text content.
* [x] Tests ensure we don't accidentally break creating new files as attachments.
* [x] Check with design if it makes sense to visually separate the attachments section and the embedding new content section.

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/attachments.spec.js
+++ b/cypress/e2e/attachments.spec.js
@@ -116,8 +116,9 @@ const checkAttachment = (documentId, fileName, fileId, index, isImage = true) =>
  * @param {string} requestAlias Alias of the request we are waiting for
  * @param {number|undefined} index of the attachment
  * @param {boolean} isImage is the attachment an image or a media file?
+ * @param {Function} check function used to check document for attachment
  */
-const waitForRequestAndCheckAttachment = (requestAlias, index, isImage = true) => {
+const waitForRequestAndCheckAttachment = (requestAlias, index, isImage = true, check = checkAttachment) => {
 	return cy.wait('@' + requestAlias)
 		.then((req) => {
 			// the name of the created file on NC side is returned in the response
@@ -125,7 +126,7 @@ const waitForRequestAndCheckAttachment = (requestAlias, index, isImage = true) =
 			const fileName = req.response.body.name
 			const documentId = req.response.body.documentId
 
-			return checkAttachment(documentId, fileName, fileId, index, isImage)
+			return check(documentId, fileName, fileId, index, isImage)
 		})
 }
 
@@ -281,6 +282,13 @@ describe('Test all attachment insertion methods', () => {
 	})
 
 	it('Create a new text file as an attachment', () => {
+		const check = (documentId, fileName) => {
+			cy.log('Check the attachment is visible and well formed', documentId, fileName)
+			return cy.get(`.text-editor [basename="${fileName}"]`)
+				.find('.text-editor__wrapper')
+				.should('be.visible')
+		}
+
 		cy.visit('/apps/files')
 		cy.openFile('test.md')
 
@@ -289,7 +297,7 @@ describe('Test all attachment insertion methods', () => {
 		cy.intercept({ method: 'POST', url: '**/text/attachment/create' }).as(requestAlias)
 		clickOnAttachmentAction(ACTION_CREATE_NEW_TEXT_FILE)
 			.then(() => {
-				return waitForRequestAndCheckAttachment(requestAlias, undefined, false)
+				return waitForRequestAndCheckAttachment(requestAlias, undefined, false, check)
 			})
 	})
 
--- a/cypress/e2e/attachments.spec.js
+++ b/cypress/e2e/attachments.spec.js
@@ -12,6 +12,7 @@ const attachmentFileNameToId = {}
 
 const ACTION_UPLOAD_LOCAL_FILE = 'insert-attachment-upload'
 const ACTION_INSERT_FROM_FILES = 'insert-attachment-insert'
+const ACTION_CREATE_NEW_TEXT_FILE = 'insert-attachment-add-text-0'
 
 /**
  * @param {string} name name of file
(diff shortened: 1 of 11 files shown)


max-nextcloud:
Dear @pbirrer 

At first sight this looks like a good addition to me. I wonder if it becomes very verbose if one has a lot of possible file types that can be created.

@marcoambrosini What do you think from a design perspective?

marcoambrosini:
I agree that this is a great addition too, should we have too many "New" menu items, we can add a "Create new" submenu in the future

pbirrer:
Yeah, a 'Create New' submenu could also work pretty well here. The number of file types you can create depends on what’s registered in the files app—it adjusts based on the apps being used in your Nextcloud instance.

max-nextcloud (changes requested):
I tested it locally and noticed that creating the new files is inconsistent with how attachments are inserted thus far.

I'll attach a screencast to illustrate what i mean by that.

If I upload a `.md` file from my computer it will show as a tiny preview with the filename and size. Same if I insert it from the files already on the Nextcloud.

However if i use the new `Create new text file` functionality the file will be embedded.

The markdown syntax matches this difference (first file was uploaded, second was created):

```md
![New text file.md](.attachments.2371/New%20text%20file.md)

[http://nextcloud.local/index.php/f/2393](http://nextcloud.local/index.php/f/2393 (preview))
```

I understand you actually want the embedding, not just an attachment. From my point of view these are two different steps:
1. Enable the creation of new files as attachments (this PR)
2. Make attachments behave (more) like previews.

Please change the behavior to match the one of other attachments for now so we can get this in. I'll add a (failing) cypress test case in the attachments test suite that should pass once the created text files behave the same as the uploaded ones.

max-nextcloud:
Here's a [screencast](https://github.com/user-attachments/assets/672dfcb3-24ed-4760-9694-bf10f5e51f67) illustrating what I am referring to.

rvjr:
Hi @max-nextcloud,

you mean you want the link to use the syntax with the global file id, instead of the relative path? Or do you mean that the file shouldn't be previewed initially (which would not make much sense, when inserting new files...)?

pbirrer:
I understand that the inconsistency between the attachment options is unfortunate. However, as it stands, there’s no way to interact with the default view created by “Upload from computer” or “Insert from file” beyond directly downloading the file or deleting the entry. 

attachment

But I’m assuming this isn’t the intended behavior, correct?

max-nextcloud:
Thanks @rvjr and @pbirrer  for getting back so swiftly!

Let me try and answer you both at once...

> you mean you want the link to use the syntax with the global file id, instead of the relative path? Or do you mean that the file shouldn't be previewed initially (which would not make much sense, when inserting new files...)?

Yes... both was my thinking - but I see how an empty attachment that cannot even be accessed makes little sense.

> But I’m assuming this isn’t the intended behavior, correct?
Well - it's not what we want in the long run - but it's the current behavior of attachments.

I'd like to see a way to convert between the three:
* plain link
* small preview that opens in new tab / downloads
* embedded content that can be edited.
I think this is out of scope for this PR.

I'd be fine with what @pbirrer currently developed if we make sure...
* [ ] the [attachment cleanup](https://github.com/nextcloud/text/blob/main/lib/Service/AttachmentService.php#L544) does not wipe the files because it thinks the attachments were removed from the text content.
* [ ] tests ensure we don't accidentally break it.
* [ ] maybe the attachments section and the embedding new content section are visually separated in the menu.

pbirrer:
Hey @max-nextcloud,

I just added a separator bar in the dropdown menu to clearly separate the new entries from the old ones.

I also attempted to update the attachment cleanup routine. However, I wasn’t able to get the previews working with filenames in the URL, so now the cleanup function will look for any references by file ID as well.

Hope this helps! Let me know if you have any questions or feedback.

lib/Service/AttachmentService.php
```diff
@@ -581,23 +581,44 @@ public function cleanupAttachments(int $fileId): int {
...
+	 *
+	 * @param string $content
+	 *
+	 * @return array
+	 */
+	public static function getAttachmentIdsFromContent(string $content): array {
+		$matches = [];
+		// matches [ANY_CONSIDERED_CORRECT_BY_PHP-MARKDOWN](ANY_URL/f/FILE_ID[ (preview)]) and captures FILE_ID
+		preg_match_all(
+			'/\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[(?>[^\[\]]+|\[\])*\])*\])*\])*\])*\])*\]\(\S+\/f\/(\d+)(?: \(preview\))?\)/',
```
max-nextcloud, line 613: We could be more permissive about the `(preview)` I think as markdown interprets this as the link title. On the other hand attachment handling usually applies to files attached via the text UI - so we can assume the links are formatted consistently.

max-nextcloud (commented):
Code looks good. I will test it one more time and hope we can merge ahead then.

lib/Service/AttachmentService.php
pbirrer: How permissive would you like it to be? just ignoring any whitespace?

rvjr: maybe just ignore anything after `file_id`? Because when it matched, we have everything we need to know which file is being used.

max-nextcloud (approved):
Looks good now. Thanks a lot for following up here!

cypress/e2e/attachments.spec.js
```diff
@@ -279,6 +280,19 @@ describe('Test all attachment insertion methods', () => {
...
+	it('Create a new text file as an attachment', () => {
+		cy.visit('/apps/files')
+		cy.openFile('test.md')
+
+		cy.log('Create a new text file as an attachment')
+		const requestAlias = 'create-attachment-request'
+		cy.intercept({ method: 'POST', url: '**/text/attachment/create' }).as(requestAlias)
+		clickOnAttachmentAction(ACTION_CREATE_NEW_TEXT_FILE)
+			.then(() => {
+				return waitForRequestAndCheckAttachment(requestAlias, undefined, false)
```
max-nextcloud, line 292: This is failing on CI right now. I suspect the dom structure of the text attachment is different from the image dom structure.

pbirrer: @max-nextcloud: Would you mind helping me with this issue? Most of the Cypress tests on my system are failing, and I currently don't have enough time to investigate this further.

pbirrer:
@max-nextcloud Do you have any suggestions on the next steps to help get this finally merged?

max-nextcloud:
> @max-nextcloud Do you have any suggestions on the next steps to help get this finally merged?

I'm sorry for taking so long to click the merge button :wink: - had a last quick review.

:bouquet: Thanks a lot for following up on this, adding and fixing the tests.

Issue #6573, opened 2024-10-30
Text inserted from the Assistant is not formatted correctly but shows raw markdown


jancborchardt:
**Describe the bug**
See title

**To Reproduce**
1. Open a new text file
2. Go to the Assistant functions on the right
   ![Image](https://github.com/user-attachments/assets/df84de1e-21f8-4d10-ba44-467bc97d936e)
3. Generate a text
4. Click "Insert"

**Expected behavior**
The text should be properly formatted with headings (or bolding) and lists.

**Screenshots**
Instead the raw markdown code is visible:
![Image](https://github.com/user-attachments/assets/c4bceef1-0d8d-4d30-af04-d5cf22660787)

**Server details:**
- Nextcloud version: Current daily server

jancborchardt:
@mejo- @max-nextcloud @julien-nc not fully sure if this is a Text issue or Assistant issue, but since Assistant just gives over the plaintext, I assume it is a Text app issue?

juliusknorr:
Should be text, we need to detect or always assume markdown like we do for smartpicker results:

https://github.com/nextcloud/text/blob/1d2f4cc4ad0c3a7a7fd4a93fd2478a84565c68eb/src/components/Suggestion/LinkPicker/suggestions.js#L91

Ref https://github.com/nextcloud/text/issues/4784

juliusknorr:
@luka-nextcloud Can you take that one?

PR #6694, opened 2024-11-25
feat: render text inserted from assistant

luka-nextcloud:
### 📝 Summary

* Resolves: #6573

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

### DIFF
--- a/src/components/Assistant.vue
+++ b/src/components/Assistant.vue
@@ -133,6 +133,8 @@ import {
 } from './Editor.provider.js'
 import { FloatingMenu } from '@tiptap/vue-2'
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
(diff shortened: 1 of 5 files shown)


src/components/Assistant.vue
```diff
@@ -302,7 +305,9 @@ export default {
 			})
 		},
 		async insertResult(task) {
-			this.$editor.commands.insertContent(task.output.output)
+			const isMarkdown = hasMarkdownSyntax(task.output.output) && isValidMarkdown(task.output.output)
```
juliusknorr, line 308: I think we could even just have one function `shouldInterpretAsMarkdown` or similarly named that does both checks, but not a blocker from my side.

juliusknorr (approved):

PR #6701, opened 2024-11-26
ci: Run cypress on ubuntu-latest

juliusknorr:
I noticed that other checks are finished a lot faster and that is due to our selfhosted runners having 24.04 already and no longer having runners for 22.04. Checking fi we can just switch to latest.

- [x] Screenshot diff on print is slightly different seems due to font rendering


mejo- (approved):

mejo-:
Thanks for looking into this @juliusknorr :heart:

Maybe worth being backported to stable30 at least?

PR #6737, opened 2024-12-05
Fix(debug): remove component in beforeDestroy hook

max-nextcloud:
* Use a set so components can only be in there once.
* Debug data can be written even if syncService is undefined.

### 🏁 Checklist

- [x] tested in the browser - this is debug functionality - not a user visible feature.


juliusknorr (approved):
Nice, thanks for pulling that out of Editor.vue as well ;)

PR #6760, opened 2024-12-09
fix(cy): avoid double upload to prevent lock error

max-nextcloud:
`isolateTest` in `beforeEach` already uploads the file.

Uploading it again in the `no actual conflict - just reload` case
sometimes led to 423 - Locked responses.

Use `cy.createTestDir()` instead and upload only once.


max-nextcloud:
https://github.com/nextcloud/text/actions/runs/12233069349/job/34124259796 is an example of a failing job.

mejo- (approved):
Thanks so much for looking into this :heart: 

Shall we backport to stable30 and stable29?

max-nextcloud:
> Shall we backport to stable30 and stable29?

This **is** stable29 :wink: 

I'll trigger backports to main and stable28.

Issue #6732, opened 2024-12-04
Build performance with vite


mejo-:
**Describe the bug**
Since the migration to vite, building the javascript assets takes much longer. This is particularly annoying during development, when using `npm run dev` or `npm run watch`.

This is mostly due to our dependencies mermaid and highlight.js being really heavy and being rebuilt each time by vite.

Some ideas we had so far:

### Build mermaid and highlight.js separately

- Create a separate JS entrypoint that is built using a dedicated vite config.
- In the main `vite.config.ts` mark these modules as external via `build.rollupOptions.external`.
 - The entrypoint registers a function in global namespace (e.g. `window.OCA.Text.modules) to initialize the modules, which loads the assets asynchronously and then returns them.

**Problem:** Building with `vite` clears the target directory upfront. So we cannot build assets into the same target directory using two separate `vite` processes.

### Use `vite serve` to provide the javascript

`vite serve` is really fast.

- Run `NODE_ENV=development npm exec vite -- --mode development serve --host` in Text starts the vite webserver that provides content on port 5173
- The reverse proxy used to serve Nextcloud would need a rule to redirect requests for `/apps/text/js/*` to the vite webserver. On `nginx` from `nextcloud-docker-dev` this can be done the following way:
  ```
  upstream vite {
      server 192.168.15.1:5173;
  }
  server {
      ...
    rewrite ^/index.php/apps/text/js/text-text.mjs /index.php/apps/text/src/main.js;
    rewrite ^/index.php/apps/text/js/text-editors.mjs /index.php/apps/text/src/editor.js;
    rewrite ^/index.php/apps/text/js/text.(.*).mjs /index.php/apps/text/src/$1.js;
      location /index.php/apps/text/src/ {
          proxy_pass http://vite/js/;
          set $upstream_keepalive false;
      }
      ...
  }
  ```

**Problem:** I didn't manage to make vite provide all required files yet.

@juliusknorr @max-nextcloud feel free to adjust and add other ideas.

mejo-:
I made the second approach work using the following nginx config snippet, which certainly can be optimized:

```
    rewrite ^(/index.php)?/apps/text/js/text-text.mjs $1/apps/text/src/main.js;
    rewrite ^(/index.php)?/apps/text/js/text-editors.mjs $1/apps/text/src/editor.js;
    rewrite ^(/index.php)?/apps/text/js/text.(.*).mjs $1/apps/text/src/$2.js;
    location /index.php/apps/text/src {
[...]

PR #6765, opened 2024-12-10
feat: Add support for `vite serve`

mejo-:
Fixes: #6732

### 🚧 TODO

- [x] Add support for a nginx config drop-in to nextcloud-docker-dev

### 🏁 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,26 @@ Follow its development setup and then continue here.
 3. ✅ Enable the app through the app management of your Nextcloud
 4. 🎉 Partytime! Help fix [some issues](https://github.com/nextcloud/text/issues) and [review pull requests](https://github.com/nextcloud/text/pulls) 👍
 
+### Use hot module replacement via `vite serve`
+
+To use the advantages of `vite serve` with hot module replacement (HMR) and not have to recompile the JS assets after each code change, do the following:
+
+1. Configure your webserver to redirect requests to `/apps/text/` to the vite serve server.
+   When using [nextcloud-docker-dev](https://github.com/juliusknorr/nextcloud-docker-dev), add the following snippet to `data/nginx/vhost.d/nextcloud.local` and restart the proxy container. You might have to replace `/apps/text` with e.g. `/apps-extra/text` depending on where the text app resides in your dev setup.
+   ```
+   location /apps/text/ {
+       proxy_pass http://host.docker.internal:5173/apps/text;
+       # fallback to nextcloud server if vite serve doesn't answer
+       error_page 502 = @fallback;
+   }
+   location @fallback {
+       proxy_pass http://nextcloud.local;
+   }
+   ```
+2. Run `npm run serve` to start the vite serve server. If text resides somewhere else than `/apps/text`, run e.g. `BASE=/apps-extra/text npm run serve`.
+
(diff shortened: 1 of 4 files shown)


max-nextcloud:
Thanks a lot for looking into this. :heart:  I'm super happy this works.

I have two suggestions:
* use a non-standart port and add it to the vite config. Then we can hardcode the same port in nextcloud-docker-dev for text. I suggest 7347 😉.
* I will look into making use of `vite/@client` instead of the middleware. It looks like that rewrites the module urls inside the browser and could replace the middleware. See step 2 of https://v5.vite.dev/guide/backend-integration.html

max-nextcloud:
I tried to follow the instructions.
It proved a bit challenging to get the changes to nginx applied. Whenever i restarted the proxy container my changes were gone. So I tried to add the locations and then ran `/etc/init.d/nginx restart`. This printed some errors about not being able to bind but then at least my config would have an effect.

However now I am seeing this in the proxy logs:
```
proxy-1      | nginx.1     | 2024/12/11 09:34:19 [error] 63#63: *9 connect() failed (101: Network is unreachable) while connecting to upstream, client: 192.168.23.1, server: nextcloud.local, request: "GET /apps/text/js/text-init.mjs HTTP/1.1", upstream: "http://192.168.15.1:5173/apps/text/js/text-init.mjs", host: "nextcloud.local"
```
Maybe the ip addresses are different for me?

mejo-:
Thanks for testing it out @max-nextcloud. I added support for a nginx config drop-in to nextcloud-docker-dev in the meantime: https://github.com/juliusknorr/nextcloud-docker-dev/pull/367

This approach is also the reason why I'd keep the default port of `vite serve`: we won't add the configuration for forwarding Text request to `vite serve` to nextcloud-docker-dev anyway.  But in future it's as simple as running `npm run serve` in Text repo and adding the small snippet to `data/nginx/vhost.d/nextcloud.local` in nextcloud-docker-dev and that's it.

> Maybe the ip addresses are different for me?

Yes, that's adjusted now as well. The PR to `nextcloud-docker-dev` adds `host.docker.internal` to the proxy container, so no longer needed to fiddle around with IPs.

mejo-:
Unfortunately the approach to make vite serve text at `/` and rewrite the URLs accordingly didn't work out well. I reverted to the earlier approach and adjusted the docs. This PR is ready to review now.

max-nextcloud (approved):

README.md
```diff
@@ -40,6 +40,26 @@ Follow its development setup and then continue here.
...
 
+### Use hot module replacement via `vite serve`
+
+To use the advantages of `vite serve` with hot module replacement (HMR) and not have to recompile the JS assets after each code change, do the following:
+
+1. Configure your webserver to redirect requests to `/apps/text/` to the vite serve server.
+   When using [nextcloud-docker-dev](https://github.com/juliusknorr/nextcloud-docker-dev), add the following snippet to `data/nginx/vhost.d/nextcloud.local` and restart the proxy container. You might have to replace `/apps/text` with e.g. `/apps-extra/text` depending on where the text app resides in your dev setup.
+   ```
+   location /apps/text/ {
+       proxy_pass http://host.docker.internal:5173/apps/text;
```
juliusknorr, line 51:
```
proxy_pass http://host.docker.internal:5173/apps/text/;
```

I needed this traling slash to make it work 🤷

PR #6767, opened 2024-12-12
chore: Bump composer files with new composer version

juliusknorr:
Make sure the autoloader check passes again on PRs with backend changes.

TO reproduce:

```
cd composer/
rm -r composer
composer install
composer dump


max-nextcloud (approved):

max-nextcloud:
Thanks a lot for looking into this!

PR #6781, opened 2024-12-14
fix(ci): do not install cypress in jest workflow.

max-nextcloud:
It is not required and attempting to install can cause failures
if the cypress server cannot be reached such as here:
https://github.com/nextcloud/text/actions/runs/12330192763/job/34415369232?pr=6780


max-nextcloud:
There's no more jest in later versions - this is against stable29.

PR #6794, opened 2024-12-17
fix(link): Don't throw exception on invalid URL href

mejo-:
When pasting strings with invalid URLs, `new URL()` in `renderHTML()` of the Link extension threw an error, which made the paste parser choke.

We should catch this exception and handle it gracefully to not break HTML parsing completely with invalid URLs.

### 🏁 Checklist

### DIFF
--- a/src/marks/Link.js
+++ b/src/marks/Link.js
@@ -60,10 +60,15 @@ const Link = TipTapLink.extend({
 
 	renderHTML(options) {
 		const { mark } = options
-		const url = new URL(mark.attrs.href, window.location)
-		const href = PROTOCOLS_TO_LINK_TO.includes(url.protocol)
-			? domHref(mark, this.options.relativePath)
-			: '#'
+		let href
+		try {
+			const url = new URL(mark.attrs.href, window.location)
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

src/marks/Link.js
```diff
@@ -60,10 +60,15 @@ const Link = TipTapLink.extend({
...
-			? domHref(mark, this.options.relativePath)
-			: '#'
+		let href
+		try {
+			const url = new URL(mark.attrs.href, window.location)
+			href = PROTOCOLS_TO_LINK_TO.includes(url.protocol)
+				? domHref(mark, this.options.relativePath)
+				: '#'
+		} catch (error) {
+			href = mark.attrs.href
```
max-nextcloud, line 70: I'd also fall back to `'#'` here as we have no way of checking if this is a protocol we actually want to link to.

max-nextcloud:
The Reuse check failure is really odd... looks like there are some licenses we sometimes use and sometimes don't.

src/marks/Link.js
mejo-: Good point, changed now.

Issue #6606, opened 2024-11-06
Block closing/navigating away with unsaved changes


juliusknorr:
After discussion with @mejo- it could be worth to implement a method in [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) to check and block navigating away with unsaved local changes.

Could prevent diverging y.js state and markdown state on the server and also make us aware if there are any saving issues that currently are left unseen.

juliusknorr:
@mejo- Assigned you as you mentioned you'd be up for looking into that

juliusknorr:
Might be an interesting read for checking if there are options to still persist on unload: https://www.speedkit.com/blog/unload-beacon-reliability-benchmarking-strategies-for-minimal-data-loss

juliusknorr:
Another thing I noticed is that when reopening such a document with unsaved changes the autosave would not happen as the frontend does not notice unsaved changes.

Easy steps to reproduce:
- Create a file
- Write something
- Save
- Write something more quickly, wait for a push to happen and force quit your browser tab
- Reopen the file and see that the save indicator is not showing any unsaved changes, but the file content is still the old one

---

This is due to the dirty flag only changing once some more typing is happening https://github.com/nextcloud/text/blob/e67590af683d7d85c9c7c49023aa4cea8abd4e0e/src/components/Editor.vue#L657

Example file has a last saved version of 128 but the first sync adds more steps:

![Image](https://github.com/user-attachments/assets/63c5e59f-9e7c-40d0-afb5-bf47537ff8aa)

I think we used to check the saved version against current version but we no longer pass the current version at load time to the frontend: https://github.com/nextcloud/text/blob/3238541dc6c1587558aa63174a96ea3cc7818cbb/lib/Service/DocumentService.php#L125

Something to think about how we can ensure to still save in such scenarios.

PR #6798, opened 2024-12-18
feat(session): Send save request via `sendBeacon` at `beforeunload`

mejo-:
### 📝 Summary

This will send a final save request on unsaved changes via the browsers native `navigator.sendBeacon()` function when navigating away from the website or the tab/browser is closed.
    
Fixes: #6606

### 🏁 Checklist


src/components/Editor.vue
```diff
@@ -675,7 +675,10 @@ export default {
 				) {
 					this.dirty = state.dirty
 					if (this.dirty) {
+						window.addEventListener('beforeunload', this.saveBeforeUnload)
```
max-nextcloud, line 678: I think this will be called multiple times per editing session. So we would end up with multiple listeners. So instead I propose we add this in the `mounted` hook and drop it in the `deleted` hook.

We only care about the case where the editor is not properly deleted here (directly closing the window for example). If the editor is closed properly we save already.

We will still need to make sure only to save if there are any changes in the undo history and `this.dirty` is true. But we can also check this in `saveBeforeUnload`.

src/components/Editor/Status.vue
```diff
@@ -89,22 +89,19 @@ export default {
 			return this.dirtyStateIndicator ? t('text', 'Saving …') : t('text', 'Saved')
 		},
 		dirtyStateIndicator() {
-			return this.dirty || this.hasUnsavedChanges
+			return this.dirty || this.document.hasUnsavedChanges
```
max-nextcloud, line 92: Does `document` really have that function? I think you defined it for the `connection` below.

max-nextcloud:
I will try it out this week and see if my comments are actually true and fix them if needed.

src/components/Editor/Status.vue
mejo-: Good catch. Changed this to `this.$syncService.hasUnsavedChanges` in all places now.

mejo-:
A good way to test this is to set a debugger breakpoint (e.g. in `saveBeforeUnload()`) and type `window.location('https://example.org')` in the developer console. That's way more handy than closing the tab/browser. Credits for this nice little trick go to @juliusknorr :wink:

src/components/Editor.vue
```diff
@@ -668,13 +668,16 @@ export default {
...
+				// ignore initial loading and other automated changes before first user change
 				if (this.$editor
 					&& (this.$editor.can().undo() || this.$editor.can().redo())
 				) {
 					this.dirty = state.dirty
 					if (this.dirty) {
+						window.addEventListener('beforeunload', this.saveBeforeUnload)
 						this.$syncService.autosave()
+					} else {
+						window.removeEventListener('beforeunload', this.saveBeforeUnload)
```
max-nextcloud, line 680: I think we still need to do this in a `dirty` watcher so it only happens if `dirty` actually changed.

max-nextcloud (approved):
Looks good now.

PR #6815, opened 2024-12-26
chore(CI): Updating reuse.yml workflow from template

nextcloud-command:
Automated update of the reuse.yml workflow from https://github.com/nextcloud/.github


juliusknorr (changes requested):
Needs to be adapted with our changes

juliusknorr (approved):

PR #6855, opened 2025-01-15
enh(ui): smart picker button at the start of line

### LINKED ISSUE #6818: Make smart picker button more prominent
We have the smart picker button in the link menu but it is too hidden, so we should show it more prominent next to the assistant button (below as a tertiary button) in addition.

Image

max-nextcloud:
Fixes #6818.

* Show a `+` button that opens the smart picker.
* Show the button when the cursor is at the end of the line.
* Do not show the button if the smart picker is already open.
* Remove the placeholder text.
* Tab navigation to the smart picker works.

## :movie_camera: Screen recording

https://github.com/user-attachments/assets/48bf53f6-882e-465f-918f-6b83a7f497fd

## TODO

* [x] check if showing the button while in the middle of the line and navigating to the end is an option.
* [x] tab navigation only works if there are no other buttons in the text (from table, preview, link paragraph)
* [ ] add tests

### DIFF
--- a/src/plugins/currentLineMenu.js
+++ b/src/plugins/currentLineMenu.js
@@ -131,7 +131,7 @@ function currentParagraphDecorations(doc, currentParagraph, editor) {
  */
 function decorationForCurrentParagraph(currentParagraph, editor) {
 	return Decoration.widget(
-		currentParagraph.pos + 1,
+		currentParagraph?.pos + 1,
 		menuForCurrentParagraph(editor),
 		{ side: -1 },
 	)
--- a/src/plugins/currentLineMenu.js
+++ b/src/plugins/currentLineMenu.js
@@ -3,8 +3,9 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { Plugin, PluginKey, EditorState } from '@tiptap/pm/state'
+import { EditorState, Plugin, PluginKey, Transaction } from '@tiptap/pm/state'
 import { Decoration, DecorationSet } from '@tiptap/pm/view'
+import { Editor } from '@tiptap/core'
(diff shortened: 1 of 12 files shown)


max-nextcloud:
I'm pondering still showing the button if one is in the middle of the line and moving the cursor to the end of the line if the button is clicked / activated. I imagine it would be nice to say change the current line into a heading with tab navigation from any point in the line.

max-nextcloud:
I updated the screen recording to reflect the latest changes.

juliusknorr:
Pushed a small commit to use NcButton, add an aria label to prevent console warnings and fix the jest tests.

juliusknorr (approved):
Very nice, gave it some testing and the code looks good 👍 

@marcoambrosini Can you also check?

I'd also be fine with merging and doing follow ups on polishing

marcoambrosini (commented):
I think that as soon as there's something in the current line, the plus button should either: 

* disappear (for now)
* Open the smart picker in a **new line**, right after the current one

src/plugins/currentLineMenu.js
```diff
@@ -131,7 +131,7 @@ function currentParagraphDecorations(doc, currentParagraph, editor) {
  */
 function decorationForCurrentParagraph(currentParagraph, editor) {
 	return Decoration.widget(
-		currentParagraph.pos + 1,
+		currentParagraph?.pos + 1,
```
max-nextcloud, line 134: This will still cause an exception if currentParagraph is undefined, right? I'd expect `undefined + 1` to raise a type error. I'll make sure this never gets called without a paragraph instead.

max-nextcloud:
merged as is and created follow up Issue in #6871

Issue #6817, opened 2025-01-02
Suggestions on smart picking


juliusknorr:
We should provider users with more examples on what they can insert into a text document using the smart picker. 

When discussing this with Jan, he mentioned squares as suggestions on what is possible to insert (like tiles in dashboard customization screen, with icons), and like Notion "Get started with". This will be a quick way to insert elements and to advertise the abilities of the smart picker more.

List of suggestions to start with:
- Add image
- Link to file or folder
- Insert table
- Smart picker

cc @jancborchardt 

## Design reference

<img width="886" alt="Image" src="https://github.com/user-attachments/assets/04a6efda-6948-4c33-ba2f-c045c3ebb18e" />

<img width="955" alt="Image" src="https://github.com/user-attachments/assets/f1ab9611-4a90-492b-9a92-af90c32516db" />

JuliaKirschenheuter:
Hi @marcoambrosini,

I have one question regarding:
> Add image

What we have now:
![Image](https://github.com/user-attachments/assets/974e44fc-6ed6-4ca0-af9a-ae4ceb789bde)

Which type should i implement as button inside of the editor?

Another question regarding "widgets". I've asked Jan and he confirmed would be better to implement buttons this way:

![Image](https://github.com/user-attachments/assets/03ce47d2-6c46-45db-8c24-e83bc9b5464d)

juliusknorr:
Discussion from the team call:

- We hide this as soon as there is some content on the page
- We call it "Upload image or file" and use the os file picker (as "Upload form computer" from the menu)
- We start with 4 buttons only
- To clarify with @jancborchardt 
  - Do we actually need the 3-dots button as we would add it as a plus button with https://github.com/nextcloud/text/issues/6818, we can also duplicate it here, but let's start the implementation without.

jancborchardt:
@juliusknorr of course it would be nice to have the "More smart picker" 3-dot button as well at some point, but yes we can first start with some essentials. :)

JuliaKirschenheuter:
Hi @marcoambrosini 

> We call it "Upload image or file" and use the os file picker (as "Upload form computer" from the menu)

Ok, this one "upload from computer" is implemented. This is actually the same function as "Link to file or folder", we don't have a function only for "add Image". What should i place instead / or replace it with some other button?

JuliaKirschenheuter:
Second thing: we have a smart picker dialog, which looks like this:

[...]

PR #6856, opened 2025-01-15
fix(files): create suggestions bar

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/6817

#### 🖼️ Screenshots

🏚️ Before 

Screenshot from 2025-01-30 14-46-31

🏡 After

Screenshot from 2025-01-30 14-46-12
Peek 2025-01-30 14-43

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -51,6 +51,7 @@
 				</template>
 				<ContentContainer v-show="contentLoaded"
 					ref="contentWrapper" />
+				<SuggestionsBar v-if="isRichEditor && contentLoaded" />
 			</MainContainer>
 			<Reader v-if="isResolvingConflict"
 				:content="syncError.data.outsideChange"
@@ -125,6 +126,7 @@ import Translate from './Modal/Translate.vue'
 import CollisionResolveDialog from './CollisionResolveDialog.vue'
 import { generateRemoteUrl } from '@nextcloud/router'
 import { fetchNode } from '../services/WebdavClient.ts'
+import SuggestionsBar from './SuggestionsBar.vue'
 
 export default {
 	name: 'Editor',
@@ -141,6 +143,7 @@ export default {
 		Status,
 		Assistant,
 		Translate,
+		SuggestionsBar,
 	},
 	mixins: [
 		isMobile,
--- a/src/components/Menu/ActionInsertLink.vue
+++ b/src/components/Menu/ActionInsertLink.vue
@@ -66,7 +66,6 @@
 <script>
 import { NcActions, NcActionButton, NcActionInput } from '@nextcloud/vue'
 import { getLinkWithPicker } from '@nextcloud/vue/dist/Components/NcRichText.js'
-import { FilePickerType, getFilePickerBuilder } from '@nextcloud/dialogs'
 import { generateUrl } from '@nextcloud/router'
 import { loadState } from '@nextcloud/initial-state'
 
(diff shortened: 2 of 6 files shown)


src/components/Editor.vue
```diff
@@ -612,6 +616,7 @@ export default {
 			this.emit('update:content', {
 				markdown: proseMirrorMarkdown,
 			})
+			this.isEmptyContent = editor.state.doc.nodeSize <= 4
```
juliusknorr, line 619: What is the background for that fixed number of 4? I'd assume we can also check the text content length directly?

src/components/Editor.vue
```diff
@@ -58,6 +58,7 @@
 				</template>
 				<ContentContainer v-show="contentLoaded"
 					ref="contentWrapper" />
+				<SuggestionsBar v-if="isEmptyContent && contentLoaded" />
```
juliusknorr, line 61: Should only be shown if a markdown file is edited, not for plain text files (e.g. `.txt`)

JuliaKirschenheuter: from the call i understood that SuggestionsBar should be shown if markdown file is not edited (empty). What should i use instead?

src/components/Editor.vue
JuliaKirschenheuter: text content length is checked by the count of `nodeSize` from `editor.state.doc`, checked with Max

src/components/Editor.vue
juliusknorr: If you create a `.txt` file or upload any other code file (like `.js` or `.php`) and open it we will not have any rich text editing, so the options there do not make sense. There is also no menu bar. Let's just not show any suggestions in this case.

src/components/Editor.vue
juliusknorr: > text content length is checked by the count of nodeSize from editor.state.doc

Yes, that is what the code does. I would still be curious about the reasoning, I assume there is one ...

JuliaKirschenheuter:
@marcoambrosini could you please have a look into current state? Thank you!

juliusknorr:
@JuliaKirschenheuter No need to commit the bundle js files in this repo, they get updated once merged.

juliusknorr:
@marcoambrosini Can you have a look UI wise? The spacing between the first line and the suggesions looks a bit off to me

@JuliaKirschenheuter Tested and functionality works as expected. Please check the remaining CI failures.

JuliaKirschenheuter:
Hi @marcoambrosini, could you please have a look?

marcoambrosini (commented):
A few points from my side: 
1- The buttons group should be left aligned  
2- This buttons flicker when opening a file that already have text
3- I think that the smart smart picker button should open the smart picker inline and not in the dialog
4- I think this buttons should be secondary
5- Possibly they should be directly below the first line, not hanging in the middle of the page

Ref
Screenshot 2025-03-20 at 14 34 38

juliusknorr:
@JuliaKirschenheuter Best also leave a comment on what you have done, otherwise others cannot follow that

As noted via chat 1,4,5 are addressed.

One cypress runner had a temporary issue before starting cypress, restarted.

> 2- This buttons flicker when opening a file that already have text
> 3- I think that the smart smart picker button should open the smart picker inline and not in the dialog

Would be fine as a follow up for me, but @JuliaKirschenheuter feel free to also push that here still.

juliusknorr:
>  3- I think that the smart smart picker button should open the smart picker inline and not in the dialog

Thinking about this one a bit more again, it might be arguable if we want that button at all, as we have the plus button on the left doing the same.

juliusknorr (approved):
👍

JuliaKirschenheuter:
> Would be fine as a follow up for me,

@marcoambrosini could i merge this PR and take points 2 and 3 (?) (it might be arguable if we want that button at all, as we have the plus button on the left doing the same) as follow-up?

Issue #882, opened 2020-06-17
RTL support


themedleb:
Solution: Please add `dir="auto"` (In the element that has the class "ProseMirror") so the text will be automatically LTR/RTL depending on the typed text.

themedleb:
Or better add `dir="auto"` to every tag (<p> <li> <h1> <h2> ...), so every tag will be aligned based on the content it has.
That way it will be perfect!

juliushaertl:
Ideally that would be something solved upstream in https://github.com/scrumpy/tiptap/issues/116 but I guess we can also add custom nodes for that as suggested there.

themedleb:
Yup, this better be implemented in the core project.
Unfortunately I can't contribute directly, since this is beyond my knowledge, all I know is HTML, CSS and a little bit of Javascript/jQuery and PHP.

themedleb:
My temporary workaround for this is to install "JSLoader" app, then go to Settings => "JavaScript loader" (Under "Administration" section) then add/paste this code:

```
setInterval(function addDirAutoAttribute() {
  var theID = document.getElementById("editor-container");
  if(theID){
    var theElement = document.getElementsByClassName("ProseMirror");
    for (var i=0; i < theElement.length; i++) {
      if (!theElement[i].hasAttribute("dir")) {
        theElement[i].setAttribute("dir", "auto");
      }
    }
  }
}, 1000);
window.onload = addDirAutoAttribute;
```
Then click "Save".

Not ideal, not efficient, ... but at least it works.

TheMadKow:
Also voting to add RTL support

juliushaertl:
Interesting pull request on tiptap https://github.com/ueberdosis/tiptap/pull/2963

amirhhashemi:
I wrote a extention that automatically adds `dir="ltr|rtl"` to nodes. It's more flexible than `dir="auto"`. It may be helpful:

https://github.com/amirhhashemi/tiptap-text-direction

selimb86:
But how to use this with NextCloud ?

juliushaertl:
Notes, Collectives and Deck use the text editor so they would directly benefit from implementing it here. Other apps would need to be handled separately.

See also https://github.com/nextcloud/server/issues/31420 for general RTL support in Nextcloud.

juliushaertl:
Good reference on how to adapt the UI bits is https://github.com/nextcloud/forms/pull/1654

haikyuu:
[...]

PR #6874, opened 2025-01-23
fix(files): add `tiptap-text-direction` extension to support RTL

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/882

Now it is possible to detect RTL writing style for languages, which requires RTL. Using https://github.com/amirhhashemi/tiptap-text-direction extension. This version works for:

- Paragraph
- Heading
- Lists

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2025-01-23 16-09-02 | Screenshot from 2025-01-23 16-07-29

### 🏁 Checklist

### DIFF
--- a/src/extensions/RichText.js
+++ b/src/extensions/RichText.js
@@ -41,6 +41,7 @@ import TaskItem from './../nodes/TaskItem.js'
 import TaskList from './../nodes/TaskList.js'
 import Text from '@tiptap/extension-text'
 import TrailingNode from './../nodes/TrailingNode.js'
+import TextDirection from 'tiptap-text-direction'
 /* eslint-enable import/no-named-as-default */
 
 import { Strong, Italic, Strike, Link, Underline } from './../marks/index.js'
@@ -111,6 +112,9 @@ export default Extension.create({
 			}),
 			LinkBubble,
 			TrailingNode,
+			TextDirection.configure({
+				types: ['heading', 'paragraph', 'listItem', 'orderedList'],
+			}),
 		]
 		const additionalExtensionNames = this.options.extensions.map(e => e.name)
 		return [
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,6 +78,7 @@
         "proxy-polyfill": "^0.3.2",
         "slug": "^10.0.0",
         "tippy.js": "^6.3.7",
+        "tiptap-text-direction": "^0.3.2",
         "uuid": "^11.0.5",
         "vue": "^2.7.16",
         "vue-click-outside": "^1.1.0",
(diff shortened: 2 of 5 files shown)


hamza221 (approved):
Looks good 🙏🏼

src/extensions/RichText.js
```diff
@@ -43,6 +43,7 @@ import TaskList from './../nodes/TaskList.js'
 import Text from '@tiptap/extension-text'
 import TrailingNode from './../nodes/TrailingNode.js'
 /* eslint-enable import/no-named-as-default */
+import TextDirection from 'tiptap-text-direction'
```
max-nextcloud, line 46: Ah... this also triggers the `import/no-named-as-default` eslint rule. So please move it one line up so it's in the `eslint-disable` block above.

Thanks!

max-nextcloud (commented):
The test failures look related as now `dir="ltr"` is include in some of the html output of the unit tests. I think you can just adjust the expectation to match the actual outcome - and we get a test for this extension for free.

Other than that and the linter fix this looks great. Thanks for looking into it!

Would be nice if we could backport it as I think it really fixes a UX bug for rtl and mixed language use.

juliusknorr (approved):
Change looks sane, @max-nextcloud could you also have a quick look/approval before merging?

max-nextcloud (approved):
Code changes look good. Thanks for taking care of this.
:rocket:

max-nextcloud:
Not sure what happened with the cypress tests. Restarted them.

max-nextcloud:
This cypress test fails fairly reliably:

Bildschirmfoto vom 2025-02-12 19-40-13

Looks like the extension might change the document to inserte the `ltr` markers and that counts as an edit that the editor allows to undo.

JuliaKirschenheuter:
> Looks like the extension might change the document to inserte the `ltr` markers and that counts as an edit that the editor allows to undo.

Thank you @max-nextcloud a lot! Does it mean, i should specify tests more clearly?

JuliaKirschenheuter:
yep, able to reproduce a problem ;((

JuliaKirschenheuter:
> yep, able to reproduce a problem ;((

should be solved now after https://github.com/amirhhashemi/tiptap-text-direction/pull/19

max-nextcloud:
> should be solved now after [amirhhashemi/tiptap-text-direction#19](https://github.com/amirhhashemi/tiptap-text-direction/pull/19)

Funny - does not seem like that has been released.

PR #6878, opened 2025-01-28
fix: attachment visibility for authenticated users via shared links

AIlkiv:
### 📝 Summary

Bug description and steps to reproduce:

admin: Created a Collective.
admin: Added a page with attached images.
admin: Enabled "access via link."
admin: Shared the link with Alice.
alice: Logged into Nextcloud.
alice: Opens the link.
Result: The Collective opens, but the images return a 403 error code.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before| after

### 🏁 Checklist

### DIFF
--- a/lib/Middleware/SessionMiddleware.php
+++ b/lib/Middleware/SessionMiddleware.php
@@ -123,12 +123,14 @@ class SessionMiddleware extends Middleware {
 	private function assertUserOrShareToken(ISessionAwareController $controller): void {
 		$documentId = (int)$this->request->getParam('documentId');
 		if (null !== $userId = $this->userSession->getUser()?->getUID()) {
-			// Check if user has access to document
-			if ($this->rootFolder->getUserFolder($userId)->getFirstNodeById($documentId) === null) {
-				throw new InvalidSessionException();
+			if (count($this->rootFolder->getUserFolder($userId)->getById($documentId)) !== 0) {
+				$controller->setUserId($userId);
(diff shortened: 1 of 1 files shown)


lib/Middleware/SessionMiddleware.php
```diff
@@ -123,12 +123,14 @@ private function assertDocumentSession(ISessionAwareController $controller): voi
 	private function assertUserOrShareToken(ISessionAwareController $controller): void {
 		$documentId = (int)$this->request->getParam('documentId');
 		if (null !== $userId = $this->userSession->getUser()?->getUID()) {
-			// Check if user has access to document
-			if ($this->rootFolder->getUserFolder($userId)->getFirstNodeById($documentId) === null) {
-				throw new InvalidSessionException();
+			if (count($this->rootFolder->getUserFolder($userId)->getById($documentId)) !== 0) {
```
juliusknorr, line 126: Any reason to move away from `getFirstNodeById`? It would be the more performant approach and I'd prefer to keep that.

juliusknorr (commented):
Thanks for your contribution. I left a small question inline, but generally the fix seems good :+1:

lib/Middleware/SessionMiddleware.php
AIlkiv: fixed

AIlkiv:
The failed tests seem unrelated to my changes. Please rerun tests.

PR #6898, opened 2025-02-12
fix/use e2e test server

max-nextcloud:
Superseeds #6813.

- **chore(deps): use @nextcloud/e2e-test-server**
- **Revert "ci: Hotfix to make login work as long as we are on older cypress version"**
- **chore(deps): use typedoc 0.27**


juliusknorr (approved):
Nice one

PR #6919, opened 2025-02-27
ci: Add codecov config

juliusknorr:
Should prevent red CI on base branches


max-nextcloud (approved):

juliusknorr:
Checked the main branch and workted

PR #6923, opened 2025-02-27
feat: Implement event so collectives can inject custom notification messages

juliusknorr:
Base pr for https://github.com/nextcloud/collectives/pull/1675 to allow the collectives app to set their own notification details for mentions within a collective.


max-nextcloud (approved):
code looks good. Makes sense to have this. did not test it though.

mejo- (approved):
Nice, works well! I guess the easiest would be to test this inside Collectives tbh.

PR #6933, opened 2025-03-06
feat: Add experimental option to allow typing during connectivity issues or being offline

juliusknorr:
Something to have merged so we can experiment on daily

### 🏁 Checklist

### DIFF
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -46,7 +46,10 @@ class ConfigService {
 
 	public function isNotifyPushSyncEnabled(): bool {
 		return $this->appConfig->getValueBool(Application::APP_NAME, 'notify_push');
+	}
 
+	public function isExperimentalOfflineTypingEnabled(): bool {
+		return $this->appConfig->getValueBool(Application::APP_NAME, 'experimental_offline_typing');
 	}
 
 	public function isFullWidthEditor(?string $userId): bool {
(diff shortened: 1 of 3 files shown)


juliusknorr:
Should be good to go to main actually :)

lib/Service/ConfigService.php
```diff
@@ -46,7 +46,10 @@ public function isRichWorkspaceEnabledForUser(?string $userId): bool {
 
 	public function isNotifyPushSyncEnabled(): bool {
 		return $this->appConfig->getValueBool(Application::APP_NAME, 'notify_push');
+	}
 
+	public function isExperimentalOffilineTypingEnabled(): bool {
```
elzody, line 51: I think there is a small typo. `Offiline` -> `Offline`.

The method is called correctly elsewhere though maybe it's just because of autocomplete :)

juliusknorr: Thanks :)

max-nextcloud:
@juliusknorr  I think you will also want to check for the flag here: https://github.com/nextcloud/text/blob/0182052c25ae96359577ca39c44ffac263dda386/src/components/Editor.vue#L364-L366

juliusknorr:
@max-nextcloud I adjusted that. Do you think it makes sense to have that given that you focus more on this the next time? Or do you want to approach this differently? Could it still be useful for testing?

I'd be fine to either merge or close

PR #6937, opened 2025-03-10
fix: Avoid pushing empty messages

juliusknorr:
Found in sentry 
Fixes a regular js console error when using notify_push where the awareness message may be pushed as empty string.

We avoid this filtering on the backend before we emit the messages over notify_push, but @max-nextcloud do you happen to know if seeing those empty awareness messages being pushed is expected?

Screenshot 2025-03-10 at 12 58 08
Screenshot 2025-03-10 at 12 58 43

Without this PR parsing the empty data fails in https://github.com/nextcloud/text/blob/ed577363f0b20e745cfe4602a18787a90dd05512/src/services/y-websocket.js#L167

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -208,7 +208,7 @@ class ApiService {
 				'message' => 'text_steps',
 				'body' => [
 					'documentId' => $document->getId(),
-					'steps' => $steps,
+					'steps' => array_values(array_filter($steps)),
 				],
 			]);
 		}


max-nextcloud (approved):

max-nextcloud:
Good catch!!

I don't know if these are expected off the top of my head. I think we still store the awareness messages in the session data. That might cause an empty string if no awareness has ever been send from the users but the session still exists.

Would be good to clean up the awareness handling. I'll create a follow up issue.

PR #6947, opened 2025-03-13
update(workflow): block-merge-freeze to latest

max-nextcloud:
Only block merge for stable branches during freeze.


max-nextcloud:
> Should be upstreamed to the .github repo

That's where it's coming from :wink:

PR #6965, opened 2025-03-17
Fix flaky cypress tests

juliusknorr:
## runner 1 Assistant - Open translate dialog

Otherwise this test is flaky as it may not focus fast enough before entering text, also clicking my interfere with the autofocus of the language selection. Certainly the test could be improved but this is a quick fix to get back to green ci

## runner 7 The session Api - signals closing connection

The issue is that the close api call was not awaited properly. This lead to situations where the push was called too early. Can be tested locally with adding a sleep on the backend in https://github.com/nextcloud/text/blob/tests/cypress-green/lib/Service/ApiService.php#L161

### DIFF
--- a/cypress/e2e/Assistant.spec.js
+++ b/cypress/e2e/Assistant.spec.js
@@ -101,18 +101,16 @@ describe('Assistant', () => {
 			.should('be.visible')
 			.click()
 
+		// eslint-disable-next-line cypress/no-unnecessary-waiting
 		cy.wait(1000)
 
 		cy.get('[data-cy="translate-input"]')
 			.should('be.visible')
 			.focus()
 
-		cy.wait(1000)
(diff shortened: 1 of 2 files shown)


cypress/e2e/Assistant.spec.js
```diff
@@ -101,13 +101,16 @@ describe('Assistant', () => {
 			.should('be.visible')
 			.click()
 
+		// eslint-disable-next-line cypress/no-unnecessary-waiting
+		cy.wait(1000)
```
juliusknorr, line 105: I did not figure out a way to make this wort otherwise. For some reason running on CI otherwise focusses the NcSelect 🤷

AndyScherzinger (approved):
🐘

Issue #6936, opened 2025-03-06
Place menubar at the bottom on mobile


juliusknorr:
This was raised by @jancborchardt as a quick enhancement for making the menu bar easier to access. We could move it to the bottom of the screen on mobile devices. Similar to other notes apps, iOS Notes. This came up when reviewing assistant usage on mobile

WDYT @max-nextcloud @marcoambrosini

marcoambrosini:
I think it's a good idea

PR #6983, opened 2025-03-21
fix(files): Place menubar at the bottom on mobile

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/6936

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2025-03-21 17-41-19 | Screenshot from 2025-03-21 17-36-55

### 🏁 Checklist


marcoambrosini (commented):
Good design wise. I think we could probably omit the emoji picker since mobile keyboards always come with one

marcoambrosini (approved):
^ that was non blocking so approving

juliusknorr:
> Good design wise. I think we could probably omit the emoji picker since mobile keyboards always come with one

Good point, we should file that as a follow up.

juliusknorr:
Restarted cypress

src/components/Editor/MediaHandler.vue
```diff
@@ -6,7 +6,7 @@
 <template>
 	<div class="editor editor-media-handler"
 		data-text-el="editor-media-handler"
-		:class="{ draggedOver }"
+		:class="{ draggedOver, 'editor-media-handler-mobile': isMobile }"
```
juliusknorr, line 9: Nitpick for the naming to fit https://getbem.com/naming/ Nothing we enforce but is used in a couple of places

```suggestion
		:class="{ draggedOver, 'editor--mobile': isMobile }"
```

juliusknorr (approved):

src/components/Editor/MediaHandler.vue
juliusknorr: To then also adjust on the css part

Issue #7022, opened 2025-03-24
Hide emoji button from the menubar on mobile


JuliaKirschenheuter:
Reference: https://github.com/nextcloud/text/pull/6983#pullrequestreview-2709878529

PR #7025, opened 2025-03-25
Hide emoji button from the menubar on mobile

JuliaKirschenheuter:
Add RegEx to check if user uses mobile device and add Emoji Picker button inside a toolbar only in this case

### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/7022

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2025-03-25 15-10-28 | Screenshot from 2025-03-25 15-08-30

### DIFF
--- a/src/components/Menu/entries.js
+++ b/src/components/Menu/entries.js
@@ -70,6 +70,8 @@ export const ReadOnlyDoneEntries = [{
 	click: ({ $readOnlyActions }) => $readOnlyActions.toggle(),
 }]
 
+const isMobileDevice = /Android|iPhone|iPad|iPod/i.test(navigator.userAgent)
+
 export const MenuEntries = [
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

src/components/Menu/entries.js
```diff
@@ -70,6 +70,9 @@ export const ReadOnlyDoneEntries = [{
 	click: ({ $readOnlyActions }) => $readOnlyActions.toggle(),
 }]
 
+// RegEx to check a real mobile device
```
juliusknorr, line 73:
```suggestion
```

Should be obvious what this does without a code comment

marcoambrosini (approved):

Issue #7024, opened 2025-03-25
Set a text on buttons within suggestions bar on mobile viewport


JuliaKirschenheuter:
Follow-up for https://github.com/nextcloud/text/issues/6817

PR #7030, opened 2025-03-26
fix(files): Add transition to buttons and take buttons away from mobile view

JuliaKirschenheuter:
Signed-off-by: julia.kirschenheuter <julia.kirschenheuter@nextcloud.com>

### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/7024

Add transition to buttons and take buttons away from mobile view

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2025-03-26 16-17-15 | Screenshot from 2025-03-26 16-08-44

### 🏁 Checklist

### DIFF
--- a/src/components/Menu/entries.js
+++ b/src/components/Menu/entries.js
@@ -42,6 +42,7 @@ import ActionAttachmentUpload from './ActionAttachmentUpload.vue'
 import ActionInsertLink from './ActionInsertLink.vue'
 
 import { MODIFIERS } from './keys.js'
+import { isMobileDevice } from '../../helpers/isMobileDevice.js'
 
 export const OutlineEntries = [{
 	key: 'outline',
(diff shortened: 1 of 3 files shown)


marcoambrosini:
@JuliaKirschenheuter I would not remove this button actually because the + button should only be visible when the cursor is active in a particular block and the mouse is hovering it

marcoambrosini (changes requested):
There's a problem with the layout braking on mobile:

https://github.com/user-attachments/assets/54e81866-2aee-417c-96b8-bd5b5184c356

marcoambrosini:
I think on mobile we should have something like this

Screenshot 2025-03-27 at 10 45 56

JuliaKirschenheuter:
@marcoambrosini done as discussed ;)

For testing with switching between desktop and mobile devices: please update a page each time

Issue #7152, opened 2025-03-16
Ctrl-K should add a link to highlighted text


atrauzzi:
It's common in markdown UIs for ctrl-k to be the shortcut for adding a link to text.

(you can even try it here in github!)

It would be nice if Notes did the same.

PR #7188, opened 2025-05-06
feat(link): Toggle link on selection with Mod-K shortcut

mejo-:
Acts only upon selection as inserting a link with empty title creates an invisible link, resulting in bad user experience in the WYSIWYG editor.

Fixes: #7152

### 🏁 Checklist


luka-nextcloud:
I've tested and it works well 👍

PR #7197, opened 2025-05-06
fix(files): Add smooth animation via clicking on heading

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/6974

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/TableOfContents.vue
+++ b/src/components/Editor/TableOfContents.vue
@@ -14,7 +14,7 @@
 					[`editor--toc__item--${heading.level}`]: true,
 					[`editor--toc__item--previous-${heading.previous}`]: heading.previous > 0,
 				}">
-				<a :href="`#${heading.id}`" @click.prevent="goto(heading)">
+				<a :href="`#${heading.id}`" class="editor--toc__item-link" @click.prevent="goto(heading)">
 					{{ heading.text }}
 				</a>
 			</li>
@@ -47,10 +47,10 @@ export default {
 	},
 	methods: {
 		goto(heading) {
-			document.getElementById(heading.id).scrollIntoView()
-
+			const element = this.$root.$el.querySelector(`#${heading.id}`)
+			element.scrollIntoView({ block: 'start', behavior: 'smooth' })
 			this.$nextTick(() => {
-				window.location.hash = heading.id
+				window.history.replaceState(window.history.state, '', `#${heading.id}`)
 			})
 		},
 		updateHeadings() {
@@ -100,6 +100,10 @@ export default {
 
 		a:hover {
 			color: var(--color-primary-element-hover);
+		}
+
+		&-link {
+			scroll-margin-top: calc(var(--default-clickable-area) + 4 * var(--default-grid-baseline));
 		}
 
 		&--1 {
--- a/src/plugins/headingAnchor.js
+++ b/src/plugins/headingAnchor.js
@@ -144,6 +144,7 @@ function anchorForHeading(heading) {
  */
 function handleClick(event) {
 	event.stopPropagation()
-	event.target.scrollIntoView()
-	window.location.hash = event.target.getAttribute('href')
+	event.preventDefault()
+	event.target.scrollIntoView({ block: 'start', behavior: 'smooth' })
+	window.history.replaceState({}, '', event.target.getAttribute('href'))
 }
--- a/src/css/prosemirror.scss
+++ b/src/css/prosemirror.scss
@@ -116,6 +116,9 @@ div.ProseMirror {
 			transition-duration: 0.15s;
 			transition-property: opacity;
 			transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+			scroll-margin-top: calc(
+				var(--default-clickable-area) + 4 * var(--default-grid-baseline)
+			);
 		}
 
 		&:hover .heading-anchor {


max-nextcloud (approved):
Tested locally and works for me. Code changes also seem fine.

mejo-:
> Mhh, interesting. For me, this PR breaks anchor links. Whenever I click on the # next to a heading, it scrolls back to the beginning of the document.

It doesn't scroll back to the top of the document, but scrolls the clicked heading into the middle of the screen if possible. I guess that's a bit counter-intuitive and I would expect the heading to scrolled to the top of the editor container.

src/components/Editor/TableOfContents.vue
```diff
@@ -47,10 +47,9 @@ export default {
 	},
 	methods: {
 		goto(heading) {
-			document.getElementById(heading.id).scrollIntoView()
-
+			document.getElementById(heading.id).scrollIntoView({ behavior: "smooth", block: "center", inline: "nearest" })
```
mejo-, line 50: I guess this should be `block: start` instead.

src/plugins/headingAnchor.js
```diff
@@ -144,6 +144,7 @@ function anchorForHeading(heading) {
  */
 function handleClick(event) {
 	event.stopPropagation()
-	event.target.scrollIntoView()
-	window.location.hash = event.target.getAttribute('href')
+	event.preventDefault()
+	event.target.scrollIntoView({ behavior: "smooth", block: "center", inline: "nearest" })
```
mejo-, line 148: Same here.

mejo- (changes requested):

src/components/Editor/TableOfContents.vue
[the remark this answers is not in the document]
mejo-: Not directly related, but `document.getElementById()` also doesn't work
1. when `Readme.md` is opened both in viewer and folder description (`getElementById()` will return the element in folder description)
2. in Collectives in edit mode (`getElementId()` will return the element in view mode)

src/components/Editor/TableOfContents.vue
```diff
@@ -47,10 +47,9 @@ export default {
...
 			this.$nextTick(() => {
-				window.location.hash = heading.id
+				window.history.replaceState({}, '', `#${heading.id}`)
```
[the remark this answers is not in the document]
JuliaKirschenheuter: because `window.location.hash` would trigger a scroll behavior: the browser would automatically scroll to the element with the corresponding id, if it exists. In this case we should avoid it, because then it would rewrite desired behaviour and browser won't jump into clicked heading

src/plugins/headingAnchor.js
JuliaKirschenheuter: found another solution with `scroll-margin-top` on the link elements

src/components/Editor/TableOfContents.vue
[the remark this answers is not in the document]
JuliaKirschenheuter: @mejo- was i able to describe my changes?

JuliaKirschenheuter:
@mejo- Hi Jonas,

could you please look into this again? Thank you!

src/components/Editor/TableOfContents.vue
```diff
@@ -47,10 +47,10 @@ export default {
 	},
 	methods: {
 		goto(heading) {
-			document.getElementById(heading.id).scrollIntoView()
-
+			const element = document.getElementById(heading.id)
```
mejo-:
```suggestion
			const contentContainer = this.$parent.$parent.$el
			const element = contentContainer.querySelector(`#${heading.id}`)
```

As mentioned, unfortunately `document.getElementById()` doesn't work if several editors with the ID are loaded. You can reproduce this by opening a `Readme.md` from files app while the same file is also loaded in the folder description editor.

My suggestion instead searches the `#${heading.id}` element from the grandparent of the Vue component, which makes sure that it's always taken from the editor instance we're operating on.

src/components/Editor/TableOfContents.vue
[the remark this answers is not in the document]
mejo-: In my tests, this somehow breaks the browser history. Using the back button no longer works as expected. Does it work for you? :thinking:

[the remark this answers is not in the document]
max-nextcloud: @mejo- where are you testing? Browser history / back button has been broken in files with viewer as far back as i can remember. It would always change the folder "behind" the viewer.

src/components/Editor/TableOfContents.vue
juliusknorr: Good point, does it make sense to use something like `$root.$el` to avoid breaking if we rearrange the components and they would be nested differently?

src/components/Editor/TableOfContents.vue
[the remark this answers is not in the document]
mejo-: True, in Files this has been broken before already. But we also have Collectives.

Before, with `window.location.hash`, a new entry was added to the browser history. Now with `widow.history.replaceState()` the current entry get's replaced. Unfortunately navigating back was already broken before, it just did nothing (and didn't revert the scroll position).

After some further testing and thinking about it again, maybe it's indeed the best to use `replaceState()` and not add a new useless entry to the history for now. Sorry for the extra loop :grimacing:

src/components/Editor/TableOfContents.vue
mejo-:
```suggestion
			const element = this.$root.$el.querySelector(`#${heading.id}`)
```

Indeed, using `$root` is much better. Here's an updated suggestion that worked in my tests.

JuliaKirschenheuter: Thanks Jonas and Julius!

mejo-:
The failing assistant cypress test seems unrelated. It also fails on other PRs.

PR #7200, opened 2025-05-07
fix: adjust table width for better view

AIlkiv:
### 📝 Summary

This fix applies only to the view mode

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before_ | after_

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/css/prosemirror.scss
+++ b/src/css/prosemirror.scss
@@ -418,3 +418,9 @@ div.ProseMirror {
 .tippy-content div {
 	visibility: visible !important;
 }
+
+div.ProseMirror[contenteditable=false] {
+	table {
+		width: 100%;
+	}
+}


mejo- (dismissed):
For the collectives app it's important to have similar layout of edit and view mode, for as less friction as possible when toggling between the two. Therefore I would prefer to keep it as is.

AIlkiv:
@mejo- 
I agree, but on the other hand, when viewing the page, it looks like an incorrect styles.

Perhaps this layout will be a compromise?
demo

mejo-:
What do our designers think? Cc @marcoambrosini

marcoambrosini:
I think ideally we want to make the table full width at all times move this "buttons column" outside. Much like the "+" button and link preview button on the left side. 

When the screen is smaller and these don't fit, I don't mind the horizontal jump when switching modes. What's more unsettling are vertical jumps. But in this case, either way is fine from my pov.

max-nextcloud:
pushed a small fix to make prettier happy.

Issue #6843, opened 2025-01-10
Loading an existing file focusses image description


juliusknorr:
**Describe the bug**
When loading an existing file that contains images the focus jumps to the alt text input

**To Reproduce**
1. Create a new file with some long text
2. Add an image after it
3. Move the cursor to the top
4. Close and reopen the file

**Expected behavior**
The cursor is at the beginning

**Actual behaviour**
The cursor is in the alt text of the image

**Additional context**

This seems like a side effect of https://github.com/nextcloud/text/pull/6609 

I'm unsure what the best way would be to catch only inserting new attachments. @max-nextcloud Any idea on how we could check that? I don't see any hooks for tiptap node or node views that would only trigger on first creation. Maybe we can override the [setImage](https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-image/src/image.ts#L107-L111) command and pass a non-persited flag to focus once? 

cc @JuliaKirschenheuter

MrRinkana:
It does not seem to be only when the file is opened, for me it happens everytime the "Start writing, or try.." Prompt appears (or dissapears) or in other words everytime enter is pressed, or line is changed from an empty. 

Steps to reproduce (nc30.0.7, Firefox):
1. Create a file.md, put some lorem ipsum text such that there is a scrollbar
2. Attach image from files (not link or smartpicker) 
3. Add some more text then another image (prob not necessary but was what I did)
4. Move cursor to top, write some radom letter then save and close file.

**Observed: everytime the file is opened, cursor jumps to the first image added. In addition everytime a row is created (enter key) or the cursor is moved such that the "Start writing, or try.." Text appears or disappears the cursor jumps to the first (first added) image but **only if one is working before** the image. 

Additionally it seems the cursor only jumps to the image that was first added, even if images were added above later. I just realised that it could also be depending on the name/sorting of files in .attachments.

This is inconvinient enough that if #6609 is the cause I would suggest it is reverted until a better solution can be made, the behaviour is inconsistent anyway, for example creating a "callout" does not focus it and the cursor just disappears.. What #6609 fixes is not really much of an inconvenience compared to this bug.

theatischbein:
Thank you sooo much! This issue was really annoying! ❤

PR #6609, opened 2024-11-06
fix(files): set focus on image description right after inserting

JuliaKirschenheuter:
### 📝 Summary

Set focus on image description right after inserting

* Part of: https://github.com/nextcloud/text/issues/6152
After inserting an attachment, focus should go after the the image -> to the image description

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-11-06 17-59-44 | Screenshot from 2024-11-06 17-59-54

### 🏁 Checklist

- [ ] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [ ] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [ ] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required

PR #7201, opened 2025-05-07
fix(image): Only focus image description on newly inserted image

mejo-:
Fixes: #6843 
Fixes: nextcloud/collectives#1740

Fixes a regression from #6609

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/MediaHandler.vue
+++ b/src/components/Editor/MediaHandler.vue
@@ -226,6 +226,9 @@ export default {
 			// Scroll image into view
 			this.$editor.commands.scrollIntoView()
 
+			// Store last inserted attachment src to focus it in ImageView.vue
+			this.$editor.commands.setMeta('insertedAttachmentSrc', { src })
+
 			emit('text:image-node:add', null)
 		},
 	},
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -170,6 +170,7 @@ export default {
 			showImageModal: false,
 			imageIndex: null,
 			isEditable: false,
+			isLastInserted: false,
 			embeddedImageList: [],
 		}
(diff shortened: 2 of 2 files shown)


mejo-:
@max-nextcloud I'm curious what you think about the approach with `setMeta()`.

The challenge is to detect whether the loaded attachment just got inserted or whether it's an existing attachment in the node view `ImageView.vue`. New attachments are uploaded and inserted into the editor content in `MediaHandler.vue`.

I had to find a way to register the newly inserted image from `MediaHandler.vue` and then check in `ImageView.vue` whether the loaded image is the one just inserted. I chose a transaction metadata property as to my knowledge that's kept in the local instance and not synchronised via the collaboration plugin, but I might be wrong here.

max-nextcloud (commented):
I think this change does what it's intended to do.
However I think the underlying problem is a different one.

I think the alt text should be focused directly after inserting the image - even if the image has not been loaded yet. But it seems we only render the alt text once the image has been loaded. We should always render the alt text. If the image fails to load or takes a long time the alt text might tell what is supposed to be displayed.

Also - if i insert an image and the processing takes a long time so I continue typing somewhere else I would not expect the focus to be moved once the image finished loading.

JuliaKirschenheuter (approved):
works, thank you!

emberfiend:
I think this patch introduced this behaviour (video below). Clicking around causes only the isLastInserted ImageView to reload. Importantly, it only happens when you click in between paragraphs...?! I don't understand enough yet about how these parts interact to suggest a fix though :sweat_smile: 

[Screencast from 2025-05-09 09-19-59.webm](https://github.com/user-attachments/assets/1847f3fb-5eae-48f7-8180-1891b72bab65)

Issue #989, opened 2020-08-26
Keymap code cleanup


juliusknorr:
:monocle_face: Minor nitpick: `Keymap` class sounds very generic and behaves generic in that all of it's behaviour is specified through the options during construction. At the same time it now implements special treatment for `ctrl-f`. Seems counter intuitive that by adding a keymap to the editor that handles `mod-s` you also get special treatment for `ctrl-f`.

Maybe put the plugin into a separate class like `BrowserSearch` or so?

_Originally posted by @azul in https://github.com/nextcloud/text/pull/988#issuecomment-680861902_

luka-nextcloud:
It's still valid.

mejo-:
Mhhh, so this is merely about renaming a javascript class? :monocle_face:

I'd suggest to either fix it straight away or close the issue :wink: 

I don't think there's much value in having issues about minor refactoring. Either it will happen along with other work anyway (and this issue will not be memorized by then) or it doesn't seem to be important enough :zany_face:

juliushaertl:
Agreed, I think Keymap still sounds sane as a naming, but we should move the Ctrl+S registration from the Editor.vue file to the actual plugin instead of using the options.

@luka-nextcloud Mind to have a look into that?

PR #7204, opened 2025-05-08
Bring back and refactor keymap, add focus trap to plain text editor

mejo-:
### 📝 Summary

* Bring back our custom Keymap extension
* Move all keymap adjustments into either their respective plugin or the Keymap extension
* Use FocusTrap also in plain text editor
* Resolves: #989

### 🏁 Checklist


max-nextcloud (approved):
Thanks a lot for tackling this. Looks really good.

Now if we could also stop handling ctrl+shift+i... then i could enable the browser dev tools even when focus is on the text.

I'll look into that and create a PR either against this PR or against main if this has already been merged.

Issue #4519, opened 2023-07-13
Only load images on demand


provokateurin:
**Is your feature request related to a problem? Please describe.**
I have a document with more than 300 images. Even on a fast connection it takes very long to load the whole document because all images are loaded before the editor is ready.

**Describe the solution you'd like**
Images should only be loaded when they are visible.

**Describe alternatives you've considered**
Make the editor ready before all images are loaded would also work. I think it's better to not load all images at once to save some network bandwith.

**Additional context**
A bit related to https://github.com/nextcloud/text/issues/4242, but not the same problem.

Issue #4242, opened 2023-06-06
Image loading moves text


provokateurin:
**Describe the bug**
I have a long document with many images.
When I scroll to the bottom it loads all the images, but then the whole text scrolls because the images are bigger than their placeholders. I have to scroll to the bottom again.

**To Reproduce**
1. Long document with many images
2. Open document, scroll to the bottom
3. Wait for the images to load
4. Text moves up

**Expected behavior**
I should still be at the bottom of the document

**Client details:**
 - OS: Linux
 - Browser: Firefox
 - Version: 113
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.7.2

**Operating system:** Alpine Linux

**Web server:** Nginx

**Database:** Postgres

**PHP version:** 8.2

**Nextcloud version:** 26.0.2

</details>

blizzz:
@provokateurin Not sure I see the same behaviour that you see. Could you perhaps do a quick screenrecording?

For me, pictures load sort of instantly, but when opening the editor I am somewhere in the middle, and certainly not where the cursor is – always at the end.

provokateurin:
Sure, my document has about 250 images so they don't load that instantly :D 

> when opening the editor I am somewhere in the middle, and certainly not where the cursor is – always at the end.

Sorry, I don't understand what you mean. Can you explain it a little bit?

blizzz:
>>  when opening the editor I am somewhere in the middle, and certainly not where the cursor is – always at the end.

> Sorry, I don't understand what you mean. Can you explain it a little bit?

When I open my test document with text, the visible area shows some middle part of the document. The cursors however, is placed at the end of the document and as such is not visible. Moving the curse by key stroke scrolls down to the bottom of the document.

provokateurin:
https://github.com/nextcloud/text/assets/26026535/5ec9b42e-e4d0-49c5-9687-eb325229e4bc

I only scroll when my mouse is on the right side, the scrolling up is due to the images loading

juliushaertl:
With files metadata we could reserve space for the image if the size is known https://github.com/nextcloud/photos/issues/2243

PR #7209, opened 2025-05-09
feature: Delay image load until visible

emberfiend:
### 📝 Summary

Hi all, first PR here. I'm trying to keep this small and self-contained. I'm new to both this codebase and Vue, please don't hesitate to give corrections or hints :)

This adds https://github.com/nextcloud/text/issues/4519 @provokateurin via an IntersectionObserver, which loads images only when they intersect the viewport.

@mejo- 's well-timed https://github.com/nextcloud/text/pull/7201 means we don't have to worry about the viewport jumping around at load time, causing unwanted loads.

#### 🖼️ Screenshots

Tricky to screenshot!

### 🚧 TODO

My notes for follow-ups:

- Find a way to do this with margin or even acceleration modelling so images start loading well before they hit the viewport.
- Reusing one IntersectionObserver across multiple components would be faster but potentially less tidy.
- As @juliusknorr said in https://github.com/nextcloud/text/issues/4242, image metadata is the more complete solution to jumpiness in general. This PR makes things worse on slow connections in a way, because scrolling beyond images to look at text below them always means a new (interrupting) wait for the document to stop changing length, instead of one longer wait when the document first loads.

### 🏁 Checklist

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -172,6 +172,7 @@ export default {
 			isEditable: false,
 			isLastInserted: false,
 			embeddedImageList: [],
+			loadIntersectionObserver: null,
 		}
 	},
 	computed: {
@@ -230,7 +231,29 @@ export default {
 				this.isLastInserted = true
 			}
 		})
-		this.loadPreview().catch(this.onImageLoadFailure)
+	},
+	mounted() {
+		this.$nextTick(() => {
+			// nextTick is necessary, intersection detection is slightly unreliable without it
+			const options = {
+				root: null,
+				threshold: 0,
+			}
+			const startImageLoad = (entries, observer) => {
+				if (entries[0].isIntersecting) {
+					observer.disconnect()
+					this.loadPreview().catch(this.onImageLoadFailure)
+				}
+			}
+			this.loadIntersectionObserver = new IntersectionObserver(
+				startImageLoad,
+				options,
+			)
+			this.loadIntersectionObserver.observe(this.$el)
+		})
+	},
+	beforeUnmount() {
+		this.loadIntersectionObserver?.disconnect()
 	},
 	methods: {
 		async loadPreview() {


max-nextcloud (approved):
Welcome @emberfiend and thanks a lot for your contribution.

You are registering the IntersectionObserver in the `mounted` function. Is there any cleanup needed when say the editor is closed and all the nodes disappear?

I noticed one thing when trying this:
When I create a document with a lot of images at the end and no other content. The document is initially scrolled to the end where a LOT of spinning wheels are shown (each 16px high). Then all the corresponding images are loaded. As they now take more space and the browser stays aligned with the topmost one (i assume) it effectively scrolls up.
I don't know how to completely prevent this. But it also seems like the 
` .image__loading` css directive that sets it to '100px' height has no effect anymore.

Would be great if you could fix this as it reduces the effect of the many small nodes that trigger many images to get loaded.

I'll approve right away as these are just two minor remarks.

emberfiend:
Thanks @max-nextcloud for the kind words and guidance :)

Added some cleanup in beforeUnmount.

Your second point is a rabbit-hole haha. So it seems a file with only images (like, not a single text node) behaves differently to one with a single line of text in it. An image-only file cannot be copied (well, it produces an empty copy) and exhibits the behaviour you describe: it's scrolled to the bottom of the document, whatever length it is, so whichever loading spinners are visible load, and scroll ends up aligned with the topmost of those images. I'm struggling to reproduce the "them being 16px high" thing though.

In contrast, adding one line of text causes the file to be scrolled to the top when opened, with much less janky behaviour following. 

So (apart from the 16px thing) I think the real question is why image-only files behave differently. My React-y gut has some guesses but I need to learn more about the whole editor structure before I can be more useful I think.

emberfiend:
Tangentially, can anyone explain why DCO is unhappy again? I was pretty sure I did my last commit in the style it liked.

juliusknorr:
Thanks for the update. The DCO is the remaining one to fix before we can merge this.

https://github.com/nextcloud/text/runs/42043204677 shows that the email does not match:

> Commit sha: ac766cf, Author: emberfiend, Committer: emberfiend; Expected "emberfiend abackhouse@proton.me", but got "Andrew Backhouse hi@andrewbackhouse.com".

Maybe you can wrap this up and squash your commits into one. Good to get merged from my perspective otherwise as well 👍

emberfiend:
@juliusknorr Thanks, I missed that it was logging errors. Squashed.

juliusknorr (approved):

Issue #7179, opened 2025-05-05
Respect default extension setting in direct editing


max-nextcloud:
**Describe the bug**
Given the default extension for text is set to 'txt'
when creating a text document in the ios client
it creates a .md file

**To Reproduce**
1. `./occ config:app:set text default_file_extension --value 'txt'`
2. Create a text document from the ios client
3. See .md extension

**Expected behavior**
The ios client should respect the configured default.

## Current Api response

/ocs/v2.php/apps/files/api/v1/directEditing

```json
{
  "textdocument": {
    "id": "textdocument",
    "editor": "text",
    "name": "text document",
    "extension": "md",
    "templates": false,
    "mimetype": "text/markdown"
  }
}
```

PR #7229, opened 2025-05-13
Respect default extension setting in direct editing

luka-nextcloud:
### 📝 Summary

* Resolves: #7179

### 🏁 Checklist

### DIFF
--- a/tests/unit/DirectEditing/TextDocumentCreatorTest.php
+++ b/tests/unit/DirectEditing/TextDocumentCreatorTest.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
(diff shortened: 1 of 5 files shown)


luka-nextcloud:
> Code looks good. Did you test it? Looks like it would be fairly easy to write a unit test for the TextDocumentCreator. Then we'd have some automated testing.

Yes, I tested it. Just added unit test for it.

tests/unit/DirectEditing/TextDocumentCreatorTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
```
juliusknorr, line 1: Needs an spdx license header

juliusknorr (approved):
php-cs is unhappy and SPDX header is missing on the test file, but otherwise good to merge 👍

PR #7237, opened 2025-05-19
fix(attachments): Adjust path to attachments folder when copying a node

mejo-:
### 🏁 Checklist

### DIFF
--- a/lib/Listeners/NodeCopiedListener.php
+++ b/lib/Listeners/NodeCopiedListener.php
@@ -13,6 +13,7 @@ use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
 use OCP\Files\Events\Node\NodeCopiedEvent;
 use OCP\Files\File;
+use OCP\Lock\ILockingProvider;
 
 /**
  * @template-implements IEventListener<Event|NodeCopiedEvent>
@@ -36,6 +37,9 @@ class NodeCopiedListener implements IEventListener {
 			&& $target->getMimeType() === 'text/markdown'
 		) {
 			$this->attachmentService->copyAttachments($source, $target);
+			$target->unlock(ILockingProvider::LOCK_SHARED);
+			AttachmentService::replaceAttachmentFolderId($source, $target);
+			$target->lock(ILockingProvider::LOCK_SHARED);
 		}
 	}
 }
--- a/lib/Service/AttachmentService.php
+++ b/lib/Service/AttachmentService.php
@@ -715,4 +715,24 @@ class AttachmentService {
 			}
 		}
 	}
+
+	public static function replaceAttachmentFolderId(File $source, File $target): void {
+		$sourceId = $source->getId();
+		$targetId = $target->getId();
+		$patterns = [
+			// Replace `[title](.attachments.1/file.png)` with `[title](attachments.2/file.png)`
+			// '/(\[[^]]+\]\(\s*\<?\.attachments\.)' . $sourceId . '(\/[^)]+\>?\s*\))/',
+			'/(\[(?:\\\]|[^]])+\]\(\s*\<?\.attachments\.)' . $sourceId . '(\/[^)]+\>?\s*\))/',
+			// Replace `[ref]: .attachments.1/file.png` with `[ref]: .attachments.2/file.png`
+			'/(\[(?:\\\]|[^]])+\]:\s+.attachments\.)' . $sourceId . '(\/[^\s]+)/',
+		];
+		$replacements = [
+			'${1}' . $targetId . '${2}',
+			'${1}' . $targetId . '${2}',
+		];
+		$content = preg_replace($patterns, $replacements, $target->getContent());
+		if ($content !== null) {
+			$target->putContent($content);
+		}
+	}
 }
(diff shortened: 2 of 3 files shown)


lib/Listeners/NodeCopiedListener.php
```diff
@@ -36,6 +37,9 @@ public function handle(Event $event): void {
 			&& $target->getMimeType() === 'text/markdown'
 		) {
 			$this->attachmentService->copyAttachments($source, $target);
+			$target->unlock(ILockingProvider::LOCK_SHARED);
+			AttachmentService::replaceAttachmentFolderId($source, $target);
+			$target->lock(ILockingProvider::LOCK_SHARED);
 		}
```
max-nextcloud, line 43: @mejo- This looks odd to me. First you unlock and then you lock. Am I missing something?

max-nextcloud: Also... maybe you can push the locking down into the `AttachmentService`? Seems like something one would always want when running `replaceAttachmentFolderId`.

lib/Service/AttachmentService.php
```diff
@@ -715,4 +715,12 @@ public function copyAttachments(File $source, File $target): void {
...
 		}
 	}
+
+	public static function replaceAttachmentFolderId(File $source, File $target): void {
+		$sourceId = $source->getId();
+		$targetId = $target->getId();
+		// Replace `](.attachments.<sourceId>/` with `](attachments.<targetId>/` in target node
+		$content = str_replace('](.attachments.' . $sourceId . '/', '](.attachments.' . $targetId . '/', $target->getContent());
+		$target->putContent($content);
+	}
```
max-nextcloud, line 725: :warning: nitpicking ahead :warning: :wink:

Getting this right without a markdown parser is quite a bit of a challenge.
Here's a few corner cases the current PR does not take into account.
Not sure which of them we need to handle.

## Not replaced but maybe should

### [Link destination](https://spec.commonmark.org/0.31.2/#link-destination) with pointy brackets
```markdown
[my attachment](<.attachments.1/image.png>)
```

### [Link destination with spaces](https://spec.commonmark.org/dingus/?text=%5Bmy%20attachment%5D(%20%20.attachments.1%2Fimage.png))
```markdown
[my attachment](  .attachments.1/image.png)
```

### [Reference links](https://spec.commonmark.org/0.31.2/#reference-link)
```markdown
[foo][bar]
[bar]: .attachments.1/image.png "title"
```


## Replaced but maybe should not

### incomplete and thus broken links
```markdown
](.attachments.1/image.png
```

### quoted links
```markdown
`[my attachment](.attachments.1/image.png)` is a link pointing to my attachment.
```

### links in code blocks
~~~markdown
## Example:
```markdown 
[my attachment](.attachments.1/image.png)
```
~~~

lib/Listeners/NodeCopiedListener.php
mejo-: Well, the problem is that there's still a shared lock on the target file, probably from the copy process. I didn't dig into it further yet, but since I only remove the lock in order to allow updating the content and don't want to interfere with the copy process otherwise, I decided to bring back the lock afterwards. That's also why I did this in the `NodeCopiedListener` directly, as it's only needed when `replaceAttachmentFolderId()` is called from inside `NodeCopiedEvent`.

lib/Service/AttachmentService.php
mejo-: Thanks, good catches. For the incomplete links that still get replaced I thought that would be an acceptable trade-off as I don't consider it likely to have `](.attachments.1/` on purpose. The advantage of using this "partial link syntax" was that I'm able to use string replacement and don't need to fiddle with regexes (less error-prone, faster).

But I agree that we should at least also catch and replace the other legitimate link syntaxes you mentioned - so I'll change this to use a regex replacement instead. Performance should not be a big deal here anyway given that it only happens on file copies.

mejo-: Btw. for quoted links etc. I would consider it acceptable that links to the attachment folder of this very page/file also get replaced, no?

lib/Listeners/NodeCopiedListener.php
max-nextcloud: ah.. that makes sense.

lib/Service/AttachmentService.php
juliusknorr: Would agree that the match is unique enough to be fine to just search and replace 👍

mejo-: I replaced the `str_replace()` with a more sophisticated `preg_replace()` now and added additional test cases.

tests/unit/Service/AttachmentServiceTest.php
```diff
@@ -104,4 +106,85 @@ public function testGetUniqueFileName() {
...
+			[ '![title](.attachments.1/)' ],
+			// Wrong fileId #1
+			[ '![image.png](.attachments.01/image.png)' ],
+			// Wrong fileId #2
+			[ '![image.png](.attachments.12/image.png)' ],
+			// Normal brackets around title
+			[ '!(image.png)(.attachments.1/image.png)' ],
+			// Square brackets in title
+			['![title [#1]](.attachments.1/image.png)' ],
+			// Reference link syntax with escaped square brackets in reference name
```
max-nextcloud, line 150:
```suggestion
			// Reference link syntax with square brackets in reference name
```

max-nextcloud (approved):
looks good to me. Thanks for the extra effort to catch markdown variations of links.

tests/unit/Service/AttachmentServiceTest.php
mejo-: Good catch :blush: 

Fixed now.

PR #7241, opened 2025-05-20
feat(psalm): Treat deprecated code as error

mejo-:
Copied from nextcloud/server#52860

### 🏁 Checklist

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,25 +1,25 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="2"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xmlns="https://getpsalm.org/schema/config"
-    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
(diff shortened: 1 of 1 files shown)


psalm.xml
```diff
@@ -37,5 +37,40 @@
 				<referencedClass name="Doctrine\DBAL\Statement" />
 			</errorLevel>
 		</UndefinedDocblockClass>
+		<DeprecatedClass>
+			<errorLevel type="suppress">
```
juliusknorr, line 41: Not sure i get this right, but wouldn't this suppress all of them directly? If we have any we should rather add to the baseline, right?

psalm.xml
```diff
@@ -1,25 +1,25 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="2"
```
juliusknorr, line 3: We are at error level 2 which should already throw errors if we have deprecations in use:

https://psalm.dev/docs/running_psalm/error_levels/

psalm.xml
mejo-: Indeed, I should have checked more thoroughly. Thanks for catching this.

PR #7275, opened 2025-05-30
use Photos metadata for pre-sized, blurhashed image placeholders

emberfiend:
### 📝 Summary

* Resolves: #4242 @provokateurin 
* Resolution to [photos/#2243](https://github.com/nextcloud/photos/issues/2243) **not** included

This introduces a cross-app "dependency" on metadata currently provided by image uploads when Photos is enabled (although it fails gracefully without it). When present in oc_files_metadata, this uses photos-size.value.width and height (and blurhash.value) to show a placeholder of correct size before the image itself loads.

Notes:
1. While it's slick on Firefox, Chrome (or Chromium anyway) has a noticeable gap between removing the placeholder and adding the loaded image. Any ideas? (I'd like to fix this before merge ideally.)
2. This adds some brittle-ness with the hardcoded values in the CSS and size calculations. Now whenever we change the ImageView we need to make sure the placeholder is the same size. Open to ideas about ways to mitigate this.
3. Many ResizeObservers running on a long document mean that this might make window resizing laggy on slower devices. You could make the case that we just size placeholders once, for the initial load's element width, or pass one in from the parent (Editor.vue already watches width).
4. The blurhash logic could be abstracted out since it's similar to what's in photos, but I feel one repetition does not justify it yet. As always, happy to accommodate if anyone feels differently.
5. Julius pointed out that this does nothing for resources AttachmentResolver doesn't provide metadata for, such as DAV-sourced bare URLs. Something to add in the future.
6. On the todo front, I started this PR moving SizeMetadataProvider out of Photos to server, but after discussing with Julius we decided to split the task and do that separately & in conversation with the involved parties. Just flagging it as a follow-up here.

General pointers about Vue best practices or local conventions also appreciated.

### 🖼️ Videos

🏚️ Before
[resize before.webm](https://github.com/user-attachments/assets/a76b02e7-ea94-46d8-b920-1fded5b2b3c9)

🏡 After (firefox)
[resize after.webm](https://github.com/user-attachments/assets/e819eee2-3415-4db7-9d32-9af0c59cc27f)

🏡 After (chromium)
[resize after chrome.webm](https://github.com/user-attachments/assets/130c155b-a66c-48e1-b65b-dd56789e4560)

### 🏁 Checklist

- [😇] [Tests](https://github.com/nextcloud/text#-testing-the-app) (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [N/A] Documentation ([README](https://github.com/nextcloud/text/blob/main/README.md) or [documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/text_configuration.rst#L2)) has been updated or is not required


src/nodes/ImageView.vue
```diff
@@ -199,6 +213,25 @@ export default {
...
 			return this.loaded && this.imageLoaded
 		},
+		canDisplayPlaceholder() {
+			return this.imageHeight > 0
+		},
+		blurhashSize() {
+			if (this.imageWidth > 0 && this.imageHeight > 0) {
+				const ratio = this.imageWidth / this.imageHeight
+				const newWidth =
+					this.wrapperWidth - 12 > this.imageWidth
```
juliusknorr, line 225: As discussed not a blocker but would be nice to base this and the paddings on the image view on the css variables so they adjust if designers change their minds about spacing.

PR #7298, opened 2025-06-12
Fix: Replace attachment file IDs when copying markdown files

pbirrer:
### 📝 Summary

When copying markdown files, the attachment IDs embedded in the file content were not updated to match the newly created attachments. This PR fixes that by:

- Updating `AttachmentService::copyAttachments()` to return a map of old to new file IDs.
- Adding `AttachmentService::replaceAttachmentFileIds()` to update these references in the file content.

### 🏁 Checklist

### DIFF
--- a/lib/Service/AttachmentService.php
+++ b/lib/Service/AttachmentService.php
@@ -693,27 +693,34 @@ class AttachmentService {
 	 * @param File $source
 	 * @param File $target
 	 *
+	 * @return array file id translation map
 	 * @throws InvalidPathException
 	 * @throws NoUserException
 	 * @throws NotFoundException
 	 * @throws NotPermittedException
 	 * @throws LockedException
 	 */
-	public function copyAttachments(File $source, File $target): void {
+	public function copyAttachments(File $source, File $target): array {
 		try {
 			$sourceAttachmentDir = $this->getAttachmentDirectoryForFile($source);
 		} catch (NotFoundException $e) {
 			// silently return if no attachment dir was found for source file
-			return;
+			return [];
 		}
 		// create a new attachment dir next to the new file
 		$targetAttachmentDir = $this->getAttachmentDirectoryForFile($target, true);
 		// copy the attachment files
+		$fileIdMapping = [];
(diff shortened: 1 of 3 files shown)


rvjr:
@mejo- this is a bugfix we discovered during our internal use of https://github.com/nextcloud/text/pull/6676. Copied files still referenced attachments in the attachments folder of the source file, when they were referred in md by file id.

lib/Service/AttachmentService.php
```diff
@@ -735,4 +742,18 @@ public static function replaceAttachmentFolderId(File $source, File $target): vo
...
 		}
 	}
+
+	public static function replaceAttachmentFileIds(File $target, array $fileIdMapping): void {
+		$patterns = [];
+		$replacements = [];
+		foreach ($fileIdMapping as $mapping) {
+			$patterns[] = '/(\[[^\]]+\]\(\S+\/f\/)' . $mapping[0] . '/';
+			// Replace `[title](URL/f/sourceId` with `[title](URL/f/targetId`
+			$replacements[] = '${1}' . $mapping[1];
```
max-nextcloud: I think this will drop the trailing `/` as it's included in the patterns but not in the replacements.

As this is a public static method and the arguments are not very complex it seems like a good candidate for a unit test. You'd have to mock `getContent` and `putContent` on target similarly to [how it's done in the `testReplaceAttachmentFolderId`  test](https://github.com/nextcloud/text/blob/main/tests/unit/Service/AttachmentServiceTest.php#L163-L170).

max-nextcloud (approved):

lib/Service/AttachmentService.php
max-nextcloud: Ahh... my bad... the `/` is ending the regexp. Unit test would still be nice - but now i see that this probably works well as is.

pbirrer: Alright, I added a unit test for replaceAttachmentFileIds. It actually helped catch a couple issues with my original regex too.

mejo-:
Thanks so much for taking care of this @pbirrer! Much appreciated :heart_hands:

PR #7300, opened 2025-06-13
fix: Do not escape possible other editors syntax when serializing

### LINKED ISSUE #4795: Escaping of square brackets breaks links of embedded attachments for external Markdown Editor
## Short summary of the issue

I'm using an external Markdown Editor (Obsidian) and sync the files via the NextCloud Desktop Client to the server. When saving a markdown file with the NextCloud text editor in the web interface, the links for embedded attachements are broken, because the square brackets are escaped with a backslash. Can this escaping for square brackets (or other chars) manually turned off?

- [ ] Markdown syntax is not properly rendered in text
- [ ] Markdown syntax is removed after closing text
- [x] Markdown syntax is changed after closing text
- [ ] Something else: Please describe more in detail

## Example markdown content

````markdown
![[Pasted image 20230908203759.png]]
````

## Resulting markdown content after saving

````markdown
![[Pasted image 20230908203759.png]]
````

## Screenshot of the rendering in text

image

## Details
- Nextcloud version: Nextcloud AIO v7.1.1

### PR DESCRIPTION
**Pull request by juliusknorr**
This is a first attempt to resolve https://github.com/nextcloud/text/issues/4795 and https://github.com/nextcloud/text/issues/6864

Mostly just a hacky fix for now, i need to think a bit more about in which cases we may want to skip the escape and in which not

- [ ] Check for side effects with cases like https://github.com/nextcloud/text/issues/931
- [ ] See how we want to handle `\[` and `\\[` in the markdown source

### DIFF
--- a/src/tests/markdown.spec.js
+++ b/src/tests/markdown.spec.js
@@ -313,6 +313,23 @@ describe('Markdown serializer from html', () => {
 			'<details>\n<summary>**summary**</summary>\n```\ncode\n```\n\n</details>\n',
 		)
 	})
+
+	const assertKeepSyntax = (source) => {
+		const tiptap = createRichEditor()
+		tiptap.commands.setContent(markdownit.render(source))
+
+		const end = tiptap.state.doc.content.size - 1
+		tiptap.commands.insertContentAt(end, ' editing')
+
+		const serializer = createMarkdownSerializer(tiptap.schema)
+		const md = serializer.serialize(tiptap.state.doc)
+		expect(md).toBe(source + ' editing')
+	}
+
+	test('keep syntax for brackets', () => {
+		assertKeepSyntax('test [[foo]] bar')
+		assertKeepSyntax('test ![[foo]] bar')
+	})
 })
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/tests/markdown.spec.js [hunk 1]
```diff
@@ -313,6 +313,28 @@ describe('Markdown serializer from html', () => {
...
+	const assertKeepSyntax = (source) => {
+		const tiptap = createRichEditor()
+		tiptap.commands.setContent(markdownit.render(source))
+
+		const end = tiptap.state.doc.content.size - 1
+		tiptap.commands.insertContentAt(end, ' editing')
+
+		const serializer = createMarkdownSerializer(tiptap.schema)
+		const md = serializer.serialize(tiptap.state.doc)
+		expect(md).toBe(source + ' editing')
```
mejo-, line 326:
```suggestion
		expect(md).toBe(source + ' editing')
		tiptap.destroy()
```

We had several test failures due to dangling editor instances in the past, so better to always explicitly destroy the editor in tests.

mejo-:
This is superseded by #8562

PR #7311, opened 2025-06-16
WIP: enh: use @nextcloud/y-http provider

### LINKED ISSUE #7283: y-http provider to replace y-websocket and websocket polyfill
The yjs provider that syncs with the server is way to complex

Have a y-http provider that takes care of the syncing and can handle being online / offline well.

## y-http

* [x] open connection
* [x] expose connection
* [x] basic syncing
* [x] basic awareness messages
* [x] include url in connection
* [x] expect client to know params to open
* [x] use ydoc client id in open
* [x] handle awareness queries (without sending them)
* [x] delay test requests
* [x] ensure awareness messages are send at least every 10 seconds
* [x] return only newer messages in DummyServer
* [x] hand version to sync request
* [x] ensure old awareness states are cleared after 30 seconds
* [x] implement closing the connection
* [x] test reopening the connection

## embedding in text js

* [x] implement client to instantiate y-http with
* [ ] extract api calls from Sync service and use connection instead
* [ ] extract document flags such as `hasOwner` from sync service and use them
* [ ] implement events in y-http provider and handle them instead of sync service events

## server side

* [ ] allow flagging updates as full compressed history
* [ ] at most send all versions since last compressed one
* [ ] clean up older versions

## later

* [ ] process queries
* [ ] send query message with every request
* [ ] send compressed history
* [ ] make the api stateless, removing the need for connection
* [ ] reset remote doc when receiving compressed history
* [ ] detect out of sync and reconnect (effectively receiving compressed history and all steps since then)
* [ ] handle sync queries (without sending them)

max-nextcloud:
This is still work in progress. Just created the PR to track TODOs and so on.

Fixes #7283.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@nextcloud/router": "^3.0.1",
         "@nextcloud/sharing": "^0.2.4",
         "@nextcloud/vue": "^8.27.0",
+        "@nextcloud/y-http": "^0.2.2",
         "@quartzy/markdown-it-mentions": "^0.2.0",
         "@tiptap/core": "^2.22.3",
         "@tiptap/extension-blockquote": "^2.22.3",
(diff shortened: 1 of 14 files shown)


max-nextcloud:
A lot of prep work landed in main since and there's no point in rebasing this PR. Better rebuild the commits on top of what we have now.

PR #7313, opened 2025-06-16
chore(migrate): useEditorMixin to useEditor composable

max-nextcloud:


src/components/Assistant.vue
```diff
@@ -271,7 +270,7 @@ export default {
 			return
 		}
 
-		this.$editor.on('selectionUpdate', this.onSelection)
+		this.editor?.on('selectionUpdate', this.onSelection)
```
mejo-, line 271: You now check in many places whether `editor` exists. This means that we cannot be sure that `useEditor` returns an editor object and always have to deal with `editor` being `undefined`? It looks a bit like an anti-pattern to me :thinking: 

If I get it right, then `editor` can be set by all the base components that call `provideEditor()` in `setup()` and then set `this.editor` in `created()` (i.e. `BaseReader`, `MarkdownContentEditor` and `Editor`). Since all consumers of `useEditor()` should be descendants (i.e child/granchild components) of the base editor components, `editor` should always already be set when used there, no?

src/components/CollisionResolveDialog.vue
```diff
@@ -29,25 +29,28 @@
...
-	mixins: [useEditorMixin, useIsRichEditorMixin, setContent, useSyncServiceMixin],
+	mixins: [useSyncServiceMixin],
 	props: {
 		syncError: {
 			type: Object,
 			default: null,
 		},
 	},
+	setup() {
+		// editor is needed for the setContent mixin
```
mejo-: `setContent()` is no longer a mixin, so the comment seems outdated.

src/components/Editor.vue
```diff
@@ -644,10 +622,7 @@ export default {
 			this.document = document
 
 			this.syncError = null
-			const editable = this.editMode && !this.requireReconnect
-			if (this.$editor.isEditable !== editable) {
-				this.$editor.setEditable(editable)
-			}
+			this.setEditable(this.editMode && !this.requireReconnect)
```
mejo-, line 627: Not directly related to this PR, but I just realized that we have a watcher for `requireReconnect` that calls `setEditable(!this.requireReconnect)`, which smells a lot like conflicting with this one.

If `this.editMode === false` and `this.requireReconnect` switches to `false`, the editor is set to editable by the watcher, but any subsequent change will switch it back to read-only here in the `onChange` handler. Or this is purely theoretical? :thinking:

src/components/Editor.vue
```diff
@@ -719,21 +694,22 @@ export default {
 		},
 
 		onStateChange(state) {
+			const editor = this.editor
```
mejo-: What's the purpose of the local variable here? Why not use `this.editor` everywhere in the handler?

src/components/Editor.vue
```diff
@@ -807,12 +783,11 @@ export default {
...
 			)
-			if (this.$editor) {
+			if (this.editor) {
 				try {
 					this.unlistenEditorEvents()
-					this.$editor.destroy()
-					this.$editor = null
-					this.hasEditor = false
+					this.editor.destroy()
+					this.editor = undefined
```
mejo-, line 790: I see that this is a moment when `this.editor` can become `undefined`. But since it's only called at the very end of `beforeDestroy()` it should not be a problem, no?

src/components/Editor.vue
```diff
@@ -905,14 +880,17 @@ export default {
 		hideTranslate() {
 			this.translateModal = false
 		},
+		applyCommand(fn) {
+			this.editor?.commands?.command(fn)
```
mejo-, line 884: This again has a little smell for me. We might expect `applyCommand()` to actually do what it's called to do and rely on it later. Just silently ignoring when `this.editor.commands` is not set looks potentially dangerous to me.

Thinking a bit more about it, I wonder whether such silent handling of undefined objects could contribute to the problem that we still have out-of-sync scenarios where steps didn't get applied.

src/composables/useEditorMethods.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import type { Editor } from '@tiptap/core'
+import escapeHtml from 'escape-html'
+import { unref, type ShallowRef } from 'vue'
+import markdownit from '../markdownit/index.js'
+
+export const useEditorMethods = (editor: ShallowRef<Editor | undefined>) => {
+	const setEditable = (val: boolean) => {
+		const ed = unref(editor)
```
mejo-: Nice, I didn't know about `unref()` so far. But it seems like this is the only place you use it and `editor.value` everywhere else :thinking:

src/components/Editor/MarkdownContentEditor.vue
```diff
@@ -88,6 +84,17 @@ export default {
 	},
 	emits: ['update:content'],
 
+	setup() {
+		const { editor } = provideEditor()
+		const { setEditable } = useEditorMethods(editor)
+		provide(editorFlagsKey, {
```
mejo-, line 92: I don't understand this. Why not use `providEditorFlags()` from `Editor.provider.ts` here as well?

mejo- (commented):
Really nice work! I didn't properly test it but read through the code changes and have some questions and remarks.

src/components/Assistant.vue
max-nextcloud: There are two things that come into play here right now:
1. `inject` returns an optional value - i.e. can always be undefined. It's possible to cast it to always be defined (see last example [in the docs for typing provide/inject](https://vuejs.org/guide/typescript/composition-api.html#typing-provide-inject) ). That implies that the component will behave in unpredicted ways - most likely throwing exceptions when being used without `provide`ing a value.
2. `editor` starts as undefined in the providing component such as `Editor.vue`. We only initialize it once we loaded the content.

2  is maybe artifact from how the content used to be loaded. I think we could be able to create the editor itself in the components setup routine these days. I'm a bit hesitant to changing the loading order right now as it feels fragile - but initializing the editor in the setup function rather than in loaded seems like it might even be a win for load times.

Thus far we hid these issues by only rendering subcomponents once the `$editor` was loaded. Whenever we removed the `v-if='hasEditor'` somewhere we'd get a lot of errors about `this.$editor` being undefined where it was not expected.

In addition the file that is rendered in an editor component never changes. If we switch between files in collective I believe the Editor component is unmounted and mounted again with a new value for the file related props. This is not how other components work, where one can just update a prop. This also means that for example the entire Menubar is recreated - with basically the same content. I was thinking that we might be able to keep the Editor component and change the props instead at some point. So far my assumption was that this would lead to `editor` being `undefined` while the new file is loaded. The `ydoc` that tracks the editing would also need to be replaced and it's part of the `collaboration` plugin config. But maybe we can even keep the editor around and "just" replace its content, the ydoc and the file related parts of the sync.

Anyway... I will take a look and see if I can initialize the `editor` in the setup function already. Then we might be able to do away with the `?` everywhere.

src/components/Editor/MarkdownContentEditor.vue
max-nextcloud: I'm trying to replace the inject that existed previously 
```js
			[IS_RICH_EDITOR]: {
				get: () => true,
			},
```

But now the flags come in one object so I cannot set just one of them.

If we'd just use `provideEditorFlags(props)` `isRichEditor` might end up being false in particular if `rich_editing_enabled` is set to false.

But I will move this to a helper function in `Editor.provider.js`. I think that might be cleaner / more clear. Something like `provideRichEditorFlag()`

src/composables/useEditorMethods.ts
max-nextcloud: Good point. The real benefit of `unref` is that it handles both refs and plain values fine. I wanted to use it somewhere else and express that in the function signature but turns out [`MaybeRef`](https://vuejs.org/api/utility-types.html#mayberef) is only available from vue 3.3 on. So I decided to just go with `.value`.

I'm somewhat torn between YAGNI and just using `.value` or being more flexible and using `unref` and more complex type signatures (`ShallowRef<Editor | undefined> | Editor` in this case).

But I agree consistency would be nice.

src/components/Editor.vue
max-nextcloud: I'll see if i can initialize the editor right in the setup function to avoid the whole problem all together.

src/components/Editor.vue
max-nextcloud: I agree. We can also just leave it and not set it to undefined. I was trying to replace `this.hasEditor = false` here. It hides a bunch of children. I wonder if they caused exceptions before because $editor was destroyed but they still existed.

src/components/Editor.vue
max-nextcloud: It's a leftover from when I believed i now had to use `this.editor.value` everywhere - which became really long. I'll clean it up.

src/components/Editor.vue
max-nextcloud: It definitely seems messy. I think we should have a single computed that determines if the editor is editable and that we hand to the editor composable where it's watched and acted upon. That way the tiptap commands would live inside the composable only and the components would only deal with vue data structures.

src/composables/useEditorMethods.ts
max-nextcloud: went with less complexity for now.

src/extensions/Session.ts
```diff
@@ -0,0 +1,55 @@
...
+export const Session = Extension.create({
+	name: 'session',
+
+	addStorage(): StoredSession {
+		return { ...emptySession }
+	},
+
+	addCommands() {
+		return {
+			setSession: (session) => () => {
```
mejo-: Could we change the passed session object here to have the same properties as the one returned from this class? E.g. `session.id` -> `session.sessionId` and `session.token` session.sessionToken`?

max-nextcloud:
Current Cypress failures:
## Runner 1

### ~~Assistant spec failure~~
Unrelated.
* https://github.com/nextcloud/assistant/pull/256 caused it.
* Fixed it in https://github.com/nextcloud/text/pull/7323.

### Menu spec failure
 Shows that the word count does not work anymore.
I experimented with this some. It seems to already be pretty unreliable in current main branch. Count sometimes does not update when typing new text for example.
Debugging showed that the content of the editor inside the `CharacterCount` component was different from the main content... do we have two editor instances there for some reason?

## Runner 5

### Share spec failing

When opening a shared file with download disabled the editor should not load.
The create request will fail and previously this led to the editor not being loaded at all.
Now it's loading as `this.editor` is already filled in the Editor.vue component.
Content is still not showing (as it's not served during create)
but it shows the read only toolbar including the `show outline` button.

We need to decide how we want the editor to behave before the create request / if it fails.
I like loading the toolbar before the content - but I'm thinking about the editing toolbar not the read only one.

src/extensions/Session.ts
```diff
@@ -0,0 +1,55 @@
...
+	sessionId: 0,
+	sessionToken: '',
+}
+
+export const Session = Extension.create({
+	name: 'session',
+
+	addStorage(): StoredSession {
+		return { ...emptySession }
+	},
```
max-nextcloud: oh no... this will require tiptap v3 as previously the extension storage is a singleton and therefor used by all loaded editor instances at the same time. ( See https://github.com/ueberdosis/tiptap/pull/6060 )

PR #7315, opened 2025-06-18
fix: prevent loading the code block extension twice for plain text

juliusknorr:
This fixes a warning that occurs as for plain text editing we include the codeBlock extension twice. I noticed this when running tests on #7300

We can only add the extension in our PlainTextExtension as it wraps everything we need for that already. In this case the addition options are just needed to pass them trough to the code block extension.

Fixes tons of warnings like:
```
stderr | src/tests/plaintext.spec.js > commonmark as plaintext > commonmark 557
[tiptap warn]: Duplicate extension names found: ['codeBlock']. This can lead to issues.

### DIFF
--- a/src/EditorFactory.js
+++ b/src/EditorFactory.js
@@ -13,8 +13,6 @@ import hljs from 'highlight.js/lib/core'
 
 import { logger } from './helpers/logger.js'
 import { FocusTrap, Mention, PlainText, RichText } from './extensions/index.js'
-// eslint-disable-next-line import/no-named-as-default
-import CodeBlockLowlight from '@tiptap/extension-code-block-lowlight'
 
 const lowlight = createLowlight()
 
@@ -65,11 +63,10 @@ const createPlainEditor = ({ language, extensions = [] } = {}) => {
 	return new Editor({
 		editorProps,
 		extensions: [
-			PlainText,
-			CodeBlockLowlight.configure({
+			PlainText.configure({
+				// Options to be passed through to `CodeBlockPlainText`
 				lowlight,
 				defaultLanguage: language,
-				exitOnTripleEnter: false,
 			}),
 			FocusTrap,
 			...extensions,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/extensions/PlainText.js [hunk 1]
```diff
@@ -14,7 +14,25 @@ import Text from '@tiptap/extension-text'
 export default Extension.create({
 	name: 'PlainText',
 
+	addOptions() {
+		return {
+			...this.parent?.(),
+			lowlight: undefined,
+			defaultLanguage: undefined,
+			exitOnTripleEnter: true,
```
mejo-: Nitpicking, but I guess in `PlainText` we don't want to exit on triple enter. I see that you overwrite this default in `EditorFactory.js` anyway, but maybe we can also default to `false` here?

src/EditorFactory.js
```diff
@@ -65,8 +63,7 @@ const createPlainEditor = ({ language, extensions = [] } = {}) => {
 	return new Editor({
 		editorProps,
 		extensions: [
-			PlainText,
-			CodeBlockLowlight.configure({
+			PlainText.configure({
```
mejo-, line 66:
```suggestion
			PlainText.configure({
					// Options to be passed through to `CodeBlockPlainText`
```

How about a short code comment here  to explain why we pass codeblock config options to the plaintext extension?

mejo- (approved):
Thanks! Didn't test but looks sensible to me.

I have two smaller remarks, none of them blocking.

src/EditorFactory.js
max-nextcloud: Alternatively the options could be set as a separate object on the code block lowlight key similar to how [tiptap defines its options in the starter kit](https://github.com/ueberdosis/tiptap/blob/main/packages/starter-kit/src/starter-kit.ts#L46-L50).

src/extensions/PlainText.js
juliusknorr: Yes, I changed it to not have an option for that in the PlainText extension now, as we can set it directly to false then and do not need to pass it through from `createPlainEditor`

src/EditorFactory.js
juliusknorr: Added a comment for now, we can move to separate options if we have some for other extensions being passed as well

PR #7347, opened 2025-06-22
bump/workflows/npm audit fix

max-nextcloud:
- **chore(workflows): update npm-audit-fix**
- **chore(workflows): no build in npm-audit-fix**

This will prevent failures on the non existing `master` branch.
Also stop building the js artifacts for the PR.


max-nextcloud:
I wonder if we should change the `build` script entry in package.json to some other name instead.
That would turn `npm run build --if-present` into a noop and remove the need to change these scripts.

mejo- (approved):

PR #7354, opened 2025-06-26
ci: update npm-audit-fix.yml workflow from template

nextcloud-command:
Automated update of the npm-audit-fix.yml workflow from https://github.com/nextcloud/.github


max-nextcloud:
we build on the integration branch with a separate job.

PR #7372, opened 2025-06-30
debug(SessionMiddleware): Log if we have an invalid session

mejo-:
We have spotted occurrences of this in combination with out-of-sync situations, so let's log it for now to better debug when this happens.

### 🏁 Checklist

### DIFF
--- a/lib/Middleware/SessionMiddleware.php
+++ b/lib/Middleware/SessionMiddleware.php
@@ -30,6 +30,7 @@ use OCP\ISession;
 use OCP\IUserSession;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Share\IManager as ShareManager;
+use Psr\Log\LoggerInterface;
 use ReflectionException;
 
 class SessionMiddleware extends Middleware {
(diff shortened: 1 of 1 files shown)


lib/Middleware/SessionMiddleware.php
```diff
@@ -171,6 +173,8 @@ public function afterException($controller, $methodName, \Exception $exception):
 		}
 
 		if ($exception instanceof InvalidSessionException) {
+			// We spotted occurrences of this in combination with out-of-sync situations, so let's log it for now.
+			$this->logger->error('Invalid session: ' . $exception->getMessage(), ['exception' => $exception]);
```
max-nextcloud: I think we can do this more specific and put it in here:
https://github.com/nextcloud/text/blob/main/lib/Middleware/SessionMiddleware.php#L94 as that's the function that's checked on push requests to a non public session and would throw an `InvalidSessionException`.

That would leave out exceptions caused elsewhere from the logs.

mejo-: Good point, pushed a change now. Is this how you meant it?

max-nextcloud: exactly!

PR #7377, opened 2025-06-30
fix/out of sync

max-nextcloud:
- **chore(debug): log pending structs in remote ydoc**
- **fix(sync): check pending structs and resync if needed**
- **fix(sync): send SyncStep1 right after step could not be applied**

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -850,6 +850,7 @@ export default {
 				filePath: this.relativePath,
 				clientId: this.ydoc.clientID,
 				pendingStructs: this.ydoc.store.pendingStructs,
+				pendingStructsRemote: this.syncProvider?.remote.store.pendingStructs,
 				clientVectors: [],
 				documentState: this.syncService?.getDocumentState(),
 			}
--- a/src/services/y-websocket.js
+++ b/src/services/y-websocket.js
@@ -289,7 +289,7 @@ export class WebsocketProvider extends Observable {
 			params = {},
 			protocols = [],
 			WebSocketPolyfill = WebSocket,
-			resyncInterval = -1,
+			resyncInterval = 30_000,
 			maxBackoffTime = 2500,
 			disableBc = false,
 		} = {},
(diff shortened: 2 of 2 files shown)


src/services/y-websocket.js
```diff
@@ -44,12 +44,22 @@ messageHandlers[messageSync] = (
...
 		provider,
 	)
+	if (
+		!pendingStructsBefore
+		&& provider.doc.store.pendingStructs
+		&& !encoder.hasContent
+	) {
+		// message could not be processed
+		encoding.writeVarUint(encoder, messageSync)
+		syncProtocol.writeSyncStep1(encoder, provider.doc)
```
mejo-, line 63: Looking at the code that handles `resyncInterva` below, there `this.ws.send(encoding.toUint8Array(encoder))` is called after `syncProtocol.writeSynStep1`. I guess it's not needed here because whatever calls this message handler will send pending requests to the websocket provider anyway?

src/components/Editor.vue
```diff
@@ -850,6 +850,8 @@ export default {
 				filePath: this.relativePath,
 				clientId: this.ydoc.clientID,
 				pendingStructs: this.ydoc.store.pendingStructs,
+				pendingStructsRemote:
+					this.$providers[0]?.remote.store.pendingStructs,
```
mejo-: `this.$provides[0]` was replaces with `this.syncProvider` in 12547cc05a3d8ce4900309db066ac7eddabcc35e, no?

mejo- (approved):
I have two comments/questions. Otherwise code changes look good to me :tada:

src/services/y-websocket.js
max-nextcloud: Exactly. The handler this is in reads from the decoder passed to it and writes to the encoder. Whatever is written to the encoder will be send over the presumed websocket.
That's also why we need to check if anything has been written already - which is unlikely as the only step that creates a response is a SyncStep1 ("Question") and that's unlikely to cause `pendingStructs`.

src/components/Editor.vue
max-nextcloud: yes... will need to adjust after rebase on main.

PR #7412, opened 2025-07-06
[stable30] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (8 unfixable, 14 only fixable manually using --force)


max-nextcloud:
Updating some tiptap packages but not others does not seem right. Let's wait for dependabot to tackle this.

PR #7418, opened 2025-07-08
fix(files): workspace implementation

skjnldsv:
Might require https://github.com/nextcloud/server/pull/53872

Master have received a few fixes, the current workspace implementation wasn't optimal, so I fixed it.

1. Now, the headers are properly ALWAYS mounted
2. We don't need to keep track of the `newWorkspaceCreated`
3. You can emit a proper `files:node:updated` to dispatch the fact that the current folder have changed (and update the workspace attributes accordingly)

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { dirname } from 'path'
 import { emit } from '@nextcloud/event-bus'
 import { getCurrentUser } from '@nextcloud/auth'
 import { getSharingToken } from '@nextcloud/sharing/public'
@@ -22,6 +21,7 @@ import axios from '@nextcloud/axios'
 import TextSvg from '@mdi/svg/svg/text.svg?raw'
 
 import { openMimetypes } from './mime.js'
+import Vue from 'vue'
 
 const FILE_ACTION_IDENTIFIER = 'Edit with text app'
 
@@ -135,8 +135,6 @@ const registerFileActionFallback = () => {
 	}
 }
 
-let newWorkspaceCreated = false
-
 export const addMenuRichWorkspace = () => {
 	const descriptionFile =
 		t('text', 'Readme') + '.' + loadState('text', 'default_file_extension')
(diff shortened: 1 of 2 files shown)


src/helpers/files.js
```diff
@@ -202,64 +200,45 @@ export const FilesWorkspaceHeader = new Header({
...
 
-		// Currently there is not much use in updating the vue instance props since render is called on every folder change
-		// removing the rendered element from the DOM
-		// This is only relevant if switching to a folder that has no content as then the render function is not called
-
 		const hasRichWorkspace = !!folder.attributes['rich-workspace-file']
-		vm.path = folder.path
-		vm.hasRichWorkspace = hasRichWorkspace
-		vm.content = folder.attributes['rich-workspace']
+		FilesHeaderRichWorkspaceInstance.hasRichWorkspace = hasRichWorkspace ?? false
```
max-nextcloud: minor nitpick:
`hasRichWorkspace` as defined above is a boolean.

```js
		const hasRichWorkspace = !!folder.attributes['rich-workspace-file']
```

So `?? false` does  nothing.

```suggestion
		FilesHeaderRichWorkspaceInstance.hasRichWorkspace = hasRichWorkspace
```

src/helpers/files.js
```diff
@@ -202,64 +200,45 @@ export const FilesWorkspaceHeader = new Header({
...
-			vm = null
+	render: async (el, folder) => {
+		// If the RichWorkspace component is already mounted, destroy it
+		if (!FilesHeaderRichWorkspaceView) {
+			FilesHeaderRichWorkspaceView = (
+				await import('../views/RichWorkspace.vue')
+			).default
+		} else {
+			FilesHeaderRichWorkspaceInstance.$destroy()
+			console.debug('Destroying existing FilesHeaderRichWorkspaceInstance')
```
max-nextcloud: This assumes that whenever `FilesHeaderRichWorkspaceView` is truthy (else case) `FilesHeaderRichWorkspaceInstance` is set.
I don't know if that's the case. I would suggest to decouple the two steps of importing and destroying like this:

```suggestion
	render: async (el, folder) => {
		if (!FilesHeaderRichWorkspaceView) {
			FilesHeaderRichWorkspaceView = (
				await import('../views/RichWorkspace.vue')
			).default
		}
		// If the RichWorkspace component is already mounted, destroy it
		if (FilesHeaderRichWorkspaceInstance) {
			FilesHeaderRichWorkspaceInstance.$destroy()
			console.debug('Destroying existing FilesHeaderRichWorkspaceInstance')
```

max-nextcloud (approved):
Two minor nitpicks - other than that looks great. Thanks for taking care of this so quickly!
❤️

skjnldsv:
Done @max-nextcloud :+1:

max-nextcloud:
Testing out if this helps with rich workspace tests failing in stable31 lately.

skjnldsv:
@max-nextcloud seems like when you load an existing workspace, clicking the content doesn't trigger the editor mode anymore :thinking:

max-nextcloud:
@skjnldsv I'll take a look. Thanks for the heads up.

max-nextcloud:
grafik
Seeing this in the console makes me think we do not extend the Vue instance with the l10n functions.

**update** adding them back removes the error but does not change the behavior.

skjnldsv:
Ah, nice catch!
Extending the vue with globals is a deprecated behaviour anyway.
I'd suggest always importing on each component to be cleaner :)

max-nextcloud:
I suspect this broke due to a later change in server. Otherwise i don't see how the cypress tests would have passed as they test editing the workspace.

**update:** most cypress tests pass locally for me as they create the workspace and then test it. - and that works fine.

:thought_balloon: *thinking out loud*
It looks like only the RichTextReader is getting mounted because `this.getFileInfo` is not getting called as `this.hasRichWorkspace` is false during onMounted.

PR #7435, opened 2025-07-13
[stable30] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (8 unfixable, 14 only fixable manually using --force)


max-nextcloud:
another partial tiptap update... let's not do that.

PR #7442, opened 2025-07-13
refactor/polling backend

max-nextcloud:
- **chore(types): type polling backend and notify service**
- **refactor(connection): use sync api instead of SessionConnection**


max-nextcloud:
superseeded by #7446

PR #7446, opened 2025-07-13
refactor/upload with separate api

max-nextcloud:
- **chore(refactor): use api functions for attachment handling**


max-nextcloud:
superseeds #7442

src/apis/Attach.ts
```diff
@@ -0,0 +1,90 @@
...
+	file: string | Blob,
+) {
+	const {
+		documentId,
+		sessionId,
+		sessionToken,
+		shareToken: token,
+	} = unref(connection)
+	const params = Object.entries({ documentId, sessionId, sessionToken, token })
+		.map(([k, v]) => v && [k, v.toString()]) // convert numbers to strings
```
mejo-: While this looks like an elegant solution, I find it hard to parse and understand. Wonder whether it wouldn't be better to use the old style and make it explicit, which search parameters are added here :thinking: 

Edit: Haha funny, just saw that you replaced it in a later commit :laughing:

PR #7447, opened 2025-07-14
chore(update): explicitly import l10n where needed

max-nextcloud:
* Follow up for #7448 

## fix(richworkspace): load after mounted.

Get file info when hasRichworkspace turns true.

The render function in `FilesWorkspaceHeader`
runs with `hasRichWorkspace` false and mounts the Component.
Then the update function runs with `hasRichWorkspace` true.
Now the component needs to load the workspace.

## chore(update): explicitly import l10n where needed
It's the recommended way to handle @nextcloud/l10n these days.

### DIFF
--- a/src/views/RichWorkspace.vue
+++ b/src/views/RichWorkspace.vue
@@ -112,6 +112,9 @@ export default {
 		},
 		hasRichWorkspace(value) {
 			this.localHasRichWorkspace = value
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Cool, thanks for the cleanup. Probably something we should do in Collectives as well.

PR #7471, opened 2025-07-20
chore/use awareness

max-nextcloud:
- **chore(move): useDelayedFlag to composables**
- **enh(status): use awareness state instead of sessions**
- **fix(css): render cursor color for logged in users**
- **fix(people): hide local client for guests**


max-nextcloud:
It's not possible to fully rely on the awareness state - in particular as it is not authorized by the server. So it would be possible to impersonate other users etc.

I looked into various options of authorizing the information in the awareness state - but most of them result in duplicating information as we need some authorative source to check against. So let's just keep the sessions as the single point of truth with respect to who is editing the content.

PR #7474, opened 2025-07-23
Several link bubble/preview fixes

mejo-:
### 📝 Summary

* fix(PreviewOptions): Pass href, so "open in new tab" option is displayed
* fix(img): Remove global CSS rules for images in prosemirror
* fix(links): Add open button to link bubble
* fix(LinkBubble): Read `editor.contentComponent` when it's ready
* enh(editorAPI): Allow to pass an openLinkHandler to editor
* fix(LinkBubble): Fix preview border overflowing link bubble
* Resolves: nextcloud/collectives#1226

### Screenshots

| :derelict_house: Before | :house_with_garden: after |
| --- | --- |
| image | image |
| image | image |

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/Links.spec.js
+++ b/cypress/e2e/Links.spec.js
@@ -43,7 +43,7 @@ describe('test link marks', function () {
 			cy.get('.link-view-bubble .widget-default', { timeout: 10000 })
 				.find('.widget-default--name')
 				.contains('Nextcloud')
-				.click({ force: true })
+				.click()
 		})
 
 		it('shows a link preview in the bubble after browsing to link', () => {
@@ -56,6 +56,16 @@ describe('test link marks', function () {
 			cy.get('.link-view-bubble .widget-default', { timeout: 10000 })
 				.find('.widget-default--name')
 				.contains('Nextcloud')
+		})
+
+		it('open button opens a new tab', () => {
+			const link = 'https://nextcloud.com/'
+			cy.insertLine(link)
+			clickLink(link)
+
+			cy.get('.link-view-bubble button[title="Open link"]').click()
+
+			cy.get('@winOpen').should('have.been.calledOnce')
 		})
 
 		it('closes the link bubble when clicking elsewhere', () => {


src/components/Link/LinkBubbleView.vue
```diff
@@ -10,6 +10,16 @@
...
 				{{ title }}
 			</div>
+			<!-- open link -->
+			<NcButton
+				:title="t('text', 'Open link')"
+				:aria-label="t('text', 'Open link')"
+				type="tertiary"
+				@click="openLink(href)">
+				<template #icon>
+					<OpenInNewIcon :size="20" />
```
max-nextcloud, line 20: I looked for a different icon briefly as this does not always open in new... but i found nothing satisfying.

src/helpers/links.js
```diff
@@ -57,4 +57,26 @@ const isLinkToSelfWithHash = function (href) {
...
+	if (
+		window.OCA.Collectives?.vueRouter
+		&& linkUrl.pathname.toString().startsWith(generateUrl(collectivesUrlBase))
+	) {
+		const collectivesUrl = linkUrl.href.substring(
+			linkUrl.href.indexOf(collectivesUrlBase) + collectivesUrlBase.length,
+		)
+		window.OCA.Collectives.vueRouter.push(collectivesUrl)
+		return
+	}
```
max-nextcloud, line 78: I don't understand what you are doing here. It looks like you are truncating the linkUrl to end with the collectivesUrlBase and then opening it. Do you want to strip the collective name, page and fragments?

max-nextcloud: ahh now i got it. You are using the `substring` **after** the base url. That **does** make sense.

max-nextcloud (approved):

src/helpers/links.js
mejo-: Nope, I want to make sure that relative links and absolute links without host also work. That's why I initiate the URL with `new URL()` with `window.location` as base.

PR #7477, opened 2025-07-24
fix(share): use relative path in share as davPath for attachments

max-nextcloud:
Allow opening attachments in public collective shares
as well as in folder descriptions of public shares.


max-nextcloud:
I tested this in collectives and in shared folders with a folder description. However I don't know if I am missing some corner cases. All in all the approach seems sound though.

mejo- (approved):
Thanks for tackling this :heart:

Does this also work with subpage shares in Collectives? I can imagine it does, as `internalPath` might be relative to the share root? :thinking:

max-nextcloud:
> Does this also work with subpage shares in Collectives? I can imagine it does, as `internalPath` might be relative to the share root? 🤔

Sadly not.... it's relative to the collective root. But maybe we can also resolve it relatively to the share root.

max-nextcloud:
@mejo-  I think I got it now. Could you double check the getShareFolder function?
It's a bit unfortunate to repeat all this code from the getTextFilePublic function. I tried to refactor it with a function that returns both the file and the folder as a `list{File, Folder | null}` - but I never managed to make psalm happy and it also introduces a fair bit of complexity.

mejo- (approved):
Nice, great that you found a generic solution that works for all kinds of public file shares :heart: 

I didn't test myself, but the code looks sensible to me.

PR #7501, opened 2025-07-27
fix(workspace): toggle right away and only render if needed

### LINKED ISSUE #7505: Cypress workspace test is flaky when switching between views
**Describe the bug**
The workspace test has been flaky recently.

It tries to render a workspace when there is none after switching to the 'Recent' view.

**To Reproduce**
See https://github.com/nextcloud/text/actions/runs/16546964854/job/46796529217 for an example of a failed run.

<details>
<summary>Cypress output</summary>

```
   1) Workspace
       Hides the workspace when switching to another view:
     AssertionError: Timed out retrying after 10000ms: Expected <div#rich-workspace> not to exist in the DOM, but it was continuously found.
      at Context.eval (http://localhost:8081/__cypress/tests?p=cypress/e2e/workspace.spec.js:639:33)

```

</details>

Image

**Expected behavior**
Workspace should only show if there is one and tests should pass reliably

**Screenshots**
Some debugging revealed that `localHasRichWorkspace` is true despite `hasRichWorkspace` having switched to false. This is most likely due to a `getFileInfo` request from the previous view finishing afterwards.

Image

**Server details:**
Text github CI

max-nextcloud:
fixes #7505

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -192,13 +192,15 @@ let FilesHeaderRichWorkspaceView
 let FilesHeaderRichWorkspaceInstance
 let latestFolder
 
+function enabled(_, view) {
+	return ['files', 'favorites', 'public-share'].includes(view.id)
+}
+
 export const FilesWorkspaceHeader = new Header({
 	id: 'workspace',
 	order: 10,
+	enabled,
 
-	enabled(_, view) {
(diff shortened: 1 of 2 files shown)


mejo- (approved):
Nice finding! I didn't test, but I guess you did. Thanks for being so persistent here!

mejo-:
Probably also a fix worth backporting, no?

PR #7514, opened 2025-07-29
fix/move wide page toggle

### LINKED ISSUE #6703: Line length toggle follow ups
### Follow up from https://github.com/nextcloud/text/pull/6569

- [ ] Hide the toggle for line length in collectives as it has its own option
- [ ] #7514 
  - Feedback from @jancborchardt 
    > I would expect a setting like this to be in the collapsed part of the toolbar along with these other more info/settings like items, as the "people" menu is more focused on collaboration and sharing.
(Also, this setting is disabled by default, right?)
    > Image

### LINKED ISSUE #7091: toggle for editor line length per user is not persistant
[feat: Add toggle for editor line length per user #6569 ](https://github.com/nextcloud/text/pull/6569)

This new feature is not persistent. I open a document, select 'Full width editor', close/re-open document, 'Full width editor' is disabled again.

So I have to enable full width editor every single time I open a document. The default view is so narrow it is completely useless for anything except a grocery list.

Not sure if this should be classified as a bug or a feature request. Depends on if it is interpreted as the all so classic "It's not a bug, it's a feature".

max-nextcloud:
* fixes #7091.
* fixes #6703.

- **chore(extract): TranslateButton from MenuBar**
- **chore(move): handleEditorWidthChange into SessionList**
- **chore(npm): prettier:change to format only changed files**
- **fix(width): load initial state in editor**
- **fix(width): remember full width on remount**
- **fix(width): use nextcloud event bus to sync width**
- **fix(width): move toggle into remaining actions menu**


src/components/Menu/MenuBar.vue
```diff
@@ -51,15 +51,7 @@
 				:force-enabled="true"
 				@click="activeMenuEntry = 'remain'">
 				<template #lastAction="{ visible }">
-					<NcActionButton
-						v-if="canTranslate"
```
mejo-, line 55: `canTranslate`, `loadState` and `showTranslate` can be removed from `MenuBar.vue` now that they got moved to the separate component. Even though I wonder whether `canTranslate` should stay here and the `v-if="canTranslate"` condition be moved to `<TranslateButton />` in this component.

src/components/Menu/TranslateButton.vue
```diff
@@ -0,0 +1,41 @@
...
+		commands,
+		view: { state },
+	} = editor.value
+	const { from, to } = state.selection
+	let selectedText = state.doc.textBetween(from, to, ' ')
+	if (!selectedText.trim().length) {
+		commands.selectAll()
+		selectedText = state.doc.textContent
+	}
+	console.debug('translation click', state.selection, selectedText)
```
mejo-: Seems like a development leftover (from earlier times) to me that can go.

package.json
```diff
@@ -15,6 +15,7 @@
     "lint": "tsc && eslint --ext .js,.ts,.vue src cypress",
     "lint:fix": "tsc && eslint --ext .js,.ts,.vue src cypress --fix",
     "prettier": "prettier --check .",
+    "prettier:change": "git diff HEAD --name-only | xargs prettier --write --no-error-on-unmatched-pattern",
```
mejo-: This doesn't work in CI, with the changed files being committed already.

src/composables/useEditorWidth.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/*
+ * Handle the state of the full editor widht toggle.
```
mejo-:
```suggestion
 * Handle the state of the full editor width toggle.
```

src/components/Menu/WidthToggle.vue
```diff
@@ -0,0 +1,30 @@
...
+import { useEditor } from '../../composables/useEditor.ts'
+import { useEditorWidth } from '../../composables/useEditorWidth.ts'
+
+const { editor } = useEditor()
+const { isFullWidth, setFullWidth } = useEditorWidth()
+
+const redrawEditor = () => {
+	nextTick(() => {
+		const { commands, view } = editor
+		view.updateState(view.state)
```
mejo-: Do you know why it's necessary to update the state after changing the width?

src/composables/useEditorWidth.ts
```diff
@@ -0,0 +1,64 @@
...
+	provide,
+	readonly,
+	ref,
+	watch,
+	type InjectionKey,
+	type Ref,
+} from 'vue'
+
+// Keep the current value around when leaving the editor and reopening
+let valueSingleton = loadState('text', 'is_full_width_editor', false)
```
mejo-, line 37: I don't understand this. How is this "kept around" when the editor gets destroyed? :thinking:

src/event-bus.d.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+declare module '@nextcloud/event-bus' {
+	interface NextcloudEvents {
```
mejo-, line 7: Is this used anywhere? :thinking:

src/components/Menu/MenuBar.vue
```diff
@@ -51,15 +51,8 @@
...
-						close-after-click
-						@click="showTranslate">
-						<template #icon>
-							<TranslateVariant />
-						</template>
-						{{ t('text', 'Translate') }}
-					</NcActionButton>
+					<TranslateButton />
+					<WidthToggle />
```
mejo-, line 55: For some reason, the clickable area doesn't span the whole menu width:

image

src/components/Editor/SessionList.vue
```diff
@@ -47,26 +47,15 @@
 							>({{ t('text', 'guest') }})</span
 						>
 					</li>
-					<li>
```
mejo-, line 50: Now that `NcPopover` doesn't have any focussable element anymore, it throws an error because it tries to add a focus trap but fails. You have to add `no-focus-trap` as property to `NcPopover` component:

image

src/components/Menu/MenuBar.vue
max-nextcloud: Kept canTranslate and did as you suggested.

max-nextcloud: adjusted the original commit with a rebase

package.json
max-nextcloud: That's correct. I introduce it to use it locally to speed up formatting before committing. But I also noticed that if I forgot to format and already commited I'd want to use a different base commit. Not sure how to tackle this best. Could always diff against `origin/main` or whatever is 'upstream'.

I had a look at how to look up the tracking branch but it seems overly complex for  a npm script command:
```
git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)"
```
https://stackoverflow.com/a/9753364

src/event-bus.d.ts
max-nextcloud: It's type hints that allow detecting from parameters given to the event:
grafik

src/composables/useEditorWidth.ts
max-nextcloud: It's on the top level of the module. So it's initialized when the file is imported. It's not exported directly but it's used in the exported `provideEditorWidth` function. So it will be kept around in that functions [closure](https://developer.mozilla.org/de/docs/Web/JavaScript/Guide/Closures#closure).
Even if the editor does not exist anymore - the `provideEditorWidth` function is still around as it could be called again by code that imported it. So the closure is also not cleaned up and `valueSingleton` keeps its value.

max-nextcloud:
The only thing that's missing from also fixing #6703 would be hiding the toggle in collectives as collectives has its own toggle. I'm unsure how best to achieve this though.
* We could hide it everywhere where `createEditor` is used - but that seems too generic.
* Collectives could specify a parameter to `createEditor`.
* Collectives could provide some data that we inject in the Editor that hides the toggle.

Alternatively we could make the toggle work on the collectives per page setting - but I wonder if that would be more confusing or less.

**update**: I found a different way (https://github.com/nextcloud/text/commit/f8ee19440106ceb93a4b7efbc0d9e807ffa870d9):
Text now detects the width has already been set and does not render the width toggle anymore.

src/composables/useEditorWidth.ts
mejo-: Ahh, thanks for the explanation. That makes a lot of sense!

src/components/Menu/MenuBar.vue
max-nextcloud: Good Catch! I was still using the old component but now we are in an NCActions container. So need to use `NcActionsCheckbox`.

src/components/Menu/WidthToggle.vue
max-nextcloud: According to my testing it's not necessary. It was introduced by the community contributor first submitting the toggle. I will remove it.

src/components/Menu/MenuBar.vue
max-nextcloud: Moving `canTranslate` into the `TranslateButton` again as we already had a test for it there and it's more cumbersome to test in the `MenuBar` component that has so many other components.

src/composables/useEditorWidth.ts
```diff
@@ -30,10 +30,28 @@ import {
...
 >
+
+const maxWidthSetOutsideOfText = () => {
+	const alreadySet = getComputedStyle(document.body).getPropertyValue(
+		'--text-editor-max-width',
+	)
+	const setByText = document.documentElement.style.getPropertyValue(
+		'--text-editor-max-width',
+	)
+	return alreadySet && alreadySet !== setByText
```
mejo-: I fear I don't fully understand the logic here. Both read the value of the css variable `--text-editor-max-width`, the first from the `<body>` element, the second from the root element (`<html>`). So Text attaches the variable to the root `<html>` element and Collectives attaches it to the `<body>` element and that's how you distinguish them?

In my tests with latest master branch of server, the latter is always empty, even if I open a text editor from the Files app. In other words I never see the full width toggle at the moment.

In general, this feels a bit brittle to me as it will make the fullWidth toggle disappear as soon as we change something with how `--text-editor-max-width` is set.

Maybe it's really better to be more explicit and make this configurable via the editorApi at `editor.js`?

mejo-: I just checked the code and it seems to me that actually both Text and Collectives set `--text-editor-max-width` on the root element: see https://github.com/nextcloud/collectives/blob/aeffe87608250a50d0eb99b526362616fa2e0eb2/src/css/editor.scss#L7 and https://github.com/nextcloud/text/blob/c74d204e8a9ddbaefc75b8af021f06060120faae/src/css/variables.scss#L7

mejo-: Also I don't understand why you use `getComputedProperties(<element>)` first and `<element>.style` second. Is this intended?

package.json
max-nextcloud: Moved to #7546

src/composables/useEditorWidth.ts
max-nextcloud: What I am trying to do here is this:
1. Read the effective value of `--text-editor-max-width` for the document body. I'm using `getComputedStyle` so it does not matter where exactly the style is set - wether on the body element or `:root` or in some other way.
2. Check if text has set a value already - for example because it was opened before. This value would always be set on `document.documentElement.style` (as implemented below).
3. If 1. and 2. both have a value set - check if it's the same to determine if 1 is only set because of 2.

Now let me figure out why this is not working as intended.

max-nextcloud: Okay... I see what's happening here. Text also set's the width on `:root` - twice in fact. So 1 returns a value even when 2 has not been set yet.

max-nextcloud: Text set the `--text-editor-max-width` in css. I removed that as it will be set either externally or by `useEditorWidth` now.

src/composables/useEditorWidth.ts
```diff
@@ -0,0 +1,86 @@
...
+let valueSingleton = loadState('text', 'is_full_width_editor', false)
+
+// This is either the reactive value of the editor width toggle
+// or null if the width has already been set outside of text.
+export const editorWidthKey = Symbol('text:editor:width') as InjectionKey<
+	Readonly<Ref<boolean> | null>
+>
+
+const maxWidthSetOutsideOfText = () => {
+	const alreadySet = getComputedStyle(document.body).getPropertyValue(
```
mejo-, line 53: Would be great to have a comment here that documents the implicit knowledge. I.e. that an app that uses Text might set the CSS variable `--text-editor-max-width` itself via an own full width toggle and that we don't expose the Text full width toggle in this case.

Issue #4687, opened 2023-08-17
Inserting Images into Table Cells


nliedmey:
Hello,

is it possible to add the feature of inserting images into table cells?

Currently, when clicking into a table cell and inserting an image via the "insert attachment" button, the image is placed below the table and another table is created below the image as well.

In the screenshot, the image was originally inserted in a cell in the middle column of the table above.

![collectives_table_error](https://github.com/nextcloud/collectives/assets/73893241/ac6a355c-66f2-424d-8334-e3ed38b1d87f)

Issue #4047, opened 2023-04-13
Pasting an image into a table destroys it


provokateurin:
**Describe the bug**
Pasting an image into a table destroys it

**To Reproduce**
1. Create a table
2. Paste an image into one of the cells

**Expected behavior**
I'm not sure if images in tables are supported in markdown, but if not it should prevent me from pasting the image, otherwise it should just end up in the cell.

**Screenshots**
![image](https://user-images.githubusercontent.com/26026535/231854225-c9d7bbe1-e0af-4897-8d58-9813f2b29eb7.png)
I pasted the image into the cell with the letter e. There were only two rows with the letters a-i before.

**Client details:**
 - OS: Linux
 - Browser: Firefox
 - Version: 112
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.7.2

**Nextcloud version:** 26

</details>

datenangebot:
Can reproduce it. A lot new rows will be created while inserting a image from the clipboard.

max-nextcloud:
> Can reproduce it. A lot new rows will be created while inserting a image from the clipboard.

We fixed the broken table state in #5461 . Now inserting an image will still split the table but the two tables will at least be in a valid state.

What remains is adding the ability to insert an image to the cell.

PR #7523, opened 2025-08-01
Multiline tables with block node content

mejo-:
### 📝 Summary

* Allow any block node type as children of table cells
* Use [markdown-it-multimd-table](https://github.com/redbug312/markdown-it-multimd-table) to parse markdown tables
* Implement new markdown serializer for whole tables to support multi-line cells
* Allow headerless tables
* Support cell alignment in serializer, but only serialize cells that contain just text, paragraph and image nodes. All other nodes can be problematic and result in broken markdown that cannot be parsed again.
* Resolves: #4047
* Resolves: #4687
* Resolves: #7582 

#### 🖼️ Screenshots

🏡 Editor | Serialized Markdown
---|---
image | image

### 🚧 TODO

- [x] Allow `block+` as content for tableCell. For some reason, allowing blockquotes results in all cell rows being serialized to markdown as blockquote. I didn't find out yet why this is the case (fixed magically, probably something upstream in Tiptap)
- [x] Parsing single-line block elements (like lists with one item, blockquotes with one line) from markdown is broken
- [x] Single-line table rows take more vertical space now
- [x] ~~2x return used to create/jump to next row. This is currently broken. Do we want it back?~~ (we don't)
- [x] ~~Optionally remove our custom newline special-handling with `<br>` in table cells~~ (let's keep it to not break backwards-compatibility when parsing existing tables)
- [x] Reset sessions on app update
- [x] Custom paste handler breaks pasting into code blocks. We need to replace it with a paste handler that prevents nested tables and allows everything else
- [x] Prevent nested tables via `filterTransaction` (as suggested [here](https://discuss.prosemirror.net/t/limit-the-number-of-nested-list-levels-to-3/4289/2))

### 🏁 Checklist

### DIFF
--- a/src/css/prosemirror.scss
+++ b/src/css/prosemirror.scss
@@ -360,6 +360,9 @@ div.ProseMirror {
 			&:first-child {
 				border-left: 1px solid var(--table-color-border);
 			}
+			* {
+				z-index: 3;
+			}
 		}
 		td {
 			border-top: 0;
@@ -403,6 +406,18 @@ div.ProseMirror {
 			td:last-child {
 				border-bottom-right-radius: var(--table-border-radius);
 			}
+		}
+
+		.selectedCell:after {
+			background: var(--color-background-dark);
+			content: '';
+			left: 0;
+			right: 0;
+			top: 0;
+			bottom: 0;
+			pointer-events: none;
+			position: absolute;
+			z-index: 2;
 		}
 	}
 }


max-nextcloud:
@mejo- Did you figure out what prevents nesting a table in a table right now?

mejo-:
> @mejo- Did you figure out what prevents nesting a table in a table right now?

Not exactly, but I guess it's rather a broken `can()` (or however this command to check whether the feature is possible at current cursor position is called). Pasting a table into a table cell works currently. That's why I added the two TODO points to make it technically impossible to add nested tables (via `filterTransaction`) and to filter out nested tables in `handlePaste()`.

Aveyron-RetD:
Will this fix as well, rich previews added with the smart picker ?
https://github.com/nextcloud/text/issues/7582

janbaum:
> 2x return used to create/jump to next row. This is currently broken. Do we want it back?

Just wanted to say that I would love this being broken :melting_face:

mejo-:
This PR is ready for review from my side @max-nextcloud @juliusknorr

cypress/e2e/nodes/Table.spec.js
max-nextcloud, line 1: You are not really testing the extensions added in this test, are you? If they are not needed I'd rather leave them out and test the bare minimum.

Or does the `Table` extensions now require them to be added to the editor?

src/markdownit/index.js
```diff
@@ -33,6 +34,12 @@ const markdownit = MarkdownIt('commonmark', { html: false, breaks: false })
 	.use(markdownitMentions)
 	.use(implicitFigures)
 	.use(mathematics)
+	.use(multimdTable, {
+		multiline: true,
+		rowspan: false,
+		headerless: true,
```
max-nextcloud: The table schema still looks like it requires a header row ...

```js
export default Table.extend({
	content: 'tableCaption? tableHeadRow tableRow*',
    ...
})
```

Does it make sense to allow headerless tables here then, if we cannot represent them in our tiptap schema?

src/nodes/Table/markdown.ts
```diff
@@ -0,0 +1,202 @@
...
+	row: Row,
+	columnWidths: number[],
+) {
+	const normalizedCells = row.cells.map((cell, cellIdx) => {
+		// Append newline to single-line cells with a block node
+		if (
+			cell.lines.length === 1
+			&& [...cell.nodeTypes].some((type) => singleLineBlockNodeTypes.has(type))
+		) {
+			cell.lines.push('')
```
max-nextcloud: This will mutate the cell in `row.cells` during the `map` iteration. The way this function is used this is not a problem - but still unexpected.

I think it would be better to pick one strategy:
* Map the existing cells to a new set of cells leaving them unchanged.
* Mutate the cells in place.

src/nodes/Table/markdown.ts
```diff
@@ -0,0 +1,202 @@
...
+			} else if (cell.align === 'right') {
+				cell.lines[lineIdx] = line.padStart(columnWidths[cellIdx])
+			} else {
+				cell.lines[lineIdx] = line.padEnd(columnWidths[cellIdx])
+			}
+		})
+		return cell
+	})
+
+	// Write row line by line
```
max-nextcloud, line 103: I'd make this a separate function. Both in terms of what they do and which arguments they need they can be separated nicely. The part above does not touch `state` at all the part below has nothing to do with `columnWidths`.

src/nodes/Table/Table.js
```diff
@@ -5,7 +5,7 @@
 
 import { mergeAttributes } from '@tiptap/core'
 import { Table } from '@tiptap/extension-table'
-import { Node } from '@tiptap/pm/model'
+// import { Node } from '@tiptap/pm/model'
```
max-nextcloud:
```suggestion
```

cypress/e2e/nodes/Table.spec.js
mejo-: Good spot. I think that was an accident when I meant to alter `src/tests/nodes/Table.spec.js`. Reverted now.

src/markdownit/index.js
mejo-: I think I initially thought that it would still be good to allow importing headerless tables and I even added a test for it. But as this will result in changed markdown anyway, let's drop it for now. Done.

src/nodes/Table/Table.js
mejo-: Leftover, removed.

src/nodes/Table/markdown.ts
mejo-: You're right. I changed this to create a clone of `cell` at the beginning of the `map` iteration now and mutate and return this clone instead.

src/nodes/Table/markdown.ts
mejo-: I outsourced the `normalizeCells()` part into an own function now, hope that's what you had in mind: 12cb94a2246fd18d61fc2b6fe5c09c16f8532811

src/nodes/Table/TableCell.js
```diff
@@ -85,31 +58,80 @@ export default TableCell.extend({
...
+						console.warn('Nested tables are not supported')
+						alert(
+							t(
+								'text',
+								'A table was pasted into a table. Nested tables are not supported.',
+							),
+						)
+
+						const newSlice = new Slice(Fragment.empty, 0, 0)
+						return newSlice
```
max-nextcloud, line 134: Do we really need to rule out multi cell pastes?
Would that effectively lead to nested tables or just paste the cells at the current cursor position?

The latter seems like it might be useful and if it's already the current behavior would be nice to keep it.

mejo-: I gave it a try by selecting two cells and pasting them and it resulted in a full new table being added to the cell, i.e. a nested table. If that's what you mean (also what we talked about last week), then I fear that's not possible out of the box and not working already now.

Issue #6938, opened 2025-03-11
Drag handle for nodes


juliusknorr:
We should add a drag handle next to nodes so they can easily be moved around. This contributes to #1914 but should be discussed in more depth before implementing especially in regards with other buttons that we already have next to the nodes (link menu, plus smart picker button, AI button).

@marcoambrosini Is this something you can come up with some design thoughts?

cc @max-nextcloud as we talked about it as well

marcoambrosini:
Hi @juliusknorr, I think:
* The drag handle should be clickable and opening a menu that makes sense for the current block 
* The link preview action can be integrated in this menu
* The Assistant can also be integrated in this menu

JuliaKirschenheuter:
seems to be that we would need to have a `tiptap-pro` license for https://tiptap.dev/docs/editor/extensions/functionality/drag-handle and https://tiptap.dev/docs/editor/extensions/functionality/drag-handle-vue

JuliaKirschenheuter:
Hi @marcoambrosini, so far i've seen from documentation i'm not sure it will be possible to:

- > opening a menu that makes sense for the current block
- > The link preview action can be integrated in this menu
- > The Assistant can also be integrated in this menu

Drag handle is visible only for dragging a node around. I will make further research.

juliusknorr:
The pro license is not suitable for us as we need to make sure our shipped libraries are compatible with our own license, which that one is not. However we can of course implement our own drag handle extension or maybe there are other 3rdparty libraries that have a compatible license.

JuliaKirschenheuter:
https://github.com/nextcloud/text/pull/7029#pullrequestreview-2717806771

JuliaKirschenheuter:
Hi @marcoambrosini,

I would like to split this issue into smaller ones and integrate those implementations into text app. This feature is large and komplex and because of it i want to discuss with you further steps.

My idea would be to have first implementation of simple nodes: paragraphs, headings, lists. Then continue with images. Not sure about tables, because data structure there is really complex and would move it to the end.

Please look into current draft: https://github.com/nextcloud/text/pull/7274. What do you think which are few next most important steps? Thank you!

marcoambrosini:
Hi @JuliaKirschenheuter, wouldn't the tiptap functionality that's being open sourced already cover this?
I don't mind if you break this up into smaller prs, but I think that from a user's perspective, we have to release this all at once. If this cannot make it for the autumn release I would hide it behind a feature flag.

PR #7537, opened 2025-08-08
feat(files): add drag handler for nodes

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/6938

#### 🖼️ Screenshots

 🏡 After

Screenshot from 2025-08-08 11-10-30
Screenshot from 2025-08-08 11-10-21
Screenshot from 2025-08-08 11-10-14
Screenshot from 2025-08-08 11-10-09

✅ Drag handler works good in collaboration

⚠️ Existing bugs:

**It is not possible to drag table:**
Screenshot from 2025-08-04 11-29-56

**Once any type of content is moved to the code block it is not possible to move it out anymore**
Screenshot from 2025-08-01 17-30-14

**Once any type of content is moved to the list (ordered or unordered) it is not possible to move it out anymore**
Screenshot from 2025-08-01 16-45-20

**Once list is moved to another list (ordered or unordered) it is not possible to move out first list anymore**
Screenshot from 2025-08-01 16-46-07


nimishavijay (changes requested):
Looks really nice! :) Looked in the original issue and I saw that there were concerns about it being too prominent, so I have some feedback to help with that:
- [ ] use a tertiary button (so there is no background)
- [ ] use --color-maxcontrast for the icon instead of --color-main-text
- [ ] Ideally we use this standard [drag icon](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&icon.query=drag&selected=Material+Symbols+Outlined:drag_indicator:FILL@0;wght@400;GRAD@0;opsz@24), but I am not able to find it in the mdi library page 🤔 if it is not possible to use that, we can use this [drag-vertical icon](https://pictogrammers.com/library/mdi/icon/drag-vertical/). 
- [ ] If possible, we move the button a few pixels up such that it is center aligned with the text like this (in the screenshot I added a background to show the position, it should be a tertiary button) 

Image

Image

max-nextcloud:
I think some of the listed issues are due to the fact that the drag handle only allows dragging top level elements. If you drag an element into another element such as a list it's not a top level element anymore and therefore cannot be dragged anymore.

At first sight there seem to be two options to solve this:
1. Do not allow dragging content into other elements - only allow sorting.
2. Allow dragging subelements such as nested lists.

I don't know if 1. is feasible and it would of course reduce the usefulness of the feature - on the other hand it would also make it easier / more clear to use.

I don't know if 2. is feasible. It would be nice to be able to sort lists with this as well for example. But it's hard to destinguish if i want to drag the first list item or the entire list. I imagine there's more such questions this would raise due to the increased complexity.

marcoambrosini:
@max-nextcloud @JuliaKirschenheuter 
I think we should do both 1) and 2), as in: we should only allow sorting of blocks and sub blocks. Otherwise it gets a bit too messy.

max-nextcloud:
grafik
found some errors in my console that seem related.

JuliaKirschenheuter:
Hi @max-nextcloud, thanks a lot for your review!

>Undoing the drag and drop does not seem to work. The elements are not moved back - which would mitigate the 'I dragged something into another element and now it's stuck situation'. So maybe we can fix this (as a follow up as well).

That works for me. `Ctrl+Z` as well as `undo` button. Could you please check again?

JuliaKirschenheuter:
Hi @max-nextcloud 

warnings, which you have mentioned are coming from third library i can't fix them and they break nothing.

max-nextcloud:
> That works for me. `Ctrl+Z` as well as `undo` button. Could you please check again?

Last time i tried it worked. I will try again and try to come up with a reproducable case - but any way that would be a follow up.

max-nextcloud (approved):
Code looks good. Cypress failures also happened elsewhere.
I think the errors should still be fixed - but that can happen as a follow up / in the upstream repo.

PR #7555, opened 2025-08-19
fix/guest user dialog

max-nextcloud:
## Description

Two small fixes for the session list:
- **fix(css): render cursor color for logged in users**
- **fix(people): hide local client for guests**

And one code cleanup:
- **chore(move): useDelayedFlag to composables**


silverkszlo:
When I open a .md file in Files app and open the same file via public link in a private browser, the display of another editor/guest overlaps with another icon (in Firefox; but maybe that's intentional :*)):

Screenshot from 2025-08-21 17-25-53

silverkszlo:
In `SessionList.vue`:

```
data() {
  return {
	  myName: '',
  }
},
```
`myName` is not being used anymore or yet.

max-nextcloud:
> When I open a .md file in Files app and open the same file via public link in a private browser, the display of another editor/guest overlaps with another icon (in Firefox; but maybe that's intentional :*)):

Yes... it's intentional. We show up to three users plus the people Icon. I see there's room for improvement :wink: - but I also have no simple idea of what else to do.

max-nextcloud:
> `myName` is not being used anymore or yet.

Great catch!

Turns out it was not used since it was introduce in *drumroll* 2019. 
:see_no_evil:

marcoambrosini (commented):
Nice improvements @max-nextcloud. Dropping a few design related comments:
 
- [x] When the guest username is not defined, instead of the question mark I suggest we use a regular `user` icon 

* In "Guest sees admin and form
now without duplicate entry for themselves"
    - [x] I would not show an input field, instead I would show a list item with avatar, the string "you", and a right align secondary "edit" button with `pencil` icon. Clicking that would activate the input.
    - [x] The input field here should be horizontally aligned with the left icon and button, right now they're slightly misaligned
   
* In "Guest starts entering new name Now with NcInputField"
     - [x] Once in edit mode, the icon here could be hidden and the input field could expand to the left
    - [x] I think it would still be good to have a submit button in the same place as the edit one, otherwise it's a bit confusing. I'd make it also primary in this case and with the icon: `check`.

max-nextcloud:
> Dropping a few design related comments:

I think I addressed them all. Let me know if I misunderstood something or something else comes up

https://github.com/user-attachments/assets/45e9a71b-1ce3-400b-9da0-52fda74721b1

silverkszlo:
Also not related to this PR but `$userName` in `SessionService.php` does not seem to be used anymore or yet: 
`$userName = $this->userId ?? $guestName;`

lib/Service/SessionService.php
```diff
@@ -223,6 +223,15 @@ public function updateSessionAwareness(Session $session, string $message): Sessi
...
+		if ($this->userId === null) {
+			return $this->getColorForGuestName($guestName);
+		}
+		$name = $this->userManager->getDisplayName($this->userId) ?? $this->userId;
+		$color = $this->avatarManager->getAvatar($this->userId)->avatarBackgroundColor($name);
+		return sprintf('#%02x%02x%02x', $color->red(), $color->green(), $color->blue());
+	}
+
 	private function getColorForGuestName(string $guestName = ''): string {
 		$guestName = $this->userId !== null ? $guestName : '';
```
silverkszlo: This sets `$guestName` to empty string when user is NOT logged in (`userId` is null), but guests are exactly when you'd expect `userId` to be `null`, right?. This seems backwards. Guests never actually use their passed name. We might want to adjust it so that guest names are respected when no user is logged in, e.g.: 
```
$guestName = $this->userId !== null ? '' : $guestName;
```

lib/Service/SessionService.php
```diff
@@ -223,6 +223,15 @@ public function updateSessionAwareness(Session $session, string $message): Sessi
 		return $this->sessionMapper->update($session);
 	}
 
+	private function getColor(string $guestName = ''): string {
```
silverkszlo, line 225: These functions appear to return different color formats: `getColor()` returns a hex string (#rrggbb), while `getColorForGuestName()` returns color names like "red" or "blue". For consistency and easier debugging, it might be better if both functions returned the same format?

silverkszlo:
In `Editor.vue` seems to be a typo in `props`. Should be `required` instead of `require`:
```
props: {
  richWorkspace: {
	  type: Boolean,
	  require: false,
	  default: false,
},
```

src/components/Editor/AvatarWrapper.vue
```diff
@@ -33,7 +39,12 @@ export default {
 		},
 		size: {
 			type: Number,
-			default: () => 32,
+			default: () =>
```
silverkszlo: This value may not be available during SSR or before CSS is loaded, and accessing the DOM on every component instantiation is unnecessary. It's safer to check for `window` and `document` first and return a fallback value otherwise, e.g.:
```
if (typeof window === 'undefined' || !document.body) {
    return 32
} else {
    // CSS variable might not be loaded/defined yet
    const value = window.getComputedStyle(document.body)
        .getPropertyValue('--default-clickable-area')
    return value ? Number.parseInt(value) : 32
}
```

lib/Service/SessionService.php
max-nextcloud: Ah funny - that's why the guest colors keep changing even when the name stays the same! Good catch! :fishing_pole_and_fish:

src/components/Editor/GuestNameDialog.vue
```diff
@@ -4,110 +4,135 @@
...
+		required: true,
 	},
+})
+const emit = defineEmits(['update:session'])
+const { connection } = useConnection()
+const { editor } = useEditor()
+const { updateUser } = useEditorMethods(editor)
+const editing = ref(false)
+const loading = ref(false)
+const guestName = ref(props.session.guestName)
```
silverkszlo, line 70: It might make sense to add a watcher for `guestName`. Right now it's only initialized at the beginning but wouldn't change if the parent-prop gets changed from above. You could add: 
```
watch(
  () => props.session.guestName,
  (newName) => {
    if (!editing.value) {
      guestName.value = newName
    }
  }
)
```

lib/Service/SessionService.php
max-nextcloud: [Server code](https://github.com/nextcloud/server/blob/a7a78054aa9fafa7117134d040cb197e8cce5816/lib/public/Color.php#L81-L88) seems like both return the same. But still would be nice to use the same format here and `$color->name()` seems much simpler than reimplementing it.

src/components/Editor/GuestNameDialog.vue
```diff
@@ -4,110 +4,135 @@
...
+		showError(t('text', 'Not connected. Cannot update guest name.'))
+		return
+	}
+	const previousGuestName = props.session.guestName
+	loading.value = true
+	try {
+		const session = await update(guestName.value, connection.value)
+		loading.value = false
+		editing.value = false
+		localStorage.setItem('nick', guestName.value)
```
silverkszlo: Access to `localStorage` can fail in private browsing mode, due to storage limits, or if disabled by the browser. Might be better to also wrap it in try/catch to avoid uncaught exceptions.

src/components/Editor/GuestNameDialog.vue
```diff
@@ -4,110 +4,135 @@
...
-					showWarning(t('text', 'Failed to update the guest name.'))
-					this.guestName = previousGuestName
-				})
-		},
-		updateBufferedGuestName() {
-			this.guestNameBuffered = this.guestName
-		},
-		t,
-	},
+const setGuestName = async () => {
```
silverkszlo, line 79: It would be a good idea to validate the `guestName` input before proceeding. Right now, empty strings or strings with only whitespace could be submitted. It could improve UX and avoid unnecessary API calls. Something like:
```
if (!guestName.value || guestName.value.trim() === '') {
  showWarning(t('text', 'Please enter a name.'))
  return
}
```

src/components/Editor/SessionList.vue
```diff
@@ -80,54 +71,29 @@ export default {
...
-				.filter(
-					(session) =>
-						session.lastContact
-							> Date.now() / 1000 - COLLABORATOR_DISCONNECT_TIME
-						&& (session.userId !== null || session.guestName !== null),
-				)
-				.sort((a, b) => a.lastContact < b.lastContact)
+		remoteSessions() {
+			return this.sessions.filter(
+				(session) => session.id !== this.currentSession.id,
```
silverkszlo: Accessing `this.currentSession.id` could throw an error if `currentSession` is `undefined`. Consider guarding against that by writing:
`(session) => !this.currentSession || session.id !== this.currentSession.id`

silverkszlo:
Typos in `Status.vue`: should be `required` instead of `require`:
```
props: {
  hasConnectionIssue: {
	  type: Boolean,
	  require: true,
  },
  dirty: {
	  type: Boolean,
	  require: true,
},
```

src/components/Editor/AvatarWrapper.vue
max-nextcloud: We're only using one size right now - so I think I'll get rid of the prop all together.

silverkszlo:
Not directly related to this PR, but `lastSavedString()` in `Status.vue` could safely check whether `document` and `lastSavedVersionTime` exist before accessing them, to avoid potential runtime errors, e.g.
```
return this.document?.lastSavedVersionTime 
  ? moment(this.document.lastSavedVersionTime * 1000).fromNow() 
  : ''
```

src/components/Editor/GuestNameDialog.vue
max-nextcloud: My understanding is that right now an empty string effectively resets the guest name. So `you` is displayed in once own view and others see a translation of `Guest`.
I think it's good to be able to unset the guest name.

You're still raising an important point though. We did not verify the length before. Longer than 64 chars would cause 500 responses from the server and longer than ~ 15 chars would break the layout. I tried to address these two issues.

src/components/Editor/SessionList.vue
max-nextcloud: I think this may also do the trick, right?
```js
(session) => session.id !== this.currentSession?.id
 ```
 The right hand side would be undefined if currentSession is undefined - which would not filter anything as all sessions have ids.

File: src/components/Editor/GuestNameDialog.vue [hunk 8]
```diff
@@ -35,7 +36,9 @@
 				<template #icon>
 					<PencilOutlineIcon :size="20" />
 				</template>
-				{{ t('text', 'edit') }}
+				<template v-if="guestName.length < 12">
```
silverkszlo: cool! Additionally I would check if `guestName` is `null` or `undefined` before trying to access its `length` property. Either with `guestName?.length < 12` here or maybe even better to do it when initializing it with `const guestName = ref(props.session.guestName || '')`.

src/components/Editor/GuestNameDialog.vue
```diff
@@ -87,7 +87,11 @@ const setGuestName = async () => {
 		const session = await update(guestName.value, connection.value)
 		loading.value = false
 		editing.value = false
-		localStorage.setItem('nick', guestName.value)
+		try {
+			localStorage?.setItem('nick', guestName.value)
```
silverkszlo: I think optional chaining on `localStorage` doesn’t really help here – it’s always defined in the browser. The actual failures we need to handle are quota issues or disabled storage, which throw at `setItem`. Using `?.` would just skip the call if `localStorage` somehow wasn’t available, silently hiding problems, whereas `try/catch` actually surfaces real errors and lets us handle them properly.

src/components/Editor/SessionList.vue
silverkszlo: I let ChatGPT and Claude weigh in: your version is shorter and looks elegant, but it relies on the assumption that every session always has an ID. The longer, explicit version is supposedly more future-proof and self-explanatory—it clearly handles the “no current session” case and won’t break if session IDs ever end up null or undefined. So while the short version works today, the explicit one might be safer if the data structure changes down the line.

silverkszlo (approved):

src/components/Editor/GuestNameDialog.vue
max-nextcloud: Hmm... we require the session prop. The underlying problem here is that the guestName is optional for sessions - because normal users do not have it. But the guest name dialog is only for guests and they should always have a guest name - even if it's an empty string.
I will see if i can define the Session type in a way that allows expressing this so we do not need to be to defensive about what we can expect.

PR #7560, opened 2025-08-22
fix(readonly-bar): Clean up read-only menu bar to behave consistently with normal menu bar

pbirrer:
### 📝 Summary

This pull request addresses several issues related to read-only mode:
- The README.md viewer in the file app no longer displays the read-only menu bar.
- The read-only menu bar now spans the full viewport width.
- Improved mobile support.

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -23,7 +23,7 @@
 			@outline-toggled="outlineToggled">
 			<MainContainer v-if="contentLoaded">
 				<!-- Readonly -->
-				 <template v-if="readOnly || (openReadOnlyEnabled && !editMode)">
+				<template v-if="readOnly || (openReadOnlyEnabled && !editMode)">
 					<slot name="readonlyBar">
 						<ReadonlyBar
 							:is-hidden="hideMenu"
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -23,11 +23,11 @@
 			@outline-toggled="outlineToggled">
 			<MainContainer v-if="contentLoaded">
 				<!-- Readonly -->
-				<div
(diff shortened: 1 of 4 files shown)


src/components/Editor.vue
```diff
@@ -23,20 +23,17 @@
 			@outline-toggled="outlineToggled">
 			<MainContainer v-if="contentLoaded">
 				<!-- Readonly -->
-				<div
+				<ReadonlyBar
 					v-if="readOnly || (openReadOnlyEnabled && !editMode)"
-					class="text-editor--readonly-bar">
-					<slot name="readonlyBar">
```
max-nextcloud, line 29: This slot is used by the collectives app to alter the content of the read only bar. I think we need to keep it so that collectives can continue doing that.

max-nextcloud (changes requested):
Thanks a lot for looking into this. 💙 

I just had a brief look at the code and so far it looks good. There's just one aspect that needs to be addressed.

I'll try this out later.

src/components/Editor.vue
pbirrer: You're right, sorry about that! I readded the slot. However, it's now a direct child of the MainContainer component, which is the same setup as here:
https://github.com/nextcloud/text/blob/f4055471ef98671df2d0240e045357d282c3c50d/src/components/Editor/MarkdownContentEditor.vue#L11-L19

Do you think that could be an issue? Personally, I feel it’s better to let the ReadonlyBar handle its own CSS styling.

mejo- (approved):
Thanks so much for tackling this @pbirrer :pray:

I tested it both in standalone Text and Collectives and it looks much cleaner now.

PR #7599, opened 2025-09-01
fix(links): do not steal focus on initial loading

max-nextcloud:
The `LinkBubbleView` is created and then mounted by `createTooltip`
in the `update` function of the `LinkBubblePluginView`.

This already happens during the initial rendering
as that is the first time the editor content is updated.

At this time the `href` is null and therefore `startEdit` is called
which in turn sets the focus in the entry field.

Instead set `edit` to `true` right away.

The behavior of `mod-k` is still inconsistent.
In particular it will not focus the link entry
if the `href` already was falsy - for example right after the start.

### DIFF
--- a/src/components/Link/LinkBubbleView.vue
+++ b/src/components/Link/LinkBubbleView.vue
@@ -153,7 +153,7 @@ export default {
 	data() {
 		return {
 			isEditable: false,
-			edit: false,
+			edit: true,
 			newHref: null,
 			referenceTitle: null,
 		}
@@ -204,10 +204,6 @@ export default {
 		this.editor.on('update', ({ editor }) => {
 			this.isEditable = editor.isEditable
 		})
-	},
-
-	mounted() {
-		this.startEditIfEmpty()
 	},
 
 	methods: {


max-nextcloud:
This issue came up in [collectives cypress tests](https://github.com/nextcloud/collectives/actions/runs/17377263565?pr=1954). While changing the title on a collective page the focus was stolen and renaming the page to `New page` fails (even though cypress focusses the text entry again):
image

I've only been able to reproduce this behavior with cypress in electron. Other browsers seem to ignore the attempt of the hidden link bubble to steal the focus.

The goal of this PR is primarily to make cypress tests in collectives reliable again. I don't feel like diving deeper into this rabbit whole as would be required in order to fix the rest of the `mod-k` behavior.

**update**: [using this branch](https://github.com/nextcloud/collectives/pull/1954/commits/81ec783d7d5f3e0b30ce82a0068777552c2a07ff) the collectives tests [now pass](https://github.com/nextcloud/collectives/actions/runs/17385166959/job/49350805894?pr=1954).

Issue #7580, opened 2025-08-25
Close button does not show when opening readonly .txt or .md files in iOS and Android Nextcloud app


mpivchev:
**Describe the bug**
Close button does not show when opening readonly .txt or .md files in iOS or Android Nextcloud app, making it impossible to close the document or go back.

**To Reproduce**
1. Open iOS or Android Nextcloud app
2. Open a readonly .md or .txt
3. There is only a checkmark icon, but no close icon
4. Window cannot be closed, app is stuck

**Expected behavior**
Close button should be shown so the app can close the document.

**Screenshots**
No close button in readonly file

<img width="376" height="199" alt="Image" src="https://github.com/user-attachments/assets/6d198b96-e293-4713-8ee3-710a70c3ab36" />

Close button appears in editable file

<img width="386" height="159" alt="Image" src="https://github.com/user-attachments/assets/fee09adc-a7cd-4431-b230-e199613ec74b" />

**Server details:**
- Nextcloud version: [e.g. 26.0.1]
- PHP Version: [e.g. 8.2]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome]
 - Browser version: [e.g. 22]
 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

mpivchev:
```
public func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
        if message.name == "DirectEditingMobileInterface" {

            if message.body as? String == "close" {
                viewUnload()
            }

            if message.body as? String == "share" {
                NCDownloadAction.shared.openShare(viewController: self, metadata: metadata, page: .sharing)
            }

            if message.body as? String == "loading" {
                print("loading")
            }

            if message.body as? String == "loaded" {
                print("loaded")
            }

            if message.body as? String == "paste" {
                self.paste(self)
            }
        }
    }
```

This is how we catch the HTML body when rendering the content

we are looking for a button named "close", which then closes the view

andersravn:
Isnt this still an issue? I still can't see the close button on iOS in a read only file. I have no updates in the App Store.

[...]

PR #7624, opened 2025-09-09
fix: add close button when open readonly files on mobile app

luka-nextcloud:
### 📝 Summary

* Resolves: #7580

Add close button when open read-only files on mobile app

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -33,6 +33,7 @@
 								:dirty="dirty"
 								:sync-error="syncError"
 								:has-connection-issue="requireReconnect" />
+							<slot name="header" />
 						</ReadonlyBar>
 					</slot>
 				</template>


src/components/Editor.vue
```diff
@@ -33,6 +33,10 @@
 								:dirty="dirty"
 								:sync-error="syncError"
 								:has-connection-issue="requireReconnect" />
+							<button
```
mejo-: @luka-nextcloud any particular reason to not use `NcButton` here?

src/views/DirectEditing.vue
```diff
@@ -95,12 +54,7 @@ export default {
 			return JSON.parse(this.initial.session) || null
 		},
 		isMobile() {
```
mejo-:
```suggestion
		isMobileApp() {
```

Nitpicking, but since you named the vue method `isMobileApp()` in `Editor.vue`, let's name it this way here as well, no?

max-nextcloud:
#7560 seems related.

mejo-:
> https://github.com/nextcloud/text/pull/7560 seems related.

Ah wow, I missed that. It seems to address my second comment from above :pray: Will take a look.

mejo-:
> While testing this, I realized that the ReadOnlyBar doesn't display well on mobile (at the bottom), it's cut off and the border line is on the wrong side. If you could take a look into improving this and harmonizing it with how MenuBar is displayed on mobile, that would be awesome. If not we can also do it in a follow-up PR.

So this is no longer necessary as it got tackled in #7560 already :pray:

PR #7627, opened 2025-09-10
fix: temporary fix useEditor composable

luka-nextcloud:
### 📝 Summary

* Resolves: # 
Temporary fix the useEditor composable to make editor work in Deck card description

### 🏁 Checklist

### DIFF
--- a/src/composables/useEditor.ts
+++ b/src/composables/useEditor.ts
@@ -4,16 +4,17 @@
  */
 
 import { Editor } from '@tiptap/core'
-import { inject, type InjectionKey, provide } from 'vue'
+import { type InjectionKey } from 'vue'
+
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
It's following the temporary fix that was mentioned here https://github.com/nextcloud/text/pull/7381#issuecomment-3270805873

max-nextcloud:
I wonder if we could go back to options api provide in the components that currently call provideEditor. It's what we used before and I think it should be possible to provide something in the options api that is returned from the setup function.

*update:* I pushed what i mean in #7628 - could you give it a try @st3iny ?

max-nextcloud:
superseeded by #7643

Issue #3915, opened 2023-03-08
Backgroundjob cleanup is incomplete


juliushaertl:
There are cases where the cleanup background job leaves documents in the database as it currently only cleans up by inactive sessions.

We should:
- Find all documents without unsaved steps and active sessions and reset them
- Find all steps that have no corresponding document table entry anymore and clean them up (as the state is gone we wouldn't recover anyways)
- Find all sessions that have no corresponding document anymore and delete them

Issue #3740, opened 2023-02-02
Database leftovers


juliushaertl:
There might be database leftovers for deleted files that we should probably clean up with a repair step or cron job

Useful query for finding orphaned steps:

```
select document_id,path from oc_text_steps as s left join oc_filecache f on f.fileid = s.document_id where path IS NULL group by document_id;
```

PR #7653, opened 2025-09-19
Fix: Add cleanup for orphaned text steps

benjaminfrueh:
### 📝 Summary

The `oc_text_steps` table was growing significantly. One reason for that were orphaned steps accumulating in the database. Steps without existing sessions were never cleaned up by the existing cleanup job.

This addresses orphaned steps only. A separate PR will add garbage collection for very old text sessions to prevent long-term accumulation.

The steps are deleted in batches of 1000 and the execution is time-limited for 30 seconds.

Fixes: #3740
Fixes: #3915

### 🏁 Checklist


lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,42 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
 	}
 
+	public function deleteOrphanedSteps(): int {
+		$startTime = microtime(true);
+		$maxExecutionSeconds = 30;
+		$batchSize = 1000;
+		$deletedCount = 0;
+
+		do {
+			$orphanedStepsBuilder = $this->db->getQueryBuilder();
```
mejo-: Nitpicking, but so far we always used `$qb` for query builders in the code. Maybe for consistency reasons call it `$orphanedStepsQb`? I personly don't have strong feelings about the naming, but think consistency will help to understand the code for new contributors.

lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,42 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
+			$stepIds = array_map(function ($row) {
+				return (int)$row['id'];
+			}, $result->fetchAll());
+			$result->closeCursor();
+
+			if (empty($stepIds)) {
+				break;
+			}
+
+			$deleteBuilder = $this->db->getQueryBuilder();
```
mejo-: Same nitpicking comment about the variable name as above.

max-nextcloud:
Thanks a lot for tackling this @benjaminfrueh !

The wording here is confusing and we've been wanting to fix this for a while - but never got to it.
A `session` is a single users editing connection. It can be ephemeral. If the network is flaky the and the user disconnects for more then 30 seconds the client will reconnect and create a new session for example. If multiple users edit the same document there are multiple sessions at the same time.
The steps are basically capturing the changes a user made with a given session. Even if the session does not exist anymore they continue to be relevant as even if i am not editing anymore and my session has been cleaned up - the changes that I made should persist.

However every 30 seconds we autosave the document including the full yjs editing history (`docuementState`). Someone who starts to edit will only fetch the latest `documentState` and all steps that have been created afterwards. So for people joining steps with an id lower than the `lastSavedVersion` in the corresponding document are not needed anymore. However there may be people who are already connected who have not yet received all steps up to the `lastSavedVersion` so we cannot clean it up right away.

I'm sorry this is such a mess to begin with. It evolved over time and we did not manage to clean it up. I'm happy to talk it through with you to figure out which steps we can remove and how to clean this up for the future.

lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,42 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
+		$startTime = microtime(true);
+		$maxExecutionSeconds = 30;
+		$batchSize = 1000;
+		$deletedCount = 0;
+
+		do {
+			$orphanedStepsBuilder = $this->db->getQueryBuilder();
+			$orphanedStepsBuilder->select('st.id')
+				->from('text_steps', 'st')
+				->leftJoin('st', 'text_sessions', 's', $orphanedStepsBuilder->expr()->eq('st.session_id', 's.id'))
```
mejo-: Ouch, indeed as @max-nextcloud pointed out, this should rather only regard steps that have no corresponding document anymore in the `text_documents` table.

lib/Db/SessionMapper.php
benjaminfrueh: Thanks for the feeback about that. I was using the variable name convention from the SessionMapper method `deleteInactiveWithoutSteps`. Using $orphanedStepsQb seems like a good balance between descriptive and consistent code.

benjaminfrueh:
Hi @max-nextcloud and @mejo- , thank you for the feedback. I've pushed some changes.
- Only orphaned steps with a version (id) < `lastSavedVersion` are deleted, with a 24h safety buffer
- Steps with no existing document are deleted
- Steps with an existing session or version >= `lastSavedVersion` are not deleted
- Added Qb suffixes to the query builder variables

If you have any more suggestions or proposed changes for the cleanup I would be happy to discuss them with you in detail.

mejo-:
Points from discussion with @benjaminfrueh:

- Condition to keep steps of active sessions doesn't help for scenario when client comes back after long time. Steps from sessions that got closed in the meantime will be missing then as well.
- What happens if client comes back after a long time and asks for steps older than last one available on server? Code seems like I get an incomplete subset of steps (if I ask for steps > 10 and server has only steps starting from 15 onwards, I get steps 15+).
 - What's the consequence client-side? Does it reconnect because applying the steps results in the yjs state that indicates missing steps?
 - Shall we return an error server-side instead if asking for steps older than those that exist?

max-nextcloud:
@mejo- You are right. I was confusing what the query does. I was thinking it only removes steps for documents where no sessions exist anymore at all. But it would remove those from a session that stopped existing even if further sessions exist.

Right now the client would fetch the steps that it can get. It may not notice that steps are missing if there are no steps arriving that depend upon them. If there are steps missing that later steps depend upon the client will notice and attempt to recover the state by loading the saved `documentState` and all steps since then. However we have observed that that can lead to small parts of the content being duplicated. So I would not want to rely on it.

Detecting the missing step server side does not seem possible right now as we are just using the auto-incrementing id to retrieve newer steps. We used to have a version count that was per document and increased in steps of one - but I thought it was a clever idea to remove that :face_in_clouds: - which I've been regretting ever since :wink: . But now you cannot tell if some step is missing in between the version provided by the client and the latest step.

lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,51 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
+		$maxExecutionSeconds = 30;
+		$batchSize = 1000;
+		$deletedCount = 0;
+		$safetyBufferTime = time() - 86400;
+
+		do {
+			$orphanedStepsQb = $this->db->getQueryBuilder();
+			$orphanedStepsQb->select('st.id')
+				->from('text_steps', 'st')
+				->leftJoin('st', 'text_sessions', 's', $orphanedStepsQb->expr()->eq('st.session_id', 's.id'))
```
max-nextcloud:
```suggestion
				->leftJoin('st', 'text_sessions', 's', $orphanedStepsQb->expr()->eq('st.document_id', 's.document_id'))
```

I think this would clear steps only for documents which have no session anymore at all.

mejo-: Removing just steps of documents where no session exists anymore at all sounds like the right approach :+1:

max-nextcloud:
@mejo- I think your change request has been addressed, right?

mejo-:
@max-nextcloud agreed, let's get this merged. I guess it should also be backported, right @benjaminfrueh?

PR #7665, opened 2025-09-20
Fix: Delete old sessions in cleanup cron

benjaminfrueh:
### 📝 Summary

The oc_text_steps table was growing significantly. This adds a step to the cleanup cron which deletes sessions with their steps older than 90 days. The sessions are deleted in batches of 1000 and the execution is time-limited for 30 seconds.

### 🏁 Checklist

### DIFF
--- a/lib/Db/SessionMapper.php
+++ b/lib/Db/SessionMapper.php
@@ -150,10 +150,10 @@ class SessionMapper extends QBMapper {
 		$ageThreshold = time() - $ageInSeconds;
 
 		do {
-			$oldSessionsBuilder = $this->db->getQueryBuilder();
-			$result = $oldSessionsBuilder->select('id')
+			$oldSessionsQb = $this->db->getQueryBuilder();
+			$result = $oldSessionsQb->select('id')
 				->from('text_sessions')
-				->where($oldSessionsBuilder->expr()->lt('last_contact', $oldSessionsBuilder->createNamedParameter($ageThreshold)))
+				->where($oldSessionsQb->expr()->lt('last_contact', $oldSessionsQb->createNamedParameter($ageThreshold)))
 				->setMaxResults($batchSize)
 				->executeQuery();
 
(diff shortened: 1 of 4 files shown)


lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,48 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
 
+	public function deleteOldSessions(int $ageInSeconds): int {
+		$startTime = microtime(true);
+		$maxExecutionSeconds = 30;
+		$batchSize = 1000;
+		$deletedCount = 0;
+		$ageThreshold = time() - $ageInSeconds;
+
+		do {
+			$oldSessionsBuilder = $this->db->getQueryBuilder();
```
mejo-: Same nitpicking comment as in the other PR about the variable name :see_no_evil:

mejo- (changes requested):
Thanks for tackling this! @max-nextcloud and me discussed this PR and think it's good to merge (except for the nitpicking comment :laughing:).

Deleting sessions is not very intrusive as the frontend will merely suggest a reload to the user if the session got removed server-side.

The thread of data loss is very minor as that would be only editing steps that exist in the frontend and were not synchronized to the server yet. Since steps get send immediately and we currently set the editor to read-only once sending steps failed for some seconds, it can only be seconds of editing changes to the document.

lib/Db/SessionMapper.php
benjaminfrueh: Thanks @mejo- , it should be fixed in the new commit, I also added logic so sessions containing step ids >= last_saved_version are never deleted.

benjaminfrueh:
@mejo- and @max-nextcloud I did make some changes shortly before your comment and pushed them. I changed the variable names to be consistent and also for safety added logic so sessions containing step ids >= last_saved_version are never deleted.

Could you review those changes and let me know if there are further todos. I also adjusted the tests accordingly.

mejo-:
Thanks @benjaminfrueh, much appreciated!

> for safety added logic so sessions containing step ids >= last_saved_version are never deleted.

I don't think that's necessary and would be in favour of reverting this change. As @max-nextcloud explained in his comment to your other PR, sessions are bound to a specific client. They're more or less merely the authentication token created when a client joins an editing session.

Any client with an active session will autosave the document every thirty seconds, what incorporates all steps into the saved document and bumps `last_saved_version`. It's not a problem if steps from deleted sessions exist. Any client applies all steps, regardless from which session they come. So I don't see extra benefit in binding session cleanup to the fact that all steps from this session were applied and autosaved already. But maybe let's wait for the thoughts from @max-nextcloud on this :wink:

benjaminfrueh:
HI @mejo-, thanks again for the detailed feedback. That makes total sense, in this case it should be save to delete any session which is older than the defined threshhold of currently 90 days. I did revert the recent change and added a seperate commit for the variable name consistency.

benjaminfrueh:
@max-nextcloud missing sign-off for revert commit is added now

lib/Db/SessionMapper.php
```diff
@@ -142,6 +142,48 @@ public function deleteInactiveWithoutSteps(?int $documentId = null): int {
...
+			$sessionIds = array_map(function ($row) {
+				return (int)$row['id'];
+			}, $result->fetchAll());
+			$result->closeCursor();
+
+			if (empty($sessionIds)) {
+				break;
+			}
+
+			$deleteStepsQb = $this->db->getQueryBuilder();
```
mejo-: This delete query should be removed. As explained earlier, we don't want to delete the steps from sessions that got deleted. This task should really merely remove sessions older than 90 days as they will be recreated easily without data loss. Removing steps more or less unconditionally on the other hand might result in data loss.

benjaminfrueh: Okay, in the recent commit I removed the step deletion part from the deleteOldSessions function.

mejo- (approved):
Very nice, now all good to go 🎉

mejo-:
@benjaminfrueh except for the now failing test :see_no_evil: :laughing:

benjaminfrueh:
> @benjaminfrueh except for the now failing test 🙈 😆

Test is updated now, thanks again for the review!

PR #7677, opened 2025-09-25
feat: Save a checksum for documents and use it to detect conflicts

benjaminfrueh:
### 📝 Summary

Text sometimes throws a `DocumentSaveConflictException` when a files mtime/etag is updated without changing the content.

**Solution:**
Compute and store a CRC32 checksum for each document. On saving or syncing, the checksum is updated, and `assertNoOutsideConflict()` uses it to detect real content changes, preventing false conflicts.

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/conflict.spec.js
+++ b/cypress/e2e/conflict.spec.js
@@ -51,7 +51,7 @@ variants.forEach(function ({ fixture, mime }) {
 		})
 
 		it(prefix + ': displays conflicts', function () {
-			createConflict(fileName, mime)
+			createConflict(fileName, 'edited-' + fileName, mime)
 
 			cy.openFile(fileName)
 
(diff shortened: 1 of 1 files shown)


max-nextcloud (approved):
Thanks a lot for looking into this!

tobiasKaminsky (approved):
Tested with Android Notes app and works 🎉

juliusknorr:
@benjaminfrueh I found that it seems less likely but still sometimes happens on our instance that we see a conflict during saves.

I managed to see this if a save request happened to take longer  (2-3s) and i meanwhile continued typing:
image


Maybe there is a way to reproduce that locally with some manual delay? I'm wondering if we can prevent throwing conflicts during times another php process is currently handling a save

Issue #7692, opened 2025-09-28
Collaborative editing unreliable (out of sync state in some clients)


mejo-:
**Describe the bug**
Sometimes people still experience out-of-sync scenarios when editing collaboratively, meaning that some session members don't see the changes from other session members anymore.

Current best guess is, that our recovery process doesn't work in all scenarios. We detect out-of-sync state clientwise using yjs `pendingStructs`.
The client then asks the server using a Yjs query and the server replies with latest saved document state and all steps since then (`lastSavedVersion`). But when the last saved document state on the server is *from* the out-of-sync client itself, it will not get resolved.

**Server details:**
- Nextcloud version: 32

@JonathanTreffler

mejo-:
## Ideas for mitigation

- #7695 Change log level in case of `SyncStep1` response from `debug` to `error` [here](https://github.com/nextcloud/text/blob/849217238608e9b111fb9c9fa7c7aabe4097621e/lib/Service/DocumentService.php#L236). This way it's easier to detect server-side, how often the scenario happens.

### Problem 1: recovery via detected `pendingStructs` and `SyncStep1` message sometimes doesn't work

1. #7694 Save requests currently set`lastSavedVersion` to `stepVersion` server-side, which is wrong and can lead to missing steps in between document state and `lastSavedVersion`.
2. #7703 Brute force fix for missing step: when applying a `SyncStep1` response in client still leaves `pendingStructs`, ask iteratively for the last 200 steps before local `version` until we don't have `pendingStructs` left. This should e.g. fix scenarios where client1 misses a step but at the same time is the client that last (auto)saved the document, in which case the document state in `SyncStep2` server response would not contain the missing step.

### Problem 2: why steps get lost

We compared network requests and responses in a session where the problem occurred and it seems like the client successfully receives all steps. So current assumption is that it gets lost while being processed in the frontend code.

The code that increments `this.version` in `SyncService.ts` doesn't check if step really got applied to Yjs document. It increments the local `version` and then merely emits an event that gets picked up by `WebsocketPolyfill.js`. If either the event listener is not registered or it fails to apply the step in some way, the version still erroneously gets incremented. Idea to tackle this in a hotfix:
[...]

PR #7695, opened 2025-09-29
chore(DocumentService): Raise log level for SyncStep2 message to error

mejo-:
This happens if the client tries to recover from a missing step, which should not happen anyway. So let's log it as error to get a better understanding of how often this happens in practice.

Related: #7692

### 🏁 Checklist


juliusknorr:
Was reverted with https://github.com/nextcloud/text/pull/7738

PR #7704, opened 2025-09-30
fix(SaveService): Don't save if yjs reports pendingStructs

mejo-:
Yjs pendingStructs mean that the local yjs document state is missing a step. So let's not allow to save such an inconstent state to the server.

### Screenshot

| Disabled save button |
| --- |
| disabled_save |

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -801,6 +801,10 @@ export default defineComponent({
 			this.translateModal = false
 		},
 
+		hasYjsPendingStructs() {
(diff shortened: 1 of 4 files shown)


max-nextcloud:
Waiting for the outcome of our attempts to prevent out of sync state and thus pending structs in the first place.

mejo-:
@max-nextcloud I'd say let's close this PR for now as the situation seems to have improved via other changes. What do you think?

mejo-:
Let's close this PR for now.

PR #7719, opened 2025-10-06
Replace cypress component tests

max-nextcloud:
- **chore(test): move yjs test from cypress to vitest**
- **chore(test): move search test from cypress to vitest**
- **chore(cleanup): drop cypress component tests in ci**


src/tests/helpers/yjs.spec.ts
```diff
@@ -26,6 +27,7 @@ describe('Yjs base64 wrapped with our helpers', function () {
 
 		const stateA = getDocumentState(source)
 		const step0A = documentStateToStep(stateA)
+		console.log(step0A)
```
mejo-: looks like a debugging leftover?

src/tests/extensions/Search.spec.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { describe, expect, it } from 'vitest'
+import Search from '../../extensions/Search'
```
mejo-, line 7: Nitpicking, but so far we have the extension in most local imports. Maybe let's stick to one syntax for now.

src/tests/extensions/Search.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+import HardBreak from '../../nodes/HardBreak.js'
+import { searchDecorationsPluginKey } from '../../plugins/searchDecorations.js'
+import lorem from '../fixtures/lorem.txt?raw'
+import createCustomEditor from '../testHelpers/createCustomEditor'
```
mejo-, line 11: Same here.

max-nextcloud: image

typescript does not like other typescript files to be imported with an extension.

src/tests/extensions/Search.spec.ts
mejo-: Ok, seems like we already use the syntax without extension in some other places. Let's ignore this for now and move forward.

Issue #7544, opened 2025-08-07
Link preview state toggle is not rendered correctly


noantiq:
**Describe the bug**
The preview options toggle for links is not filled, no matter which option is selected.

**To Reproduce**
1. Edit a page and add a link in a new line
2. click on the three dot menu to change its preview state
3. Observe that none of the two options is filled (visually selected)
4. click on an option
5. observe that the option gets filled just before that menu closes and the change is applied to the link preview
6. Open the three dots menu again
7. There doesn't seem to be a selected option again

**Expected behavior**
The selected option is filled, so the state is visually reflected in the menu.

**Screenshots**

<img width="1498" height="436" alt="Image" src="https://github.com/user-attachments/assets/accd2eb4-3cb1-4667-a4f8-6657c0973d1d" />

**Server details:**
- Collectives app version: 3.0.3

mejo-:
Transferred the issue to the Text app, as it's a bug there.

PR #7750, opened 2025-10-09
fix(PreviewOptions): set active radio button on open

janbaum:
Fixes: #7544

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/PreviewOptions.vue
+++ b/src/components/Editor/PreviewOptions.vue
@@ -17,7 +17,7 @@
 				data-text-preview-option="text-only"
 				name="preview-option"
 				value="text-only"
-				:checked="type === 'text-only'"
+				:model-value="type"
 				@change="(e) => toggle(e.currentTarget.value)">
 				{{ t('text', 'Text only') }}
 			</NcActionRadio>
@@ -25,7 +25,7 @@
 				data-text-preview-option="link-preview"
 				name="preview-option"
 				value="link-preview"
-				:checked="type === 'link-preview'"
+				:model-value="type"
 				@change="(e) => toggle(e.currentTarget.value)">
 				{{ t('text', 'Show link preview') }}
 			</NcActionRadio>


max-nextcloud:
@janbaum Thanks a lot for your PR!

The linter is unhappy. I think the message is self explanatory:
image

Could you try and fix this?

After fixing you can confirm with `npm run lint` that your fix worked and did not trigger anything else.

Feel free to amend the existing commit (`git commit --amend --no-edit src/components/Editor/PreviewOptions.vue`) and force push it to your branch.


**update:** Just noticed that the error is maybe not so self-explanatory. So I will elaborate a bit.

`v-model` is a two way binding - it sets the value in the child component and when the child emits an event it changes it in the parent.

But here `type` is a prop and props are not expected to be changed. We already handle the `@change` event for the toggles. So instead of the two way binding you only want a one way binding.

The `checked` prop that we used before is deprecated:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/d747fe3358cdb0a84aa409686ebdd73595c32316/src/components/NcActionRadio/NcActionRadio.vue#L101-L119

Recommendation is to use `modelValue` instead - which is also what `v-model` uses under the hood for providing the value to the child component ( so the one way we still need ).

mejo-:
Thanks for the comments @max-nextcloud! We actually struggled with this problem while trying to find a proper fix. But you're right, as [documented here](https://v2.vuejs.org/v2/guide/components-props#One-Way-Data-Flow), component props in Vue.js must not be mutated inside the component.

So since we still have to use `v-model` for the `NcActionRadio` component, we probably have to introduce a local [component property](https://v2.vuejs.org/v2/api/#data) in the `data()` object of the component (e.g. `selectedType`) and set it to the value of the `type` component prop in `created()`.

Me and @janbaum actually already did this first while working on the problem yesterday, but I didn't like it because it left us with a duplicated data structure. But apparently that's the way to go then, no?

@janbaum could you imagine to update the PR accordingly?

max-nextcloud:
@mejo- I think the crucial part of your fix is:
* that `v-model` uses `modelValue` rather than `checked` internally.
* that you hand it a string rather than a boolean so it can do the comparison with the value itself.

I think you could still achieve both without using `v-model` by setting the `modelValue` prop to `type`. But I have not tried it though.

I also considered the `data` option you suggest and thought it was overly complex. But if that's the only way to go that's also fine with me.

janbaum:
@mejo- @max-nextcloud Thank you both for the hints, I'll try :)

max-nextcloud:
Nice! Confirmed it works. Thanks a lot! :heart:

max-nextcloud:
@janbaum The fix has been merged and backported to Nextcloud 32. The backport to 31 [failed](https://github.com/nextcloud/text/pull/7750#issuecomment-3396152608) and needs to be done by hand.

Would you have some time until Wednesday to give this a try? Otherwise we'll take care of it to make sure it lands in the next patch release.

janbaum:
> @janbaum The fix has been merged and backported to Nextcloud 32. The backport to 31 [failed](https://github.com/nextcloud/text/pull/7750#issuecomment-3396152608) and needs to be done by hand.
> 
> Would you have some time until Wednesday to give this a try? Otherwise we'll take care of it to make sure it lands in the next patch release.

Tbh I am not sure if its correct what I did :D

https://github.com/nextcloud/text/pull/7781

max-nextcloud:
> Tbh I am not sure if its correct what I did :D

Thanks a lot for giving it a try! mejo already got back to you. If you need further advice feel free to mention me in the new PR.

PR #7783, opened 2025-10-14
fix(sync): Reset document sessions on upgrade

### LINKED ISSUE #7692: Collaborative editing unreliable (out of sync state in some clients)

mejo-:
A bug in synchronizing the document state might have resulted in a Yjs document state with missing steps being persisted into the Ydoc that we store on the server. If no client that had the missing step did an autosave afterwards, future sync session members will never recover from this situation.

We implemented several improvements recently that should result in less out-of-sync situations where a step becomes missing. To start with fresh editing sessions and not carry on with possibly unrecoverable document states, let's reset all editing sessions with the next app update.

See #7692 for further context.

Resolved #7692

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -15,7 +15,7 @@
 - **💾 Open format:** Files are saved as [Markdown](https://en.wikipedia.org/wiki/Markdown), so you can edit them from any other text app too.
 - **✊ Strong foundation:** We use [🐈 tiptap](https://tiptap.scrumpy.io) which is based on [🦉 ProseMirror](https://prosemirror.net) – huge thanks to them!
 	]]></description>
-	<version>7.0.0-dev.1</version>
+	<version>7.0.0-dev.2</version>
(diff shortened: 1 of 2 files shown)


max-nextcloud:
We already discussed this and the code changes are minimal. Would like to wait for a review from @juliusknorr

max-nextcloud (approved):

PR #7787, opened 2025-10-15
fix(sync): send first update without initial document state

max-nextcloud:
Process initial document state as incoming message.
This way it will be tracked as received from the remote server by y-websocket
and the update send to the server will exclude it.

Also use the sync services `opened` event inside WebsocketPolyfill
rather than mixing async `syncService.open().then` and event handling.
This allows us to emit `opened` and then `sync` from `syncService.open()`.

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -211,7 +211,6 @@ class DocumentService {
 		$stepsToInsert = [];
 		$stepsIncludeQuery = false;
 		$documentState = null;
-		$newVersion = $version;
 		foreach ($steps as $step) {
 			$message = YjsMessage::fromBase64($step);
 			if ($readOnly && $message->isUpdate()) {
@@ -228,7 +227,7 @@ class DocumentService {
 			if ($readOnly) {
 				throw new NotPermittedException('Read-only client tries to push steps with changes');
 			}
-			$newVersion = $this->insertSteps($document, $session, $stepsToInsert);
+			$this->insertSteps($document, $session, $stepsToInsert);
 		}
 
 		// By default, send all steps the user has not received yet.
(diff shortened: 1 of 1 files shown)


max-nextcloud:
Cypress failures are due to https://github.com/nextcloud/server/pull/55726

lib/Service/DocumentService.php
```diff
@@ -265,7 +265,7 @@ public function addStep(Document $document, Session $session, array $steps, int
 
 		return [
 			'steps' => $stepsToReturn,
-			'version' => $newVersion,
+			'version' => isset($documentState) ? $document->getLastSavedVersion() : 0,
```
juliusknorr, line 267: Can you tell if this has any other side effects? We now no longer return the latest version (number of steps in the db) but just the last saved. We used this in the past to compare and see if the document had unsaved changes, but I think that part was changed already.

juliusknorr: Otherwise `$newVersion` is no longer used as a variable

src/services/SyncService.ts
```diff
@@ -238,15 +248,12 @@ class SyncService {
 		})
 			.then((response) => {
 				this.#outbox.clearSentData(sendable)
-				const { steps, documentState } = response.data as {
+				const { steps, documentState, version } = response.data as {
 					steps: Step[]
 					documentState: string
```
juliusknorr, line 256:
```suggestion
					documentState: string
					version: number
```

src/services/SyncService.ts
```diff
@@ -198,6 +201,13 @@ class SyncService {
 		}
 	}
 
+	_emitDocumentStateStep(documentState: string, version: number) {
```
juliusknorr, line 207: This is typescript so we could do `private emitDocumentStateStep` as far as I see we mostly used `_` as a convention for private methods

juliusknorr (commented):
General approach makes sense but only looked at the code, not tested

lib/Service/DocumentService.php
max-nextcloud: Good catch wrt `$newVersion`. We were not using the version in the response on the client side anymore.

The unsaved changes check is on the server when cleaning up docs, right? That should still work. this is not changing the data structures there.

src/services/SyncService.ts
max-nextcloud: Yes. We do have `_emitError` already and as we'd want to backport this I thought I stick to the existing style.

lib/Service/DocumentService.php
juliusknorr: Right there we still have it, i just remembered that we also used it in the frontend in the early days

Issue #7573, opened 2025-08-21
‘Add folder description’ should not appear if ‘Show folder description’ is unchecked in the Files settings.


Jerome-Herbinet:
‘Add folder description’ should not appear if ‘Show folder description’ is unchecked in the Files settings, to consistant from a UI/UX POV. I don't see the value in keeping this option if the setting is unchecked.

### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #7314, opened 2025-06-12
[Bug]: Hide "new folder description" button on a upload only share link (file request link)


julien21110:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [x] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

It seems this button appeared with nextcloud 31.
With an upload only shared link, this button have no sense : the user cannot edit any readme file in a file request link.
If the user clicks on this button, he fall in a no mans land.
Moreover, if he try again clicking the button, he see the read me file (through a request file only link...)

### Steps to reproduce

1. Be on nextcloud 31
2. Create a file request link / upload only public link
3. Click on the add folder description button

![Image](https://github.com/user-attachments/assets/1fc25a88-eea0-4f86-af30-b23dae7e5167)

### Expected behavior

Button not available with upload only link.

### Nextcloud Server version

31

### Operating system

RHEL/CentOS

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 31 to 32)

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [x] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```json

```

### List of activated Apps

```shell

```

### Nextcloud Signing status

```shell

```

### Nextcloud Logs

```json

```

janbaum:
I had a look at this and have a question.

First, I could reproduce the issue as described. After once reproducing it, I don't get the "Add folder description" anymore but "Create new file". I couldn't figure out why. Cleared browser cache, deleted the created folder and created a new one, but since the first time repsoducing the issue, I only get "Create new file" 🤷 

For the "Add folder description":
- click it
[...]

PR #7788, opened 2025-10-15
fix(files.js): check,ether "show folder descr." is set or not

janbaum:
### 📝 Summary

* Resolves: #7573

With this PR, the option 'Add folder description' is not more visible, if 'Show folder description' is turned off.
The user has to reload the page though, since the function, that checks the state of this setting, is somehow only called when the site gets reloaded.

### 🚧 TODO

- [ ] check value of `"files-setting-richworkspace"` (FilesSettings.vue) when '+New' menu is opened ?

### 🏁 Checklist

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -31,6 +31,9 @@ export const addMenuRichWorkspace = () => {
 		displayName: t('text', 'Add folder description'),
 		category: NewMenuEntryCategory.Other,
 		enabled(context) {
+			if (!window?.OCA?.Text?.RichWorkspaceEnabled) {
+				return false
+			}
 			if (Number(context.attributes['rich-workspace-file'])) {
 				return false
 			}


janbaum:
> The user has to reload the page though, since the function, that checks the state of this setting, is somehow only called when the site gets reloaded.

I dont fully understand why, this is just an observation...

This also results in some weired behaviour of some edge cases, e.g:

1. setting is enabled
2. disable seeting and close settings with upper-right `X`
3. open `+New` and click 'Add folder description' -> this will create a `Readme.md`
4. repeat step 3
5. observe, that option 'Add folder description' is still available, but when you click, it fails, because the file already exists

So it should actually also be prevented, that the option shows off, when a `Readme.md` already exists, right?

This also works the other way around, because I am pretty convinced that the problem is in the inital state, that is not updated, other than from reloading

mejo-:
@janbaum if you're motivated, you could look into https://github.com/nextcloud/text/issues/7314 as well, which is much related to this issue.

Issue #7693, opened 2025-09-28
read-only link sharing of .md should not allow changing links from text to link-preview mode


alexanderdd:
**To Reproduce**
1. create a .md file with a link like [example](https://example.com/)
2. share the file externally via link, but with read-only permission
3. access the link from different browser/computer
4. click the three dots to the left of the link, change it from "text" to "preview"
5. go back to logged-in user and refresh: the preview mode of the link was changed

**Expected behavior**
in read-only mode, nothing in the document should be allowed to change, also not the way links are displayed

**Server details:**
- Nextcloud version: "31.0.8"

let me know if you need more details

Issue #6814, opened 2024-12-23
Link menu button in lists


mejo-:
**Describe the bug**
When a list entry contains only a link, the three-dot menu button to switch between text and preview is displayed left to it. The alignment is off and overlaps the menu bullet point:

![Image](https://github.com/user-attachments/assets/ab7f5c4a-9482-470a-870e-e2ce138bbfa3)

But probably we should not show the menu at all in lists, as switching a link to preview breaks the entire list.

**To Reproduce**
1. Add a standalone text link inside a link item
2. See the broken alignment of the three-dot menu button
3. Opne three-dot menu and select 'preview'
4. See broken list

**Expected behavior**
Either disable the menu to switch between text and preview link inside lists, or fix both showing previews in lists and alignment of the button in lists.

**Server details:**
- Nextcloud version: 29

janbaum:
I noticed somethink else, but was unsure is this is related or should be in a new issue:

if you create a list item and insert a link:

<img width="755" height="142" alt="Image" src="https://github.com/user-attachments/assets/9facf897-9d8a-48ce-af56-f51fe96b7e8b" />

Now say you want to move the list one list item lower, to place something else above the link, in the very same list. For that you move the curser to the beginning of the link _(right before the `https`)_.

If you now press enter, to move the list item one further below, the corresponding PreviewOption _(3 dots)_ doesn't move along, but it stays at the upper bulletpoint:

<img width="735" height="82" alt="Image" src="https://github.com/user-attachments/assets/a32961e8-1054-4c9a-9fec-0aec9e7dd008" />

And funny side note, the link preview doesn't work anymore in this state:

<img width="821" height="282" alt="Image" src="https://github.com/user-attachments/assets/26732dd5-3ad5-4292-96bd-b22edfdafe53" />

Issue #7272, opened 2025-05-27
Buttons outside editor (smart picker, assistent, link settings) invisible in mobile and embedded editors


mejo-:
**Describe the bug**
We have several UI elements positioned outside the editor: assistant button on the right, link settings three-dot-menu on the left and smart picker button also on the left.

All these buttons are invisible and thus inaccessible on mobile view as the editor takes full width there.

Same problem exists when the text editor gets embedded in a container that has 670px width or less - as is currently the case in Collectives. While we can fix this in Collectives by providing more horizontal space to the parent container, we will still run into this problem again when text gets embedded in other places that don't have so much space available.

## Discussed next steps (from discussion during company week 09/2025)

* [ ] Harmonise visibility of smart picker plus button and drag handle (either follow cursor or mouse)
* [x] #7793 
  * For text-only links: move options into link bubble
  * For link previews: move menu to top right corner of the preview
* [ ] Move drag handle and smart picker buttons closer to text after three-dot menu vanished

marcoambrosini:
Hi @mejo- 
I would put all this functionality in buttons at the bottom of the screen in a toolbar. Right above the mobile keyboard.

mejo-:
We discussed this issue during the contributors week. Here's the notes:

The basic issue is that we have, on the left, from left to right: a plus button, a drag handle, and a 3-dots-hamburger menu and the assistant button.

- The plus follows cursor on desktop? (Was on cursor at mobile because mobile has no hover)
- AI button prominence maybe not so important any more? Can the button be demoted to the topbar?
- On desktop, maybe one context menu? But it was removed historically for a11y reasons.
- Talked about whether the three-dots were actually necessary, possible strategies to remove it.
- Option: move options into the current bubble menu, or add an inline button after the link text.
- Question: should clicking on the text open the link it points to? Or the bubble (current behaviour)?
-  Decent consensus: Easy first step to move the three-dots stuff into the bubble.
- Move assistant button into the toolbar (as high priority so it doesn't get hidden)
- Another low-hanging fruit: Max: move the + inline with text, somewhere between left-aligned and centre-aligned, only visible when the cursor is hovering an empty para. Marco disagrees.
- Low hanging fruit: make drag handle full height of node
- Julia: selection-based contextual tool tray thingy would be a good endgame
  - Good consensus, but more longterm
- Drag handle should be closest to text. maybe make + behave the same for now

PR #7793, opened 2025-10-16
fix(PreviewOptions): Move three-dot menu into preview box or link bubble

mejo-:
Fixes several UX bugs that stem from the existence of the three-dot menu for link preview options. It also will allow us to move drag handle and smart picker link closer to the editor container.

Contributes to #7272 

* Fixes: #6814 (broken alignment of preview options in lists)
* Fixes: #7693 (preview options displayed in read-only documents)
* Fixes: nextcloud/collectives#1920 (hidden preview options on narrow screens)

Also fixes the bug that the link is displayed above the preview in read-only view in Collectives.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image
image | image

### 🏁 Checklist


src/nodes/Preview.vue
```diff
@@ -57,13 +91,11 @@ export default {
...
-/* Top align with preview image:
- * - unset center alignment
- * - margin-top of widget-default
- */
-:deep(.preview-options-container) {
-	top: unset;
-	transform: unset;
-	margin-top: calc(var(--default-grid-baseline, 4px) * 3) !important;
+// Align in upper right corner of preview image
+.link-preview-header {
```
benjaminfrueh, line 95: Hi @mejo- thanks for this change, I reviewed and tested it, feels great and also looks much better now.

One minor issue, when the link title is long, it is very difficult to click the three-dot menu. It only works in a tiny area at the very top-right corner, because this is positioned absolute and behind the title. If this three-dot would have a higher z-index here, then this would also feel good to click.

Screenshot from 2025-10-21 11-50-09

mejo-:
> One minor issue, when the Link title is long, it is very difficult to click the three dot menu. It only works in a tiny area at the very top-right corner, because this is positioned absolute and behind the title. If this three-dot would have a higher z-index here, then this would also feel good to click.

Great catch! Updated now and I also added a `background-color` so the button is more visible.

benjaminfrueh (approved):

Issue #7604, opened 2025-09-03
Buttons left of text input are cut off on smaller window sizes / mobile


noantiq:
**Describe the bug**
It's much more noticeable in Deck and Collectives where I already created issues (https://github.com/nextcloud/deck/issues/7185, https://github.com/nextcloud/collectives/issues/1920), but also in a standalone text document the buttons left of the input (plus button and three dots button next to links) are cut off on smaller screen sizes.

**To Reproduce**
1. Open any markdown file on mobile
2. notice that there is no plus button
3. insert a link
4. notice that there is no three dots menu next to the link
---
1. Open any markdown file on desktop with the window in fullscreen
2. notice that there is a plus sign next to the text input area
3. insert a link
4. notice that there is a three dots menu next to the link
5. slowly decrease the window size
6. notice that first the plus button and later the three dots menu are cut off as they are simply pushed out of the visible area

**Expected behavior**
No buttons are cut off or hidden, I can still use the plus button and the three dots menu

**Screenshots**

<img width="1572" height="1378" alt="Image" src="https://github.com/user-attachments/assets/22c6ff01-f76a-4bea-a673-435959476ea5" />

<img width="1430" height="1378" alt="Image" src="https://github.com/user-attachments/assets/7124d9c6-94da-4f23-81f8-d5c32a9f6361" />

<img width="1190" height="1374" alt="Image" src="https://github.com/user-attachments/assets/2b377aae-a1a3-4ba0-9aac-62768915381b" />

**Server details:**
- Nextcloud version: 31.0.8 (Text 5.0.0)

PR #7818, opened 2025-10-22
fix(FloatingButtons): group smartpicker button and drag handle together

mejo-:
- The button group follows the cursor now
- The button group is top-aligned to hovered paragraph node
- The drag handle button has cursor type `grab`
- Not displayed on mobile and full width view

Also removes block margin from link previews. We already have margin between paragraphs and it makes the floating buttons look off otherwise.

Fixes: #7272
Fixes: #7604

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

#### 🖼️ Screencast

| Screencast |
| --- |
| recording1 |

### 🏁 Checklist


src/components/Editor/FloatingButtons.vue
```diff
@@ -0,0 +1,112 @@
...
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<DragHandle
+		:editor="editor"
+		class="floating-buttons"
+		:class="{ heading: isHeadingNode }"
+		:on-node-change="onNodeChange">
+		<NcButton
```
juliusknorr, line 12: Seems a bit weird to have the plus button wrapped in the drag handle itself. I guess that is a bit of a shortcut to have a more straight forward implementation. Only "issue" i could see is that you can use the plus button as a drag handle as well. Not a blocker for me though

benjaminfrueh (approved):

src/components/Editor/FloatingButtons.vue
mejo-: Hehe, first I thought the same. Then I realized that it's exactly what the "notion style" Tiptap example does it as well :zany_face: 

I didn't see a good reason to not do it and actually it results in cleaner code, as I can reuse the `node` and `pos` returned by the drag handler callback for the plus button action.

Issue #7874, opened 2025-10-30
Remove suggestions buttons from text preview


JuliaKirschenheuter:
**To reproduce:**

1. Create README.md withing text app
2. Leave it empty
3. Close a file and open a files overview

**How it is:**

<img width="1280" height="394" alt="Image" src="https://github.com/user-attachments/assets/05689476-c05a-402e-b1d6-51a9b4e9adc7" />

**How it should be:**

<img width="2053" height="632" alt="Image" src="https://github.com/user-attachments/assets/753c8b70-114e-402b-b341-36ff6a09469b" />

PR #7875, opened 2025-10-30
fix(files): Disable suggestions bar output for not rich workspace

JuliaKirschenheuter:
### 📝 Summary

* Resolves: https://github.com/nextcloud/text/issues/7874

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -56,7 +56,8 @@
 					v-show="contentLoaded"
 					ref="contentWrapper"
 					:read-only="!editMode" />
-				<SuggestionsBar v-if="isRichEditor && contentLoaded" />
+				<SuggestionsBar
+					v-if="isRichEditor && contentLoaded && !isRichWorkspace" />
 			</MainContainer>
 			<Reader
 				v-if="isResolvingConflict"


juliusknorr:
@marcoambrosini I know we removed the placeholder text, but now nothing is indicating to the user that they can click in there and type something if it is empty and not in focus.

marcoambrosini:
@juliusknorr what about re-introducing the placeholder? Why was it taken out? If accessibility is the concern, I think we could solve it by having an action that  focuses the `textarea` within the `new` menu. It could read: "Add folder description"

mejo-:
> @juliusknorr what about re-introducing the placeholder? Why was it taken out? If accessibility is the concern, I think we could solve it by having an action that focuses the textarea within the new menu. It could read: "Add folder description"

It was removed in https://github.com/nextcloud/text/pull/6855 when the smart picker button was added to the side. I also think we should bring it back, not only in empty rich workspace documents, but in general.

max-nextcloud:
> Why was it taken out?

* It seemed redundant to write... "Type `/` to add" when there's a plus button right next to it.
* Not everyone can mention people (I think it's not possible on public shares). So the "`@` to mention" part would have needed to become context dependent.
* The placeholder we used did not just show on an empty document but also in empty lines, which seemed broken to me as placeholders usually go away when something is typed into the textfield and do not show again on new emtpy lines.

marcoambrosini:
@max-nextcloud to address those points: 
1) I think It's still nice a nice invitation and it inform users about the existence of the keyboard shortcut
2) We can skip the @ mention part
3) This is just a bug that can be fixed

max-nextcloud:
Sounds good to me. I wonder if it might be easier to use the approach of the suggestion bar of checking for an empty editor for the placeholder as well.
Previously we used  the tiptap placeholder extension if i remember correctly. Fixing that seems more involved.

juliusknorr:
https://tiptap.dev/docs/editor/extensions/functionality/placeholder Does behave as I would expect to only render if the document is fully empty by using css rules as in the example

```
 /* Placeholder (at the top) */
  p.is-editor-empty:first-child::before {
    color: var(--gray-4);
    content: attr(data-placeholder);
    float: left;
    height: 0;
    pointer-events: none;
  }
```

mejo-:
I checked Notion and Coda, they both show the placeholder on each new paragraph. I wonder why we would want it only on empty files, but don't have strong feelings about it.

@JuliaKirschenheuter, do you plan to bring back the placeholder as part of this PR? Or would you prefer if someone else takes over? I could do so if you like.

marcoambrosini:
> I wonder why we would want it only on empty files

I think it's fine to show the placeholder text on each new line (not the suggestion bar though)

mejo-:
Let's bring this in. I opened https://github.com/nextcloud/text/pull/8075 for bringing back the placeholder.

PR #7922, opened 2025-11-12
Cleanup API and Document service

max-nextcloud:
- ApiService#create() with getOrCreateDocument()
- More strict types on save function
- enh(logging): Warn if a state file is missing


max-nextcloud:
autoloader check is also failing on main and seems unrelated. So I will merge ahead.

PR #7956, opened 2025-11-18
Convert versions spec to playwright

max-nextcloud:
- add versions spec
- versions with ViewerSection
- move close() into ViewerSection
- convert version spec from cy to playwright


max-nextcloud:
Cypress is failing because their download server responds with a 500. Seems unrelated.

Issue #6935, opened 2023-10-27
Support Latex Math commands 


giannistolou:
Hello :wave:,

I started using Nextcloud and Collectives a few weeks ago, and I love it. It's incredibly user-friendly, and the markdown support is fantastic. However, I've encountered a critical problem because it doesn't support LaTeX and mathematical symbols. If you work or study in a scientific field, you will need this support to write and share mathematical symbols.

I'm wondering if you can add LaTeX support to help everyone who uses it. This is such an important feature for education, mathematicians and more.

Unfortunately, I don't know much about the infrastructure of Collectives and Nextcloud, but I'd like to suggest some libraries that can work on the frontend using the $$ math symbol $$ or $ math symbol $: [KaTeX](https://katex.org/) and [MathJax](https://www.mathjax.org/).

I've also noticed a similar [issue](https://github.com/nextcloud/notes/issues/345) in the Notes app.

henrysheehy:
This is essential. KaTeX is favourable as it renders vectorised equations, rather than raster copies like Mathjax.

juliusknorr:
There is actually https://tiptap.dev/docs/editor/extensions/nodes/mathematics these days, so this could be a good first contribution to integrate this extension into text. The important part in addition to integrating the extension is that the node needs to be extended to be able to have a markdown representation with a `toMarkdown`. For loading existing files we need markdown it to handle transform this syntax to html which can then again be picked up by the tiptap node in the `parseHTML` function. A good sample of how this is done with other nodes is the callout node type.

https://www.npmjs.com/package/@iktakahiro/markdown-it-katex#block may be a good option for the markdown representation.

PR #7957, opened 2025-11-18
Add LaTex Support

moodyjmz:
### 📝 Summary

* Resolves: #6935 [](https://github.com/nextcloud/text/issues/6935)

This PR adds LaTeX math support to the text app.

it uses:

- Inline typing `$…$`
- Block typing `$$…$$`
- Pasting of inline
- Pasting of blocks
- Editing of both - I did consider using the same approach as inline links, the bubble, but decided to use the modal approach for both
- I added a feature into the menu to change existing content into maths - under the markup section

Code:
- I used a factory for making math nodes to avoid duplicating code
- It adds a modal, following the pattern of Image

Tests:
It adds some tests covering the features added, and issues I encountered while adding the feature

Limitations:

The existing modal doesn’t appear to be resizable - therefore here is set to largest

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|image
image|image
image|image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Math/ShowMathModal.vue
+++ b/src/components/Math/ShowMathModal.vue
@@ -1,3 +1,8 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
 <template>
 	<NcDialog
 		v-if="show"
--- a/src/css/extensions.scss
+++ b/src/css/extensions.scss
@@ -1,1 +1,6 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
 @import 'katex/dist/katex.min.css';
--- a/src/nodes/MathematicsView.vue
+++ b/src/nodes/MathematicsView.vue
@@ -1,3 +1,8 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
 <template>
 	<NodeViewWrapper :as="isBlock ? 'div' : 'span'" :class="wrapperClass">
 		<span ref="mathEl" @click="onMathClick"> </span>


moodyjmz:
I wanted to see if it would push the expected files - not ready for review yet

moodyjmz:
@max-nextcloud Sorry about the confusion, hopefully this PR is better to check. It only has one commit as I took the files from the other branch.

README.md
```diff
@@ -19,9 +19,28 @@
 - **🙋 Work together:** Share and collaborate with friends and colleagues, no matter if they use Nextcloud or not!
 - **💾 Open format:** Files are saved as [Markdown](https://en.wikipedia.org/wiki/Markdown), so you can edit them from any other text app too.
 - **✊ Strong foundation:** We use [🐈 tiptap](https://tiptap.dev) which is based on [🦉 ProseMirror](https://prosemirror.net) – huge thanks to them!
+- **🧮 Math support:** Write mathematical formulas using LaTeX syntax with inline (`$E=mc^2$`) and display (`$$...$$`) math rendering powered by KaTeX.
 
 Nextcloud Text is the default text editor since Nextcloud 17. To start editing just open an existing markdown or plaintext file or create a new one.
 
+### Mathematical formulas
```
juliusknorr, line 26: So far we do not really have much end user documentation. I'd say we could keep it in here for now still. We probably want this to go to docs.nextcloud.com but just the math section might not be that useful.

@max-nextcloud @mejo- Something we could discuss further as you were thinking about enhancing documentation for collectives, but there may be some overlap with what we could make use of for text itself.

package.json
```diff
@@ -124,6 +127,7 @@
     "prettier-plugin-organize-imports": "^4.3.0",
     "prosemirror-test-builder": "^1.1.1",
     "rollup-plugin-webpack-stats": "^2.1.7",
+    "sass-embedded": "^1.93.3",
```
juliusknorr: Could you elaborate what this is used for? From the package description I do not see why this would be needed

juliusknorr (commented):
Very nice, I left one inline comment and there are two CI checks that need some more work to pass.

When testing the modal seems to have an issue with not taking up the full height of the textarea, but I'm unsure if that is a component issue or could be addressed here as well. Would be fine to track as a follow up for polishing:

Screenshot 2025-11-18 at 20 54 36

juliusknorr:
Also cc @marcoambrosini for a design review

package.json
moodyjmz: Was being nosey and forgot I added it, I removed it.

moodyjmz:
> Very nice, I left one inline comment and there are two CI checks that need some more work to pass.
> 
> When testing the modal seems to have an issue with not taking up the full height of the textarea, but I'm unsure if that is a component issue or could be addressed here as well. Would be fine to track as a follow up for polishing:
> 
> Screenshot 2025-11-18 at 20 54 36

Thanks :) 

I addressed those issues.

Yes, the modal, I was unsure what caused it, but felt it was out of scope to address in this PR.

mejo-:
@moodyjmz thanks a lot for the contribution, much appreciated :heart: 

I just reviewed the PR (was on vacation last week) and realized that it adds `@tiptap/extension-mathematics` as NPM dependency but doesn't actually use it. I guess that it's a leftover from your development process? Also, looking at the upstream `@tiptap/extension-mathematics` code I wondered what was the reason why you decided to implement the whole mathematics node type yourself instead of using and extending the upstream `@tiptap/extension-mathematics` one :blush:

moodyjmz:
> @moodyjmz thanks a lot for the contribution, much appreciated ❤️
> 
> I just reviewed the PR (was on vacation last week) and realized that it adds `@tiptap/extension-mathematics` as NPM dependency but doesn't actually use it. I guess that it's a leftover from your development process? Also, looking at the upstream `@tiptap/extension-mathematics` code I wondered what was the reason why you decided to implement the whole mathematics node type yourself instead of using and extending the upstream `@tiptap/extension-mathematics` one 😊

@mejo- You are right about the npm dependency - that shouldn't be there. I looked at callouts.js for guidance and followed that. I'm not sure how many changes would be needed to use the tiptap approach, although that is likely a better implementation and reduces maintenance overhead. I can see, looking at other files, that tiptap extensions are extended from elsewhere.

moodyjmz:
@mejo- Removing the dependency is easy. However, I would say it should be used instead. However this also raises the question of backward compatibility  - if this feature is implemented differently in terms of processing to/from markdown than what is already there - this needs to be determined. Based on that, then the amount of changes can be determined. Is this feature already in the in the wild?

mejo-:
@moodyjmz thanks a lot for considering my concerns. I agree that we should use the upstream Tiptap extension and expand on it in case we need extra functionality. The `toMarkdown()` functions need to be added for sure, and if we still want/need the custom NodeView representation, then the NodeView vue file as well.

I just opened https://github.com/nextcloud/text/issues/7977 to track follow-up topics for this PR. If you don't mind, we can move the conversation there.

PR #7962, opened 2025-11-19
Fix: initialize undomanager for empty document

benjaminfrueh:
The first paste action in an empty document was not being tracked by the Yjs UndoManager, making it impossible to undo and not tracking it as a change.

This adds initUndoManagerForEmptyDocument() which primes the undo manager by inserting and deleting a space, then clearing the undo stack. This ensures user actions are properly tracked.


src/components/Editor.vue
```diff
@@ -485,6 +485,28 @@ export default defineComponent({
 			this.idle = false
 		},
 
+		initUndoManagerForEmptyDocument() {
+			if (
```
juliusknorr, line 489: Is this something that gets added to the history and pushing steps to the server? As discussed I'd be fine with this as a workaround but we should file an upstream issue at least.

benjaminfrueh: @juliusknorr I clear the history after deleting the space, and in my test there were no steps added to the database in this process. I agree that an upstream issue for yjs should be created and this workaround should be removed once the issue is fixed or a better workaround is found.

juliusknorr: Ok, would be fine with me then.

However tests seem to fail quite heavily. Maybe it has some unexpected side effects?

benjaminfrueh: Some failing tests, such as the ones counting characters seem to be timing-related and caused by this change. Making the tests even more flaky with this workaround doesn't seem ideal. Maybe we should discuss and explore an alternative less invasive workaround for this.

benjaminfrueh:
After more detailed debugging of the undoManager, I found the root cause and also was able to find a solution. Closing this PR and creating a new one with a fix.

benjaminfrueh:
Fixed in https://github.com/nextcloud/text/pull/8063

Issue #7963, opened 2025-11-20
i18n: Need context and grammar needs to be fixed


rakekniven:
https://github.com/nextcloud/text/blob/b35d0e95f2e4ece58d6377867b853da55b0db6d2/src/components/Math/ShowMathModal.vue#L11

Translators did not get the meaning

rakekniven:
@moodyjmz please enlight us

moodyjmz:
Display is for the block variant of the latex math equation


Inline is for the inline version

rakekniven:
Please add this information to the vue files.

See https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#vue for details.
Without this information, I doubt that good translations can be produced.

moodyjmz:
> Please add this information to the vue files.
> 
> See https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#vue for details. Without this information, I doubt that good translations can be produced.

Ah cool, thanks for the pointer - I will do that

moodyjmz:
I added a PR for this. As per the PR, I wasn't able to get the notes next to the content as it triggered the linter - happy to move them if possible.

moodyjmz:
I moved the notes per PR feedback, so hopefully this will be resolved soon.

PR #7965, opened 2025-11-21
:globe_with_meridians: i10n Add notes for maths dialog translations

moodyjmz:
### 📝 Summary

* Resolves: #7963

Added notes for translators to better understand the context of what they are translating

> ~~I have added the notes slightly above as I got linter warnings when I added them next to the content as the content is in a tenary~~

> I also changed the text from Display >>> Block as it better matches how the content is displayed, ie inline or block

#### 🖼️ Screenshots
N/A

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Math/ShowMathModal.vue
+++ b/src/components/Math/ShowMathModal.vue
@@ -4,11 +4,7 @@
 -->
 
 <template>
-	<NcDialog
-		v-if="show"
-		:name="dialogTitle"
-		size="large"
-		@closing="$emit('close')">
+	<NcDialog v-if="show" :name="dialogTitle" size="large" @closing="$emit('close')">
 		<NcTextArea
 			v-model="localLatex"
 			:label="t('text', 'LaTeX formula')"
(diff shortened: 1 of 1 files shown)


rakekniven:
I do not think this will work and the text will be displayed at Transifex.
@nickvergessen Your opinion?

src/components/Math/ShowMathModal.vue
```diff
@@ -4,6 +4,10 @@
 -->
 
 <template>
+	<!-- TRANSLATORS LaTeX formula, LaTeX is a system to display complex maths and scientific notation/formulae -->
+	<!-- TRANSLATORS There are two ways to show these formulae, Inline, or Display -->
+	<!-- TRANSLATORS Inline means it appears with other content, eg "something [the formula] something" -->
+	<!-- TRANSLATORS This is the title of a modal dialog to edit a LaTeX formula -->
```
nickvergessen: I don't think this works.
The translation strings need to be moved to computed or other direct js code in that case

moodyjmz: I have changed it - thanks for the suggestion

mejo-: @nickvergessen @rakekniven is the PR good to go from your side now?

src/components/Math/ShowMathModal.vue
```diff
@@ -62,6 +58,18 @@ export default {
...
+	computed: {
+		dialogTitle() {
+			// TRANSLATORS LaTeX formula, LaTeX is a system to display complex maths and scientific notation/formulae
+			// TRANSLATORS There are two ways to show these formulae, Inline, or Block
+			// TRANSLATORS Inline means it appears with other content, eg "something [the formula] something"
+			// TRANSLATORS Block means it appears on its own, akin to a paragraph
+			// TRANSLATORS This is the title of a modal dialog to edit a LaTeX formula
+			return this.isBlock
+				? t('text', 'Edit block formula')
+				: t('text', 'Edit inline formula')
```
nickvergessen: Needs to be a single line, otherwise the line above the translation is no comment:
```suggestion
			return this.isBlock ? t('text', 'Edit block formula') : t('text', 'Edit inline formula')
```

My real suggestion to be more clear would be:

```suggestion
			if (this.isBlock) {
				// TRANSLATORS LaTeX formula, LaTeX is a system to display complex maths and scientific notation/formulae
				// TRANSLATORS There are two ways to show these formulae, Inline, or Block
				// TRANSLATORS Inline means it appears with other content, eg "something [the formula] something"
				// TRANSLATORS Block means it appears on its own, akin to a paragraph
				// TRANSLATORS This is the title of a modal dialog to edit a LaTeX formula
				return t('text', 'Edit block formula')
			}
			return t('text', 'Edit inline formula')
```

Results in:
Image

But that should also make it clear to you, that you need to duplicate it for both strings, if the hint should be shown on both.

moodyjmz: @nickvergessen That's good - thanks for the screenshot of how translators see it - that's useful. I have made the changes you suggested.

nickvergessen: Looks good now:
grafik

Issue #7977, opened 2025-11-24
Bugs with new mathematics node types


mejo-:
**Describe the bug**

Follow-up issue for #7957.

- Inline formulas enforce a newline at their end, which is unexpected for inline nodes.
- When inserting an inline formula at the end of the document, it seems impossible to position the cursor after the node, so impossible to insert content behind it.
- When inserting a block type formula and placing the cursor in the next line, typing will remove the added formula.
- In menubar, "Display formula" seems a bit unclear to me as node type. Maybe "block math formula" or just "block math" would be better? Similarly, "inline formula" could be changed to "inline math formula" or just "inline math".
- In menubar, "Inline formula" is placed in the block node submenu, next to blockquotes, codeblocks and callouts. I would not expect it there. Not sure though what would be the best alternative place 🤔 
- Probably we should use and expand https://tiptap.dev/docs/editor/extensions/nodes/mathematics instead of re-implementing it all ourselves.
- Simple playwright tests to add both node types would be great.

moodyjmz:
I think some of these bugs might be addressed by using the tiptap mathematics extension

moodyjmz:
I think changing the menu names first and deciding where the feature lives is one step.

Migrating over to the tiptap extension is another.

moodyjmz:
@mejo- Do we need to consider backward compat?

It looks like tiptap just needs `div[data-type="block-math"]` to be picked up (and inline too) without content matching etc

So we can do:

```js
         parseHTML() {                                                                                                     │
              return [                                                                                                        │
              // Standard Tiptap format                                                                                     │
              { tag: 'div[data-type="block-math"]' },
}
```

Otherwise, we need to pick up the katex block as well

mejo-:
> @mejo- Do we need to consider backward compat?

Sorry I totally forgot to comment on that. Nope, we don't. So far, the feature is only merged into the development branch of Text (`main`) and no new major release got branched off from there yet. The release of Nextcloud 33 (next major release) is scheduled for 2026-02-18. Before this date, there's no need to consider backwards (in)compatibility.

moodyjmz:
@mejo- awesome!

moodyjmz:
I think I would change to use tiptap, then leave the renaming for another PR - I imagine where it sits in menus involved more people. @mejo-

mejo-:
> I think I would change to use tiptap, then leave the renaming for another PR - I imagine where it sits in menus involved more people. @mejo-

[...]

PR #7978, opened 2025-11-24
Update maths to use tiptap vs custom implementation for latex

moodyjmz:
### 📝 Summary

* Resolves: #7977

Summary

  Migrated the mathematics nodes to use the official @tiptap/extension-mathematics package instead of our custom
  implementation. This brings us in line with Tiptap while maintaining all existing functionality.

  Changes Made

  Core Migration

  - Extended Tiptap's math nodes instead of building from scratch (src/nodes/Mathematics.js)
    - Used .extend() pattern to customize BlockMath and InlineMath nodes
    - Added VueNodeViewRenderer to keep our Vue component UI
    - Followed the same pattern as CodeBlock and Image nodes

  Backward Compatibility

  - Added parseHTML rules to handle KaTeX HTML from markdown-it
    - When loading markdown files, `$E=mc^2$` gets converted to KaTeX HTML by markdown-it
    - Our custom rules extract the LaTeX from <annotation> tags so it parses correctly
    - This ensures existing markdown files with math formulas load properly

  Bug Fixes

  - Fixed inline math pushing to new line (src/nodes/MathematicsView.vue)
    - Removed space from `<span ref="mathEl"> </span>` that was causing line breaks
    - Cleaned up modal rendering with v-if instead of wrapper div
    - Inline formulas now stay inline as expected

  Menu Updates

  - Updated menu commands to use new Tiptap signatures (src/components/Menu/entries.js)
    - Changed node names: `math_inline` >>> `inlineMath`, `math_block` >>> `blockMath`
    - Updated command calls to use object parameters: insertInlineMath({ latex })
    - Fixed menu actions to access editor state correctly

  Tests

  - Updated all test expectations (src/tests/nodes/Mathematics.spec.js)
    - Changed node name assertions to match new Tiptap names
    - Updated command calls to use new signatures
    - All 20 tests passing ✅

  What Still Works

  - Modal editing with auto-open for empty formulas
  - Both inline and block math rendering
  - Markdown roundtrip (load and save)
  - All existing UI/UX unchanged from user perspective

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/Menu/entries.js
+++ b/src/components/Menu/entries.js
@@ -355,9 +355,12 @@ export const MenuEntries = [
 				label: t('text', 'Inline math'),
 				icon: Sigma,
 				isActive: 'inlineMath',
-				action: (command, editor) => {
-					const { from, to, empty } = editor.state.selection
-					const latex = empty ? '' : editor.state.doc.textBetween(from, to)
+				action: (command, editor = null) => {
+					let latex = ''
+					if (editor) {
+						const { from, to, empty } = editor.state.selection
+						latex = empty ? '' : editor.state.doc.textBetween(from, to)
+					}
 					return command.insertInlineMath({ latex })
 				},
 			},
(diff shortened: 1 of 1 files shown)


moodyjmz:
@mejo- I think this should address the issues you mentioned, sans the inline menu question

mejo- (approved):
Tested and works really well now!

Thanks so much for going the extra mile to address all the issues @moodyjmz :pray:

I pushed a small commit to make prettier happy.

mejo-:
For some reason some playwright tests fail reliably now :thinking:

mejo-:
> For some reason some playwright tests fail reliably now 🤔

Found the reason and pushed a fix. the `editor` property to `action` callback in menu is not always passed, so the callback needs to take this into account.

When CI runs through now I will merge. Thanks again @moodyjmz for the contribution, much appreciated :pray:

moodyjmz:
That's great! Thanks for finding the issue and fixing it

PR #7998, opened 2025-12-01
test: Use example.org instead of nextcloud.com for testing links

mejo-:
CI often runs into rate limits when testing against nextcloud.com and we're asked to not use nextcloud.com in tests anyway.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/npm-audit-fix.yml
+++ b/.github/workflows/npm-audit-fix.yml
@@ -59,13 +59,12 @@ jobs:
         id: npm-audit
         uses: nextcloud-libraries/npm-audit-action@1b1728b2b4a7a78d69de65608efcf4db0e3e42d0 # v0.2.0
 
-      - name: Run npm ci and npm run build
+      - name: Run npm ci
         if: steps.checkout.outcome == 'success'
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-audit-fix.yml
```diff
@@ -59,13 +59,12 @@ jobs:
...
         uses: nextcloud-libraries/npm-audit-action@1b1728b2b4a7a78d69de65608efcf4db0e3e42d0 # v0.2.0
 
-      - name: Run npm ci and npm run build
+      - name: Run npm ci
         if: steps.checkout.outcome == 'success'
         env:
           CYPRESS_INSTALL_BINARY: 0
         run: |
           npm ci
-          npm run build --if-present
```
max-nextcloud, line 68: I think the idea here is to make sure it still builds after the npm audit update. But I guess the same would also happen in the PR created below.

mejo-: We had this removed in the past as well, see c253378969e08dc10727653547ed91f590fad507 :wink:

PR #8001, opened 2025-12-01
chore(test): format text with playwright

max-nextcloud:
## Summary

Migrating some more tests to playwright.


max-nextcloud:
Print spec looks like somehow the table width shrank
grafik

mejo-:
> Print spec looks like somehow the table width shrank 

The table width in edit mode is smaller than in readonly mode nowadays, as the edit buttons don't exist in readonly mode and we use the available extra space there. That was requested by the design team some months back and I implemented it.

max-nextcloud:
> The table width in edit mode is smaller than in readonly mode nowadays

Nice! That explains it as i just made the share editable - but maybe we should not apply this in print view as no buttons are visible there either.

max-nextcloud:
> I wonder whether the cypress test "Initializes the workspace without errors" is necessary as I would expect the workspace to not be visible if an error was thrown, but maybe I'm missing something there.

We are logging some exceptions rather than throwing them. I think this is a regression test for a fix that removed one of these errors.

PR #8002, opened 2025-12-01
feat: Add table-only editor API

silverkszlo:
### 📝 Summary

* Related to and needed for: https://github.com/nextcloud/whiteboard/pull/816

Add createTable() API to enable embedding markdown tables in other apps without rich text editing features. Introduces PlainTableDocument node that restricts content to tables only, and PlainTable extension that bundles required functionality (table editing, markdown serialization).

- Add PlainTableDocument node with table-only content model
- Add PlainTable extension (bundles Markdown, EditableTable, etc.)
- Add PlainTableContentEditor.vue component
- Expose window.OCA.Text.createTable() API in editor.js
- Export PlainTable from extensions/index.js

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/PlainTableContentEditor.vue
+++ b/src/components/Editor/PlainTableContentEditor.vue
@@ -4,9 +4,7 @@
 -->
 
 <template>
-	<Wrapper
-		:content-loaded="true"
-		:show-outline-outside="false">
+	<Wrapper :content-loaded="true" :show-outline-outside="false">
 		<MainContainer>
 			<ContentContainer :read-only="readOnly" />
 		</MainContainer>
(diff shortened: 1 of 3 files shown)


src/extensions/PlainTable.js
```diff
@@ -0,0 +1,33 @@
...
+import Text from '@tiptap/extension-text'
+
+export default Extension.create({
+	name: 'PlainTable',
+
+	addOptions() {
+		return {
+			...this.parent?.(),
+		}
+	},
```
max-nextcloud: I think you do not need this for two reasons:
* Since you are creating a new extension i don't think parent is defined.
* You are not adding any options. So even if you were extending another extension there would be no need to overwrite `addOptions`.

max-nextcloud (commented):
Looks good but I have not tested it.

Since you are not using collaboration might be possible to write a playwright component test for this that calls `createTable()` for a div or so. That would make testing it easier.

src/components/Editor/PlainTableContentEditor.vue
```diff
@@ -0,0 +1,142 @@
...
+	props: {
+		content: {
+			type: String,
+			required: true,
+		},
+		readOnly: {
+			type: Boolean,
+			default: false,
+		},
+		showOutlineOutside: {
```
juliusknorr: I think we can remove the outline option for the table editor

juliusknorr (approved):

Issue #8031, opened 2025-12-10
Many warnings in the browser due to missing aria label


susnux:
**Describe the bug**
There are many warnings in the browser because there are buttons which have no accessible name.
All interactive elements need either a visual name (e.g. the button text) or an accessible aria name.

In this component all buttons only have a title but no label - neither visible nor aria:
https://github.com/nextcloud/text/blob/f8fb432627db6b9216d10ab331c9f62c402b434e/src/components/Editor/FloatingButtons.vue

**Expected behavior**
Either visible text or set the same as you have as `title` as `aria-label` (basically put it for both title and aria-label).

**Server details:**
- Nextcloud version: 33 and 32

PR #8032, opened 2025-12-10
fix(FloatingButtons): Add missing `aria-label` to NcButton components

mejo-:
Fixes: #8031

### 🏁 Checklist


max-nextcloud:
I tried to reproduce the problem in main branch but could not see the warnings so could not test this.

susnux:
@max-nextcloud the funny thing is I did not see this in the text app but in the tables app which seems to have a text integration?

susnux:
(anyways missing aria labels when there is no visual text are always a reason to fail WCAG so good that this is fixed now :) )

Issue #7954, opened 2025-11-16
Code snippets with long line cause overflow in full width view


adamkelly86:
**Describe the bug**
When in full width mode, code snippets seem to cause overflow. 
In Nextcloud 31, this did not occur. 

**To Reproduce**
1. Go to Collectives
2. Create a new page
3. Turn on full width view
4. Add a code block which will overflow the full width 

The same also happens in a single markdown file. The overflow also seems to add extra padding on the left of the text.

**Expected behavior**
Code block should wrap or be scrollable without causing the rest of the document to be out of the view

**Screenshots**
Full Width - Non Overflow
<img width="1920" height="929" alt="Image" src="https://github.com/user-attachments/assets/25b6732f-f940-4a8b-9b18-03a948715158" />

Full Width - Overflow
<img width="1920" height="921" alt="Image" src="https://github.com/user-attachments/assets/8c500b3f-fda3-41a3-b08f-6cd8dd6a6b9a" />

Markdown file in viewer
<img width="955" height="992" alt="Image" src="https://github.com/user-attachments/assets/35070559-1c20-49d5-97a6-ad4dd3e44733" />

**Server details:**
- Collectives app version: 3.3.0
- Nextcloud version: Nextcloud Hub 25 Autumn (32.0.1) - AIO instance
- PHP Version: 8.3.27
- Database: PostgreSQL 17.6 on x86_64-pc-linux-musl, compiled by gcc (Alpine 14.2.0) 14.2.0, 64-bit 

**Client details:**
 - OS: Windows, Ubuntu
 - Browser: Firefox, Chrome
 - Browser version: 
Firefox Browser 145.0 (64-bit)
Chrome Version 142.0.7444.162 (Official Build) (64-bit)

 - Device: Desktop (100% zoom on 1920x1080 resolution)

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)

[nextcloud.log](https://github.com/user-attachments/files/23571224/nextcloud.log)

#### Browser log

[nextcloud-browser.log](https://github.com/user-attachments/files/23571226/nextcloud-browser.log)

</details>

max-nextcloud:
This also seems to be an issue with the full width view in the text app itself (limited width is fine).

So I will move the issue to the text repository and add some details to the description.

Issue #7522, opened 2025-07-29
Outline / table of contents overflow into notes field


andreas-1107:
The outline overflows into the notes field.

## Steps to reproduce
1. On desktop, create a new note
2. Add four levels of headings with medium-long titles
3. Enable the outline by ctrl/command + alt/option + H
4. See the outline flow into the notes field

### Expected behaviour
The outline either cuts the titles in the outline at the border of the outline bar, or it wraps the text

### Actual behaviour
The outline flows beyond the border of the outline bar into the notes field.

### Screenshots

<img width="1091" height="462" alt="Image" src="https://github.com/user-attachments/assets/8c80be0b-5620-4a53-80da-b29346344bdf" />

## Server
Please complete the following information.

- Notes app version: 4.12.2
- Nextcloud version: AIO 31.0.7
- OS: Mac
- Web server: default AIO 31.0.7
- PHP version: default AIO 31.0.7
- Database: default AIO 31.0.7

**Nextcloud configuration:**

No special config, using defaults of AIO 31.0.7

## Client
Please complete the following information.

- Browser (incl. version): Firefox 141.0 (aarch64)
 - OS: Mac OS Version 15.5 (24F74)

## Log files

No log

Issue #7228, opened 2025-05-13
Markdown table will breake the whole design, if wide-text is enabled and you make the table wide enough to have a scroll bar


fabiscafe:
**Describe the bug**
If you manage to create a scrollbar on the markdown table, while in wide-text mode it'll break the whole side design. The table will "walk" to the right with every inch you widen the table

**To Reproduce**
1. Create a new markdown file
2. Create a table
3. Write to header 1-3 this `Lorem ipsum dolor sit amet, consecteturadipiscingelit,seddoeiusmodtemporincididuntutlabore`
4. If that doesnt create the issue for you, add a 4th, 5th, 6th, …  column with the same text

**Expected behavior**
It should add a scrollbar and not break the design

**Screenshots**
Normal, while not having text wide enough to make it scroll
![Image](https://github.com/user-attachments/assets/8d1e7d7f-383a-4947-8399-076a378774e6)

with enough room for a long scrollbar
![Image](https://github.com/user-attachments/assets/044681a5-107f-4dba-b31f-9f7fc5840848)

**Server details:**
- Nextcloud version: 31.0.4
- PHP Version: 8.3.21
- Database: Postgres 16.9

**Client details:**
 - OS: Debian
 - Browser: Firefox
 - Browser version: 132
 - Device: Desktop

mejo-:
This seems to be related to `.content-wrapper` container using `display: grid` and having a `<table>` element inside.

Issue #6876, opened 2025-01-27
Make the index in other views apart from the shared link


SamuXzX:
**Is your feature request related to a problem? Please describe.**
When viewing a Markdown file using a share link generated in Nextcloud, it appears a button to show an index of the file in a sidebar, generated from the headings:

![Image](https://github.com/user-attachments/assets/80922e9e-baf2-4848-abc5-83c49a0d7225)
(The button says "Hide index")

I don't get why this does not appear when editing the file from Nextcloud itself:

![Image](https://github.com/user-attachments/assets/cec2812a-be95-4b8a-9d16-f8b18d727218)

Another way of viewing the file is opening it directly from the system, with the Nextcloud Desktop Client installed:

![Image](https://github.com/user-attachments/assets/c24c2492-fe5e-4f3d-af97-dfcb7b235031)
(The button says "Open in browser")

![Image](https://github.com/user-attachments/assets/831e2ff7-e822-4a79-aea1-752bc050da90)

**Describe the solution you'd like**
I would like to be able to see such index in any view in Nextcloud.
One problem seems to be that the toolbar is full, but this seems like an entirely different problem (which I think needs to be solved, as being constrained like this could be a problem for Text).

The button could just appear at the left side of the top bar, just like the buttons for the sidebar are aligned to the right:

![Image](https://github.com/user-attachments/assets/e8656b0c-d264-4a86-b4ca-aa8823342448)

There would obviously be a problem with small screens, but maybe both this new "Show index" button and the buttons for the sidebar should receive their own row when in mobile view:

![Image](https://github.com/user-attachments/assets/38cf4a94-1940-44c3-8bb9-81913f546406)
I understand this would take more screen space, but if we want to have all the options in-view, for now I see no other solution.

mejo-:
Thanks for your report @SamuXzX. What you describe is the read-only mode of the text editor, where the toolbar is hidden and the button to toggle the outline is displayed instead. In read-write mode you find the same option at the bottom of the headings menu in the toolbar ("Show outline"). Does this solve the issue for you?

SamuXzX:
Thank you, it does in fact clarify it. I suppose the button is there due to a lack of space in the toolbar, as it is the most obvious place if someone has to fit it into the a submenu. I would like to change the title of the request: maybe we should think of a better place for that button, outside of the toolbarm as it is not an action to format or add something. The same would apply to the three-dot button at the end of the toolbar, which contains the guide and the character count:

[...]

PR #8033, opened 2025-12-11
Rework outline / table of contents

mejo-:
### 📝 Summary

See https://github.com/nextcloud/text/issues/7741#issuecomment-3607495936

* Resolves: #3586 
* Resolves: #5379
* Resolves: #6876
* Resolves: #7228 
* Resolves: #7522 
* Resolves: #7954 
* Resolves: nextcloud/collectives#1624 
* Resolves: nextcloud/collectives#1715 
* Resolves: nextcloud/collectives#1777

## Screenshots desktop

Outline | Outline zoomed | Table of contents | Table of contents zoomed
---|--- | --- | ---
image | image | image | image

## Screenshots mobile

Menu | Table of contents
--- | ---
image | image

## Screencast desktop

https://github.com/user-attachments/assets/c7c41ca9-c5fa-4296-9054-f027894d26c8

## Screencast mobile

https://github.com/user-attachments/assets/1cce768f-430d-4c12-aa4f-b736a7d894fb

### 🚧 TODO

- [x] Outline not displayed in view mode in Collectives
- [x] RichWorkspace: Hide outline action in headings menu
- [x] Update headings only when a heading changed
- [x] Update headings also in MarkdownContentEditor
- [x] Somehow `useTableOfContents` flag from `useEditorFlags` is always false in `useEditorWidth` when loading Collectives
- [x] Fix and extend tests
- [x] Test manually in a lot of scenarios:
  - [x] text standalone, collectives, (embedded?)
  - [x] collectives landing page with widgets above
  - [x] wide screen desktop, narrow screen desktop, mobile
  - [x] full width view and not 
  - [x] sidebar open and not 
  - [x] public share
  - [x] print view
  - [x] long headings
  - [x] long heading list (vertical scroll)
  - [x] conflict handling view
  - [x] read-only files

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/MenuBar.spec.js
+++ b/cypress/e2e/MenuBar.spec.js
@@ -63,14 +63,14 @@ describe('Test the rich text editor menu bar', function () {
 
 		it('applys default', function () {
 			cy.openTestFile()
-			cy.get('@maxWidth').should('equal', '80ch')
+			cy.get('@maxWidth').should('equal', 'min(80ch, (100% - 2 * 40px))')
 		})
 
 		it('toggles value', function () {
 			cy.openTestFile()
 			cy.getActionEntry('remain').click()
 			cy.contains('Full width editor').click()
-			cy.get('@maxWidth').should('equal', '100%')
+			cy.get('@maxWidth').should('equal', 'calc(100% - 2 * 40px)')
 		})
 
 		it('preserves on reopen', function () {
@@ -79,7 +79,7 @@ describe('Test the rich text editor menu bar', function () {
 			cy.contains('Full width editor').click()
 			cy.closeFile()
 			cy.openTestFile()
-			cy.get('@maxWidth').should('equal', '100%')
+			cy.get('@maxWidth').should('equal', 'calc(100% - 2 * 40px)')
 		})
 
 		it('preserves on reload', function () {
@@ -88,7 +88,7 @@ describe('Test the rich text editor menu bar', function () {
 			cy.contains('Full width editor').click()
 			cy.visit('/apps/files')
 			cy.openTestFile()
-			cy.get('@maxWidth').should('equal', '100%')
+			cy.get('@maxWidth').should('equal', 'calc(100% - 2 * 40px)')
 		})
 
 		it('does not interfere if width is already set', function () {


marcoambrosini (commented):
Really nice @mejo- :) 

One point from my side: 

* I would trigger the outline on **hover**, not **click**
* Once hovered, I would provide a [**pin button**](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:keep:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=pin&icon.size=24&icon.color=%231f1f1f) in the same location where the `x` button is right now. Once the pin button is pressed, we can then display the `x` button to unpin

jancborchardt (commented):
Agree with @marcoambrosini about triggering it on hover. And when leaving the area, it should close again.

Instead of automatically staying open and needing to click a "close" button, we can have a "pin" icon that can be toggled, as proposed before. :)

mejo-:
> I would trigger the outline on hover, not click

Thanks for the feedback. I was unsure about that one, but now that I implemented it, I agree it's much better. I updated the desktop screencast, please have a look :relaxed:

src/components/Editor.vue
```diff
@@ -522,6 +527,9 @@ export default defineComponent({
 				}
 			})
 			this.updateUser(session)
+			this.updateHeadings(
+				[...headingAnchorPluginKey.getState(this.editor.state)?.headings] ?? []
+			)
```
max-nextcloud: minor nitpick: I don't think `Editor.vue` should 'know' how to extract the headings from the editor. I'd prefer `this.updateHeadings(editor)` and then `useEditorHeadings` also wraps that knowledge.

One could even take this one step further and have an editor argument in `useEditorHeadings` and add a `update` event handler to the editor that sets the reactive headings according to the editors state. That way `Editor.vue` would not need to worry about `useEditorHeadings` at all.

marcoambrosini (approved):
🚀

src/components/Editor/ContentContainer.vue
```diff
@@ -73,40 +66,13 @@ export default {
 
 <style scoped lang="scss">
 .editor__content {
-	max-width: min(var(--text-editor-max-width), calc(100vw - 16px));
+	max-width: min(var(--text-editor-max-width), 100vw);
+	width: min(var(--text-editor-max-width), 100vw);
```
max-nextcloud: Those seem redundant. Are you setting both on purpose?

src/components/Editor/EditorOutline.vue
```diff
@@ -4,81 +4,155 @@
...
+			visibleHeadingIds: new Set(),
+			activeHeadingId: null,
+		}
+	},
+	watch: {
+		headings() {
+			this.$nextTick(() => {
+				this.setupIntersectionObserver()
+			})
+		},
```
max-nextcloud: I think there's no need to teardown and rebuild the intersection observer here. Should be enough to update the observed entries.
Something along these lines:
```
		headings(newHeadings, oldHeadings) {
			this.updateIntersectionObserver(newHeadings, oldHeadings)
		},
...

		updateIntersectionObserver(newHeadings, oldHeadings)
			const newIds = new Set(newHeadings.map((h) => h.id))
			const oldIds = new Set(oldHeadings.map((h) => h.id))
			const added = Set.difference(newIds, oldIds)
			const removed = Set.difference(oldIds, newIds)
			removed.forEach((id) => this.unobserve(id))
			this.$nextTick(() => {
				added.forEach((id) => this.observe(id))
			})
		},
		observe(id) {
			const el = document.getElementById(id)
			if (el) {
				this.intersectionObserver.observe(el)	
			}
		},
 .... unobserve like above - but i am not sure if that's even needed given that the element will be removed.
```

max-nextcloud: This would also make the watcher robust against overwriting the headings array with the same content - it would just do nothing as no ids were added or removed.

src/components/Editor/TableOfContents.vue
```diff
@@ -121,67 +169,69 @@ export default {
...
-		&--5 {
-			--padding-left: 4rem;
+		&.level5 {
+			--padding-inline-start: calc(4 * var(--level-padding));
 		}
 
-		&--6 {
-			--padding-left: 5rem;
+		&.level6 {
+			--padding-inline-start: calc(5 * var(--level-padding));
```
max-nextcloud: I think you could use an scss for loop here.
https://sass-lang.com/documentation/at-rules/control/for/

src/components/Editor.vue
mejo-: Sorry, this was refactored in a later commit. Now `editor` is passed as argument to `useEditorHeadings()`.

src/components/Editor/ContentContainer.vue
mejo-: Same here, I removed `width` again in a subsequent commit.

src/editor.js
```diff
@@ -190,7 +188,8 @@ window.OCA.Text.createEditor = async function ({
 	onCreate = ({ markdown }) => {},
 	onLoaded = () => {},
 	onUpdate = ({ markdown }) => {},
-	onOutlineToggle = (visible) => {},
+	onTocToggle = (visible) => {},
```
max-nextcloud, line 196: `onTocToggled`, right?

src/composables/useEditorWidth.ts
```diff
@@ -43,11 +44,13 @@ export const editorWidthKey = Symbol('text:editor:width') as InjectionKey<
...
 >
 
-// desktop: leave space for floating buttons and outline (2 * 40px)
+// desktop: leave space for floating buttons and outline (2 * 40px) if necessary
 // wide screen: 80ch content, narrow screen: full width
-const defaultEditorWidthDesktop = 'min(80ch, (100% - 2 * 40px))'
-const defaultEditorWidthMobile = 'min(80ch, 100%)'
-const fullEditorWidthDesktop = 'calc(100% - 2 * 40px)'
+const editorWidthDesktop = 'min(80ch, 100%)'
+const editorWidthDesktopEnhanced = 'min(80ch, (100% - 2 * 40px))'
```
max-nextcloud, line 50: I wonder if we could express this in css - as in using some sibling selector.

src/components/Editor/MarkdownContentEditor.vue
```diff
@@ -118,14 +118,15 @@ export default {
...
-		return { editor, setContent }
+		return { editor, setContent, updateHeadings }
 	},
 
 	created() {
 		// Set content after the setup function
 		// as it may render other vue components such as preview toggle
 		// which breaks the context of the setup function.
 		this.setContent(this.content, { addToHistory: false })
+		this.updateHeadings()
```
max-nextcloud, line 129: is this even needed given that the composable also seems to initialize the headings?

src/components/Editor/EditorOutline.vue
max-nextcloud: I think I figured out a way to even do without the set difference. I'll give it a try and let you know.

src/components/Editor/TableOfContents.vue
mejo-: Good point, changed now :blush:

src/editor.js
mejo-: Well, I changed it to `onTocToggle` everywhere now, given that we have `onOutlineToggle` (without the `d`) already.

mejo-:
Psalm failures are unrelated.

PR #8037, opened 2025-12-12
chore(split): SourceView from ViewerComponent

max-nextcloud:


src/components/ViewerComponent.vue
```diff
@@ -121,66 +101,21 @@ export default {
...
-				type: response.headers['content-type'],
-			})
-			const reader = new FileReader()
-			reader.readAsText(blob)
-			return new Promise((resolve) => {
-				reader.onload = () => {
-					resolve(reader.result)
-				}
-			})
+			this.onLoaded()
```
mejo-: It seems new that `toggleEdit()` now calls `this.onLoaded()` which emits `update.loaded`. Was this change in behaviour intended?

src/components/SourceView.vue
```diff
@@ -0,0 +1,148 @@
...
+			return this.$attrs.e2EeIsEncrypted || false
+		},
+
+		isMarkdown() {
+			return (
+				this.mime === 'text/markdown' || this.mime === 'text/x-web-markdown'
+			)
+		},
+
+		/** @return {boolean} */
```
mejo-: The annotation already existed before the refactoring, but the return value of `readerComponent` seems to be rather a Vue component and not a boolean a boolean, right? Maybe a good opportunity to fix (or remove?) the annotation?

mejo- (approved):
Thanks for the refactoring! Out of curiosity, why did you do it? Is it part of a bigger plan? Or just something that caught your eye and cried for a component split? 😊

Code changes look good to me. I didn't test but guess you did. I have two small comments/questions.

max-nextcloud:
> Thanks for the refactoring! Out of curiosity, why did you do it? Is it part of a bigger plan? Or just something that caught your eye and cried for a component split? 😊

I'm looking into a way to make the ViewerComponent create a separate vue instance from the viewer itself. This should do away with some warnings about attrs and listeners being readonly and so on.

grafik


So this is a first step in that direction by making the ViewerComponent less complex.

src/components/ViewerComponent.vue
max-nextcloud: Good point. I was thinking... what if `update:loaded` has not been called during `mouted()` because we are in the `source view` and then edit is toggled before `onLoaded` fired. The `update:loaded` would never be called.
But my understanding is that the editor won't render anything - so also won't render the edit button until `update:loaded` has been emitted so this scenario cannot happen.

max-nextcloud:
Versions playwright spec is known to fail due to https://github.com/nextcloud/viewer/issues/3046

PR #8052, opened 2025-12-13
fix(PlainTable): Add paragraph node, required for table cells

mejo-:
With table cells now requiring a block node as content, the paragraph node is needed to allow any content in a table cell.

Also slightly improves `create-table-api.spec.ts`. No need to open Text in Viewer to test `window.OCA.Text.createTable()`.

### 🏁 Checklist


silverkszlo:
thank you for the test fixes!

PR #8058, opened 2025-12-13
fix(useEditorWidth): no full width toggle in rich workspace and on mobile

mejo-:
There's no use in toggling full width in either of them.

### 🏁 Checklist

### DIFF
--- a/src/composables/useEditorWidth.ts
+++ b/src/composables/useEditorWidth.ts
@@ -22,6 +22,7 @@ import axios from '@nextcloud/axios'
 import { emit, subscribe } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 import { generateUrl } from '@nextcloud/router'
+import { useIsMobile } from '@nextcloud/vue/composables/useIsMobile'
 import {
 	computed,
 	inject,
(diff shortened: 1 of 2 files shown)


src/composables/useEditorWidth.ts
```diff
@@ -83,9 +85,14 @@ export const provideEditorWidth = () => {
...
 export const useEditorWidth = () => {
+	const isMobile = useIsMobile()
+	const { isRichWorkspace } = useEditorFlags()
+
 	// This will be null if the width is already configured outside of text.
 	const isFullWidth = inject(editorWidthKey)
-	if (isFullWidth === null) {
+
+	// Disable the toggle in rich workspace, mobile view and when width is set outside of text.
+	if (isFullWidth === null || isRichWorkspace || isMobile.value) {
```
max-nextcloud: It would be nice if the comment and the condition matched. I think reordering would help but I am also unsure about `isFullWidth === null` and "when width is set outside of text". That looks like two different conditions to me. `isFullWidth === null` happens when `null` is injected for the `editorWidthKey`. For example in the `createTable` and `createEditor` functions. But I think it's independend of wether the width is set outside as in Collectives.

max-nextcloud (approved):
One minor comment about a comment. Other than that seems good.

src/composables/useEditorWidth.ts
mejo-: Yeah, it' a bit hard to describe. I adjusted the comment now to be closer to what the code does and also to have same order as the conditions.

PR #8072, opened 2025-12-15
test(playwright): add basic tests for editor API

mejo-:
### 🏁 Checklist

### DIFF
--- a/playwright/support/fixtures/editor-api.ts
+++ b/playwright/support/fixtures/editor-api.ts
@@ -25,7 +25,7 @@ export interface EditorApiFixture {
  */
 export const test = base.extend<EditorApiFixture>({
 	createEditor: async ({ page }, use) => {
-		const createComponent = async (
+		const createEditor = async (
 			options: {
 				type: 'editor' | 'table',
 				content?: string,
@@ -58,17 +58,6 @@ export const test = base.extend<EditorApiFixture>({
 				},
 				{ containerId, type: options.type, content: options.content, fileId: options.fileId, readOnly: options.readOnly },
 			)
-		}
-
-		const createEditor = async (
-			options: {
-				type: 'editor' | 'table',
-				content?: string,
-				fileId?: number,
-				readOnly?: boolean,
-			},
-		) => {
-			await createComponent(options)
 		}
 
 		await use(createEditor)


playwright/e2e/editor-api.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+				})
+			},
+			{ containerId, ...options },
+		)
+	}
+
+	fileTest.use({ fileContent: 'some content\n' })
+
+	fileTest.beforeEach(async ({ open, page }) => {
+		await open()
```
max-nextcloud: I think this will open the viewer in the background. Is that what you were trying to do? Two editors in the same tab?

mejo-: Ah no. My intention was merely to upload a file and get the `fileId` of it to pass it to the `createEditor()` function. You're right, it's overkill to keep the file open in the background.

mejo-: Adjusted now.

mejo-:
> All in all looks good. The create...Editor functions could be moved into fixtures for easier reuse. But we can also do that later, just need to make sure not to copy them anywhere else / rewrite them.

That's a good point. I refactored the tests with help of copilot and think they're much cleaner structured now. Let me know what you think :blush:

playwright/support/fixtures/editor-api.ts
```diff
@@ -45,17 +47,32 @@ export const test = base.extend<EditorApiFixture>({
...
+			options: {
+				type: 'editor' | 'table',
+				content?: string,
+				fileId?: number,
+				readOnly?: boolean,
+			},
+		) => {
+			await createComponent(options)
+		}
+
```
max-nextcloud: This seems like `createEditor` does the same as `createComponent` or am I missing something?

playwright/support/fixtures/editor-api.ts
```diff
@@ -45,17 +47,32 @@ export const test = base.extend<EditorApiFixture>({
...
 		await use(createEditor)
 	},
+	containerId: async (_, use) => {
+		await use('test-editor-api')
+	},
```
max-nextcloud: I think you can just do
```js
containerId: 'test-editor-api',
```

Or if you want to make it customizable with `test.use({ containerId: 'something-else' })`:
```js
  containerId: ['test-editor-api', { option: true }]
```

playwright/support/fixtures/editor-api.ts
```diff
@@ -3,36 +3,38 @@
...
 			options: {
+				type: 'editor' | 'table',
 				content?: string,
 				fileId?: number,
 				readOnly?: boolean,
-				containerId?: string,
 			},
 		) => {
-			const containerId = options.containerId ?? 'test-editor-api'
+			const containerId = 'test-editor-api'
```
max-nextcloud: You're not using the containerId fixture from below. So I think you can either do away with it or add it above to the `{ page }`.

playwright/support/fixtures/editor-api.ts
max-nextcloud: oh... and i think it doesn't even return anything.

playwright/support/fixtures/editor-api.ts
mejo-: Indeed, good catch. I made `containerId` a global const now and just passed it here.

playwright/support/fixtures/editor-api.ts
mejo-: I don't see a reason to make it configurable. At the same time it needs to be exposed by the fixture as the tests need it in order to use it in CSS selectors.

playwright/support/fixtures/editor-api.ts
mejo-: True. That's a leftover from an earlier refactoring where I still had separate `createEditor()` and `createTable()` functions. Fixed now :blush:

PR #8077, opened 2025-12-17
Fix intersection observer life cycle

max-nextcloud:
- cleanup unused heading.previous
- headings: keep anchors and intersection observer
- headings: lookup anchors within current editor

### DIFF
--- a/playwright/support/fixtures/Node.ts
+++ b/playwright/support/fixtures/Node.ts
@@ -77,9 +77,11 @@ export class Node {
 	}
 
 	async open() {
+		// loading the file list may take a while
+		const timeout = 10_000
 		await this.page.goto(`f/${this.id}`)
 		await expect(this.page.getByLabel(this.name, { exact: true }))
-			.toBeVisible()
+			.toBeVisible({ timeout })
 	}
 
 	async move(newName: string) {


src/plugins/headingAnchor.ts
```diff
@@ -130,6 +130,10 @@ function decorationForHeading(heading: Heading) {
  * @return {HTMLElement}
  */
 function anchorForHeading(heading: Heading) {
+	const existing = document.getElementById(heading.id)
```
mejo-: Interesting. Does this mean that we kept adding anchor elements for existing headings and they piled up? 🤔

src/plugins/headingAnchor.ts
```diff
@@ -4,11 +4,11 @@
...
 
 import { t } from '@nextcloud/l10n'
+import type { Node } from '@tiptap/pm/model'
 import { Plugin, PluginKey, Transaction } from '@tiptap/pm/state'
-import { Decoration, DecorationSet } from '@tiptap/pm/view'
-import extractHeadings from './extractHeadings.js'
+import { Decoration, DecorationSet, EditorView } from '@tiptap/pm/view'
 import type { Heading } from '../composables/useEditorHeadings'
-import type { Node } from '@tiptap/pm/model'
+import extractHeadings from './extractHeadings.js'
```
mejo-: nitpick: I think you renamed it to `extractHeadings.ts` earlier.

src/plugins/headingAnchor.ts
max-nextcloud: I think they were at least removed from the dom when they were removed from the decoration set... and then maybe garbage collection... hopefully :shrug: 
At least we were creating them for all headings when one changed.

max-nextcloud: This is actually still relevant for the anchor that are left behind when a heading is changed / removed.
One can define a [destroy function](https://prosemirror.net/docs/ref/#view.Decoration^widget^spec.destroy) for the widget. This might be relevant as it's still referenced by the intersection observer. Which might interfere with garbage collection.

src/components/Editor/TableOfContents/TableOfContentsEntry.vue
```diff
@@ -0,0 +1,93 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<!-- :class="{ '--initial-render': initialRender }" -->
```
mejo-: This line can go, it's a leftover that I already removed in the toc PR.

src/plugins/headingAnchor.ts
max-nextcloud: I investigated this with chromiums detached elements profiler some and it looks like we had some detached elements before - but not anymore after this fix.

mejo-: Really nice sleuthing work!

PR #8079, opened 2025-12-17
fix(table): use small buttons for table actions

mejo-:
Leaves more space and focus for the table itself, especially on small screens like mobile.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

### 🏁 Checklist


max-nextcloud (approved):
code looks good to me.

PR #8134, opened 2026-01-06
fix(playwright): use server master branch for unreleased version

max-nextcloud:
Fixed by copying the [start script from `end_to_end_encryption`](https://github.com/nextcloud/end_to_end_encryption/blob/980440486cd6289fa03183d48caa1cfe829662b0/tests/playwright/start-nextcloud-server.js).
Also renders the setup steps superfluous.


max-nextcloud:
Playwright is missing the open sidebar menu entry in the viewer:
grafik

I am checking in with files team about what's up with that:
* https://github.com/nextcloud/viewer/issues/3046

PR #8153, opened 2026-01-13
fix(viewer): mount text vue instance explicitely

max-nextcloud:
The component configured in the viewer handler
is instantiated from the vue module inside the viewer.

We want to use our own vue module so create an instance explicitely and mount it.

This fixes warnings in development like $attrs is readonly.

It also makes the vue dev tools work!

### Todo

* [ ] merge #8178

### DIFF
--- a/src/components/ViewerComponent.vue
+++ b/src/components/ViewerComponent.vue
@@ -36,6 +36,7 @@ export default defineComponent({
 		SourceView,
 		Editor,
 	},
+	inheritAttrs: false,
 	provide() {
 		return {
 			isEmbedded: this.isEmbedded,
@@ -65,10 +66,6 @@ export default defineComponent({
 		mime: {
 			type: String,
 			default: null,
-		},
-		permissions: {
-			type: String,
-			default: '',
 		},
 		source: {
 			type: String,


max-nextcloud:
#### 🖼️ Dev tools Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

juliusknorr:
I'm unsure if we have a test for https://github.com/nextcloud/text/issues/5531 these days but should double check this case

max-nextcloud:
> I'm unsure if we have a test for #5531 these days but should double check this case

We do have this:
https://github.com/nextcloud/text/blob/302dcc7b31f065b2cb1b8871fa29f780df7e8b75/cypress/e2e/attachments.spec.js#L301-L330

( I know because it was broken initially.) The editor for the text file in a text file renders fine.

silverkszlo:
unrelated but it seems that `permissions` is being defined but not used in `ViewerComponent.vue`:
```
permissions: {
	type: String,
	default: '',
},
```

silverkszlo:
also unrelated but it seems that `fileVersion` and `isEncrypted` are not defined anywhere in `ViewerComponent.vue` (not in props, data, computed or methods). So will always be `undefined` I guess. Old code?
```
useSourceView() {
	return (
		this.source
		&& (this.fileVersion
			|| !this.fileid
			|| this.isEmbedded
			|| this.isEncrypted)
		&& !this.hasToggledInteractiveEmbedding
	)
},
```

silverkszlo:
> I'm unsure if we have a test for #5531 these days but should double check this case

So as advised I have reproduced the steps from #5531 in the UI and what happened was the following:
1. Bluntly pasting the internal link for an image, e.g. "http://nextcloud.local/index.php/f/37" and then clicking on "Show link preview". Result was: no preview.
Screenshot from 2026-01-21 16-21-53

Then clicking "Text only" and then "Show link preview" again. Result: not a real preview but the link again.
Screenshot from 2026-01-21 16-22-48

What did work: Adding the link via` /` → `Files` → image. Then "Show link preview" shows the real image preview. Same for when I do` /` → `Any link` → [internal link].
Screenshot from 2026-01-21 16-35-37

BUT then I deleted all the links, added the normal internal link again, clicked on "Show link preview" and the preview works fine. Cached somewhere? 🤷

max-nextcloud:
> unrelated but it seems that `permissions` is being defined but not used in `ViewerComponent.vue`:
> 
> ```
> permissions: {
> 	type: String,
> 	default: '',
> },
> ```

This is a really good catch :fishing_pole_and_fish: - in particular as it points to another issue.

We're inheriting the attributes when the component is rendered in viewer:
grafik

So we can remove this - but we also need to add [`inheritAttrs: false`](https://v2.vuejs.org/v2/guide/components-props#Disabling-Attribute-Inheritance) to the component so it does not get rendered in the dom instead.

PR #8156, opened 2026-01-15
fix: preserve OCA.Text properties in public share script

silverkszlo:
This PR makes insertion and editing of tables in public shares of Whiteboards possible: https://github.com/nextcloud/whiteboard/pull/941

Prevent race condition where public.js would overwrite createEditor/createTable methods if editors.js loaded first. Uses spread operator to merge instead of replace, matching the pattern already used in editor.js.

### 🏁 Checklist

### DIFF
--- a/playwright/support/fixtures/editor-api.ts
+++ b/playwright/support/fixtures/editor-api.ts
@@ -49,7 +49,6 @@ export const test = base.extend<EditorApiFixture>({
 					document.body.appendChild(container)
 
 					const method = type === 'editor' ? 'createEditor' : 'createTable'
-					// @ts-expect-error - OCA.Text is a global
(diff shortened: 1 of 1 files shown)


src/public.js
```diff
@@ -6,6 +6,7 @@
 import { loadState } from '@nextcloud/initial-state'
 
 OCA.Text = {
+	...OCA.Text,
```
juliusknorr: Can we be sure that OCA.Text is always defined before as well? Otherwise we may need to safeguard this with an empty object to merge into

juliusknorr:
```suggestion
	...(window?.OCA?.Text ?? {}),
```

silverkszlo: Good idea, thanks!

PR #8157, opened 2026-01-15
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency

### DIFF
--- a/lib/Db/Document.php
+++ b/lib/Db/Document.php
@@ -38,7 +38,6 @@ class Document extends Entity implements \JsonSerializable {
 	protected ?string $checksum = null;
 
 	public function __construct() {
-		$this->addType('id', 'integer');
 		$this->addType('currentVersion', 'integer');
 		$this->addType('lastSavedVersion', 'integer');
 		$this->addType('lastSavedVersionTime', 'integer');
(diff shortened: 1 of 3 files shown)


lib/Db/Document.php
```diff
@@ -27,6 +27,7 @@
  * @method setChecksum(?string $checksum): void
  */
 class Document extends Entity implements \JsonSerializable {
+	public $id = null;
```
CarlSchwan, line 30: this is not required

lib/Db/Document.php
```diff
@@ -37,6 +38,7 @@ class Document extends Entity implements \JsonSerializable {
 	protected ?string $checksum = null;
 
 	public function __construct() {
+		$this->addType('id', 'integer');
 		$this->addType('currentVersion', 'integer');
 		$this->addType('lastSavedVersion', 'integer');
 		$this->addType('lastSavedVersionTime', 'integer');
```
CarlSchwan: same

lib/Db/Document.php
max-nextcloud: grafik

But psalm fails without it.

CarlSchwan: sight :/

lib/Db/Document.php
max-nextcloud: I can remove this and psalm is still happy. So I guess I will but I'll keep the $id = null

PR #8168, opened 2026-01-18
[main] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (2 unfixable, 24 only fixable manually using --force)


max-nextcloud:
plawright failure is not related.

PR #8178, opened 2026-01-22
fix(source-view): compute isEncrypted in ViewerComponent

max-nextcloud:
Follow up for #8037.


max-nextcloud:
Cypress failes because inserting file links is currently broken as `OC.Files` is undefined since NC 34.

PR #8196, opened 2026-01-26
perf: Decrease amount of filecache SQL call

CarlSchwan:
### 📝 Summary

Instead of checking for the existance of the file and then getting it, directly get it and catch the exceptions. Still called too many times in my opinion, but that should help a bit :)

image

### 🏁 Checklist


juliusknorr:
Good to see that, previously the get() call used to be more expensive 🎉

PR #8223, opened 2026-01-31
Migrate to playwright to update cypress

max-nextcloud:
Cypress 13 uses a browser so outdated it's not supported by Nextcloud anymore.

Many cypress tests fail with cypress 15 with browser crashes.
Backport the commits that migrate these tests to playwright.
Upgrade cypress to version 15.

- chore(test): Migrate cypress offline tests to playwright
- chore(test): split setOnline and setOffline
- chore(test): add offline fixture to playwright
- chore(test): get request token from /csrftoken
- chore(test): split request token fixture from random user
- chore(test): cleanup unused type and context
- chore(test): mimetype change in playwright
- chore(test): basic editor page object model
- chore(test): file class for handling uploaded file
- chore(test): file.open()
- chore(test): with file.move() and file.close()
- chore(test): with editor.typeHeading()
- chore(test): change mimetype both ways
- chore(test): move File class to separate file
- chore(test): use EditorSection in offline test
- chore(test) .withOpenMenu() to access submenus
- chore(test): write offline and come back online
- chore(test): reduce nesting in offline test
- chore(test): wait for close request when closing
- chore(test): migrate mobile spec from cy to playwright
- chore(test): convert smart picker test to playwright
- chore(test): convert print spec to playwright
- chore(test): new File() with params object
- chore(test): handle request token as part of login
- chore(test): User class to wrap Account and Page
- chore(test): print with shared file
- chore(test): no need to merge randomUserTest
- chore(test): rename editor.locator to editor.el
- chore(test): make File attributes read only
- chore(test): use parameter properties
- chore(test): separate screenshots based on host
- chore(test): add versions spec
- chore(test): versions with ViewerSection
- chore(test): move close() into ViewerSection
- chore(test): convert version spec from cy to playwright
- chore(test): folder description with playwright
- chore(playwright): Fix prettier warnings
- test(playwright): Add regression test for CJK IME input as first character
- chore(test): format text with playwright
- fix(css): full width table in print view
- chore(deps): update dependency cypress to v15

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,3 +6,8 @@ coverage:
                 paths:
                     - 'src'
                 only_pulls: true
+        patch:
+            default:
+                paths:
+                    - 'src'
+                only_pulls: true


mejo- (approved):
Thanks for tackling this! Nice side-effect is that we can mark playwright as required in the branch protection rules once we dropped 31 support (next month 31 becomes EOL).

max-nextcloud:
> Thanks for tackling this! Nice side-effect is that we can mark playwright as required in the branch protection rules once we dropped 31 support (next month 31 becomes EOL).

How about we backport this to stable31 as well? It's 99% changes to tests and might be helpful to keep tests running on stable31 in case we need to backport fixes to it later on.

PR #8257, opened 2026-02-09
[stable31] Fix/hide link decoration in read-only mode

silverkszlo:
### 📝 Summary

Note: This bug only happens in Nextcloud 31 and already has been fixed in NC 32.

#### Problem
When a Text file is shared in read-only mode, users can access and interact with link preview options (toggle preview mode, delete link, open in new tab). Changes appear to work but don't persist after reopening the file, creating a confusing UX.

#### Solution
Fixed the `previewOptions` plugin to check `editor.options.editable` at the beginning of the `apply()` function, before checking if the document changed. This ensures decorations (and thus the PreviewOptions component) are never created when the editor is read-only.

### 🏁 Checklist

### DIFF
--- a/src/plugins/previewOptions.js
+++ b/src/plugins/previewOptions.js
@@ -27,7 +27,7 @@ export default function previewOptions({ editor }) {
 		state: {
 			init(_, { doc }) {
 				if (!editor.options.editable) {
-					return { decorations: DecorationSet.create() }
+					return { linkParagraphs: [], decorations: DecorationSet.empty }
 				}
 				const linkParagraphs = extractLinkParagraphs(doc)
 				return {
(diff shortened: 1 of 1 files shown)


silverkszlo:
> Code changes look sensible to me. I didn't test myself but guess you did so.
> 
> Small remark: I wonder whether it would be better to simply not load the `previewOptions` plugin in the Tiptap paragraph node if editor is read-only, but guess that would be more complicated to implement in a way that it is reactive in a way that it takes into account if the editor becomes editable/non-editable after being initialized.
> 
> Given that it's a fix for a stable release that will become EOL soon, I'd not spend too much energy into optimizing this 😉

Thank you! I'll merge that for now so that it gets into the upcoming release. I can refactor later on if we want that :)

Issue #5597, opened 2024-04-03
Open text documents with lazy session initialization


juliushaertl:
Given the new approach to load a document state from the file content, we should consider adding a way to initialize an editing session lazy. This could be useful for various cases:
- In a read only document we likely don't need to update as often -> can be a large impact on instances that share a lot of read only files that get accessed frequently
- In rich workspaces we could only lock the file once someone actually starts editing
- Collectives could rather use this instead of requiring to switch between different editor components

We'd probably need:
- Have an API to fetch the document content or yjs state without creating a new session, maybe even fetch steps for read only sessions
- Some less frequent polling/notify push could then be used to check if a user editing joins or some steps were pushed
- Think about how a handover can work from the UX perspective
- Think about how the session can be initialized in the frontend code base

PR #8284, opened 2026-02-19
fix: Exclude Readme.md files from file locking in the text app

rikled:
### 📝 Summary

* Resolves: [README.md is almost always locked](https://github.com/nextcloud/files_lock/issues/101)

This PR excludes Readme files from file locking. Without it is almost impossible to delete these files as they are opened as soon as a user visits the directory of the file. This is meant as an intermediate solution until https://github.com/nextcloud/text/issues/5597 is implemented.

### 🏁 Checklist

### DIFF
--- a/lib/Service/ApiService.php
+++ b/lib/Service/ApiService.php
@@ -131,8 +131,7 @@ class ApiService {
 
 		// Disable file locking for Readme.md files, because in the
 		// current setup, this makes it almost impossible to delete these files.
-		/** @psalm-suppress RedundantCastGivenDocblockType */
-		if (!$readOnly && strcasecmp((string)$file->getName(), 'Readme.md') !== 0) {
+		if (!$readOnly && strcasecmp($file->getName(), 'Readme.md') !== 0) {
 			$isLocked = $this->documentService->lock($file->getId());
 			if (!$isLocked) {
 				$readOnly = true;
--- a/tests/unit/Service/ApiServiceTest.php
+++ b/tests/unit/Service/ApiServiceTest.php
@@ -76,6 +76,7 @@ class ApiServiceTest extends \PHPUnit\Framework\TestCase {
 		$file->method('getStorage')->willReturn($storage);
 		$file->method('getId')->willReturn($id);
 		$file->method('getOwner')->willReturn($owner);
+		$file->method('getName')->willReturn('name');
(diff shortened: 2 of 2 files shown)


lib/Service/ApiService.php
```diff
@@ -129,7 +129,10 @@ public function create(?int $fileId = null, ?string $filePath = null, ?string $b
 
 		$hasOwner = $file->getOwner() !== null;
 
-		if (!$readOnly) {
+		// Disable file locking for Readme.md files, because in the
+		// current setup, this makes it almost impossible to delete these files.
+		/** @psalm-suppress RedundantCastGivenDocblockType */
+		if (!$readOnly && strcasecmp((string)$file->getName(), 'Readme.md') !== 0) {
```
juliusknorr:
```suggestion
		if (!$readOnly && strcasecmp($file->getName(), 'Readme.md') !== 0) {
```

I think if the phpunit tests fail on that it is rather due to the file mock not having calls of getName mocked.

Probably just a matter of addding in https://github.com/nextcloud/text/blob/16f94d837bd129f9c4fda74e5d51d51e45b4c5f4/tests/unit/Service/ApiServiceTest.php#L77

```php
$file->method('getName')->willReturn('MyFile.md');
```

juliusknorr: When running with the real implementation it is always returning a string

https://github.com/nextcloud/server/blob/0d6c5516947cbe0837ab4731b00bbd60f37908d0/lib/public/Files/Node.php#L214
http://github.com/nextcloud/server/blob/0d6c5516947cbe0837ab4731b00bbd60f37908d0/lib/private/Files/Node/Node.php#L313
Especially since basename always returns a string, not null or anything else: https://www.php.net/manual/en/function.basename.php

rikled: So, shall I fix the tests then?

juliusknorr: Yes, that would be better as then the mocked unit test is closer to how a real code execution flow would be.

JuliaKirschenheuter:
@mejo- could you please create a manual backport for stable31 too? Thanks a lot!

mejo-:
> @mejo- could you please create a manual backport for stable31 too? Thanks a lot!

already happened: https://github.com/nextcloud/text/pull/8366

Issue #4900, opened 2023-10-25
Links removed on text formatted as inline code


Imberflur:
## Short summary of the issue

Opening markdown with links on text containing inline code formatting with backticks will remove the link or move it to the first portion of text before the inline code formatting starts. I believe this is a specific case of https://github.com/nextcloud/text/issues/593 but I haven't been able to find any discussion/information on this particular case.

- [ ] Markdown syntax is not properly rendered in text
- [x] Markdown syntax is removed after closing text
- [x] Markdown syntax is changed after closing text
- [ ] Something else: Please describe more in detail

## Example markdown file

````markdown
[`SomeCode`](https://link.to/something)
[One `two` three](https://link.to/something)
````

## Resulting markdown file after opening in text

````markdown
`SomeCode`
[One](https://link.to/something) `two` three
````

## Screenshot of the rendering in text

![image](https://github.com/nextcloud/text/assets/2002109/bd8edc8e-7f86-40ef-a78b-ab6088a12f63)

## Details
- Nextcloud version: 27.1.2

mejo-:
Confirmed, I can reproduce.

PR #8318, opened 2026-03-08
fix(links): allow inline code within links

mejo-:
The Tiptap inline code mark extension excludes all other mark types from coexisting with itself. This makes sense for most mark types, but link is an exception.

Unfortunately there's no way to dynamically create a list of all marks when `excludes` is initialized, as this happens when compiling the editor schema, so the latter is not available yet.

Thus to fix the bug while keeping the side effects small, we have to explicitely list all mark types except link and code in `excludes` in the code mark definition.

Fixes: #4900

### 🏁 Checklist


max-nextcloud:
grafik

Hum... so the serialization looks okay:

```code
Hello you `inline `[`code`](https://nextcloud.com)` how are what about https://autolink.com ? yoy` ?

```

PR #8320, opened 2026-03-09
Debug flaky cypress test in stable33

max-nextcloud:
Tests have been passing since i pushed this. And I tried them several times. But anyways - with these changes the next test failure will hopefully have a more meaningful output:
* Avoid retries that fail for unrelated reasons ( not being able to create an already existing directory )
* Compare file ids of the created file and the moved file. Failures look like these might be changing :exploding_head: 

## Failures before

Note how there **is** an attachments folder - just with a different id.

Test all attachment insertion methods -- test if attachment folder is moved with the markdown file (failed)

### DIFF
--- a/cypress/e2e/attachments.spec.js
+++ b/cypress/e2e/attachments.spec.js
@@ -345,40 +345,44 @@ describe('Test all attachment insertion methods', () => {
 			})
 	})
 
-	it('test if attachment folder is moved with the markdown file', () => {
-		const fileName = 'moveSource.md'
-		cy.createFolder('subFolder')
-		cy.createMarkdown(
-			fileName,
-			'![git](.attachments.123/github.png)',
-			false,
-		).then((fileId) => {
-			const attachmentsFolder = `.attachments.${fileId}`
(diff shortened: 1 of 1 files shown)


cypress/e2e/attachments.spec.js
```diff
@@ -345,40 +345,44 @@ describe('Test all attachment insertion methods', () => {
...
+				})
+
+			cy.visit('/apps/files')
+			cy.openFolder('subFolder')
+			cy.getFile('test.md')
+				.should('exist')
+				.should('have.attr', 'data-cy-files-list-row-fileid')
+				.then((documentId) => {
+					const files = attachmentFileNameToId[documentId]
+					cy.get('@createdFileId').should('eq', parseInt(documentId))
```
max-nextcloud, line 375: In the failing tests it looks as if the `.attachments` folder has a different id then expected. So compare the ids first to see if the id changed when moving the file.

mejo-:
> I wonder whether we should also forward-port this to main?

Ah just saw you already requested this :wink:

PR #8341, opened 2026-03-16
Backport/8284/stable31

JuliaKirschenheuter:
### 📝 Summary

Backport for https://github.com/nextcloud/text/pull/8284

### 🏁 Checklist

### DIFF
--- a/tests/unit/Service/ApiServiceTest.php
+++ b/tests/unit/Service/ApiServiceTest.php
@@ -72,6 +72,7 @@ class ApiServiceTest extends \PHPUnit\Framework\TestCase {
 		$file->method('getStorage')->willReturn($storage);
 		$file->method('getId')->willReturn($id);
 		$file->method('getOwner')->willReturn($owner);
+		$file->method('getName')->willReturn('name');
 		return $file;
 	}
 


lib/Service/ApiService.php
```diff
@@ -143,7 +143,10 @@ public function create(?int $fileId = null, ?string $filePath = null, ?string $b
 
 		$hasOwner = $file->getOwner() !== null;
 
-		if (!$readOnly) {
+		// Disable file locking for Readme.md files, because in the
+		// current setup, this makes it almost impossible to delete these files.
+		/** @psalm-suppress RedundantCastGivenDocblockType */
+		if (!$readOnly && strcasecmp((string)$file->getName(), 'Readme.md') !== 0) {
```
rikled, line 149: I think we don't need the string cast anymore since I added a name in the ApiServiceTest.php

rikled: Then you should also be able to remove `/** @psalm-suppress RedundantCastGivenDocblockType */`

JuliaKirschenheuter: thank you @rikled! Then it will look like:

```
if (!$readOnly) { 
```

rikled: No, I think it should look like this:
```
		// Disable file locking for Readme.md files, because in the
		// current setup, this makes it almost impossible to delete these files.
		if (!$readOnly && strcasecmp($file->getName(), 'Readme.md') !== 0) {
```

so just remove the string cast and the psalm suppress command. Then hopefully the php unit tests will still pass.

mejo-:
I backported the original PR to stable33, then stable32 and then stable31. So closing in favour of https://github.com/nextcloud/text/pull/8366.

PR #8342, opened 2026-03-16
chore(cleanup): replace markdownit-katex with @mdit/tex

max-nextcloud:
The latex does not need to be rendered during the markdown processing.
It was extracted from the annotation afterwards anyways.

This avoids a duplicate import of KaTex which shrinks our largest chunk
from 2.4 Mb to 2.1 Mb.


package.json
```diff
@@ -101,12 +100,13 @@
...
     "vue-click-outside": "^1.1.0",
     "vue-material-design-icons": "^5.3.1",
-    "webdav": "^5.9.0",
+    "webdav": "^5.8.0",
     "y-prosemirror": "^1.3.7",
     "y-protocols": "^1.0.7",
     "yjs": "^13.6.29"
   },
   "devDependencies": {
+    "@mdit/plugin-tex": "^0.24.1",
```
mejo-: This should be a dependency, not a dev dependency, I guess.

mejo- (approved):
Code changes look good apart from my comment. Thanks for tackling this ❤️

PR #8352, opened 2026-03-17
fix(editor): make editor container grow on empty/short content

mejo-:
Also tested with folder descriptions.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/playwright/e2e/design.spec.ts
+++ b/playwright/e2e/design.spec.ts
@@ -1,0 +1,31 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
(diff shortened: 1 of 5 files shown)


max-nextcloud:
I wonder if the buttons were meant to be up there and that's why the text was shrinked. @marcoambrosini What's your take on this?

We definitely want to expand the text area all the way to the bottom so the mouse cursor changes and content can be dragged into it. So for me the question is more if we want the buttons to be floating above the text area - which would then be a follow up.

mejo-:
> I wonder if the buttons were meant to be up there and that's why the text was shrinked. @marcoambrosini What's your take on this?

We already had this topic in the past and I'm pretty sure that Marco said they should be at the bottom, no?

PR #8353, opened 2026-03-17
feat(table): allow sort by column

### LINKED ISSUE #5993: Table: allow to sort by column
It would be nice, if a table could be sorted by selecting the column you want to sort by.

Click on a colum header and select, if you want to sort the table by this column ascending or descending.

There is no alternative.

**Additional context**
I would need different sorting of the table in different situations.

luka-nextcloud:
### 📝 Summary

* Resolves: #5993

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/nodes/Table/Table.js
+++ b/src/nodes/Table/Table.js
@@ -183,6 +183,118 @@ export default Table.extend({
 					}
 					return true
 				},
+			sortColumn:
+				(direction = 'asc', cell = null) =>
+				({ state, tr, dispatch }) => {
+					if (
+						cell?.type?.name !== 'tableCell'
+						&& cell?.type?.name !== 'tableHeader'
+					) {
+						return false
+					}
+
+					// find the table, its position and the column index of the cell
+					let table = null
+					let tablePos = null
+					let columnIndex = -1
+
+					state.doc.descendants((node, pos) => {
+						if (node.type.name !== 'table') {
+							return true
+						}
+
+						for (
+							let rowIndex = 0;
+							rowIndex < node.childCount;
+							rowIndex += 1
+						) {
+							const row = node.child(rowIndex)
+							for (
+								let currentColumnIndex = 0;
+								currentColumnIndex < row.childCount;
+								currentColumnIndex += 1
+							) {
+								if (row.child(currentColumnIndex) === cell) {
+									table = node
+									tablePos = pos
+									columnIndex = currentColumnIndex
+									return false
+								}
+							}
+						}
+
+						return true
+					})
+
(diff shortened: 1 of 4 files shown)


luka-nextcloud:
#### 🖼️ Screenshots
[demo.webm](https://github.com/user-attachments/assets/34b4699c-8c0b-4a12-ad17-ea2179c64e5f)

max-nextcloud:
@luka-nextcloud Thanks a lot! That looks really great. I'll take a look at the code. Could you rebase and `npm run prettier:fix` on the branch to make prettier happy?

Thanks!

src/tests/nodes/Table.spec.js
```diff
@@ -236,8 +236,78 @@ describe('Table extension', () => {
...
+			return false
+		}
+		currentIndex += 1
+		return true
+	})
+	if (selectionPos != null) {
+		editor.commands.setTextSelection(selectionPos)
+	}
+	return selectionPos
+}
```
max-nextcloud: This should not need to set the text selection but you can use it to get the header cell position to provite to the command.

src/nodes/Table/Table.js
```diff
@@ -181,6 +183,94 @@ export default Table.extend({
 						)
 						dispatch(tr.setSelection(selection).scrollIntoView())
 					}
+					return true
+				},
+			sortColumn:
+				(direction = 'asc', explicitColumnIndex = null) =>
```
max-nextcloud: It would be nice if this command did not depend on the cursor position. This would remove the need to move the cursor around just to change the sorting of a table.

So instead of getting the `$cell` from the state below we'd need to provide it as a function argument. I think the easiest would be to just provide the `position` of the cell as returned by `this.getPos()` inside the cell node and then find the cell from there.

This would also remove the need for the explicitColumnIndex - as the index would just be the index of the cell identified by the position.

max-nextcloud (dismissed):
All in all looks good.

I think it would be nice to not move the cursor around to change the sort order. I added some more detailed comments explaining how I think this would be feasible.

src/nodes/Table/Table.js
```diff
@@ -72,6 +72,31 @@ function findSameCellInNextRow($cell) {
...
+		}
+		return null
+	}
+
+	return (
+		tryResolve(position)
+		|| tryResolve(typeof position === 'number' ? position + 1 : position)
+	)
+}
+
```
max-nextcloud: Why is this needed?

benjaminfrueh (approved):

PR #8405, opened 2026-03-24
[stable32] fix(SessionService): sanitize displayName to utf8 encoding

silverkszlo:
### 📝 Summary

A user reported that they can not open text/md files and were getting the following error message: "Malformed UTF-8 characters, possibly incorrectly encoded", while most other users could open the same file without a problem.

This PR uses `EncodingService` to convert display names from external backends (e.g. LDAP/AD) to UTF-8 before including them in JSON responses.

### 🏁 Checklist

### DIFF
--- a/lib/Service/SessionService.php
+++ b/lib/Service/SessionService.php
@@ -32,6 +32,7 @@ class SessionService {
 	private IAvatarManager $avatarManager;
 	private ?string $userId;
 	private ICache $cache;
+	private EncodingService $encodingService;
 
 	/** @var ?Session cache current session in the request */
 	private ?Session $session = null;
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Interesting corner case and I wonder whether this would need to be applied in other places as well. Searching for `[dD]isplayName` revealed at least `lib/Notification/Notifier.php` and `lib/Controller/UserApiController.php`.

But fine with merging as is. A comment in the code would be nice to make it clear for the future why we do the extra sanitizing there.

mejo-:
Ahh, I just now realized that this is supposed to be merged to `stable32` directly. Any particular reason why the fix is not targeting `main` first (and backported from there)?

silverkszlo:
> Ahh, I just now realized that this is supposed to be merged to `stable32` directly. Any particular reason why the fix is not targeting `main` first (and backported from there)?

No good reason, I just started developing in 32 as the user reported their issue in that version and then I have just noticed, after creating the PR, that it would have made more sense to apply that in main directly. I'll add the change to the other branches soon.

max-nextcloud:
Asked around as I assumed people would run into the same problem in other apps.
@blizzz responded:
> I don't believe we do anything about the encoding in the LDAP backend (did not double check, i think we just take what we receive from the server, apart of the user id is made of). But it would make more sense to fix this on the source rather than on any other place.

I tend to agree.

mejo- (approved):

PR #8446, opened 2026-04-10
#8332: Tag files when AI generated content gets added

### LINKED ISSUE #8332: Tag files when AI generated content gets added
Whenever AI generated content is added to a text file we need to add the standard system tag to it:

https://github.com/nextcloud/server/pull/58836 can be used for that.

samin-z:
### 📝 Summary

* Resolves: #8332 

#### 🖼️ Screenshots
Screenshot 2026-04-10 at 10 49 51

### 🚧 TODO

- [x] Add 'Generated By AI' tag to text files using Ai tools.

### 🏁 Checklist


tests/unit/Controller/AiControllerTest.php
```diff
@@ -0,0 +1,54 @@
...
+use OCP\AppFramework\Http\DataResponse;
+use OCP\IRequest;
+use PHPUnit\Framework\MockObject\MockObject;
+use Test\TestCase;
+
+class AiControllerTest extends TestCase {
+	/** @var IRequest|MockObject */
+	private $request;
+	/** @var AiTagService|MockObject */
+	private $aiTagService;
```
CarlSchwan:
```suggestion
	private IRequest&MockObject  $request;
	private AiTagService&MockObject $aiTagService;
```

lib/Service/AiTagService.php
```diff
@@ -0,0 +1,42 @@
...
+		private LoggerInterface $logger,
+	) {
+	}
+
+	/**
+	 * @param int $fileId
+	 *
+	 * @return void
+	 *
+	 * @throws \Exception
```
blizzz: that's not true, is it?

lib/Service/AiTagService.php
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @since 34.0.0 (ISystemTagObjectMapper::assignGeneratedByAITag)
+	 */
+	public function tagFileAsAiGenerated(int $fileId): void {
+		try {
+			$this->systemTagObjectMapper->assignGeneratedByAITag((string)$fileId, 'files');
+		} catch (\Exception $e) {
```
blizzz, line 32: There is no signature in the called function that says it throws an Excpetion, or is there? Would then not do it, and avoid catching an Exception that should not be caught here.

blizzz (commented):
Nitpicking, but otherwise looks good, did not test.

lib/Service/AiTagService.php
samin-z: you're right, I added @throws thinking the underlying call (assignTags) might throw at runtime but since I'm already catching everything in the try/catch it never reaches the caller anyway. i'll rwmove the annotation.

blizzz (approved):

PR #8476, opened 2026-04-16
Security: Unbounded `limit` parameter in user search can be abused for resource exhaustion

tomaioo:
## Summary

Security: Unbounded `limit` parameter in user search can be abused for resource exhaustion

## Problem

**Severity**: `Medium` | **File**: `lib/Controller/UserApiController.php:L39`

The `index(string $filter = '', int $limit = 5)` method accepts client-controlled `limit` and passes it directly to collaborator search. Without an upper bound, an attacker can request very large limits, causing expensive directory lookups and increased response size.

## Solution

Enforce a strict maximum (e.g., 10-50) for `limit` before invoking search, and consider server-side rate limiting for this endpoint.

## Changes

- `lib/Controller/UserApiController.php` (modified)

### DIFF
--- a/lib/Controller/UserApiController.php
+++ b/lib/Controller/UserApiController.php
@@ -55,6 +55,7 @@ class UserApiController extends ApiController implements ISessionAwareController
 
 		if (!$this->getSession()->isGuest()) {
 			// Add other users to the autocomplete list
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Thanks @tomaioo 🙏

PR #8507, opened 2026-04-24
fix(workspace): reuse WorkspaceService file lookup in direct()

joshtrichards:
## 📝 Summary

The `direct()` action had a private `getFile()` helper that duplicated file-discovery logic already implemented in `WorkspaceService::getFile()`.

This caused `direct()` to behave differently from `folder()` and `publicFolder()` in one important way: the private helper used `continue` rather than an early `return`, so it returned the *last* matched file in the priority-ordered filename list rather than the first. `WorkspaceService::getFile()` correctly returns on the first match.

By switching `direct()` to `WorkspaceService::getFile()`, it now also benefits from the service’s cache-based lookup and storage-invalid handling.

### Changes

- Remove `WorkspaceController::getFile()`
- Replace its use in `direct()` with `$this->workspaceService->getFile()`
- Remove the now-unused `use OCP\Files\File` import
- Collapse the redundant legacy `@copyright`/`@author`/`@license` docblock into
  the existing SPDX header

### Testing

Add minimal unit coverage for `WorkspaceService::getFile()` to verify:

- first-match priority is preserved
- directory cache entries are ignored
- invalid storage returns `null`

`direct()` now reuses the same lookup implementation as `folder()` and `publicFolder()`.

P.S. The namespaces used for our unit test classes in this repo are inconsistent; not a functional problem but some are in a clearly dedicated "test" space while others use production code namespaces.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Controller/WorkspaceController.php
+++ b/lib/Controller/WorkspaceController.php
@@ -1,31 +1,10 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
- */
-
-declare(strict_types=1);
-/**
- * @copyright Copyright (c) 2019 Julius Härtl <jus@bitgrid.net>
- *
- * @author Julius Härtl <jus@bitgrid.net>
- *
- * @license GNU AGPL version 3 or any later version
- *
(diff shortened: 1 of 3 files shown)


mejo- (approved):
Thanks @joshtrichards, code changes make sense and good catch 💙

PR #8517, opened 2026-04-27
fix(editorApi): fix callbacks on editor events

mejo-:
The new Editor.js wrapper component that got introduced in the indexedb PR broke event propagation from Editor.vue component to editorApi consumers.

Fixes loading editor and updating read-only content in Collectives.

The `resolved` emit was a leftover (dead code) so I removed it.

### 🏁 Checklist

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -7,7 +7,7 @@ import { defineComponent, h, nextTick, ref, watch } from 'vue'
 import Editor from './Editor.vue'
 
 export default defineComponent({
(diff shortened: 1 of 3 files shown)


src/components/Editor.js
silverkszlo: shouldn't `'create:content', 'update:content'` be added here?

mejo-: Good catch! Fixed now.

silverkszlo (approved):

PR #8520, opened 2026-04-27
fix(links): allow to transform inline links to previews

### LINKED ISSUE #8331: Link options: preview/text options visible for inline links
**Describe the bug**
The options to toggle a link between preview and text type is visible for links even if they're inline links, i.e. not alone in an otherwise empty paragraph. Trying to change such a link to preview doesn't do anything though.

**To Reproduce**
1. Create a link surrounded by other text
2. Click on the link, open the three-dot menu in the link bubble
3. See the preview options
4. Select "Show link preview", nothing happens

**Expected behavior**
I'm unsure what's the best fix:

1. Remove the preview options from the three-dot menu when the link is inline. This means inconsistent UI for inline and stand-alone links
2. Allow to transform inline links into a preview, which will split the paragraph and insert the preview as a new block node. This might be unexpected as well, but people always can undo this change.

I tend to prefer the second option.

@marcoambrosini @max-nextcloud what do you think?

**Screenshots**
Image

mejo-:
Fixes: #8331

#### 🖼️ Screenshots

| 🏡 After |
|--- |
| recording |

### 🏁 Checklist

### DIFF
--- a/src/nodes/Preview.js
+++ b/src/nodes/Preview.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { Node, getNodeType, isNodeActive } from '@tiptap/core'
+import { Node, getMarkRange, getNodeType, isNodeActive } from '@tiptap/core'
 import { VueNodeViewRenderer } from '@tiptap/vue-2'
 import { domHref, isLinkToSelfWithHash, parseHref } from './../helpers/links.js'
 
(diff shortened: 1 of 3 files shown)


src/nodes/Preview.js
```diff
@@ -78,15 +78,49 @@ export default Node.create({
...
 								this.name,
 								previewAttributesFromSelection(state),
 							)
 							.run()
-					)
+					}
+
+					if ($from.parent.type.name !== 'paragraph') {
+						return false
+					}
```
max-nextcloud: So this does not work in headings - but for example in lists it would work because the list items contain paragraphs, right?

max-nextcloud: Solving this by also allowing the conversion in headings.

PR #8526, opened 2026-04-29
Center inserted images

Jerome-Herbinet:
### 📝 Summary

This PR centers the inserted images and aligns them with their captions, which are already centered by default.
The issue was not apparent with large images that spanned the full width of the editing and display area.
However, with small images, the display was inconsistent.

This improvement will also have a positive impact on, for example, the content of the Collectives pages.

⚠️ **Not tested : Test from reviewers is required.**

#### 🖼️ Screenshots

| Before | After |
| ---|--- |
|  |  |

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -511,6 +511,7 @@ export default {
 
 .image__main {
 	max-height: calc(100vh - 50px - 50px);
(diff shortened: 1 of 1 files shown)


silverkszlo (approved):
looks good, thank you! :)

PR #8534, opened 2026-05-06
chore(performance): use async components properly

max-nextcloud:
No need to await the import the moment we import the component.
Vue handles this fine when the component is actually needed.


mejo- (approved):
Nice 🚀 

Shall we backport this to stable33 at least?

PR #8566, opened 2026-05-11
fix(orderdList): preserve non-1 start numbers

### LINKED ISSUE #4828: Ordered list resets starting number unexpectedly
**Describe the bug**
The ordered list I created started at 42. and when I saved it, the numbers in the list looked good. The next day, the list was reset and started at 1. 

I have no idea why or how it happened. 

**To Reproduce**
1. Create a readme.md for a folder. 
1. Type 42. and space, add text. 
2. Click 'Enter' and an ordered list is created with the next line starting at 43. 
3. Create a few more list items. 
4. Save file. 
5. The list should now appear correctly in the top of the folder. 
6. something happens (?)
7. Checking back the list now starts at 1. 

**Expected behavior**
The list numbering doesn't change. 

**Server details:**
- Nextcloud version: 27.0.2

mejo-:
Also fixes preserving `0` as first list number.

Fixes: #4828

### 🏁 Checklist

### DIFF
--- a/src/nodes/OrderedList.ts
+++ b/src/nodes/OrderedList.ts
@@ -3,6 +3,9 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+import type { Node } from '@tiptap/pm/model'
+import type { MarkdownSerializerState } from 'prosemirror-markdown'
+
 import { OrderedList as TiptapOrderedList } from '@tiptap/extension-list'
 import { toggleListCommand } from '../commands'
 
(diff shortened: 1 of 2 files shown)


src/nodes/OrderedList.ts
```diff
@@ -23,6 +26,18 @@ const OrderedList = TiptapOrderedList.extend({
...
+	// @ts-expect-error - toMarkdown is a custom field not part of the official Tiptap API
+	toMarkdown(state: MarkdownSerializerState, node: Node) {
+		const start = node.attrs.start ?? 1
+		const maxW = String(start + node.childCount - 1).length
+		const space = state.repeat(' ', maxW + 2)
+		state.renderList(node, space, (i) => {
+			const nStr = String(start + i)
+			return state.repeat(' ', maxW - nStr.length) + nStr + '. '
+		})
+	},
```
max-nextcloud, line 40: This looks very much like the prosemirror-markdown version only that our attr is called start while theirs is called order.

https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/to_markdown.ts#L101-L109

Did you consider renaming the attr / having a duplicate one / calling the prosemirror code from here?

mejo-: You're right, but the attribute inherited by Tiptap is `start`. We'd have to add the `order` attribute to our node extension as well as overwrite `renderHTML()`, `addInputRules()` and many more when we want to rename the attribute to `order` for us.

Also, I replaced `node.attrs.order || 1` with `node.attrs.start ?? 1` to allow lists to start with `0. `.

I opened https://code.haverbeke.berlin/prosemirror/prosemirror-markdown/pulls/130 now to fix the latter in ProseMirror upstream, let's see whether it gets accepted. For the meantime I still suggest to keep the solution as is. If the ProseMirror PR gets accepted we indeed could add a new attribute `ordered` with `rendered: false` that mirrors the current `start` attribute. What do you think about that?

max-nextcloud: Thanks for digging into this and opening the upstream PR!

max-nextcloud (approved):

PR #8571, opened 2026-05-12
fix(links): allow to pass custom link handler into editor and use it

mejo-:
### 📝 Summary

The return of the link handler 👻
    
Required as we want to pass a link handler from Collectives to be used for the "Open link" button and when opening a link via Ctrl-click.

### 🏁 Checklist

### DIFF
--- a/playwright/e2e/conflict.spec.ts
+++ b/playwright/e2e/conflict.spec.ts
@@ -335,12 +335,14 @@ test('conflict dialog is sticky when scrolling', async ({
 	const pushPromise = page.waitForRequest(/push/)
 	await editor.typeHeading('Long content\n')
 	await pushPromise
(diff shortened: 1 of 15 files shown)


max-nextcloud (commented):
Looks great. Now we have two ways of injecting the link hander:

By hand in `Editor.vue` - and with the `useOpenLinkHandler` mixin in `PreviewOptions` and `LinkBubbleView`.

Could you turn them into a composable and use that everywhere for consistency?

mejo-:
> Could you turn them into a composable and use that everywhere for consistency?

Very good idea, done now in eb33685871d69caa8e82dde6771ebf12505d1fb0

max-nextcloud (approved):

PR #8599, opened 2026-05-16
fix: treat log files as plaintext

### LINKED ISSUE #5253: Color markings for .log files, not for .txt files
**Describe the bug**
Allright guys, I'm at a loss. I've spend the better part of today trying things, but I'm not getting this done. My problem is that Text opens my .log files correctly, but with colored text and also parts are italic. It recognizes the .log files as plain text correctly, because I have added the log file type as text/plain to mimetypemapping.json. So no rich editing, like it's supposed to do. But it appears to be formatting the text in some way anyway. The .log files are essentially just text files, but with a .log extension. If I change the extension in Nextcloud to .txt, the file loads correctly as plain black text without any formatting. 

**To Reproduce**
1. Open file with .log extension.
2. See parts of text are colored, and some text in italic and maybe some other formatting. 
3. Close file, and change extension to .txt.
4. Open file again, and behold! A plain text file without any formatting. 

**Expected behavior**
Any file type that has been added as [text/plain] should be shown just like that: as plain text. (right?)

**Screenshots**
_Log file but with .txt extension_
Schermafbeelding 2024-01-13 171509
_Log file but with .log extension_
Schermafbeelding 2024-01-13 171425

**Server details:**
- Nextcloud version: 28.0.1
- PHP Version: 8.2.13
- Database: Mysql 8.2.0

**Client details:**
 - OS: Windows, Linux
 - Browser: Chrome
 - Browser version: 120.0.6099.217
 - Device: Laptop, phone, desktop

#### Nextcloud log (data/nextcloud.log)
Nothing is showing in the logs

#### Browser log
I'll add this when needed.

SSDWGG:
## Summary
- map `.log` files to plaintext highlighting
- add a regression test for the extension mapping

Fixes #5253

## Tests
- npx vitest run src/tests/helpers/mappings.spec.ts


mejo-:
@SSDWGG thanks as well for this contribution. As far as I can see, Text doesn't open `.log` files at all per default. So I wonder whether it's useful to add syntax highlighting conditions for it. Can you explain your usecase a bit?

Regarding the unit test, I think it can be dropped in any case. It doesn't test much except that the object key you introduced to `mappings.ts` has the value your put there.

PR #8619, opened 2026-05-20
chore/upgrade vue 3

max-nextcloud:
- upgrade dependencies to vue3
- replace new Vue with createApp
- `<var>.sync` to `v-model:<var>`
- fix imports
- Vue.observable to reactive
- from beforeDestroy() to beforeUnmount()
- key attribute in v-for component

## Todos

* [x] Table action buttons broken
* [x] Editor and Menubar alignment in Collectives broken
* [x] Menubar actions keep getting reloaded endlessly
* [x] Check if InfoBar properties from Collectives are still reactive
* [x] Check if collaboration cursor from other session members show their username
* [x] `$on` no longer exists in Vue 3. We still use it in DirectEditing
* [ ] Test DirectEditing
* [x] Table doesn't spawn full width. Claude says:
      <details>
      image
      image
      </details>


src/components/Editor/MediaHandler.vue
```diff
@@ -173,23 +173,28 @@ export default {
...
-				true,
-				undefined,
-				this.startPath,
-			)
+			const filePicker = getFilePickerBuilder(t('text', 'Insert an attachment'))
+				.startAt(this.startPath)
+				.allowDirectories(false)
+				.setMultiSelect(false)
+				.setButtonFactory((nodes) => [{
+					label: t('text', 'Choose'),
```
max-nextcloud: The [design guidelines](https://docs.nextcloud.com/server/latest/developer_manual/design/writing.html#button-and-action-labels) recommend:
> Use verb + noun for buttons that trigger an action. The noun can be omitted when context makes it obvious.

I'd propose `Insert file`. I'm also fine with `Insert`. `Choose` is very generic and does not say anything about what will happen.

src/helpers/files.js
```diff
@@ -117,22 +119,25 @@ export const FilesWorkspaceHeader = {
...
-		FilesHeaderRichWorkspaceInstance = createApp(FilesHeaderRichWorkspaceView, {
-			content,
-			hasRichWorkspace,
-			path,
-		}).$mount(el)
 
+		// Create a new instance of the RichWorkspace component
+		FilesHeaderRichWorkspaceApp = createApp({
+			render: () => h(FilesHeaderRichWorkspaceView, { ...FilesHeaderRichWorkspaceState }),
+		})
```
max-nextcloud, line 138: I wonder if we could do:
```js
FilesHeaderRichWorkspaceApp = createApp(FilesHeaderRichWorkspaceView, FilesHeaderRichWorkspaceState)
```

Doesn't the deconstruction and rebuilding of the second argument break reactivity? Or is that the point?

max-nextcloud: https://vuejs.org/guide/essentials/reactivity-fundamentals.html#limitations-of-reactive

mejo- (approved):
Ready for merge from my side 🚀

src/helpers/files.js
mejo-: Unfortunately this didn't work, so I reverted the change. Apparently passing the reactive proxy object directly as `rootProps` to `createApp` breaks its reactivity.

src/TextEditorEmbed.ts
```diff
@@ -0,0 +1,225 @@
...
+		// eslint-disable-next-line no-console
+		console.warn(
+			'%c%s\n%c%s\n%s',
+			styleBold,
+			intro,
+			styleItalic,
+			introChrome,
+			introFirefox,
+		)
+	}
```
enjeck: This is the same functionality in `src/helpers/debug.ts#debugYjs()`? Do we need both?

src/composables/useConnection.ts
```diff
@@ -78,14 +82,28 @@ export function provideConnection(
...
 	const connection = inject(connectionKey)
 	const openData = inject(openDataKey)
 	return { connection, openData }
 }
 
+/**
+ * Mimic axios error for a conflict while creating the session.
+ *
+ * This will be emitted from the SyncService
+ * and trigger conflict handling in Editor.vue
```
enjeck:
```suggestion
 * and trigger conflict handling in CollaborativeEditor.vue
```

enjeck (commented):
non blocking comments:

src/components/Menu/ToolBarLogic.js
```diff
@@ -92,8 +92,9 @@ export default defineComponent({
 				this.$refs.remainingEntries?.focusButton?.()
 			} else {
 				// The ref is in no order (ordered by the time they needed to mount), so we need to order them like they are shown on the menu
-				const entries = [...this.$refs.menuEntries].sort((a, b) => this.visibleEntries.findIndex(({ key }) => key === a.$vnode.data.key)
-					- this.visibleEntries.findIndex(({ key }) => key === b.$vnode.data.key))
+				const entries = [...this.$refs.menuEntryRefs]
```
benjaminfrueh: ~~I think this should be `this.menuEntryRefs` here now with the function ref as it doesn't register anything in $refs.~~ (Already fixed in a later commit)

Also sorting should no longer be necessary, as it is stored by `v-for` index in display order, unlike before in mount order.

Maybe this whole else block could just become:
```js
this.menuEntryRefs[this.activeMenuEntry]?.focusButton()
```

src/nodes/Table/TableView.vue
```diff
@@ -213,11 +213,16 @@ export default {
 		position: absolute;
 		left: 0;
 		bottom: calc(2 * var(--default-grid-baseline));
-		width: calc(100% - var(--clickable-area-small) - 4px) !important;
+		// width: calc(100% - var(--clickable-area-small) - 4px) !important;
```
benjaminfrueh: Maybe can delete this commented-out line?

src/createTable.ts
```diff
@@ -27,17 +27,17 @@ interface CreateTableOptions {
  *
  * @param options Options for the created table
  */
-export function createTable(options: CreateTableOptions): TextEditorEmbed {
-	const PlainTableContentEditor = defineAsyncComponent(() => import('./components/Editor/PlainTableContentEditor.vue'))
+export async function createTable(options: CreateTableOptions): TextEditorEmbed {
```
benjaminfrueh: Async functions always return a promise in Typescript. With this being async the return type should be `Promise<TextEditorEmbed>` now instead of `TextEditorEmbed` - did typescript compile with this?

For the caller in whiteboard async should be fine as whiteboard already awaits this function.

src/views/FilesSettings.vue
```diff
@@ -4,7 +4,7 @@
 -->
 <template>
 	<div id="files-setting-richworkspace">
-		<NcCheckboxRadioSwitch v-model="showWorkspace" @update:checked="toggle">
+		<NcCheckboxRadioSwitch v-model="showWorkspace" @update:modelValue="toggle">
```
benjaminfrueh, line 7: This is two listeners on the same event: `v-model` is just a shorthand for `modelValue` + handler `@update:modelValue`. Should work as long as they are in this order in the template. `toggle()` reads `this.showWorkspace` and relies on it already being updated.

Maybe it would be safer to use the event payload directly, the toggle function could become:
```ts
toggle(enabled) {
	emit(enabled ? 'Text::showRichWorkspace' : 'Text::hideRichWorkspace')
	axios.post(generateUrl('/apps/text/settings'), {
		key: 'workspace_enabled',
		value: enabled ? '1' : '0',
	})
}
```

src/composables/useLinkFile.ts
```diff
@@ -37,17 +37,9 @@ export function useLinkFile({
...
-		const node = await pickFile(startPath.value).catch(() => {
-			// TODO: check if this works.
-			// return focus to the trigger.
-			if (event.target instanceof HTMLElement) {
-				event.target.focus()
-			}
-		})
+	const linkFile = async () => {
+		const node = await pickFile(startPath.value)
 		if (!node) {
```
benjaminfrueh, line 44: The catch was removed and it seems like `pickFile` can only succeed or reject on dismissal. This guard should currently not be reachable.

Maybe in the `pickFile()` function we could wrap the `filePicker.pick()` in a try/catch and return null.
```ts
let file
try {
	file = await filePicker.pick()
} catch {
	return null
}
```
That would make the guard here more meaningful again.

src/views/DirectEditing.vue
```diff
@@ -128,6 +128,7 @@ export default {
 		async close() {
 			this.saving = true
 			setTimeout(async () => {
+				await this.$refs.editor.save?.()
```
benjaminfrueh: If `save()` rejects here then `close` and `callMobileMessage('close')` never run. The `beforeUnmount` in `CollaborativeEditor` handles this with a timeout to handle rejection and hanging. Maybe the same pattern could also be used here to make sure the close always completes? Alternatively the save could be wrapped in a try/catch block.

src/TextEditorEmbed.ts
mejo-: Good catch. Actually it's dead code and the one in `src/helpers/debug.ts` is all we need.

src/components/Menu/ToolBarLogic.js
mejo-: Wow, good catch. Indeed the sorting is no longer necessary. Fixed.

src/composables/useLinkFile.ts
mejo-: You're right. Refactored now as suggested by you.

src/views/DirectEditing.vue
mejo-: Good catch! Refactored and fixed as well.

src/composables/useLinkFile.ts
```diff
@@ -57,8 +58,13 @@ export function useLinkFile({
  *
  * @param startPath path to start the file picker on
  */
-async function pickFile(startPath: string) {
-	const filePicker = buildFilePicker(startPath)
+async function pickFile(startPath: string): Promise<Node | undefined> {
+	let filePicker
+	try {
+		filePicker = buildFilePicker(startPath)
```
max-nextcloud: Now I noticed that @benjaminfrueh proposed to wrap the `filePicker.pick()` in a try catch rather than the build:
> Maybe in the pickFile() function we could wrap the filePicker.pick() in a try/catch and return null.

I think that makes more sense as the building should be deterministic, whereas the `.pick()` is the user interaction.

mejo-: Yep, in 08cf15740ef913fdbb28bbfe834a6ce6b6bef145 I changed the try/catch to contain both `buildFilePicker()` and `filePicker.pick()`.

PR #8723, opened 2026-06-09
fix(autofocus): persist cursor positions in localStorage

### LINKED ISSUE #8699: Remember where the user left off

I use the text editor to create long-running todo lists for projects. I keep them open in tabs on my laptop for planning and I open them in the NextCloud app on my phone to check things off as I do them (and to make small edits/additions "in the field").

The problem is: when the app requires a "reconnect" (which effectively happens every time I open it), it jumps you back up to the top of the document. I then have to scroll all the way down the todo list to pick up where I left off.

It would be great if the text editor remembered where the user's cursor was in a given session, and then bring them back to that position when they reconnect.

Ideally this would be device-specific, not user-specific, so that the same user can have the same document open in multiple devices in different cursor positions (eg: at the end of a todo list on my desktop, in the middle of a todo list on my phone). I imagine that means the position would be stored locally in the client, rather than on the server, so each device can maintain its own "memory" of where it left off.

I searched the queue for existing issues, but couldn't find one - although the terminology is a bit subjective, so I may have missed one. Feel free to close this if an issue already exists.

mejo-:
Fixes: #8699

This makes the autofocus feature much more usable, as it persists the cursor position per device, not only per browser tab.

Also migrates Autofocus extension to TypeScript.

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -85,7 +85,7 @@ import { Collaboration } from '@tiptap/extension-collaboration'
 import { useElementSize } from '@vueuse/core'
 import { defineComponent, inject, ref, shallowRef, watch } from 'vue'
 import { Doc, logUpdate } from 'yjs'
-import Autofocus from '../extensions/Autofocus.js'
+import Autofocus from '../extensions/Autofocus.ts'
 
 import { provideEditor } from '../composables/useEditor.ts'
 import { provideEditorFlags } from '../composables/useEditorFlags.ts'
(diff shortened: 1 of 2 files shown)


max-nextcloud (approved):
Code changes look good. I wonder if the playwright failures are related though.

max-nextcloud:
Saw the same timeout failure elsewhere. So I suspect it's not related.

max-nextcloud:
Playwright test is known to fail.

PR #8753, opened 2026-06-16
Fix: show Assistant icon if only translation provider available

janepie:


blizzz:
> /compile

no compile in text needed – assets are automatically build after merge

src/components/Menu/entries.ts
```diff
@@ -118,7 +118,7 @@ export const getAssistantMenuEntries = (): MenuEntry[] => {
 		component: AssistantAction,
 		priority: 7,
 	}
-	const hasAssistantTaskTypes = loadState('text', 'taskprocessing', []).length > 0
+	const hasAssistantTaskTypes = loadState('text', 'taskprocessing', []).length > 0 || loadState('text', 'translation_available', false)
```
blizzz: is this complete? My gut feeling says, the next other thing that is neither taskprocessing nor translation will potentially end up in the same state?

janepie:
> no compile in text needed – assets are automatically build after merge

Oh good to know!
For testing: task types are cached so if you disable them it may take a moment for the change to reach text

blizzz:
> > no compile in text needed – assets are automatically build after merge
> 
> Oh good to know! For testing: task types are cached so if you disable them it may take a moment for the change to reach text

What is `a moment`? Generally acceptable, and it is highly unlikely that someone just sits there and toggles task types all day long.

src/components/Menu/entries.ts
janepie: Well that's a problem of the hardcoded task type list in https://github.com/nextcloud/text/blob/678b78d451c5557269998e01d42c9b98d7e949d6/lib/Service/InitialStateProvider.php
I actually didn't check if we have new ones that are not in there. Maybe it would be nicer to filter for task types that have Text as input? But that would be a bigger change, not for tonight :D

janepie:
> What is `a moment`? 

1 min :)

blizzz (approved):
makes sense to me

janepie:
Playwright tests are breaking everywhere and should be unrelated

PR #8796, opened 2026-06-30
fix(sync): reset #sending flag when no active connection on sendStepsNow

omnizs38:
### 📝 Summary

* Resolves:

When `sendStepsNow()` is called without an active connection, it sets `this.#sending = true` and then returns early — but never resets the flag to `false`.

The `sendSteps()` interval guards its body with `!this.#sending`, so once this early-return path is hit, the interval fires forever but `sendStepsNow()` is never called again. All locally typed changes accumulate in the outbox and are silently never pushed to the server, even after the connection recovers resulting in data loss.

**Fix:** reset `this.#sending = false` before the early return.

### 🏁 Checklist

### 🤖 AI (if applicable)


src/services/SyncService.ts
```diff
@@ -236,6 +236,7 @@ class SyncService {
 			this.bus.emit('stateChange', { dirty: true })
 		}
 		if (!this.hasActiveConnection()) {
+			this.#sending = false
```
mejo-, line 239: This looks like a sensible fix to me. We return early but don't reset `this.#sending`.

src/tests/services/SyncService.spec.ts
```diff
@@ -42,6 +42,17 @@ const initialData = {
 const openResult = { connection, data: initialData }
 
 describe('Sync service', () => {
+	it('resets #sending when there is no active connection', async () => {
```
mejo-, line 45: The test looks useless to me, I'd drop it.

mejo- (changes requested):
@omnizs38 did you run into an actual bug that you're fixing here, or was it a random finding by AI?

The fix in SyncService.ts looks sensible to me. Do you agree @max-nextcloud?

src/tests/services/SyncService.spec.ts
omnizs38: Okay, but I'm not at home right now. When I am, I'll definitely make some edits based on your suggestions. Thank you very much!

mejo-: @omnizs38 can you simply drop the test altogether from the PR?

Also, out of curiosity:

> Did you run into an actual bug that you're fixing here, or was it a random finding (by AI)?

max-nextcloud:
> 

> The fix in SyncService.ts looks sensible to me. Do you agree @max-nextcloud?

Yes. Also makes sense to me.

mejo-:
I opened #8963 as I lack permissions to push into the branch of this PR. The updated PR contains just the fix in SyncService, dropping the useless test. Thanks again for your contribution @omnizs38, much appreciated.

Issue #8774, opened 2026-06-22
Auto-scale embedded images


alimmroth:
### Use Case
Embedded images seem not to be scaled automatically. Also the frame around it is in a fixed ratio, which wastes a lot of space when embedding very wide images (like screenshots).

### Proposal

- Automatically scale images, so that the whole image is visible.
- Automatically adjust the frame around the embedded image
- Extra: I think the visible frame line is unnecessary and not looking very good and could be removed. Frameless embedded pictures combine with the background much more nicely.

Related issue, going further, asking for the image to be freely scalable by the user: #4556 

#### Nextcloud right now (wide screenshot with visible frame)

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/2da38d0b-7d0d-4464-b1d2-320fb888b389" />

mejo-:
Thanks for reporting the issue @alimmroth.

What do you mean with "Automatically scale images, so that the whole image is visible."? I wasn't able to reproduce a situation where big images are cut off, they're already auto-scaled thanks to `max-width: 100%` in my tests. Also I cannot see a cut off image in your screenshot.

Also the frame you mentioned is already gone at least starting with Nextcloud 33. Which Nextcloud version where you using when testing?

alimmroth:
Hello mejo,
I found out, this only happens when inserting the image via the file picker. With drag-and-drop, auto scaling works.

You're right, in 33.0.6 I can also reproduce, there's no frame und the ratio is auto-adjusted.

### 33.0.3
<img width="450" alt="Image" src="https://github.com/user-attachments/assets/be1a21ce-aec5-4b88-afde-4c85ec3cda81" />

### 33.0.6
<img width="450" alt="Image" src="https://github.com/user-attachments/assets/91030e86-09c2-4ae9-899f-81cdaa2e37bc"/>
<img width="450"  alt="Image" src="https://github.com/user-attachments/assets/631fbdba-3561-4fc9-8b72-99e0feea2ae7" />

mejo-:
Now I can reproduce: this is about linking an image via "Link to file or folder", not about adding the image as attachment. This probably needs to get fixed in ReferenceFileWidget.vue in Files app. Or maybe in Viewer, as the styles that force the image to get displayed full-size seem to stem from `src/components/Images.vue` in Viewer 🤔

PR #8800, opened 2026-07-02
fix(image): auto-scale embedded images to fit their container

omnizs38:
Embedded images were not reliably scaled to fit the available width, and tall images could overflow the surrounding frame instead of being shrunk to be fully visible.

Constrain the rendered image with max-width: 100% and object-fit: contain while letting width and height stay auto, so the whole image is always visible and the frame adjusts to the image's real aspect ratio without reserving fixed empty space.

### 📝 Summary

* Resolves:  The PR resolves issue [Auto-scale embedded images and adjust frame #8774
](https://github.com/nextcloud/text/issues/8774)

Specifically, it addresses the first two of the three points in that issue:

1. Auto-scale embedded images so the whole image is visible images now get max-width: 100% + object-fit: contain with auto width/height.
3. Auto-adjust the frame around the image — height: auto lets the frame collapse to the image's real aspect ratio instead of reserving fixed empty space.

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -516,7 +516,11 @@ export default {
 }
 
 .image__main {
+	max-width: 100%;
 	max-height: calc(100vh - 50px - 50px);
+	width: auto;
+	height: auto;
+	object-fit: contain;
 	margin: auto;
 }
 
@@ -536,6 +540,7 @@ export default {
 
 	img {
 		max-width: 100%;
+		height: auto;
 	}
 
 	&:hover {


mejo-:
Thanks for the PR @omnizs38. I tested it locally on latest Text main branch but was unable to reproduce the original issue even without the PR. Could you elaborate on how to reproduce the problem you aim to fix with this PR?

In my tests, image width is already limited to 100% thanks to the `.image__view img { max-width: 100%; }` rule in `ImageView.vue`.

mejo-:
This doesn't solve the problem it intends to solve. Feel free to adjust the PR @omnizs38 and we'll reopen it. For now I'll close it.

PR #8813, opened 2026-07-04
fix(ImageView): use `a` element for clickable items

### LINKED ISSUE #8801: Hovering attachments doesn't change cursor to pointer
**Describe the bug**
When hovering an attachment (not an image), the cursor doesn't switch to pointer. This gives the impression that the attachment is not clickable, which leads to confusion.

**Server details:**
- Nextcloud version: 33.0.6

mejo-:
Fixes the cursor not being a pointer on a clickable item.

Also move the click handler for non-image attachments from the full-width container box to the attachment box with a visible boundary.

Fixes: #8801 

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/nodes/ImageView.vue
+++ b/src/nodes/ImageView.vue
@@ -23,9 +23,10 @@
 						<div
 							v-if="isMediaAttachment"
 							contenteditable="false"
-							class="media"
-							@click="handleAttachmentClick">
-							<div class="media__wrapper">
+							class="media">
+							<a
+								class="media-wrapper media-wrapper__attachment"
+								@click="handleAttachmentClick">
 								<img
 									v-show="loaded"
 									:src="imageUrl"
@@ -36,7 +37,7 @@
 									<span class="name">{{ alt }}</span>
 									<span class="size">{{ attachmentSize }}</span>
 								</div>
-							</div>
+							</a>
 							<div v-if="showDeleteIcon" class="buttons">
 								<NcButton
 									:aria-label="t('text', 'Delete this attachment')"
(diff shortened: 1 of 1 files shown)


src/nodes/ImageView.vue
```diff
@@ -23,9 +23,10 @@
 						<div
 							v-if="isMediaAttachment"
 							contenteditable="false"
-							class="media"
-							@click="handleAttachmentClick">
-							<div class="media__wrapper">
+							class="media">
+							<div
+								class="media__wrapper clickable"
+								@click="handleAttachmentClick">
```
max-nextcloud: How about using an `<a>` tag instead. That seems semantically closer to what the element is.
The `href` could even point to the url `handleAttachmentClick` opens if it does not open the viewer.

src/nodes/ImageView.vue
```diff
@@ -53,7 +54,7 @@
 								v-show="loaded"
 								:src="imageUrl"
 								:alt="alt"
-								class="image__main"
+								class="image__main clickable"
```
max-nextcloud: Maybe also turn the surounding `<div>` into an `<a>` and move the click handler there?

src/nodes/ImageView.vue
mejo-: Good point, adjusted now.

src/nodes/ImageView.vue
mejo-: Same. I had to adjust the DOM and CSS class structure slightly to retain the current behaviour that non-image attachments are left-aligned while images are center-aligned.

max-nextcloud (approved):
Code looks good. Did not test though.

PR #8823, opened 2026-07-06
chore(refactor): aviod extending serializer options

### LINKED ISSUE #8821: refactor reference definitions in Markdown serializer
In #8717 we introduced a `referenceDefinitions` object that is passed in the options of the prosemirror-markdown serialize option. In the Vue 3 PR we migrated the markdown extension to Typescript which revealed that `referenceDefinitions` is not a valid object property. We made Typescript ignore this for now, but this needs to be fixed properly.

One idea is to instead of passing and filling the `referenceDefinitions` map in the serializer introduce something like a `getReferencesFromContent()` function that is called *after* the Markdown serializer.

max-nextcloud:
Only tested with the automated tests for now - but they cover the case pretty well.
* fixes #8821


mejo- (approved):
Looks good, thanks a lot 🙏

PR #8840, opened 2026-07-08
Add footnotes support

### LINKED ISSUE #283: Footnote support
In GitLab by @rotdrop on Feb 1, 2022, 15:52

Hi,

I'm currently trying to migrate a few DokuWiki pages to NC collectives. However, I am missing support for footnotes. Of course, this would also require corresponding changes in the text-editor, see https://github.com/nextcloud/text/issues/2142

There already is a plugin for markdown-it

https://github.com/markdown-it/markdown-it-footnote

which supports the syntax described here

https://www.markdownguide.org/extended-syntax#footnotes

But then, footnotes are non-standard in MD. Still it would be nice.

### LINKED ISSUE #2142: Footnote support would be nice

mejo-:
### 📝 Summary

* Resolves: #2142
* Resolves: nextcloud/collectives#283

Some implementation details:
- In UI, the reference IDs are hidden on purpose. Instead, all references and their corresponding footnotes get a sequential ordinal that is displayed. The footnotes are ordered by their ordinal. This is how Obsidian and others do it as well and it avoids confusion regarding manually used and out-of order numeric references.
- References can be inserted via menubar entry, via `[^]` input rule or via <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>F</kbd> shortcut.
- Deleting a reference also removes it's corresponding footnote *if no other reference to the footnote exists*.

#### 🖼️ Screenshots

Editor | Markdown
---|---
image | grafik

### 🏁 Checklist

### 🤖 AI (if applicable)


marcoambrosini (commented):
Very nice! A couple of comments from a quick first pass: 

- I would avoid using a container and gray out the text instead
- I would put this in a new add menu in the toolbar, same as the + in the document, where we can access everything from one bucket. And with a search-bar on top of the list

mejo-:
> I would avoid using a container and gray out the text instead

Good idea, implemented now. See updated screenshot.

> I would put this in a new add menu in the toolbar, same as the + in the document, where we can access everything from one bucket. And with a search-bar on top of the list

I added it to the menubar now, which will automatically add it to the `+` (add) menu in the document.

I'm unsure about the icon to use though. For now I used https://pictogrammers.com/library/mdi/icon/asterisk/ which seems the most fitting to me.

Other icons I considered: 
- https://pictogrammers.com/library/mdi/icon/numeric-1-box-outline/
- https://pictogrammers.com/library/mdi/icon/format-superscript/ looks too mathematical to me
- https://pictogrammers.com/library/mdi/icon/pin/ might also be an option
- https://pictogrammers.com/library/mdi/icon/bookmark/ not really fitting

src/nodes/FootnoteReference.ts
```diff
@@ -43,6 +58,100 @@ const FootnoteReference = Node.create({
...
+				const footnoteType = state.schema.nodes.footnote
+				const paragraphType = state.schema.nodes.paragraph
+
+				let c = chain()
+					.insertContent({ type: 'footnoteReference', attrs: { referenceId } })
+
+				if (!existingFootnote) {
+					const newFootnote = footnoteType.create({ referenceId }, paragraphType.create())
+					const lastChild = state.doc.lastChild
+					const hasFootnotesBlock = lastChild?.type === footnotesType
```
max-nextcloud, line 99: This relies on the document schema. What if we add something after the footnotes?

max-nextcloud (approved):

src/nodes/FootnoteReference.ts
mejo-: Turns out this becomes more complex than expected, as we cannot do several transactions in a row and still rely on the positions calculated upfront. So this would be refactored to become a single transaction, which rules out using the Tiptap `insertContentAt()` command, which in turn means we have to take care of moving the cursor ourselves. So I decided to leave it as is for now.

PR #8875, opened 2026-07-14
fix(filepicker): hide menu

max-nextcloud:
We use the file picker in two scenarios:
* Linking to a file.
* Adding a file as an attachment.

In both cases creating new folders does not make sense.


mejo- (approved):
Thanks, makes sense.

PR #8876, opened 2026-07-14
chore(refactor): loop over possible footnotes once

max-nextcloud:
I initially cached the `referenceLabels` in case there are multiple `footnotes` nodes.
But that increased complexity quite a bit and seems like an unneccessary optimization.


mejo- (approved):
Much cleaner indeed, thank you ❤️

PR #8884, opened 2026-07-16
upgrade/typescript v6

max-nextcloud:
- **chore(deps): update dependency typescript to v6**
- **chore(deps): limit typescript to v6**


mejo- (approved):
Thanks for tackling this 🙏 Unfortunately it seems like this will need another manual rebase due to a merge conflict.

PR #8891, opened 2026-07-19
[stable32] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


max-nextcloud:
Merging despite RC freeze as ocp updates do not alter the app itself.

PR #8898, opened 2026-07-22
feat: add per-user raw Markdown editing (#3570)

### LINKED ISSUE #3570: Markdown edit source
Sometimes markdown has non-standard elements that Text renders as it thinks is standard.

A button "Edit source" that would allow editing raw markdown.

**Additional context**

soapergem:
### 📝 Summary

* Resolves: #3570

Adds a per-user **"Edit Markdown source"** toggle that lets a user switch the current document into a raw Markdown editor on the fly, without disrupting other collaborators.

**Why the per-user, detached approach.** The rich editor and the plain editor bind to the same shared yjs document but use incompatible ProseMirror schemas (rich node tree vs. a single plain-text document). Two users on different schemas against one shared document would corrupt each other — which is exactly the concern raised in the issue discussion. So raw editing runs on a **detached** plain-text editor (no `Collaboration` extension): the collaborative rich editor stays mounted and connected in the background, so the shared document never sees the plain-text schema and other participants are unaffected.

**Round-trip and conflicts.** On leaving raw mode the edited source is applied back to the live document. If the document drifted in the meantime (concurrent edits from other users), the existing `CollisionResolveDialog` is reused so the user can keep their raw version or the remote one, instead of silently clobbering either. If the user made no changes, or the document did not drift, the edits are applied directly with no prompt.

**Interplay with `rich_editing_enabled`.** The toggle is gated on rich Markdown editing being active, so when an admin has set `rich_editing_enabled=0` (the whole instance already edits raw) the toggle simply does not appear — no "raw within raw", no schema mixing. The mode is ephemeral per-user view state and is not persisted, which deliberately sidesteps the open "user vs. file vs. instance setting" question from the issue.

#### What's included

* `useRawEditing` composable — enter/exit state machine with drift detection
* `RawMarkdownEditor` — detached plain-text editor seeded with the serialized Markdown
* `RawEditingToggle` — overflow-menu entry, gated to rich Markdown editing
* Autosave suppression while detached; unload warning when raw edits are unsaved
* A "Discard changes" affordance to leave raw mode without applying edits

#### 🖼️ Screenshots

**1. Toggle in the ⋯ overflow menu**

Edit Markdown source entry in the editor overflow menu

**2. Editing the raw Markdown source**

The current user is switched to a detached plain-text editor showing the document's raw Markdown. **Rich text** returns to the WYSIWYG view.

Detached raw Markdown source editor

**3. Conflict resolution when the document changed concurrently**

If another collaborator changed the document while it was being edited raw, the existing collision dialog is reused to keep either the local (raw) version or the remote one.

Collision dialog offering to keep local raw edits or the remote version

### 🚧 TODO

- [ ] Confirm desired wording for the menu entry / buttons with maintainers

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` passes; `tsc` clean under TypeScript 6)

#### Testing notes

* **Unit** (`src/tests/composables/useRawEditing.spec.ts`, 7 cases): enter/gating, direct apply when there is no drift, conflict raised on drift, no-op when unchanged, discard, and conflict clearing.
* **End-to-end** (`cypress/e2e/RawEditing.spec.js`, 5 cases): toggling into the source view, applying raw edits back to the rich editor, discarding, and both branches of the conflict dialog when the document changed concurrently. All passing against a local Nextcloud instance.

### 🤖 AI (if applicable)

### DIFF
--- a/cypress/e2e/RawEditing.spec.js
+++ b/cypress/e2e/RawEditing.spec.js
@@ -1,0 +1,99 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { randUser } from '../utils/index.js'
+
+const user = randUser()
+
+describe('Raw Markdown editing', function() {
+	before(function() {
+		cy.createUser(user)
+	})
+
+	beforeEach(function() {
+		cy.login(user)
+		cy.uploadTestFile()
+		cy.visit('/apps/files')
+		cy.openTestFile()
+		// Seed some rich content to round-trip.
+		cy.getContent().type('# Hello{enter}World')
+		cy.getContent().find('h1').should('contain.text', 'Hello')
+	})
+
+	const openRawEditor = function() {
+		cy.getActionEntry('remain').click()
+		cy.contains('Edit Markdown source').click()
+		return cy.get('[data-text-el="raw-markdown-editor"]').should('exist')
+	}
+
+	const rawContent = function() {
+		return cy.get('[data-text-el="raw-markdown-editor"] .ProseMirror')
+	}
+
+	it('shows the Markdown source when toggled on', function() {
+		openRawEditor()
+		rawContent().should('contain.text', '# Hello')
+		// The rich formatting toolbar is hidden while editing raw.
+		cy.get('div[data-text-el="menubar"] .text-menubar__entries').should('not.exist')
+	})
+
+	it('applies raw edits back into the rich editor', function() {
+		openRawEditor()
+		rawContent().type('{selectall}{del}## Changed heading{enter}{enter}body text')
+		cy.get('[data-cy="exitRawEditing"]').click()
+		cy.get('[data-text-el="raw-markdown-editor"]').should('not.exist')
+		cy.getContent().find('h2').should('contain.text', 'Changed heading')
+		cy.getContent().should('contain.text', 'body text')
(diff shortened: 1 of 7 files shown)


max-nextcloud:
Hi @soapergem 

Thanks for your contribution. I like the general approach of a per session single user switch. That seems useful also for inspecting the markdown.

I'm unsure about adding another scenario that runs into a conflict resolution screen. We've observer that it's overwhelming in particular with large files as we currently have no indication of what changed. We want to improve upon that - but we are not there yet.

Technically we could also implement the same approach without a conflict screen and leave the resolution to yjs. We'd need to keep a copy of the ydoc at the start of the editing session. Then compute an yjs update that transforms the ydoc into one that corresponds to the document after the editing session. This update could be applied to the main editing session just like any other yjs update. In cases of small edits that create no actual conflicts this would be ideal. However if people are editing the same section it may lead to counterintuitive results.

I'd be curious what @marcoambrosini thinks from a design point of view.

I also have not looked at the code changes yet at all.

soapergem:
Hi @max-nextcloud. I appreciate the feedback. I could definitely whip up a version that drops the conflict-resolution screen and leans on yjs for the merge instead, based on what you just described. Should I wait for @marcoambrosini 's feedback first? Or should I just continue forward on that path?

max-nextcloud:
@soapergem Let's wait for the feedback to make sure we know what we are aiming for in terms of UX. This may take a while though - as far as I can tell marco is not available right now. I'll also talk to some other folks to gather some feedback.

PR #8901, opened 2026-07-22
feat: inline comments

### LINKED ISSUE #185: Comments in lines
No, but it would be great to have the ability to comment text directly in the line.

Similar to gdocs and word, marking a text and writing a comment on the side.

Linking a specific part of the text in the comment section in the panel on the right could also be a solution, maybe for the start, being able to link headlines `#H2`

Just an idea, might be too ambitious, but maybe sth. for the backlogs.

mejo-:
### 📝 Summary

This PR implement inline point-anchor comments as discussed in #185.

- Comments can be inserted via toolbar submenu, by typing `[?]` or by pressing <kbd>Ctrl</kbd>-<kbd>Alt</kbd>-<kbd>M</kbd>
- A superscript chat bubble icon (💬) marks the comment anchor in the text. Clicking it opens the comment bubble
- Inside the comment bubble, one can add, edit and remove replies
- Comments and footnotes can be hidden to improve legibility via "Hide annotations" in the toolbar submenu
- Resolves: #185

#### Markdown representation

Comments are stored using the footnote syntax already supported by Text, with a comment- prefix to distinguish them from regular footnotes:

```markdown
The quick[^comment-1] brown fox.

[^comment-1]:
    - @[jane](mention://user/jane) *(2026-05-21T10:30Z)*
      Is "quick" the right word here?
    - @[bob](mention://user/bob) *(2026-05-21T12:45Z)*
      I think "fast" is more accurate.
```

Existing Markdown editors with footnote support (Obsidian, Ghostwriter) render the comment definitions as footnotes and make them accessible.

#### 🖼️ Screenshots

Toolbar submenu | Comment bubble | Entering first comment reply | Editing a reply
---|---|---|---
image | grafik | grafik | grafik

#### 🖼️ Screencast

recording

### 🚧 TODO

- [x] ~~Regression: Undo is now always disabled in toolbar~~
- [x] ~~Would be great if adding/editing/deleting replies could be undoable as well~~ edit: turns out it already is :tada:
- [ ] Support to mention users in comments
- [ ] Optionally replace NcRichContenteditable with a stripped-down Text editor for editing comments and replies

### 🏁 Checklist

### 🤖 AI (if applicable)


src/components/Comment/CommentBubbleView.vue
```diff
@@ -0,0 +1,453 @@
...
+			authorLabel: item.attrs.authorLabel,
+			timestamp: item.attrs.timestamp ? new Date(item.attrs.timestamp) : null,
+			body: wrapper.innerHTML,
+			markdownBody: mdSerializer.serialize(item).trim(),
+		})
+	})
+	return result
+})
+
+const isFirstComment = computed(() => items.value.length === 0 || (items.value.length === 1 && !items.value[0].body))
```
silverkszlo: Image

The button only ever renders "Reply" and never "Comment", even when it's the first comment. Claude says, that `isFirstComment` compares serialized HTML ("\<p>\</p>") and not real emptiness, so it's always false and therefore only shows "Reply".

silverkszlo:
Screenshot from 2026-07-28 11-02-36

the two close-icons get very close to each other, when there is multiple comments

src/nodes/Comment.ts
```diff
@@ -0,0 +1,67 @@
...
+			const item = node.child(i)
+			const { author, authorLabel, timestamp } = item.attrs
+			const authorMarkdown = author
+				? `@[${authorLabel}](mention://user/${encodeURIComponent(author)})`
+				: `@${authorLabel || ''}`
+			const ts = timestamp ? ` *(${timestamp})*` : ''
+			return `- ${authorMarkdown}${ts}\n`
+		})
+		state.delim = savedDelim
+		// state.wrapBlock('    ', marker, node, () => state.renderContent(node))
```
silverkszlo: old comment?

silverkszlo:
Adding a very long comment breaks it and one also can not scroll down inside the comment bubble anymore. Maybe it would make sense to set a max number for characters?

Screenshot from 2026-07-28 12-47-07

silverkszlo:
When being a guest, I can not edit previous comments before I enter a guestname, but I can delete them. Is that intentional?

Screenshot from 2026-07-28 13-07-46

src/components/Comment/CommentBubbleView.vue
mejo-: Good catch. Fixed in 1525c9f2d2e593ffe959712ffa84376fbfaa439f.

mejo-:
> the two close-icons get very close to each other, when there is multiple comments

Good catch. I fixed this by switching to `placement: 'left-start'` for the floating-ui bubble. It's cleaner anyway in my opinion.

> Adding a very long comment breaks it and one also can not scroll down inside the comment bubble anymore. Maybe it would make sense to set a max number for characters?

Another good catch. Fixed by setting a `max-height` for the multiline input field.

> When being a guest, I can not edit previous comments before I enter a guestname, but I can delete them. Is that intentional?

Indeed, that's inconsistent. I changed this now that deleting is only possible after adding a guest name as well for consistent UX.

I also did the following:

- Visually highlight the reference of the currently open comment bubble.
- Add navigation buttons to comment bubble to jump to previous/next comment.
- Improve keyboard navigation: 
  - allow to navigate to references via tab
  - open comment bubble on Enter/Space
  - close comment bubble on Esc
  - place cursor in editor content after reference after closing comment bubble via Esc or button

jancborchardt (changes requested):
Very cool! Some first feedback:

- [x] Showing/hiding annotations changes the text flow a little bit, which makes things look unstable/brittle. That might be because the comment emoji is aligned a bit to the top? It’s fine if it’s vertically centered, it would be more important for the text not to move vertically. :) 
- [x] The comment field is quite large / high, which makes it a bit massive and imposing. It would be ok to reduce it to the height of 3.5 lines of text, or even a single line if we use autosize.
- [x] The "Edit comment" and "Delete comment" actions can go into a 3-dot action menu. Then they can have proper text and are not so in-your-face.
- [x] A quick animation moving the comment field down when sending the comment would be nice. Better than an interface jump
- [ ] Right now it is only possible to show a single comment thread, right? Other solutions make it possible to view all comments at the same time, which is especially useful if there are multiple and not really many replies going on.

mejo-:
>  Right now it is only possible to show a single comment thread, right? Other solutions make it possible to view all comments at the same time, which is especially useful if there are multiple and not really many replies going on.

I would like to keep this for a follow-up, as I have to say I find it quite confusing how others display all the comments per default and I think we should think this through properly.

All other comments addressed @jancborchardt :relaxed:

src/components/Menu/utils.js
```diff
@@ -61,7 +62,8 @@ function getKeys(isMobile, { keyChar, keyModifiers }) {
  * @param editor
  */
 function isDisabled(actionEntry, editor) {
-	return actionEntry.action && !actionEntry.action(editor.can(), editor)
+	return (actionEntry.action && !actionEntry.action(editor.can(), editor))
+		|| (actionEntry.isAnnotation && annotationsHidden.value)
```
max-nextcloud: minor nitpick:
I don't like this annotation specific logic in the generic menu handler.

I think the check could be moved into the action function of both footnote and comment buttons. But then I am not sure if it wouldn't be better to make annotationsHidden a part of the editor itself. That way one could also prevent creating annotations with keyboard shortcuts, when they are hidden.

src/components/Comment/CommentBubbleView.vue
```diff
@@ -40,6 +40,23 @@
 				</div>
 			</div>
 		</div>
+		<div class="comment-bubble__composer">
+			<textarea
```
max-nextcloud: Maybe use `NcTextArea` ?

src/components/Comment/CommentBubbleView.vue
```diff
@@ -59,7 +76,26 @@ const props = defineProps<{
...
+onMounted(() => {
+	props.editor.on('update', onUpdate)
+	// Focus input field when comment bubble is opened.
+	// The ProseMirror transaction that opens the bubble steals focus after finishing, so focus with a delay.
+	setTimeout(() => replyInput.value?.focus(), 50)
+})
+onUnmounted(() => props.editor.off('update', onUpdate))
+
 const commentNode = computed<Node | null>(() => {
+	void editorVersion.value
```
max-nextcloud: Never saw that syntax before. I assume it's for making the commentNode computed depend on editor version, right?

max-nextcloud: That seems like a fairly complex way to trigger vue computed updates from prosemirror to then trigger a prosemirror processing to get the node and update the computed.

How about having a `Ref<Node|null>`, moving the computed function into the onUpdate and setting the ref explicitly?

max-nextcloud: Ah... I see you also want it to depend on `props.referenceId` 🤔

max-nextcloud: How about having a reactive map (either an actual Map or an object) from referenceId to the node in question that is updated for all comment nodes in the update handler and then a computed that picks the one for `props.referenceId`?

That way the prosemirror part would be separated from the vue reactivity part.

src/nodes/CommentReference.ts
```diff
@@ -113,11 +117,84 @@ const CommentReference = Node.create({
...
+					return true
+				})
+
 				return c
-					.focus()
 					.scrollIntoView()
+					.openCommentBubble(referenceId)
 					.run()
 			},
+			addCommentReply: (referenceId: string, text: string) => ({ state, dispatch }) => {
```
max-nextcloud: minor nitpick:
I think I would expect this to take either the comment position or the comment itself as the first argument. The comment node is readily available in the bubble view and that way the command would not have to know about reference ids.

src/components/Comment/CommentBubbleView.vue
```diff
@@ -141,8 +141,10 @@ const props = defineProps<{
 	referenceId: string
 }>()
 
+const DRAFT_KEY_PREFIX = 'text-comment-draft-'
+
 const itemsContainer = ref<HTMLElement | null>(null)
-const replyText = ref('')
+const replyText = ref(sessionStorage.getItem(`${DRAFT_KEY_PREFIX}${props.referenceId}`) ?? '')
```
max-nextcloud, line 189: Is the reference id unique across different files? Or is sessionStorage somehow scoped to one file?

src/plugins/CommentBubblePluginView.ts
max-nextcloud, line 1: Have you considered using a node view for the Comments Node instead?

At first sight that seems simpler / cleaner than to not show the node at all and instead have a plugin that renders it's content separately.

src/nodes/CommentReference.ts
mejo-: I agree, that will be cleaner. Changed now in 7cb621b600ef12e80f6c3eec8e023135165f4405

src/components/Menu/utils.js
mejo-: I agree and I had a similar sentiment regarding where to track the state of hiding annotations. I moved this to a dedicated editor extension now. It was not an easy one, but I think I found a good solution. Curious what you think: 77bfff602f55f183342eab2d68d5a8f5b1c69a7c

jancborchardt:
> I would like to keep this for a follow-up, as I have to say I find it quite confusing how others display all the comments per default and I think we should think this through properly.

@mejo- thanks! :) Of course fine to have it in a follow-up. But I’d say it’s definitely very important, as long comment-threads are not something very common. And if you have to go through a lot of small feedback / grammar fixes in one document, it gets cumbersome if you have to manually select the comments.

src/components/Comment/CommentBubbleView.vue
mejo-: I agree that trying to hold a reactive editor object is ugly. I like your last suggestion to use a reactive map from referenceId to the node and implemented this in 63ec2514619297c1c638c09b0fa3eac43ea97432

mejo-:
> @mejo- thanks! :) Of course fine to have it in a follow-up. But I’d say it’s definitely very important, as long comment-threads are not something very common. And if you have to go through a lot of small feedback / grammar fixes in one document, it gets cumbersome if you have to manually select the comments.

@jancborchardt in general I agree. Just a short addition: there's navigation buttons to navigate to previous/next comment in the comments bubble, that should help a bit with it.

src/components/Comment/CommentBubbleView.vue
mejo-: You're right, there's room for conflicts. I tried to make sure the sessionStorage entry gets cleaned up wherever necessary, so in practice there should be no conflicts. We could add the `fileId` to the key string, but this adds extra complexity as the `fileId` is not easily accessible and in MarkdownContentEditor not available at all.

src/components/Comment/CommentBubbleView.vue
```diff
@@ -58,26 +58,20 @@
...
-						v-if="isEditable && !isGuestWithoutName"
-						variant="tertiary"
-						size="small"
-						:title="t('text', 'Delete')"
-						@click="deleteItem(i)">
-						<template #icon>
-							<DeleteIcon :size="16" />
-						</template>
-					</NcButton>
+					<NcActions v-if="isEditable && !isGuestWithoutName" :container="commentBubble ?? undefined">
```
max-nextcloud: Is that `commentBubble ?? undefined` on purpose? That does not seem like it would do anything.

jancborchardt:
@mejo- yeah I did see those – that's still manual work though (and needs expecting and then finding them first as well). :) Anyway, smth for a follow-up as you said.

src/plugins/CommentBubblePluginView.ts
mejo-: I haven't considered it so far and I'm not sure whether it's the better approach. Given that the Comments node holds all comments, I find it a bit counter-intuitive to add a node view to it that only renders one of them, and in a bubble that needs to positioned relatively to the corresponding CommentReference node.

src/components/Comment/CommentBubbleView.vue
mejo-: Fixed already in bf427f29815f5e95b77eb119c426fe1fdcda70e1

src/nodes/CommentReference.ts
```diff
@@ -200,21 +200,18 @@ const CommentReference = Node.create({
...
 
 				let commentPos = -1
-				let targetComment: ProseMirrorNode | null = null
 				state.doc.descendants((node, pos) => {
-					if (targetComment) {
+					if (commentPos !== -1) {
 						return false
 					}
-					if (node.type.name === 'comment' && node.attrs.referenceId === referenceId) {
+					if (node === comment) {
```
max-nextcloud, line 207: Ah... I see, you still need to loop through the doc descendants to get the `commentPos`. I thought we could do away with the entire loop here as we already loop through it when getting `commentNode` in the BubbleView.

max-nextcloud (approved):
Had a look at all commits. Looks good to me.

ntnsndr:
This is so exciting, thank you!

PR #8916, opened 2026-07-27
Use insertAttachmentPreview instead of insertAttachment on pasting or…

### LINKED ISSUE #8745: Paste PDF files with preview by default + add option to toggle mode
### Problem / Usecase
When adding PDF's to a text via drag and drop, no preview window is generated. It also isn't possible to optionally enable the preview. There isn't even a "three-dot-menu" when the PDF is embedded just as a file.

When adding a PDF via the smartpicker, the PDF is embedded into the text with a preview window.

This is not satisfactory, because adding files via drag and drop is a very common usecase, I think.

### Suggested solutions
- Make embedding with preview the default, for either of the two methods of pasting a PDF
- Make it optional as a point in the "preview options" in the three-dot-menu
- Show the menu for embedded plain files aswell

### Further ideas
- Maybe it could be nice to have a context menu appear when inserting a file by drag and drop, that shows two clickable options: Generate preview window / paste as a simple file

##### Inserted by drag and drop
Image

##### Inserted with smartpicker
Image #/

##### Three-dot-menu
Image

hiobjan:
### 📝 Summary

* Resolves: #8745 #1882

**Problem / Usecase**
When adding a file to a text via drag and drop, no preview window is generated. It also isn't possible to optionally enable the preview.

**Solution**
Use the same behaviour/code as by inserting via smartpicker. I didnt change the behaviour of images, because the attachment-view is better than the attachment-preview. #8774

#### 🖼️ Screenshots

🏚️ Before 
image
 🏡 After
image

### 🚧 TODO

- [ ] #8774

### 🏁 Checklist

### 🤖 AI (if applicable)


mejo-:
Thanks a lot for the contribution @hiobjan. I wholehearted agree that it's confusing to have different representations of attachments in Text and that it's not possible to toggle between them.

I see that your PR improves the UX when a user drags and drops a file into the editor and expects a file preview. On the other hand, this means that inserting attachments via drag and drop behaves differently than inserting attachments via menubar, which is something we should avoid in my opinion.

I'd prefer to first discuss this topic in and issue and decide on a way forward, ideally together with the design team. I opened https://github.com/nextcloud/text/issues/8948 and would be curious about your thoughts there.

I'll therefore close this PR for now.

PR #8917, opened 2026-07-27
fix/autosave timer

max-nextcloud:
- **fix(autosave): every time the typing stops**
- **chore(refactor): document tracking into provideSyncService**
- **fix(autosave): only save when server is ready**
- **fix(autosave): better dirty tracking with versions**
- **chore(refactor): Save service with its own bus**
- **chore(refactor): separate SaveService from SyncService and ydoc**
- **fix(autosave): exponential delay for retries on error**
- **fix(autosave): retry if server throttled save request**
- **fix(autosave): handle out of sync clocks**


src/services/SaveService.ts
```diff
@@ -70,7 +75,6 @@ class SaveService {
 				force,
 				manualSave,
 			})
-			this.emit('stateChange', { dirty: false })
```
mejo-, line 73: Was this removal intentional? It seems unrelated to the commit and I didn't check whether it's save to be removed.

src/services/SaveService.ts
```diff
@@ -115,7 +119,16 @@ class SaveService {
...
 
 	_autosave() {
-		logger.debug('_autosave')
+		const lastSave = this.document.value?.lastSavedVersionTime ?? 0
+		const now = Date.now() / 1000
+		// Server won't accept autosaves yet
+		if (now < lastSave + SERVER_AUTOSAVE_INTERVAL) {
+			logger.debug('Not autosaving as last save is recent', { lastSave, now })
+			const nextSave = lastSave + SERVER_AUTOSAVE_INTERVAL + Math.random() * MAX_RANDOM_AUTOSAVE_DELAY
+			setTimeout(() => this.autosave(), (nextSave - now) * 1000)
```
mejo-: It took me some time to understand the logic here. If I got it right, it means that the next save will happen sometime between 0 and 13 seconds from now but not before lastSave + 10s.

What's the purpose of randomising it in a range of 3 seconds? To make autosaves more useful because other clients save at a different point in time? Or to lower the load on the server?

src/composables/useSyncService.ts
```diff
@@ -47,7 +49,25 @@ export function provideSyncService(
...
 
-	return { document, syncService }
+	const versionWithChanges = ref<number>(0)
+	/**
+	 * Update the tracked version based on the one in the event
+	 *
+	 * @param event that triggered the update
+	 * @param event.version with changes pushed to the server
+	 */
+	function updateVersionWithChanges(event: { version: number }) {
```
mejo-: I always forget about the logic, but can we be sure that a remote higher version always includes our changes? Or can a remote client push a higher version than our local version because they added changes and our changes get lost that way?

src/services/SaveService.ts
```diff
@@ -115,15 +118,21 @@ class SaveService {
...
+		return this.save({ manualSave: false })
+			.then(() => {
+				this.autosaveErrorCount = 0
+			})
+			.catch((error) => {
+				logger.error('Failed to autosave document.', { error })
+				// double the delay on every failed attempt
+				const delay = Math.pow(2, this.autosaveErrorCount) * RETRY_TIMEOUT
+				setTimeout(this.autosave, delay * 1000)
+				this.autosaveErrorCount++
```
mejo-: This can grow infinitely, no? Wouldn't it be better to clamp this to a maximum value?

Also, `Math.pow(2, this.autosaveErrorCount) * RETRY_TIMEOUT` is a bit hard to parse in the first moment, even though I must admit that it's a clever way to achieve "doubly on each increment" 😉

src/composables/useSaveService.ts
```diff
@@ -63,7 +63,15 @@ export function provideSaveService(
 	 * @param event.document latest state of the document
 	 */
 	function updateDocument(event: { document: Document }) {
-		document.value = event.document
+		// Limit lastSavedVersionTime to now. No saving from the future.
```
mejo-, line 66: Given this change I now wonder whether it's clever to calculate delays by comparing server and client timestamps at all. It's probably pretty common that the clocks differ by a few seconds, no?

src/services/SaveService.ts
```diff
@@ -143,7 +140,7 @@ class SaveService {
 				// server did not save due to throttling
 				if (saved === false) {
 					// Make sure to not hammer the server if clocks are out of sync.
-					setTimeout(this.autosave, SERVER_AUTOSAVE_INTERVAL * 1000)
+					setTimeout(this.autosave, (SERVER_AUTOSAVE_INTERVAL - AUTOSAVE_DEBOUNCE) * 1000)
```
mejo-: What's the purpose of subtracting one second here?

mejo- (commented):
I have quite some comments, but most of them are probably indicators that I'm uncertain about the impact of the code changes. To be honest I would prefer pairing review in a call 😆

src/services/SaveService.ts
max-nextcloud: You're right. It's in the wrong commit. Should have been part of [3355f46](https://github.com/nextcloud/text/pull/8917/commits/3355f46c247c01213ed3a6b852ded8f3efbfa3a6) .

It's intentional. Before we toggled the `dirty` state by emitting these events. But that does not work for some cases. Now we compare versions which covers these events:
* Server response happily to save but does not actually save.
  The server will only save new versions every 10 seconds.
  If the latest save just happened it will still respond with 200
  but list the outdated version in the response.
  Comparing the versions shows that our changes have not been saved yet.
* Other user already saved the file with our changes.
  So far dirty would stay true until WE save our changes.
  Comparing the versions also shows the file was saved
  when it was saved by someone else.

src/services/SaveService.ts
max-nextcloud: Without the random range all clients would attempt to save at the same time. That seemed to be asking for trouble. I had race conditions in mind - but also what you are saying.

I later removed both the attempt at calculating the time the server is available again and the randomness for the sake of simplicity and deterministic behavior of the client for easier testing.

src/composables/useSyncService.ts
max-nextcloud: There are two different mechanisms:
* For the **sync** everyone pushes their changes, the server assigns incremental ids to them and yjs builds a consistent document from them. Our local changes are first aggregated in the editor ydoc and then send as an update (step) to the server. The update is independent from `document.version`. It is a diff against a ydoc that is build based on all steps we received from the server. The server then responds with a version which is the version used to store these steps. Any later version means the corresponding steps have been saved later. They do not necessarily build upon our own changes - but yjs handles that.
* For the **save** everyone pushes a serialization of their current editor doc alongside the last version they received and processed. The serialized content will always include all changes up to the version specified and may include local changes that have not been pushed yet.

src/services/SaveService.ts
max-nextcloud: Good points. We do not use the counter anywhere else so we could as well just start with an instance variable and then double that up to a maximum. Also... if the user continues typing autosave will be triggered again anyway. If the server is in maintenance mode and does not respond we will keep trying as the `lastVersionSavedTime` does not increment and this error handling does not block further requests. So we will need something else anyway.

src/composables/useSaveService.ts
max-nextcloud: I did remove some of the computations later on. But maybe we should fully rely on the clients clock.

src/services/SaveService.ts
max-nextcloud: `this.autosave` will trigger a debounce and only save afterwards. So if we want to save after `SERVER_AUTOSAVE_INTERVAL` we need to take that into account.

src/services/SaveService.ts
max-nextcloud: Made it more simple yet. Throttling all autosave request attempts to one request in 10 seconds - no matter if there are errors or not.

src/composables/useSaveService.ts
max-nextcloud: Did so now.

mejo- (approved):

PR #8922, opened 2026-07-28
chore(npm): allow scripts of known NPM dependencies

mejo-:
Starting with NPM v12, scripts defined by dependencies need to be allowed explicitely to protect against supply-chain attacks.

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,5 +149,14 @@
       "email": "jus@bitgrid.net",
       "role": "Developer"
     }
-  ]
+  ],
+  "allowScripts": {
+    "@parcel/watcher": true,
+    "core-js": true,
+    "cypress": true,
+    "esbuild": true,
+    "protobufjs": true,
+    "ssh2": true,
+    "cpu-features": true
+  }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -149,5 +149,14 @@
...
+  ],
+  "allowScripts": {
+    "@parcel/watcher": true,
+    "core-js": true,
+    "cypress": true,
+    "esbuild": true,
+    "protobufjs": true,
+    "ssh2": true,
+    "cpu-features": true
+  }
```
max-nextcloud, line 161: I think I only use `cypress` and `esbuild`. Do we actually need the others?

mejo-: The others are indirect dependencies and we have them since a long time. Until now (with NPM < v12), their scripts get executed all the time you run `npm i` or `npm ci`. If we don't add them, `npm` will mourn and show a warning each time. I think it's acceptable to add them.

max-nextcloud (approved):

PR #8962, opened 2026-08-04
fix(editor): set `translate=yes` for read-only editor

### LINKED ISSUE #2666: Browsers can't translate page content automatically
**Describe the bug**
The built in browser translations only translate the sidebar (page titles), the content of the page itself is untranslated.

**To Reproduce**
1. Go to any collectives page
2. Activate the built in browser translation (tested on Firefox and Chromium)
3. Wait for the translation to complete and observe

**Expected behavior**
The whole page, including the main content should be translated into the selected language.

**Screenshots**
The original content in english:
Image

The content with translation enabled to german:
Image

**Server details:**
- Collectives app version: 4.4.2
- Nextcloud version: 34.0.2
- PHP Version: 8.3.32
- Database: MariaDB 10.11.18

**Client details:**
 - OS: Fedora/Linux
 - Browser: Firefox 153.0 & Chromium  150.0.7871.124
 - Device: Desktop

mejo-:
Tells browsers to translate the editor content despite it being a `contenteditable` element.

Fixes: nextcloud/collectives#2666

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/extensions/RichText.ts
+++ b/src/extensions/RichText.ts
@@ -54,6 +54,7 @@ import Markdown from './Markdown.js'
 import Mention from './Mention.js'
 import Search from './Search.ts'
 import TextDirection from './TextDirection.ts'
+import TranslateAttribute from './TranslateAttribute.ts'
 import Typography from './Typography.ts'
 
 const lowlight = createLowlight(common)
@@ -169,6 +170,7 @@ export default Extension.create<RichTextOptions>({
 					'taskItem',
 				],
 			}),
+			TranslateAttribute,
 			Typography,
 			MathInline,
 			MathBlock,
--- a/src/extensions/TranslateAttribute.ts
+++ b/src/extensions/TranslateAttribute.ts
@@ -1,0 +1,26 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { Extension } from '@tiptap/core'
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+
+const translateAttributeKey = new PluginKey('translateAttribute')
+
+export default Extension.create({
+	name: 'TranslateAttribute',
+
+	addProseMirrorPlugins() {
+		return [
+			new Plugin({
+				key: translateAttributeKey,
+				props: {
+					attributes: () => ({
+						translate: this.editor.isEditable ? 'no' : 'yes',
+					}),
+				},
+			}),
+		]
+	},
+})


grnd-alt (approved):

src/composables/useEditorMethods.ts
```diff
@@ -10,13 +10,16 @@ import type { Session } from '../services/SyncService.ts'
...
+import translatePlugin from '../plugins/translateAttribute.ts'
 import { isUser } from '../services/SyncService.ts'
 
 /**
  *
  * @param editor to apply methods to
  */
 export function useEditorMethods(editor: Editor) {
+	editor.registerPlugin(translatePlugin(editor))
+
```
max-nextcloud: That seems like an odd place to register a plugin. If I remember correctly `useEditorMethods` is called in multiple components such as the `GuestNameDialog` . So will most likely attempt to register the plugin multiple times - which might not be a problem.

My first intuition would be to add this to the `RichText` and `PlainText` extensions.

I'm not sure about `PlainText` - do we want source code to be translatable in case a language is set? Maybe that's a follow up or we start with RichText only.

src/plugins/translateAttribute.ts
```diff
@@ -0,0 +1,27 @@
...
+ * translation despite contenteditable. Falls back to translate="no" when editable.
+ *
+ * @param editor Tiptap editor instance
+ */
+export default function translatePlugin(editor: Editor) {
+	return new Plugin({
+		key: translateAttributeKey,
+		props: {
+			attributes: () => ({
+				translate: editor.isEditable ? 'no' : 'yes',
```
max-nextcloud: I think it would be better to use the `editorState` that is given to the [attributes function](https://prosemirror.net/docs/ref/#view.EditorProps.attributes) here.

That should make it adjust in case the editor switches between being editable and read only.

Currently I think it will only take into account the state of the editor when `translatePlugin` is called.

max-nextcloud (changes requested):
Thanks for tackling this! ❤️ 

I have two requests for changes.

src/composables/useEditorMethods.ts
mejo-: Agreed, changed now.

src/plugins/translateAttribute.ts
mejo-: I don't think that [EditorState](https://prosemirror.net/docs/ref/#state.EditorState) holds the information whether the editor is editable. For that, the [EditorView](https://prosemirror.net/docs/ref/#view.EditorView) would need to be available, as it holds the `editable` property.

I tested this by turning off the network and waiting for an editable editor to turn read-only. Afterwards, the `translate` attribute was changed from `no` to `yes`. After turning on the network again, it switched back to `no`. So it seems to be reactive and work as expected with the current approach.

max-nextcloud (approved):
Looks good now.

Just a linter issue remaining.

PR #9021, opened 2026-08-09
test(playwright): migrate propfind tests from Cypress to Playwright

mejo-:

### 🏁 Checklist

### 🤖 AI (if applicable)


playwright/e2e/propfind.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+		test('always adds rich workspace property', async ({ page, user }) => {
+			const properties = [PROPERTY_WORKSPACE_FLAT, PROPERTY_WORKSPACE_FILE_FLAT]
+
+			await page.goto('/apps/dashboard')
+			await user.uploadFile({ name: 'Readme.md', content: '' })
+
+			const [root1] = await propfindFolder(user, '/', 0, properties)
+			expect(root1).toHaveProperty(PROPERTY_WORKSPACE_FLAT, '')
+
+			await deleteWebDAVResource(user, '/Readme.md')
```
mejo-: This line most likely fixes the flakiness of the test compared to the Cypress one. Without this line, this test succeeds reliably locally but fails reliably (so far) in CI.

Claude assumes that this is because the storage cache doesn't get invalidated properly  in CI when only overwriting the file compared to deleting it and uploading a new file (which results in new fileId etc). It may well be that the missing cache invalidation is due to different APCu caching setup in CI.

max-nextcloud: I wonder if we could fix that different setup. Would such a setup have visible consequences in actual setups or is this a CI artifact we are observing?

mejo-: I don't have hard feelings here. I didn't dive into investigating the actual differences with caching setup between CI and local environment, I merely was happy to have found an explanation for the flaky test and decided to move on :laughing:

mejo-: @max-nextcloud do you see a problem with adding the delete in the test? It arguably changes what is being tested: updating a file no longer is tested. But do you think it should be tested in the propfind tests?

max-nextcloud: Well... this [started out](https://github.com/nextcloud/text/commit/38e95a6da869eae360e4d905e23a9e7be4705259) as a regression test. Would be interesting to see if this actually represents a user visible bug. So if one user creates an empty readme.md and then another one adds content... does the content show? I'll try.

max-nextcloud: So I tried this on cloud.nextcloud.com and it works as intended. However we may not have the problematic caching config and I think I remember that the propfind was also used by the desktop client.
In the browser it basically directly opens an editing session - editable or read only and dynamically syncs the content.

max-nextcloud: Yes... this initially was about android: https://github.com/nextcloud/server/pull/32548#issuecomment-1137225927

max-nextcloud: I was hoping the backports would tell us if this is present in previous versions as well. But they do not work automatically.

max-nextcloud: [test is passing in stable34](https://github.com/nextcloud/text/actions/runs/31523474669/job/93885997032?pr=9035) so this seems to be a recent regression.

max-nextcloud: The test was just being flaky. The underlying cause is that Etags will not be updated if the mtime does not change. So if the file was initially written in the same second the update happened the update will not be propagated.

This problem is specific to our test setup - as usually one does not update a file multiple times per second. I'll adjust the test to account for this.

max-nextcloud (approved):

PR #9028, opened 2026-08-11
refactor/file service

max-nextcloud:
- **refactor(backend): introduce FileService**
- **chore(refactor): separate file getters for share and user**
- **chore(refactor): get file in controller for apiService->create**
- **chore(refactor): use constructor promotion for DocumentServce**
- **chore(refactor): move file getters into FileService**


lib/Controller/PublicSessionController.php
```diff
@@ -64,21 +72,35 @@ protected function isPasswordProtected(): bool {
...
+		/*
+			* Check if we have proper read access (files drop)
+			* If not then well 404 it is.
+			*/
+		try {
+			$this->fileService->checkSharePermissions($token, Constants::PERMISSION_READ);
+		} catch (NotFoundException) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		} catch (NotPermittedException) {
+			return new DataResponse(['error' => $this->l10n->t('This file cannot be displayed as download is disabled by the share')], Http::STATUS_NOT_FOUND);
```
grnd-alt: I would say this code path is dead if the [other comment](https://github.com/nextcloud/text/pull/9028/changes#r3803690630) of mine applies

also: didn't think much more of it but wouldn't a difference in response allow some kind of enumeration on existing public share paths, or is the error not part of the http response?

lib/Service/FileService.php
```diff
@@ -0,0 +1,236 @@
...
+		return $shareAttribtues !== null
+			&& $shareAttribtues->getAttribute('permissions', 'download') === false;
+	}
+
+	/**
+	 * @param $shareToken
+	 *
+	 * @return void
+	 *
+	 * @throws NotFoundException|NotPermittedException
```
grnd-alt: I can't see where this throws NotPermittedException.

`getShareByToken` throws only ShareNotFound which is translated to NotFoundException
and `getPermission` throws NotFoundException and InvalidPathException

grnd-alt (approved):

lib/Controller/PublicSessionController.php
max-nextcloud: Good catch... this scenario was removed ... [three years ago...](https://github.com/nextcloud/text/pull/3765). Time to clean it up 😄.

lib/Service/LockService.php
```diff
@@ -0,0 +1,84 @@
...
+			));
+		} catch (NoLockProviderException|PreConditionNotMetException) {
+		} catch (OwnerLockedException) {
+			return false;
+		}
+		return true;
+	}
+
+	public function unlock(File $file): void {
+		if (!$this->lockManager->isLockProviderAvailable()) {
```
grnd-alt, line 70: nitpick: is there an advantage to calling this method here?

lockManager checks if there's a lockprovider as well, so if there is a lockProvider  `lockManager->getLockProvider` is called twice?

same in the `lock` method

lib/Service/FileService.php
```diff
@@ -0,0 +1,240 @@
...
+
+		// Ideally we'd optimize this part in the future by storing the path and getting the acutal target directly
+		$files = $userFolder->getById($fileId);
+		if (count($files) === 0) {
+			throw new NotFoundException();
+		}
+
+		// Workaround to always open files with edit permissions if multiple occurrences of
+		// the same file id are in the user home, ideally we should also track the path of the file when opening
+		usort($files, static fn (Node $a, Node $b) => ($b->getPermissions() & Constants::PERMISSION_UPDATE) <=> ($a->getPermissions() & Constants::PERMISSION_UPDATE));
```
grnd-alt, line 105: I think this can be done more efficient. If I get this right this sorts the array by comparing if the PERMISSION_UPDATE is present, so essentially is a search for an element in the array which has the PERMISSION_UPDATE without control over which of those comes first?

as afterwards it's checked if there's also PERMISSION_READ  I think this can be simplified into a for loop that searches for a file with PERMISSION_READ and PERMISSION_UPDATE.

something like this:
```
foreach ($files as $file) {
    if ($file->getPermissions() & (Constants::PERMISSION_UPDATE | Constants::PERMISSION_READ) === 3) {
        return $file;
}

grnd-alt (approved):

lib/Service/LockService.php
max-nextcloud: Will address as part of #9062

lib/Service/FileService.php
max-nextcloud: Will address as part of #9062

lib/Service/LockService.php
max-nextcloud: Addressed in a35187c34e155c5d79cf223b6f61e6fd4205d11a

lib/Service/FileService.php
max-nextcloud: Addressed in ab2acac442a3e9bf2d222ca692be8b30fcb0b8fe

PR #9040, opened 2026-08-13
Remove smart picker dialog from text

### LINKED ISSUE #7795: Remove smart picker dialog from text
Smart picker buttons should just open the smart picker inline searchable menu and not the big blocking dialog

xhon-pelushi:
Fixes #7795


max-nextcloud (commented):
Hi @xhon-pelushi 

Thanks a lot for your contribution. Could you add some screenshots to illustrate the effective change in the UI? That will make the design review easier.

The failing `static analysis` test is unrelated. You do not need to worry about that.

xhon-pelushi:
Thanks @max-nextcloud — happy to add screenshots. Summary of the UI change while I gather them:

**Before:** choosing a smart-picker result could open an extra dialog layer on top of the editor.
**After:** the smart picker stays inline in the editor flow (no separate dialog), matching the issue goal of removing the smart picker dialog from Text.

I'll follow up with before/after screenshots as soon as I can capture them from a running instance.

xhon-pelushi:
Thanks @max-nextcloud — before/after screenshots for the smart picker UX change:

**Before** (blocking dialog via `getLinkWithPicker`):

Smart picker dialog

**After** (inline `/` menu — same path as typing slash in the editor):

Inline smart picker menu

Static HTML repros rendered with headless Chromium; code change removes the dialog call and focuses the editor with `/` instead. Can remove `pr-screenshots/` before merge if preferred.

PR #2700, opened 2022-07-09
Do not unintended remove or modify link titles

susnux:
* Resolves: #2699
* Target version: master 

### Summary
Keep the link title set in markdown as `text` is not using it, there is no reason to modify it.
Also added a test case for this.

Issue #599, opened 2020-01-21
no "new" in creator capability


tobiasKaminsky:
Currently there is "new text document" in creators endpoint on text app.
Collabora only shown "presentation", "document", …

As clients might show it different, creator should only reflect to type, so that clients can choose what to show.

Server:
![image](https://user-images.githubusercontent.com/5836855/72802670-98629500-3c4c-11ea-98ea-2f6646a1a171.png)

Android:
![image](https://user-images.githubusercontent.com/5836855/72802759-c3e57f80-3c4c-11ea-8b79-17bd9a0d9773.png)

--> I would add "Create new" to every creator entry, once a new text version is released.

tobiasKaminsky:
@juliushaertl ^

juliushaertl:
Yes, lets do this for 18.0.1 then.

tobiasKaminsky:
Isn't this "just" in Text?

juliushaertl:
Yes, but text is not released in the app store but shipped with the release. That means at the moment we are unfortunately bound to the server release cycle.

PR #9047, opened 2026-08-14
Derive semantic Markdown changes

hweihwang:
Part 1 of the version-comparison work for nextcloud/collectives#599: the semantic model that turns two Markdown revisions into a typed list of changes.

- Two documents in, typed changes out, built on ProseMirror ChangeSet
- Changes classified as text, formatting, attribute, structure or unknown, with conservative move detection
- An explicit complexity limit: you get a bounded result, never a silent partial list

First of six. Followed by #9048, #9049 and #9050 in this repo, then nextcloud/collectives#2699 and #2700.

Full story, screenshots and the test matrix live in the #599 implementation update: https://github.com/nextcloud/collectives/issues/599#issuecomment-5295582982

AI disclosure: developed with OpenAI Codex (gpt-5.6-sol). Codex generated the implementation and tests under my direction; I reviewed, cleaned and verified the code, ran the full test and CI matrix, and signed the commits personally.

### DIFF
--- a/src/comparison/comparisonAlignment.ts
+++ b/src/comparison/comparisonAlignment.ts
@@ -1,0 +1,392 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { ComparisonCoarseReason as CoarseReason } from './markdownComparisonTypes.ts'
+
+export const DEFAULT_COMPARISON_CELL_LEDGER = 40_000
+export const DEFAULT_COMPARISON_TOKEN_LEDGER = 84_000_000
+
+export interface ComparisonWorkLedger {
+	remainingCells: number
+	remainingTokenComparisons: number
+}
+
+export interface ComparisonAlignmentOptions<T> {
+	work: ComparisonWorkLedger
+	fingerprint: (item: T) => string
+	profile: (item: T) => readonly string[]
+	compatible: (before: T, after: T) => boolean
+}
+
+export interface ComparisonAlignmentStep {
+	before: number | null
+	after: number | null
+}
+
+export interface ComparisonCoarseAlignmentRegion {
+	before: { from: number, to: number }
+	after: { from: number, to: number }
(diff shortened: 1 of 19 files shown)


benjaminfrueh:
@hweihwang Tested: 3-column table (headers A/B/C, data x/x/x), delete column C. Header change is reported correctly as C, but the data row shows column A as changed instead of column C. I think it could be because all three "x" cells are identical.

hweihwang:
Thanks for the concrete reproduction — this was exactly the case T01 now pins.

At the commit you reviewed, the three identical body cells (`x/x/x`) gave the column alignment no evidence to tell which column was deleted, so the body rows could be attributed to the wrong column even though the header row resolved correctly.

The fix lands as regression test T01 (`attributes the duplicate-body column deletion`) in the current tip: deleting column C from a 3-column header with two `x/x/x` body rows must emit a single `table-column` delete whose descriptors point at column 2 (C) for the header and both body rows.

Reproduced against the current tip (signed commit `c59958172e`):

- 1 edit, kind `table-column`, operation delete
- descriptor 1: header cell, before path `[0, 0, 2]`
- descriptor 2: body row 1, before path `[0, 1, 2]`
- descriptor 3: body row 2, before path `[0, 2, 2]`

All three descriptors now point at column C — the header and both data rows agree. Tests T01/T02 also cover insertion and the `first/middle/last` variants.

hweihwang:
Thanks for the concrete reproduction. This is exactly the case T01 now pins.

At the commit you reviewed, the three identical body cells (x/x/x) gave the column alignment no evidence to tell which column was deleted, so the body rows could be attributed to the wrong column even though the header resolved correctly.

The fix landed as regression test T01 (attributes the duplicate-body column deletion) at the current tip: deleting column C from a 3-column header with two x/x/x body rows must emit a single table-column delete whose descriptors point at column C for the header and both data rows.

Reproduced against the current tip (c59958172e): one table-column delete, descriptors at before paths [0,0,2], [0,1,2], [0,2,2]. Header and both rows agree on column C. T01/T02 also cover insertion and first/middle/last variants.

benjaminfrueh (approved):

PR #9062, opened 2026-08-15
refactor/introduce IContext

max-nextcloud:
- **chore(refactor): Introduce IContext and FileContext**
- **chore(refactor): simplify IContext API with helper classes**

### Prerequisits

* [x] #9153 
* [x] #9149 

### Backend

- [x] Handle `sync` requests with IContext.
- [x] get text file in `AttachmentService` based on context.
- [x] Fix unit tests for `ApiService`.
- [x] Fix 412 resonses to `push` requests.
- [x] handle saving in IContext.
- [x] Address https://github.com/nextcloud/text/pull/9028#discussion_r3811351730
- [x] Address https://github.com/nextcloud/text/pull/9028#discussion_r3811480603
- [x] check for other requests that rely on the context.
- [x] Migrate existing documents.
- [x] Add index for context_type, context_id.
- [x] Have a lazy context that does not setup the filesystem unless needed.
- [ ] always create the context on the same layer.

### Frontend

- [x]     Add context arguments to createCollaborativeEditor.
- [x]     Pass these arguments to the CollaborativeEditor component as props.
- [x]     Use these props to connect.
- [x]     ~~Provide these props when saving.~~ We still use the `documentId` to save and resolve the context on the backend.
- [ ] Look into attachment handling / attachment resolver - in particular with non-file context

### Testing

- [x] Test with public file shares
- [x] Check playwright tests
- [x] Fix playwright
  - [x] print test since 5cca04cad6416039ecb3102318234d0204eb5e12 shows the placeholder at the bottom
- [ ] Fix cypress tests
- [ ] Test with collectives

### Feedback

- [x] Separate classes
- [x] specify strict types on new classes
- [x] double check: a call to `removeInactiveSessionsWithoutSteps()` within the create flow was removed, and not introduced back elsewhere.
- [ ] change from autoincrement to snowflake ids
- [ ] consider to mark the new event unstable or experimental (not aware of a phpdoc or internal established annotation. )

### Documentation

- [ ] document how to implement and register a Context


max-nextcloud:
## Todo

### Backend

* [ ] Add index for context_type, context_id.
* [ ] Migrate existing documents.
* [ ] handle saving in IContext.
* [ ] check for other requests that rely on the context.

### Frontend

* [ ] Add `context` arguments to `createCollaborativeEditor`.
* [ ] Pass these arguments to the `CollaborativeEditor` component as props.
* [ ] Use these props to connect.
* [ ] Provide these props when saving.

blizzz:
So far I read up to commit `chore(tweak): return first editable file early`. I love your iterative style and the discipline you keep with having small, digestible and sensible commits! My impression is also that with having the abstraction layers, there are many places where it just gets cleaner, and having specific logic better capsuled!

What I stumble over is that the IContext file contains the additional classes `DocumentData`, `SessionInfo`, `NewSessionData`. Usually this is something we keep separate and I am not sure it will always works with autoloading? Also about fellow developer expectations. Maybe something to consider. 

Maybe this is also OK, just to double check: a call to `removeInactiveSessionsWithoutSteps()` within the `create` flow was removed, and not introduced back elsewhere. Alright if intentional of course.

No biggie: new classes miss the `declare(strict_types=1);` statement. 

Can consider to mark the new event unstable or experimental, though I am not aware of a phpdoc or internal established annotation.

blizzz:
On my dev instance I get an error during Migration:

```json
{
  "reqId": "miQrQfvExjkhgL8Z3V6T",
  "level": 3,
  "time": "2026-08-19T14:14:45+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "updater",
  "method": "",
  "url": "--",
  "scriptName": "occ",
  "message": "\\OC\\Updater::failure: Exception: Database error when running migration 090000Date20260817110024 for app text\nAn exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  column \"context_type\" of relation \"oc_text_documents\" contains null values",
  "userAgent": "--",
  "version": "35.0.0.3",
  "occ_command": [
    "occ",
    "upgrade"
  ],
  "data": {
    "app": "updater"
  }
}
```

DB is postgres.

Probably happens because the column is added as `NOT NULL` but the rows would be empty (=> NULL). Maybe set the default to `"file"`?

**UPDATE**: same with context_id. Default to 0?

max-nextcloud:
> Probably happens because the column is added as `NOT NULL` but the rows would be empty (=> NULL). Maybe set the default to `"file"`?

I'm currently thinking we should do a step by step migration:
1) Introduce the new columns - without `NOT NULL`.
2) Migrate the old data.
3) Change the columns to be `NOT NULL`.

The `file` default might work well - but I think it's better to have it set explicitely. For the `contextId` it's very important it's not `0` because we also lookup documents by `contextId` when creating a new session.

blizzz:
> > Probably happens because the column is added as `NOT NULL` but the rows would be empty (=> NULL). Maybe set the default to `"file"`?
> 
> I'm currently thinking we should do a step by step migration:
> 
>     1. Introduce the new columns - without `NOT NULL`.
> 
>     2. Migrate the old data.
> 
>     3. Change the columns to be `NOT NULL`.
> 
> 
> The `file` default might work well - but I think it's better to have it set explicitely. For the `contextId` it's very important it's not `0` because we also lookup documents by `contextId` when creating a new session.

This is definitely the cleaner approach and more robust approach.

max-nextcloud:
@blizzz could you try again?

blizzz:
> @blizzz could you try again?

Works with these changes:

```diff
diff --git a/lib/Migration/Version090000Date20260817110024.php b/lib/Migration/Version090000Date20260817110024.php
index 0a911de25..8abb1d7c2 100644
--- a/lib/Migration/Version090000Date20260817110024.php
+++ b/lib/Migration/Version090000Date20260817110024.php
@@ -45,8 +45,8 @@ class Version090000Date20260817110024 extends SimpleMigrationStep {
                        $table->modifyColumn('id', [
                                'autoincrement' => true,
                        ]);
-                       return $schema;
                }
-               return null;
+
+               return $schema;
        }
 }
diff --git a/lib/Migration/Version090000Date20260819110024.php b/lib/Migration/Version090000Date20260819110024.php
index 6ba16154c..4ae46d98f 100644
--- a/lib/Migration/Version090000Date20260819110024.php
+++ b/lib/Migration/Version090000Date20260819110024.php
@@ -36,11 +36,11 @@ class Version090000Date20260819110024 extends SimpleMigrationStep {
        public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
                $qb = $this->connection->getQueryBuilder();
                $qb->update('text_documents', 'd')
-                       ->set('d.context_type', $qb->createNamedParameter('file'))
+                       ->set('context_type', $qb->createNamedParameter('file'))
                        ->where($qb->expr()->isNull('context_type'))
                        ->executeStatement();
                $qb->update('text_documents', 'd')
-                       ->set('d.context_id', 'd.id')
+                       ->set('context_id', 'd.id')
                        ->where($qb->expr()->isNull('context_id'))
                        ->executeStatement();
        }
```

The first one: migration only succeeds when the autoincrement has to be created. I did not bother undoing this change locally, more robust to just return the schema i think.

The second change is necessary, as it otherwise fails with 

image

src/services/WebSocketPolyfill.ts
```diff
@@ -42,14 +38,16 @@ export default function initWebSocketPolyfill(
 		#onSync
 		#onOpened
 		#processingVersion = 0
+		#documentId = 0
```
blizzz, line 41: TIL [Private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements).

lib/Controller/AttachmentController.php
```diff
@@ -302,6 +286,36 @@ public function getMediaFilePreview(string $mediaFileName, string $shareToken =
 		return new DataResponse('', Http::STATUS_NOT_FOUND);
 	}
 
+	private function getAuth(string $shareToken, bool $updatePermissionRequired = true): IShare|IUser {
```
blizzz, line 289: wonder whether it makes sense to have an umbrella interface here to use as return type hint. Or a wrapper class. Probably overkill. Benefit would be not to have a union return, but at some point you've got to figure out when to work with a share or user at the callee, so… just thinking loudly.

lib/Service/DocumentService.php
```diff
@@ -361,6 +361,10 @@ public function autosave(Document $document, IContext $context, int $version, st
 		if ($lastMTime > 0 && !$force && !$this->cache->get('document-save-lock-' . $document->id)) {
 			$updatedDocument = $context->updateDocument($document);
```
blizzz: `$document` and `$updatedDocument` are they same. Objects are passed by reference, and FileContext does not clone `$document` but just returns it (provided the checks pass and don't return null)

lib/Listeners/DocumentContentUpdatedListener.php
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		$updatedDocument = $context->updateDocument($document);
+		if (!$updatedDocument) {
+			// nothing changed.
+			return;
+		}
+
+		$oldChecksum = $document->getChecksum();
+		$newChecksum = $updatedDocument->getChecksum();
```
blizzz: here as well, same document instance that is acted upon

lib/Listeners/DocumentContentUpdatedListener.php
```diff
@@ -0,0 +1,74 @@
...
+use OCA\Text\Service\DocumentService;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Files\NotFoundException;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @template-implements IEventListener<Event|DocumentContentUpdated>
+ */
+class DocumentContentUpdatedListener implements IEventListener {
```
blizzz, line 24: It needs to be registered yet

lib/Migration/Version090000Date20260820132113.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('text_documents');
+		if (!$table->hasIndex('text_documents_context_index')) {
+			$table->addIndex(['context_type', 'context_id'], 'text_documents_context_index');
```
blizzz: Is it intentionally not a unique index? Allows to have to entries against the same context (e.g. race conditions).

lib/Service/DocumentService.php
```diff
@@ -96,42 +102,57 @@ public function isSaveFromText(): bool {
...
 			$document = $this->documentMapper->insert($document);
-			$this->cache->set('document-version-' . $document->getId(), 0);
+			$this->cache->set('document-version-' . $document->id, 0);
 		} catch (Exception $e) {
 			if ($e->getReason() !== Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
 				throw $e;
 			}
 			// Document might have been created in the meantime
-			$document = $this->getDocument($file->getId());
+			$document = $this->getDocument($document->id);
```
blizzz: or should this be `documentMapper->load()`? `$document` is likely not set here.

lib/Listeners/DocumentContentUpdatedListener.php
```diff
@@ -0,0 +1,74 @@
...
+		$newChecksum = $updatedDocument->getChecksum();
+		if ($oldChecksum !== null && $newChecksum !== null && $oldChecksum === $newChecksum) {
+			// Same content: no need to reset document session. Still update document mtime and etag as they might have changed
+			$this->documentMapper->update($updatedDocument);
+			return;
+		}
+
+		// Reset document session to avoid manual conflict resolution if there's no unsaved steps
+		try {
+			$this->documentService->resetDocument($document->id, true);
```
blizzz: does `$document->id` still match the fileId in this case?

If not, maybe also double check usage in Command\ResetDocument

lib/Middleware/SessionMiddleware.php
```diff
@@ -94,7 +94,8 @@ private function assertDocumentSession(ISessionAwareController $controller): voi
 		$documentId = (int)$this->request->getParam('documentId');
 		$sessionId = (int)$this->request->getParam('sessionId');
 		$token = (string)$this->request->getParam('sessionToken');
-		$shareToken = (string)$this->request->getParam('token');
+		$shareToken = (string)$this->request->getParam('token')
+			|| (string)$this->request->getParam('shareToken');
```
blizzz: unused

src/apis/connect.ts
```diff
@@ -27,16 +34,33 @@ export interface OpenData {
...
  *
  * @param params Parameters identifying the document
  */
-export async function open(params: OpenParams): Promise<{ connection: Connection, data: OpenData }> {
-	const _baseUrl = params.token
-		? generateUrl('/apps/text/public')
-		: generateUrl('/apps/text')
-	const url = `${_baseUrl}/session/${params.fileId}/create`
+export async function openShare(params: OpenShareParams): Promise<{ connection: Connection, data: OpenData }> {
+	const url = generateUrl('/apps/text/public/session/123/create')
```
blizzz: debug leftover?

lib/Service/AttachmentService.php
```diff
@@ -514,48 +419,21 @@ private function isDownloadDisabled(File $file): bool {
 	 * @throws NotFoundException
 	 * @throws NotPermittedException
 	 */
-	private function getTextFile(int $documentId, string $userId): File {
-		$userFolder = $this->rootFolder->getUserFolder($userId);
-		$file = $userFolder->getFirstNodeById($documentId);
+	private function getTextFile(int $documentId, IShare|IUser $auth): File {
+		$document = $this->documentMapper->find($documentId);
```
blizzz: could also result in a DoesNotExistException, which is not handled

lib/Context/ContextManager.php
```diff
@@ -0,0 +1,62 @@
...
+		$this->logger->debug('Registering context for type "' . $type . '".');
+		if (array_key_exists($type, $this->contexts)) {
+			$this->logger->warning('Context of type "' . $type . '" was already registered!');
+			return;
+		}
+		$this->contexts[$type] = $factoryClassName;
+	}
+
+	public function getContext(string $type, int $id, IShare|IUser $auth): IContext {
+		$factoryClassName = $this->getContexts()[$type];
```
blizzz:
```suggestion
		$factoryClassName = $this->getContexts()[$type] ?? null;
```

to avoid log spam on non-existing keys

src/extensions/Autofocus.ts
```diff
@@ -6,7 +6,7 @@
 import { Extension } from '@tiptap/core'
 
 export interface AutofocusOptions {
-	fileId: number | null
+	id: number | null
```
blizzz: is now context type + id, innit?

src/views/CollaborativeEditorApp.vue
```diff
@@ -17,27 +17,20 @@
...
-	fileId = undefined,
-	filePath = undefined,
-	autofocus = false,
-	noLazyImages = false,
-	readonlyBarComponent = undefined,
-	readonlyBarProps = {},
-	shareToken = undefined,
-} = defineProps<{
-	fileId?: number
+defineProps<{
```
blizzz, line 32: I get the hint:

> ViewerComponent.vue migrated to type-only defineProps without withDefaults, making active and isEmbedded required and dropping the onLoadedHandler: () => {} default; Viewer mounts this externally via ViewerView.js (props: ViewerComponent.props), so expect missing-required-prop warnings and @loaded="undefined". And createEditor.ts still gates on Boolean(options.fileId), so a caller using the newly-advertised context option falls through to createMarkdownContentEditor.

But I do not fully understand it 🙈

blizzz (changes requested):

lib/Db/DocumentMapper.php
```diff
@@ -40,6 +41,23 @@ public function find(int $documentId): Document {
 		return Document::fromRow($data);
 	}
 
+	public function load(string $type, int $id): ?Document {
+		/* @var $qb IQueryBuilder */
+		$qb = $this->db->getQueryBuilder();
+		$result = $qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('context_type', $qb->createNamedParameter($type)))
+			->andWhere($qb->expr()->eq('context_id', $qb->createNamedParameter($id)))
```
blizzz, line 50: The optional query type is not specified, at least for context_id it should be `IQueryBuilder::PARAM_INT`. `IQueryBuilder::PARAM_STR` is the default.

lib/Migration/Version090000Date20260817110024.php
```diff
@@ -0,0 +1,51 @@
...
+
+		if (!$table->hasColumn('context_id')) {
+			$table->addColumn('context_id', Types::BIGINT, [
+				'notnull' => false,
+				'unsigned' => true,
+			]);
+		}
+
+		$column = $table->getColumn('id');
+		if (!$column->getAutoincrement()) {
```
blizzz, line 44: actually, I am not sure whether this is safe, also across DB types. I do not see similar code in Nextcloud. Maybe double check with @nickvergessen

nickvergessen: Adding autoincrement does not work as far as I recall. Had it in another app when we had to add the primary keys to all databases.

max-nextcloud: What else can we do? Drop the table and create it from scratch?
That would kill all ongoing editing sessions but would also make sure all entries follow the new schema and remove the need for a migration.

lib/Controller/AttachmentController.php
max-nextcloud: I would like to create the `IContext` at the controller layer everywhere. That would at least avoid passing the Union. We could also have a `SharedFileContext` and an `OwnFileContext` or so. That way we'd always have an IContext to pass around.

Will see if i get to this or do it as a follow up.

lib/Context/ContextManager.php
max-nextcloud: TIL 😄

src/apis/connect.ts
max-nextcloud: I wonder why this did not cause CI to fail more heavily.

lib/Migration/Version090000Date20260817110024.php
nickvergessen: Create a new table with correct structure, copy the entries over, new migration file drop the old table

max-nextcloud: Current plan is to use snowflake ids instead of autoincrement.

lib/Listeners/NodeWrittenResetDocumentListener.php
```diff
@@ -41,37 +44,16 @@ public function handle(Event $event): void {
...
-			try {
-				$this->documentService->resetDocument($node->getId(), true);
-			} catch (DocumentHasUnsavedChangesException|NotFoundException $e) {
-				// Do not throw during event handling.
-				// DocumentHasUnsavedChangesException: A document editing session is likely ongoing, someone can resolve the conflict
-				// NotFoundException: The event was called on a file that was just created so a NonExistingFile object is used that has no id yet
-				$this->logger->warning('Reset document skipped in NodeWrittenResetDocumentListener', ['exception' => $e]);
-			}
-		}
+		$this->lockService->unlock($node);
```
blizzz, line 53: Claude claims the unlock happens now before `isSaveFromText()' is being called which was guarding it. Hence it claims after an autosave it is not lock-protected in the session any longer.

lib/Command/ResetDocument.php
blizzz, line 115: `$fileId` (from input) is meant to be `$id` (from foreach) here in this blocks? There are a couple lines that use `$fileId`, e.g. `$mounts = $this->userMountCache->getMountsForFileId($fileId);` right at the start of the loop, or `$file = $this->fileService->getFileById($fileId, $userId);` that fetches always the same file and unlocks it afterwards.

PR #9071, opened 2026-08-17
[stable29] Update autoloader according to CI

max-nextcloud:
The check-autoloader CI job has been failing for a long time.


max-nextcloud:
Follow up for #9070.

composer/composer/installed.php
max-nextcloud, line 1: Not sure where these changes are coming from. `stable30` is using the same php and composer versions in the CI and still does not have these.

PR #9078, opened 2026-08-19
fix[stable33]: add initial state for translation provider

janepie:
Initial state for `translation_available` is missing in this branch for some reason, so the button doesn't appear
Follow up for #8753


max-nextcloud (approved):
code looks good. Also matches current main branch.

PR #9101, opened 2026-08-25
[stable34] fix(FloatingButtons): hide btns until first mousemove

silverkszlo:
### 📝 Summary

Floating buttons overlapped the placeholder when opening Text editor on Android ~tablet-size. This fix hides them on mount until first mousemove. The Vue 3 wrapper of @tiptap/extension-drag-handle does that internally, so this is only needed for the stable branches.

The buttons only render when the viewport is >= 1024px: they are gated by `showFloatingButtons` in `ContentContainer.vue`, which uses `useIsMobile()`, a pure width check. The `DragHandle` plugin positions and reveals the buttons only in its `mousemove` handler. Until then they sit unpositioned at the editor's top-left corner, on top of the placeholder. Touch devices never fire `mousemove`, so on tablet-size touch screens (wide enough to pass the width check) the buttons stayed overlapping the placeholder permanently.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
Attachment file 1 Bug7222980_1785523514966!screen1
|
Screenshot from 2026-08-25 09-20-15

B | A

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/components/Editor/FloatingButtons.vue
+++ b/src/components/Editor/FloatingButtons.vue
@@ -68,6 +68,13 @@ export default {
 		},
 	},
 
+	mounted() {
+		// Start hidden until the drag handle plugin positions the buttons on
+		// first mousemove. The Vue 3 wrapper of @tiptap/extension-drag-handle
+		// does that internally, so this is not needed on NC v35 onwards.
+		this.$el.style.visibility = 'hidden'
+	},
+
 	methods: {
 		onNodeChange({ node, pos }) {
 			this.node = node


max-nextcloud (commented):
So this means these buttons will stay hidden on touch devices no matter what, right?

They do seem useful though. Can't we trigger the positioning differently from the mouse move?

silverkszlo:
> So this means these buttons will stay hidden on touch devices no matter what, right?
> 
> They do seem useful though. Can't we trigger the positioning differently from the mouse move?

As far as I understand it, they were always hidden on touch devices: https://github.com/nextcloud/text/blob/313312bf9365d074baa5871399820f9cf804a750/src/components/Editor/ContentContainer.vue#L65

max-nextcloud:
I think `isMobile` only checks the screen size. It has the wrong name. At least your PR description says:

> The buttons only render when the viewport is >= 1024px: they are gated by showFloatingButtons in ContentContainer.vue, which uses useIsMobile(), a pure width check.

silverkszlo:
> I think `isMobile` only checks the screen size. It has the wrong name. At least your PR description says:
> 
> > The buttons only render when the viewport is >= 1024px: they are gated by showFloatingButtons in ContentContainer.vue, which uses useIsMobile(), a pure width check.

Ah, you mean touch devices with big screens, now I get it :)

So according to my UI testing in the browser, the floating buttons appear on touch devices with >=1024px screens when you tap on a paragraph, which should be fine, right?

I have tested that with with Chrome's DevTools → Sensors → Show Sensors → Touch: Force enabled

max-nextcloud (approved):

PR #9102, opened 2026-08-25
fix(sync): do not mistake own save requests for outside changes

silverkszlo:
### 📝 Summary

The bug: whenever working on a Text document and after saving a couple of times, the conflict view would appear, although no other user edited the document.

The document is loaded at the start of each request. A sync request compares its etag and checksum against the file only after several other queries. If a save request finishes in between, the file has changed but the loaded document does not reflect that yet, so the sync request then returned 409 and the client showed the conflict view, even though nobody else edited the file.

Reload the document before raising the conflict to compare against the latest saved state. Genuine outside changes still mismatch after the reload and keep raising the conflict.

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/DocumentService.php
+++ b/lib/Service/DocumentService.php
@@ -307,6 +307,7 @@ class DocumentService {
 
 	/**
 	 * @throws DocumentSaveConflictException
+	 * @throws DoesNotExistException
 	 * @throws InvalidPathException
 	 * @throws NotFoundException
 	 */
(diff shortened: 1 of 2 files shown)


max-nextcloud (commented):
I only looked at the code changes thus far. They make sense to me. At least I cannot see how they would do harm. At the same time cypress and playwright are failing - which I don't quite understand. So I rebased the PR to see if that's still the case.

I have no idea if the changes to the additional unit test tests anything meaningful.

max-nextcloud (approved):
Now CI is green and I had a look at the test and it actually tests what it is supposed to.

PR #9149, opened 2026-09-01
fix/no conflicts on sync

max-nextcloud:
fix(sync): do not check for conflicts during sync

The document may be in the middle of getting saved.

When the file has been updated
but the metadata in the documents table has not
it may seem like the file was overwritten
even though text is saving the file.

This does not indicate an actual conflict.
Avoid triggering the conflict screen in that race condition.

- **Revert "fix(sync): do not mistake own save requests for outside changes"**
- **fix(sync): do not check for conflicts during sync**


AndyScherzinger (approved):
🐘

PR #9155, opened 2026-09-02
chore(test): avoid duplicate extensions in testEditor

max-nextcloud:
Filter the default extensions by name so they can be replaced in the test.

Avoids warnings in footnotes and comments spec.

## Before

grafik


mejo- (approved):
Thanks for looking into this! I pushed one tsc error fix.

PR #9156, opened 2026-09-02
Fix link bubble behaviour

As3r1408:
### 📝 Summary

* Resolves: -

This improves link interaction in read-only mode.

Previously, clicking a link in read-only mode could open the link bubble instead of directly navigating to the linked destination. This change makes links behave more naturally in read-only mode by opening the destination directly when clicked.

The existing link bubble behaviour in editing mode is preserved, including the ability to use the bubble for interacting with links without immediately navigating away.

While making this change, an issue was also encountered where toggling between the link display options could leave the context menu open and briefly show both options as selected. This was fixed as part of the same change so that the menu closes correctly after changing the link display option.

<s>Related to [nextcloud/collectives#1026](https://github.com/nextcloud/collectives/issues/1026).</s>

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
imageimage image| image image

### 🏁 Checklist

### DIFF
--- a/src/components/Editor/PreviewOptions.vue
+++ b/src/components/Editor/PreviewOptions.vue
@@ -4,6 +4,7 @@
 -->
 <template>
 	<NcActions
+		ref="actions"
 		v-model="open"
 		data-text-link-options="select"
 		class="link-options"
@@ -16,7 +17,7 @@
 			data-text-preview-option="text-only"
 			name="preview-option"
 			value="text-only"
-			:modelValue="type"
+			:modelValue="selectedType"
 			@change="(e) => toggle(e.currentTarget.value)">
 			{{ t('text', 'Text only') }}
 		</NcActionRadio>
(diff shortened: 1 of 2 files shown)


max-nextcloud:
Hi @As3r1408 
Thanks for your contribution and welcome!

max-nextcloud:
I think we discussed this and the main reason for having the link bubble in read only mode was consistency with the edit mode.

Briefly checked the linked issue but could not find anything about the discussion.

@mejo- I think you looked into this more than I did. What's your take on it?

As3r1408:
> I think we discussed this and the main reason for having the link bubble in read only mode was consistency with the edit mode.
> 
> Briefly checked the linked issue but could not find anything about the discussion.
> 
> @mejo- I think you looked into this more than I did. What's your take on it?

I just checked the related issue, turns out I misread something in there, the issue was actually talking about link previews, not the link bubbles like I thought, that's my bad 😅

I understand the thing about consistency, I just thought it would be better UX-wise

mejo-:
Thanks for your contribution @As3r1408, it's much appreciated. We decided to change the behaviour to always follow the link on click (edit and view mode) instead. See https://github.com/nextcloud/text/pull/9202. And don't hesitate to give it a try and report back what you think about it.

PR #9159, opened 2026-09-02
fix(image): keep data: URI images on markdown round-trip

### LINKED ISSUE #9108: [Bug]: Editing a Markdown file with inline (data: URI) images deletes them from the file
### ⚠️ This issue respects the following points: ⚠️

- [x] This is **not** a troubleshooting question, general support matter, or webserver/proxy problem, but likely a bug _(if unsure, ask the [Community Help Forum](https://help.nextcloud.com))_.
- [x] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aissue+label%3Abug) OR solved at the [Community Help Forum](https://help.nextcloud.com/) _(I've searched!)_.
- [x] I'm using a maintained major version of Nextcloud Server **and** tested against the latest patch level. _([Supported major versions and current patch levels](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule))_.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).
- [x] I've tried my best to provide clear reproduction steps that someone unfamiliar with this bug could use to reproduce it.

### Bug description

Text does not render Markdown images whose source is a `data:` URI, and — more seriously — once
such a document is opened and edited, **the images are removed from the stored file**. They are not
shown as broken, or as a placeholder: the image is absent from the document Text builds, so the
next save writes the file back without it. A one-word text edit is enough to destroy every picture
in the document.

Inline `data:` images are valid Markdown (CommonMark places any URL in `alt`), and they are
the natural format for self-contained documents — a single file that stays complete when it is
moved, copied, shared, or read offline. Several tools export chats, notes and reports this way.

### Steps to reproduce

1. Create a Markdown file containing one inline image, e.g.
   `a picture` — a normal JPEG, nothing exotic.
2. Upload it to Nextcloud (WebDAV or the Files web UI). Confirm the file on the server still
   contains the `data:` URI.
3. Open it in Files. Text displays the document — **the image is missing**, with no indication
   that anything was omitted.
4. Change one word of the text and let Text save.
5. Download the file, or inspect it on the server.

### Expected behavior

Either the image renders, or — at minimum — content Text cannot represent is **preserved on save**.
Silently discarding part of a user's file is the part that matters: the document was complete when
it was opened and is not when it is closed.

### Actual behaviour

The image is gone from the stored file.

Measured on a live instance (sizes and MD5s of the same document before and after):

| copy | size | md5 |
|---|---|---|
| as uploaded | 56344 | `ad73141d3179f82d35e97bf482fe23b7` |
| after opening in Text and editing one line | **2888** | `97f7f445021d9b82fbaa4ff015ae3757` |

The `` line is not altered or replaced — it is deleted. Nextcloud's
own file versioning kept the original, which is the only reason nothing was lost; a user without
versions enabled, or one who notices after the retention window, loses the images permanently.

### Where it comes from

Text builds image nodes from markdown-it tokens via the image node's `parseMarkdown`, and the
TipTap `Image` extension it uses declares:

```js
addOptions() { return { inline: false, allowBase64: false, HTMLAttributes: {}, resize: false } }
…
parseHTML() { return [{ tag: this.options.allowBase64 ? "img[src]" : 'img[src]:not([src^="data:"])' }] }
```

`allowBase64` is TipTap's own default and Text never sets it, so `data:` sources are excluded. The
node is therefore never created; the document Text holds has no image in it; and the Markdown
serializer, which writes what the document contains, emits a file without the image.

Worth noting that every *other* layer already accepts these images:

- **markdown-it**, which Text bundles, explicitly permits them — its own validator is
  `/^data:image\/(gif|png|jpeg|webp);/`, i.e. raster images pass and `data:image/svg+xml` does not.
- **Nextcloud's CSP** already allows `data:` images (`img-src * data: blob:`).

So the exclusion is not a considered security position — it is an unset option, inherited from a
library default, that happens to also delete user data.

### Suggested fix

1. **Never drop unrepresentable content on save.** Whatever is decided about rendering, a document
   that arrives with content Text cannot display should not be written back without it. This is the
   data-loss half and it is independent of the rest.
2. **Allow inert raster data URIs.** Enable base64 images for `data:image/png`, `image/jpeg`,
   `image/gif`, `image/webp` and keep rejecting `data:image/svg+xml`, which is the only genuinely
   dangerous case (SVG can carry script). markdown-it's existing allow-list is exactly this set, so
   the two layers would finally agree.
3. If inline images are unwanted as a matter of product design, then **say so in the UI** — a
   placeholder reading "1 image not displayed" would at least warn the user before they edit and
   lose it.

### Workaround (for anyone finding this issue)

Removing the guard from the built chunk restores both rendering and round-trip safety:

```
:not([src^="data:"])   →   (removed, 3 occurrences)
```

in `apps/text/js/Wrapper-<hash>.chunk.mjs`. It is wiped by every update, it fails the code
integrity check, and it allows SVG data URIs too — so it is a stopgap, not a fix.

### Nextcloud Server version

34

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.3

### Web server

Nginx

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [x] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```json

```

### List of activated Apps

```shell
### Steps to reproduce

1. Create a Markdown file containing one inline image, e.g.
   `![a picture](data:image/jpeg;base64,/9j/4AAQSkZJRgABAQ…)` — a normal JPEG, nothing exotic.
2. Upload it to Nextcloud (WebDAV or the Files web UI). Confirm the file on the server still
   contains the `data:` URI.
3. Open it in Files. Text displays the document — **the image is missing**, with no indication
   that anything was omitted.
4. Change one word of the text and let Text save.
5. Download the file, or inspect it on the server.

### Expected behaviour

Either the image renders, or — at minimum — content Text cannot represent is **preserved on save**.
Silently discarding part of a user's file is the part that matters: the document was complete when
it was opened and is not when it is closed.

### Actual behaviour

The image is gone from the stored file.

Measured on a live instance (sizes and MD5s of the same document before and after):

| copy | size | md5 |
|---|---|---|
| as uploaded | 56344 | `ad73141d3179f82d35e97bf482fe23b7` |
| after opening in Text and editing one line | **2888** | `97f7f445021d9b82fbaa4ff015ae3757` |

The `![…](data:image/jpeg;base64,…)` line is not altered or replaced — it is deleted. Nextcloud's
own file versioning kept the original, which is the only reason nothing was lost; a user without
versions enabled, or one who notices after the retention window, loses the images permanently.

### Where it comes from

Text builds image nodes from markdown-it tokens via the image node's `parseMarkdown`, and the
TipTap `Image` extension it uses declares:


addOptions() { return { inline: false, allowBase64: false, HTMLAttributes: {}, resize: false } }
…
parseHTML() { return [{ tag: this.options.allowBase64 ? "img[src]" : 'img[src]:not([src^="data:"])' }] }


`allowBase64` is TipTap's own default and Text never sets it, so `data:` sources are excluded. The
node is therefore never created; the document Text holds has no image in it; and the Markdown
serializer, which writes what the document contains, emits a file without the image.

Worth noting that every *other* layer already accepts these images:

- **markdown-it**, which Text bundles, explicitly permits them — its own validator is
  `/^data:image\/(gif|png|jpeg|webp);/`, i.e. raster images pass and `data:image/svg+xml` does not.
- **Nextcloud's CSP** already allows `data:` images (`img-src * data: blob:`).

So the exclusion is not a considered security position — it is an unset option, inherited from a
library default, that happens to also delete user data.

### Suggested fix

1. **Never drop unrepresentable content on save.** Whatever is decided about rendering, a document
   that arrives with content Text cannot display should not be written back without it. This is the
   data-loss half and it is independent of the rest.
2. **Allow inert raster data URIs.** Enable base64 images for `data:image/png`, `image/jpeg`,
   `image/gif`, `image/webp` and keep rejecting `data:image/svg+xml`, which is the only genuinely
   dangerous case (SVG can carry script). markdown-it's existing allow-list is exactly this set, so
   the two layers would finally agree.
3. If inline images are unwanted as a matter of product design, then **say so in the UI** — a
   placeholder reading "1 image not displayed" would at least warn the user before they edit and
   lose it.

### Workaround (for anyone finding this issue)

Removing the guard from the built chunk restores both rendering and round-trip safety:


:not([src^="data:"])   →   (removed, 3 occurrences)


in `apps/text/js/Wrapper-<hash>.chunk.mjs`. It is wiped by every update, it fails the code
integrity check, and it allows SVG data URIs too — so it is a stopgap, not a fix.

### Server configuration

- Nextcloud: **34.0.3**
- Text app: **7.0.1**
- PHP: 8.3.6
- OS: Ubuntu 24.04 LTS
- Web server: nginx
- Database: MariaDB 10.11

### Client

Chromium-based browser, desktop. Not browser-specific — the content never reaches the DOM.
```

### Nextcloud Signing status

```shell

```

### Nextcloud Logs

```json

```

### Additional info

_No response_

guillaume-flambard:
Fixes #9108.

Editing a markdown file that contains base64 images (data: URI) deleted those images from the file on the next save. The root cause sits in the image parse rules: both the block `Image` and `ImageInline` extensions exclude `img[src^=data:]` when the TipTap `allowBase64` option is false, and nothing in this app turns it on. markdown-it parses the data: URI fine, but the img never becomes a document node, so the save writes the file without it. That exclusion is the upstream TipTap default, not a decision this app ever made, and there is nothing in the git history suggesting it was deliberate here.

The change enables `allowBase64` on both extensions. The file content is the user's own and the editor's job is to preserve it, not to filter it. Rendering already goes through the `ImageView` node view, which handles the src as-is.

Regression test added in `markdown.spec.js`, covering both paths:

- a standalone data: URI image (block level, wrapped in a figure)
- an inline data: URI image inside a paragraph

Both round-trip through the editor byte-identical. Full unit suite: 1582/1582 across 50 files. Before the change the new test fails (the image is dropped), which is the reported data loss.

### DIFF
--- a/src/nodes/Image.ts
+++ b/src/nodes/Image.ts
@@ -54,6 +54,10 @@ const Image = TiptapImage.extend<ImageOptions>({
 		return {
 			...this.parent?.() as ImageOptions,
 			noLazyImages: false,
+			// Markdown files can legitimately contain base64 data: URI images.
+			// Parsing them is required, otherwise they are silently dropped on
+			// the next save (issue #9108).
+			allowBase64: true,
 		}
 	},
 
--- a/src/nodes/Image.ts
+++ b/src/nodes/Image.ts
@@ -35,12 +35,15 @@ const Image = TiptapImage.extend<ImageOptions>({
 	},
 
 	parseHTML() {
+		if (!this.options.allowBase64) {
+			return [{ tag: 'figure img[src]:not([src^="data:"])' }]
+		}
+		// With base64 parsing on, admit a data: URI only when its mime type is an
+		// image one. A data:text/html src then never becomes a node, so the parse
+		// rule does not have to rely on the content security policy alone.
 		return [
-			{
-				tag: this.options.allowBase64
-					? 'figure img[src]'
-					: 'figure img[src]:not([src^="data:"])',
-			},
+			{ tag: 'figure img[src]:not([src^="data:"])' },
+			{ tag: 'figure img[src^="data:image/"]' },
 		]
 	},
 
--- a/src/nodes/ImageInline.ts
+++ b/src/nodes/ImageInline.ts
@@ -49,6 +49,9 @@ const ImageInline = TiptapImage.extend<ImageOptions>({
 			...this.parent?.() as ImageOptions,
 			noLazyImages: false,
 			inline: true,
+			// See Image.ts: data: URI images must survive an edit round-trip
+			// instead of being dropped on save (issue #9108).
+			allowBase64: true,
 		}
 	},
 
--- a/src/nodes/ImageInline.ts
+++ b/src/nodes/ImageInline.ts
@@ -35,12 +35,15 @@ const ImageInline = TiptapImage.extend<ImageOptions>({
 	},
 
 	parseHTML() {
+		if (!this.options.allowBase64) {
+			return [{ tag: 'img[src]:not([src^="data:"])' }]
+		}
+		// With base64 parsing on, admit a data: URI only when its mime type is an
+		// image one. A data:text/html src then never becomes a node, so the parse
+		// rule does not have to rely on the content security policy alone.
 		return [
-			{
-				tag: this.options.allowBase64
-					? 'img[src]'
-					: 'img[src]:not([src^="data:"])',
-			},
+			{ tag: 'img[src]:not([src^="data:"])' },
+			{ tag: 'img[src^="data:image/"]' },
 		]
 	},
 
--- a/src/tests/markdown.spec.js
+++ b/src/tests/markdown.spec.js
@@ -38,6 +38,13 @@ describe('Markdown though editor', () => {
 		expect(markdownThroughEditor('~~Test~~')).toBe('~~Test~~')
 		expect(markdownThroughEditor('Have an `inline code` element')).toBe('Have an `inline code` element')
 	})
+	test('images with data: URI survive a round-trip (#9108)', ({ markdownThroughEditor }) => {
+		const dataUri = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=='
+		// standalone image (block level, wrapped in a figure)
+		expect(markdownThroughEditor(`![pixel](${dataUri})`)).toBe(`![pixel](${dataUri})`)
+		// inline image inside a paragraph
+		expect(markdownThroughEditor(`Before ![pixel](${dataUri}) after`)).toBe(`Before ![pixel](${dataUri}) after`)
+	})
 	test('ul', ({ markdownThroughEditor }) => {
 		expect(markdownThroughEditor('+ foo\n+ bar')).toBe('+ foo\n+ bar')
 		expect(markdownThroughEditor('* foo\n* bar')).toBe('* foo\n* bar')
--- a/src/tests/markdown.spec.js
+++ b/src/tests/markdown.spec.js
@@ -45,6 +45,15 @@ describe('Markdown though editor', () => {
 		// inline image inside a paragraph
 		expect(markdownThroughEditor(`Before ![pixel](${dataUri}) after`)).toBe(`Before ![pixel](${dataUri}) after`)
 	})
+	test('a non image data: URI does not become an image node (#9108)', ({ markdownThroughEditor }) => {
+		// allowBase64 on its own admits any mime type. The parse rules narrow it to
+		// data:image/, so the syntax stays literal text instead of turning into an
+		// image whose src could never render. The user's characters are kept either
+		// way, which is the point of #9108.
+		const htmlUri = 'data:text/html;base64,PHNjcmlwdD5hbGVydCgxKTwvc2NyaXB0Pg=='
+		expect(markdownThroughEditor(`![x](${htmlUri})`)).toBe(`\\![x](${htmlUri})`)
+		expect(markdownThroughEditor(`Before ![x](${htmlUri}) after`)).toBe(`Before \\![x](${htmlUri}) after`)
+	})
 	test('ul', ({ markdownThroughEditor }) => {
 		expect(markdownThroughEditor('+ foo\n+ bar')).toBe('+ foo\n+ bar')
 		expect(markdownThroughEditor('* foo\n* bar')).toBe('* foo\n* bar')


max-nextcloud (commented):
Hi @guillaume-flambard 

Thanks for tackling this. The approach makes sense to me.

We will need to check if this has any security implications. `data:` uris may contain other mime types such as `text/html` or `application/javascript`.

I hope our CSP have us covered there - but we need to confirm that.

My understanding is that you are only trying to preserve the images in the file - not render them. Is that correct?

guillaume-flambard:
Hi @max-nextcloud, thanks for looking at this.

On your second question first, because it changes the answer to the first one: no, preserving is not the whole intent. These images are meant to render, and the code for that is already there and predates this PR.

`AttachmentResolver.isDirectUrl()` already lists `data:` alongside `http://` and `https://` (`src/services/AttachmentResolver.js:140`). A `data:` src goes through the direct URL branch and comes back as `previewUrl` and `fullUrl` untouched. So the rendering path was built for this. The only missing piece was the parse rule, which dropped the node before it ever reached the resolver. That is why the image disappeared on save: it was never in the document to begin with.

On the security side, three things, and one caveat I would rather raise myself.

The CSP does cover it. The server default in `lib/public/AppFramework/Http/ContentSecurityPolicy.php` is `$allowedImageDomains = ["'self'", "data:", "blob:"]`, so `data:` is already an allowed image source. It is not an allowed script, object or frame source: `$allowedScriptDomains` is `["'self'"]`, `$inlineScriptAllowed` is false, and both `$allowedObjectDomains` and `$allowedFrameDomains` default to empty. A `data:text/html` or `data:application/javascript` payload has nowhere to load from under that policy.

The src also only ever reaches one sink. In `ImageView.vue` it is bound to `` in both branches (lines 32 and 56). The `<a>` wrappers around them have no `href`, only click handlers. Nothing puts the src into `v-html`, an iframe or a navigation target. An `` with a non-image mime type fails to decode, `@load` never fires, and `v-show="loaded"` keeps it hidden. You get a broken image, not an execution path.

The caveat: TipTap's `allowBase64` is all or nothing. Upstream it flips the parse selector from `img[src]:not([src^="data:"])` to `img[src]`, with no mime filtering, so the parse rule now accepts `data:` of any type. I think that is fine given the two points above, and given the file content is the user's own. But if you want defense in depth, I am happy to add a `parseHTML` override on both extensions that only admits `data:image/...`, so a malformed mime type never becomes a node in the first place. That is a small change and it keeps the fix for #9108 intact.

Tell me which way you prefer and I will push it.

guillaume-flambard:
Rather than leave you with a question, I pushed the stricter option so you can see both and pick.

Both extensions now narrow the permissive branch instead of accepting any data: URI:

```ts
parseHTML() {
	if (!this.options.allowBase64) {
		return [{ tag: 'figure img[src]:not([src^="data:"])' }]
	}
	return [
		{ tag: 'figure img[src]:not([src^="data:"])' },
		{ tag: 'figure img[src^="data:image/"]' },
	]
}
```

`ImageInline` gets the same without the `figure` scope. A `data:text/html` src now never becomes a node, so the parse rule no longer leans on the CSP and the sink analysis alone.

Writing the test for it turned up something worth knowing: the payload is not lost either. A non image data: URI round-trips as escaped literal text, `\x`, so the user's characters survive exactly as #9108 asks. The stricter rule only moves where the boundary sits between an image and text, it does not reintroduce data loss. The test asserts that string, which is also what distinguishes this behaviour from the previous commit.

Full suite is green locally, 1583 tests, plus eslint and tsc on the changed files.

If you prefer to keep it minimal and rely on the CSP, drop the last commit and the two line version stands on its own.

PR #9160, opened 2026-09-02
fix(collaboration): override addProseMirrorPlugins to prevent default caret user

benjaminfrueh:
### 📝 Summary
Fixes the collaborator cursor showing a id (e.g.  `User: 123`) instead of the actual name in stable35.

## What happened
Tiptap sets its own default user when the caret feature is setup. Since a recent tiptap update (>= 3.24.0), which we use now with stable35, this happens later and ends up overwriting the name we set.

## Fix
Override tiptap's `addProseMirrorPlugins` in our caret extension so it no longer sets a default user, leaving our user data as the only source.

Related tiptap commit which introduced this incompatibility: 
https://github.com/ueberdosis/tiptap/commit/976b8e39a5d45e7141b7690aa58c18a6b45dbfdb

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/extensions/CollaborationCaret.ts
+++ b/src/extensions/CollaborationCaret.ts
@@ -5,6 +5,7 @@
 
 import { t } from '@nextcloud/l10n'
 import { CollaborationCaret as TiptapCollaborationCaret } from '@tiptap/extension-collaboration-caret'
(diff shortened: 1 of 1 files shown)


max-nextcloud (approved):
Code changes look good.

PR #9166, opened 2026-09-03
perf(cleanup): Optimize cleanup job

Koc:
### 📝 Summary

This is continuation of the #8421.

#### Problem:
The text app cleanup background job took roughly 800 seconds per run. The loop over documents with no active session unconditionally called `cleanupAttachments`, even for documents that had unsaved changes and could not be reset. The same ~8,000 documents kept re-entering the cleanup loop every cron run, and `cleanupAttachments` is expensive because it resolves the file, reads Markdown content, scans the database, and iterates the attachment directory.

#### Solution:
- Attempt to `resetDocument` before running `cleanupAttachments`.
- Catch `DocumentHasUnsavedChangesException`:
  - If unsaved changes are recent (last step within 30 days), skip the document. Attachment references may only exist in the unsaved state, so cleaning them now would be incorrect.
  - If unsaved changes are older than 30 days, force-reset the document, log a warning, and then clean up attachments.
- Added a helper to fetch the latest step timestamp for a document.

Impact:
- Cleanup cron runtime dropped from ~800 seconds to ~20 seconds.
- Abandoned unsaved changes older than 30 days are now cleaned up instead of accumulating forever.

### 🏁 Checklist

### 🤖 AI (if applicable)


lib/Cron/Cleanup.php
```diff
@@ -34,7 +37,21 @@ public function __construct(
 	protected function run($argument): void {
 		$this->logger->debug('Run cleanup job for text documents');
 		foreach ($this->documentService->getAllWithNoActiveSession() as $document) {
-			$this->attachmentService->cleanupAttachments($document->getId());
+			$documentId = $document->getId();
+			try {
+				$this->documentService->resetDocument($documentId);
```
Koc, line 42: need to investigate why `resetDocument` was removed before

PR #9186, opened 2026-09-07
enh/snowflake ids

max-nextcloud:
enh(document): use snowflake ids

Do not rely on autoincrement ids,
which require shuffling tables around in the migration
as some databases cannot change columns to autoincrement.

`$document->id` is now a string,
which conflicts with the definition in the Entity class.
Mute the psalm warning about this for now.

Proper solution will be to migrate to the new ORM.
https://carlschwan.eu/2026/08/18/modern-stable-apis-for-your-nextcloud-application/

### Todo

* [x] AttachmentService::cleanupAttachments(int $fileId) is being called with the documentId from Cron/Cleanup:37
* [x] The change to auto-increment in the migration is still in place
* [x] ResetDocument - no file ids.
* [x] Front-end types are not migrated, they still advertize number


lib/Command/ResetDocument.php
```diff
@@ -75,7 +75,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		if ($all) {
 			$fileIds = [];
 			foreach ($this->documentService->getAll() as $document) {
-				$fileIds[] = $document->getId();
+				$fileIds[] = $document->id;
```
blizzz: this not a file id though? (Claude finding)

blizzz (changes requested):
Changes look good to me. Claude found a few items:

- `AttachmentService::cleanupAttachments(int $fileId)` is being called with the documentId from `Cron/Cleanup:37` 
- Front-end types are not migrated, they still advertize `number`
- The change to auto-increment in the migration is still in place

blizzz:
OK, i think my attempt was utter nonsense. But the issue should not really be dependent upon the snowflake ID, but the change in general, and the frontend passing a file id here :thinking: 

– after all, attachment resolver was left unticked in https://github.com/nextcloud/text/pull/9062#issue-5160607973, but the Controller changed, so I guess we have to make it work.

Max, feel free to drop my silly commit.

PR #9202, opened 2026-09-14
fix(links): open links on click and link bubble on hover

### LINKED ISSUE #1227: Open links to other collective pages without popup
When I clink a link to another collective page in "view" mode, the page is not opened. Instead a popup opens with some information about this page and a "copy" icon. To open the page, you have to click the "info" box below. This is quite confusing as you don't really understand why the link does not just bring you to the linked page and what to do.

image

Leave links as they are and let the browser decide what to do. When clicking/tapping it, it will open the linked page, but people can also use their right mouse button or tap and hold it to open the context menu of the **browser** to choose what to do with it.

Make the behaviour configurable as global administrative setting.

In view mode show a popup when hovering the link, but still allow the link to be opened by just clicking it. In edit mode show the popup always when clicking the link, so positioning the cursor into the by clicking it, is still possible.

**Additional context**
Changing the behaviour of well known elements like links is bad for usability. People know what a "link" is: when a text is underlined and usually colored you can click or tap on it will open. This is how links work almost everywhere. There are also other well known patterns: clicking while holding [Ctrl] or [Cmd] will open the link in a new tab. Right-clicking it or tap and hold will open a context menu where you can decide what to do with the link - like copy it, open it in a new tab etc..

By inventing a new behaviour like "open a popup when you click a link" you force people to learn this new behaviour. It is at least confusing but may also be a real problem since the new popup even does not contain any clear indication where to click to open the linked page eventually.

One exception is when _editing_ documents, where you may want to be able to edit the link as well and not just open it. That's the reason why there are popups in _editors_ so one can choose what to do. But when _viewing_ documents, the behaviour of links should not be changed without a very good reason.

### LINKED ISSUE #2713: Allow internal Collectives links to open directly
This issue was split from #2708 as requested.

I use Collectives as a structured wiki / knowledge base. The wiki is divided into several Collectives according to the type of information they contain.

For example, I have a main **Wiki** page which serves as an entry point and contains links to pages in different parts of the knowledge base.

One of these is a separate Collective called **Slovník (Glossary)**. It contains explanations of individual terms used throughout the wiki. When a term needs further explanation, I can simply link the word on another wiki page to its corresponding page in the Glossary.

I use the same approach for other parts of the wiki. For example, **Návody (Guides)** is a separate Collective containing general guides and procedures. Other areas are separated in a similar way, for example IT infrastructure and other categories of documentation.

This allows me to keep different types of information logically separated while still connecting everything through internal links.

The following screenshot shows an example of this structure and the internal links used on the main Wiki page:

Image

The problem occurs when following these links.

After clicking an internal link, the linked page does not open. Instead, a preview is displayed:

Image

To actually navigate to the page, I then have to click the **open-page icon in the upper-right corner of the preview**:

Image

Only after this second click is the actual Collectives page opened.

Image

So navigation currently works like this:

`click internal link → preview → click open-page icon → target page`

With a wiki structured in this way, internal links are the main method of navigating between related information. Having to go through the preview every time therefore adds an unnecessary step, especially when following several related terms or documentation pages.

I would like to have an option to open internal Collectives page links directly:

`click internal link → target page`

The existing preview does not need to be removed. It could remain available for users who find it useful. A per-user setting to choose between the current preview behavior and direct navigation would be sufficient.

For example:

`Open internal page links directly`

This would make it possible to use Collectives more naturally as an interconnected wiki while still keeping different areas of the knowledge base separated into their own Collectives.

### LINKED ISSUE #5528: A11y: Link Bubble does not open on focus via tab key
**Describe the bug**
When a link in a read only document is focussed with the tab key the link bubble does not open - even not on pressing space or enter.

**To Reproduce**
1. Go to https://cloud.nextcloud.com/s/edBtfaJSJbNAnnx
2. tab until focus is on the link
3. press space and enter
4. See no link bubble and no way to open the link.

**Expected behavior**
Have a way to open the link bubble and follow the link.

**Screenshots**
grafik

### LINKED ISSUE #7600: Always focus href entry after using mod-k keyboard shortcut
**Describe the bug**
The behavior of `mod-k` is still inconsistent.
In particular it will not focus the link entry
if the `href` already was falsy - for example right after the start.
(Follow up for #7599 )

**To Reproduce**
1. Open a new text document.
2. Enter some word.
3. Select the word (mod-a) and turn it into a link (mod-k)
4. See that the link entry does not receive focus.

It works after showing the link bubble for a link with an actual href.
It does not work directly after showing the link bubble for a link with an empty href.

**Expected behavior**
Whenever the shortcut is used link bubble should open with the link text entry focussed.

**Server details:**
- Nextcloud version: 31, 32

**Client details:**
 - Browser: tested with Electron and FF
 - Device: Desktop

There are no errors in the relevant logs.
It's just that `key` inside the LinkBubbleView stays the same so the watcher does not trigger.

### LINKED ISSUE #9110: Link pop-up sometimes disappears when Editor is in full-width mode, the document jumps
**Describe the bug**

When working in full diff mode, the Link pop-up often vanishes shortly after opening it, making it not possible to operate. This behaviour does not occur in the default narrow-width mode. Also, the document jumps down probably with the height of the pop up. Seen on c.nc.c

https://github.com/user-attachments/assets/35c3ec1e-77e8-49e6-babd-5c2c252715b4

mejo-:
### 📝 Summary

* Open links straight away on click
* On desktop open link bubble on hover
* Add edit button to open link bubble after each link

Fixes: nextcloud/collectives#1227
Fixes: nextcloud/collectives#2713

---

Also fixes a11y issue with keyboard navigation: with focussed link, the Enter key now opens the link. The link bubble can be opened by, focussing the edit button behind the link and pressing Enter/Space.

Fixes: #5528

---

fix(linkBubble): focus href input field after Ctrl-K

Fixes: #7600

---

fix(LinkBubble): set tippy root position to absolute

Take the link bubble out of the flow before popper positions it, otherwise it briefly narrows the editor content and scroll anchoring scrolls the page.
    
Fixes: #9110

### Screenshots

Before | After
--- | ---
image | image
image | image

### 🏁 Checklist

### 🤖 AI (if applicable)


marcoambrosini (approved):
very nice improvement!

max-nextcloud (approved):
Looks good to me.

Some of the added playwright tests are failing - so that still needs to be fixed.

PR #9213, opened 2026-09-16
docs(text): require JSDoc for public exports

hweihwang:
### Summary

Require JSDoc for public exports in production code and document the 42 affected exports. Keep the rule's existing coverage of function types.

### Tests

Local lint passed with no warnings. All 12 rule checks passed. The 15 changed source files compile to the same JavaScript as before.

### AI

- [x] The content of this PR was partly or fully generated using AI tools.
- [x] The AI-generated content was reviewed, comprehended and tested by a human.


eslint.config.js
```diff
@@ -10,6 +10,14 @@ import { defineConfig, globalIgnores } from 'eslint/config'
 export default defineConfig(
 	...recommended,
 	globalIgnores(['src/tests/fixtures/*']),
+	{
+		name: 'production-public-jsdoc',
+		files: ['src/**/*.{js,ts,vue}'],
+		ignores: ['**/*.test.*', '**/*.spec.*', '**/*.cy.*', '**/test/**', '**/tests/**', '**/__tests__/**', '**/__mocks__/**'],
```
max-nextcloud, line 16: I think we can do without the ignores list.

My understanding is that this configuration object is relaxing the rule to only warn on public functions.
So the ignores means we are more strict when it comes to tests. Or are tests already ignored?

max-nextcloud (approved):
Just one small comment. Other than that looks good to me.

I did not read all the jsdoc comments as I lack the context to verify them (that's what they are for after all).

eslint.config.js
max-nextcloud: Ah... i see... this is matching `GLOB_FILES_TESTING` in `@nextcloud/eslint-config`.

PR #9228, opened 2026-09-20
[stable33] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (6 unfixable, 24 only fixable manually using --force)


mejo-:
Bumps Tiptap and likely needs code changes we did in stable34 and newer when bumping Tiptap there some versions back.

